PR #3, opened 2021-03-22
cokpituous adjustments

martinpitt:
After that we ought to be able to make a first upstream release, and then request the distro packaging.


martinpitt:
@KKoukiou was faster in #2.

PR #11, opened 2021-03-24
Generate and download pre-built dist tarballs

martinpitt:
Copy and adjust the dist tarball workflows from cockpit. These tarballs
are required for packit (as its sandcastle containers are too small to
run webpack), and convenient for livetest and PR testing.

Also copy, simplify, and adjust cockpit's make_dist.py as
test/download-dist, and provide a `make download-dist` shortcut for it.

-----

I tested this on my fork: The [actions ran fine](https://github.com/martinpitt/cockpit-machines/actions), created a [dist repo](https://github.com/martinpitt/cockpit-machines-dist), and I can do this:
```
git clean -ffdx
GITHUB_BASE=martinpitt/cockpit-machines make download-dist
TEST_OS=fedora-34 make vm
```
This will download/unpack the pre-built dist tarball and use it for VM construction without `npm install` and webpack.

### DIFF
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -1,0 +1,36 @@
+name: build-dist
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          # need this to also fetch tags
+          fetch-depth: 0
+          # avoid checking out the default pull/NNN/head ref, as that gives
+          # mismatching SHAs in generated tarballs
+          ref: "${{ github.event.pull_request.head.sha || github.sha }}"
+
+      - name: Set up dependencies
+        run: |
+          sudo apt update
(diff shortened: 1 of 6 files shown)


martinpitt:
I set up a -dist repo: https://github.com/cockpit-project/cockpit-machines-dist

After this lands and the first commit lands, I need to rename the branch to `main`. Apparently creating a new repo on GitHub still defaults to master, and you can't set a default branch before the first commit.

martinpitt:
I'm still not completely happy about this -- `make download-dist` feels both right and wrong. Perhaps we want `make` to always try and download a tarball if available, then CI and local testing would automatically use it. But then again locally we would always get production dist/ instead of a development one. This is all still very experimental, maybe for now I'll just change test/run to do `make download-dist || true` ?

KKoukiou:
@martinpitt So I am actually happy with this as is. One question. Do we want to extract also node_modules as we do in the make_dist script in cockpit?

KKoukiou (approved):

martinpitt:
> Do we want to extract also node_modules

I think eventually we want to ship a few of them like we do in cockpit, to easily run tests. Right now the tarball doesn't contain any.

PR #12, opened 2021-03-24
build: Opportunistically try to download pre-built tarballs

martinpitt:
Eliminate the separate `make download-dist` rule again, and make the
main tarball build rule try download-dist first. If that fails, fall
back to building it locally.

This will make CI and local testing do the right thing: Building
tarballs happens with `NODE_ENV=production` anyway, building will run
`npm install` and thus also download from the internet, and for local
development one usually invokes `make` or `make watch`, not the tarball
build rule.

Drop the static `$(WEBPACK_TEST)` dependency, as in the case of
downloading the tarball, the rule will *produce* that file instead of
requiring it. In the build fallback case, programmatically build it with
another `make` invocation, as a dynamic dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,24 +99,24 @@ devel-install: $(WEBPACK_TEST)
 dist-gzip: $(TARFILE)
 	@ls -1 $(TARFILE)
 
-# ensure that LIB_TEST and spec are older than the unpacked tarball
-download-dist: $(LIB_TEST) cockpit-machines.spec
-	test/download-dist
-	@ls -1 $(TARFILE)
-
+# when downloading a tarball, ensure that LIB_TEST and spec are older than the unpacked tarball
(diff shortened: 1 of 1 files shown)


martinpitt:
`git clean -ffdx; TEST_OS=fedora-34 make srpm` in my branch before sending the PR: builds locally.  Same command after a few minutes when this PR is done with the build/publish workflows downloads a tarball and is done in seconds. :+1:

martinpitt:
Bah, it's really hard to get over this [lifecycle flake](https://logs.cockpit-project.org/logs/pull-12-20210324-134513-f129647b-debian-stable/log.html#25-2) :sob:

KKoukiou (approved):

martinpitt:
That test fails in #5 as well, and I really don't see how I could have broken it here. Let's debug this asap.

PR #13, opened 2021-03-24
packit: Enable for Fedora 33

martinpitt:
Now that we have pre-built dist tarballs, use them for packit. Start
with Fedora 33, other releases will be enabled step by step.


martinpitt:
Hmm, perhaps packit.yaml needs to be on origin before this works? I'll re-submit that as origin branch.

PR #16, opened 2021-03-25
tests: Invoke run-tests with correct main branch

martinpitt:
Otherwise it complains about

    fatal: ambiguous argument 'origin/master': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'
    fatal: bad revision 'origin/master'

which also breaks the "retry affected tests" logic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ codecheck:
 
 # run the browser integration tests; skip check for SELinux denials
 check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
-	test/common/run-tests
+	test/common/run-tests --base main
 
 # checkout Cockpit's bots for standard test VM images and API to launch them
 # must be from master, as only that has current and existing images; but testvm.py API is stable


marusak (commented):
Why is this needed? CI should correctly set up `BASE_BRANCH`.

marusak:
Is this when you run it locally? I wonder if bots would ever pass different BASE_BRANCH than `main`.

martinpitt:
@marusak: Maybe, but `make check` doesn't, see description. But maybe a better fix would be to teach run-tests to auto-detect main vs. master?

martinpitt:
I think adding this to run-tests could work:

    git branch --list --format='%(refname:short)' main master

It shows "master" for bots, cockpit etc,. and "main" for cockpit-machines.

marusak:
so right now BASE_BRANCH will always be `main`. But if we ever make stable branch here, then this would not behave as expected.

> But maybe a better fix would be to teach run-tests to auto-detect main vs. master?

It is about against what branch the PR is opened, not just main vs master.

martinpitt:
Good point -- how do we fix this then? Shall we teach run-tests to simply not do the "affected tests" stuff if `$BASE_BRANCH` is not set?

marusak:
So if this PR is just about local runs, then I think we can check if BASE_BRANCH is defined and if not then for `make check` just define it to `main`.

martinpitt:
Matej and I quickly discussed that on IRC, let's rather fix this in run-tests.

martinpitt:
I sent https://github.com/cockpit-project/cockpit/pull/15589 to fix that in a better way IMHO.

PR #18, opened 2021-03-25
test: Fix TestMachinesNetworks.testNetworkSettings

martinpitt:
This was getting seriously confused because the test removed the
test_network before detaching it from the domain. That seems to be a
corner case in libvirt, and different versions have subtly different
behaviour wrt. sending events. In that case, the original test_network
was removed from the UI very late, at the time when the test already
tried to add a new one, and thus the UI raced on the network's "Edit"
button.

Fix that by detaching the network from the VM before deleting the entire
network. As this happens on a shut-off VM, we don't get an event/UI
update for this when it's done on the CLI. So do that on the UI instead,
and ensure that it updates correctly.

### DIFF
--- a/test/check-machines-networks
+++ b/test/check-machines-networks
@@ -624,6 +624,12 @@ class TestMachinesNetworks(VirtualMachinesCase):
         b.wait_in_text("#vm-subVmTest1-network-1-type", "network")
         b.wait_in_text("#vm-subVmTest1-network-1-source", "test_network")
 
+        # Remove test_network from the VM again
+        b.click("#delete-vm-subVmTest1-iface-1")
+        b.wait_in_text(".pf-c-modal-box__title-text", "Delete Network Interface")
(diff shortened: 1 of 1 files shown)


martinpitt:
[example 1](https://logs.cockpit-project.org/logs/pull-17-20210325-075620-62129586-centos-8-stream/log.html), [example 2](https://logs.cockpit-project.org/logs/pull-17-20210325-072657-62129586-rhel-8-4/log.html).

I locally ran this 5 times against rhel-8-4, centos-8-stream, and debian-stable, and it passed every time. This should actually tickle the "affected test" logic here, so let's see how it fares.

KKoukiou (approved):

martinpitt:
Landing with the debian-stable lifecycle failure -- that's the other number 1 test failure, and this PR at least addresses the other one.

PR #19, opened 2021-03-25
Some test cleanups

martinpitt:


martinpitt:
I am not sure if we actually want to keep rawhide (I think we should at least give it a try, and if necessary ignore failures in PRs), but I at least want one run to validate the "drop version override" commit.

martinpitt:
I don't have high hopes for packit now, as it's sufficiently in the evening for the TF instances to really suck. But at least our CI should be happy.

martinpitt:
Oh wow, that went much better than I hoped for :+1:

KKoukiou (approved):

PR #24, opened 2021-03-26
test: Skip creation of NFS storage pool for fedora-34

skobyda:
On fedora-34, there is a bug which causes nfs-mountd service to fail and
therefore NFS server cannot be set up. This causes NFS storage-pool test
to fail.
Usually naughty is created for that. But then all the other storage pool
creation test cases would be skipped too. That is too many test cases which
would end up being ignored.

https://bugzilla.redhat.com/show_bug.cgi?id=1943157

Fixes failure at https://github.com/cockpit-project/bots/pull/1835/

### DIFF
--- a/test/check-machines-storage-pools
+++ b/test/check-machines-storage-pools
@@ -440,13 +440,19 @@ class TestMachinesStoragePools(VirtualMachinesCase):
             xfail=True,
         ).execute()
 
-        StoragePoolCreateDialog(
-            self,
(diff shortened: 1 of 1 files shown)


marusak:
Can you please also test with bots PR? Thanks!

martinpitt:
As mentioned in the image refresh, let's rather wait for a fixed f-34 and not land this.

PR #26, opened 2021-03-29
workflows: ensure there is a dist build available to consume when creating a tag for the release

KKoukiou:
I have not tested it - I will do it tomorrow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,5 +28,7 @@ jobs:
         run: |
           # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
           export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
+          # Wait for build-dist and publish-dist workflows to complete
+          export DOWNLOAD_DIST_OPTIONS=--wait
           cd /build
           release-runner -r https://github.com/$GITHUB_REPOSITORY -t $(basename $GITHUB_REF) ./cockpituous-release
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -3,6 +3,9 @@ on:
   pull_request:
   push:
     branches: [main]
+    tags:
+      # this is a glob, not a regexp
+      - '[0-9]*'
 
 jobs:
   build:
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -1,5 +1,8 @@
 name: release
 on:
+  workflow_run:
+    workflows: publish-dist
+    types: [completed]
   push:
     tags:
```
martinpitt, line 7: This doesn't quite work -- the workflow will run *both* on pushing the tag, *and* for *every* PR opening and landing.

Rather, keep the `push: tags` trigger and teach the download-dist call in `Makefile` something like `${DOWNLOAD_DIST_OPTIONS:-}` and in this workflow set `DOWNLOAD_DIST_OPTIONS=--wait`.

martinpitt (changes requested):

martinpitt:
FTR, quick hack if this somehow doesn't work: add `GITHUB_BASE=fake/nonexisting` to the release workflow to force the dist download to fail.

.github/workflows/release.yml
KKoukiou: I tested it locally, outside of the release workflow and it works like a charm

I am not sure how I should actually test it by triggering the real release-workflow, I don't want to publish something accidentally from a run from my fork.

.github/workflows/release.yml
```diff
@@ -27,6 +27,6 @@ jobs:
       - name: Run cockpituous
         run: |
           # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
-          export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
+          export HOME=$(getent passwd $(id -u) | cut -f6 -d:) DOWNLOAD_DIST_OPTIONS=--wait
```
martinpitt, line 30: I'd prefer this on a separate line to make it easier to read, but not a blocker.

martinpitt (approved):
Looks good, thank you!

martinpitt:
@KKoukiou : Hmm, is this complete? I thought you also needed to add that auto-generated license file to `%license`?

KKoukiou:
> @KKoukiou : Hmm, is this complete? I thought you also needed to add that auto-generated license file to `%license`?

That was fixeed in another PR. See

commit dfd041034a52905c62a3bcc4a3e94ef6687cb8a6
Author: Katerina Koukiou <kkoukiou@redhat.com>
Date:   Mon Mar 29 13:49:17 2021 +0200

    Fix License field in the package spec file

martinpitt (approved):
Of course, sorry for mixing these up. Thanks!

PR #28, opened 2021-03-30
Makefile: set the name of the extracted tar.gz in download-dist

KKoukiou:


martinpitt:
The "how" looks ok, but I don't understand the "why" yet? I suppose that's some collision between naming them after the SHA vs. after the release tag? How does that collision look like?

I'm a bit worried that the version is not only in the tar name, but also ends up somewhere in the webpack, spec, and other places.

I thought for releases we'd skip the download for now?

KKoukiou:
@martinpitt so indeed I went with skipping download for releases for now. I tested on your repo, with fake GITHUB_BASE, the release workflow run is https://github.com/martinpitt/cockpit-machines/runs/2228831901?check_suite_focus=true

It created correctly the tar file artifact and failed as expected later on.

martinpitt (approved):
Ack, thank you!

PR #34, opened 2021-04-05
download-dist: Rename tarball for tagged HEAD

martinpitt:
After adding a release tag, Makefile expects the tarname to be
`cockpit-machines-<tag>.tar.gz`, instead of the long git format
with `version-N-SHA`. Rename the tarball in that case so that CI,
`make srpm` etc. still work.

This is a bit of a hack for CI, these should not be used for official
release tarballs as they still contain the "long git format" version in
the shipped .spec file. We need to think about how to resolve this.

### DIFF
--- a/test/download-dist
+++ b/test/download-dist
@@ -109,6 +109,17 @@ def download_dist(wait=False):
         ftar.extract(names[0])
         tar_path = os.path.realpath(names[0])
 
+    # Rename tar to what Makefile expects for a release
+    try:
+        tag = subprocess.check_output(['git', 'describe', '--exact-match'],
+                                      stderr=subprocess.DEVNULL, universal_newlines=True).strip()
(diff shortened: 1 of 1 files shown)


martinpitt:
This fixes failures like in https://github.com/cockpit-project/bots/pull/1865 ([log](https://logs.cockpit-project.org/logs/pull-1865-20210405-092118-e15ad857-fedora-34-cockpit-project-cockpit-machines/log.html)) which currently happen as main HEAD is a tag 242.1.

Easy to reproduce locally with `make srpm`.

Better ideas much appreciated! E.g. I wonder if we could simply not ship the pre-generated .spec file perhaps? It's the only place where the version is actually being used. I haven't checked how that plays with our release script.

marusak (approved):
I don't have any better idea TBH. Now at least 3 images are blocked due to this so its fine to unblock us and then we can think about some better solution.

martinpitt:
... of course landing *anything* will side-step that problem, until the next release :) But right, let's.

PR #35, opened 2021-04-06
Always just unpack dist/ and package-lock.json from downloaded dist tarballs

martinpitt:
Downloading and directly using the actual dist tarballs used to be
problematic for releases: After pushing a tag, the downloaded tarball
still had the wrong name such as cockpit-machines-241.1.g1234abc.tar.gz,
and inside it the pre-built .spec file's `Version:` was the same.
This led to various crappy workarounds like commit 9b4a1773dd and
e69a0a857804, and not using the pre-built tarballs at all for releases.

The expensive (for building) and precious (for re-using a webpack that
was previously validated) part is the dist/ directory, *not* the tar
file itself -- once dist/ exists, the remaining version-dependent files
(only the .spec for now) and the .tar can be built in a split second.

So change download-dist to always just unpack dist/ and
package-lock.json and not keep the actual tar file. Also refuse to do
anything if dist/ already exists or NODE_ENV==development, like in
cockpit [1].

Call download-dist from the Makefile's main "build webpack" rule.
Structurally this makes a lot more sense, as now `npm run build` and
`download-dist` do the same -- they produce dist/. This greatly
simplifies the `dist-gzip` rule, as this once again always does the same
thing: Ensure the .spec file is up to date and build a tar from the
sources and dist/.

Change Makefile's NODE_MODULES_TEST to actually test something in
node_modules/, to ensure that `npm install` will actually run when
necessary -- package-lock.json is not good enough, as that gets
extracted from the downloaded tarballs also.

Bring back commit d9b3b601679 to wait for the downloads to exist for
packit and releases.

[1] https://github.com/cockpit-project/cockpit/commit/5434aeebee5e0fe

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -6,7 +6,7 @@ downstream_package_name: cockpit-machines
 actions:
   post-upstream-clone: make cockpit-machines.spec
   create-archive:
-    - test/download-dist --wait
+    - make dist-gzip DOWNLOAD_DIST_OPTIONS=--wait
     - find -name 'cockpit-machines-*.tar.gz'
 jobs:
   - job: tests
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
         run: |
           # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
           export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
-          # This will make download-dist to fail in order to create dist manually instead
-          export GITHUB_BASE=fake/nonexisting
           cd /build
+          # Wait for build-dist and publish-dist workflows to complete
+          export DOWNLOAD_DIST_OPTIONS=--wait
           release-runner -r https://github.com/$GITHUB_REPOSITORY -t $(basename $GITHUB_REF) ./cockpituous-release
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ TARFILE=cockpit-$(PACKAGE_NAME)-$(VERSION).tar.gz
 RPMFILE=$(shell rpmspec -D"VERSION $(VERSION)" -q `ls cockpit-machines.spec.in cockpit-machines.spec 2>/dev/null | head -n1`).rpm
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 # stamp file to check if/when npm install ran
-NODE_MODULES_TEST=package-lock.json
+NODE_MODULES_TEST=node_modules/.bin/webpack
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/manifest.json
 # one example file in src/lib to check if it was already checked out
@@ -75,8 +75,8 @@ download-po: $(WEBLATE_REPO)
 %.spec: %.spec.in
 	sed -e 's/%{VERSION}/$(VERSION)/g' $< > $@
 
-$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(LIB_TEST) $(shell find src/ -type f) package.json webpack.config.js
-	NODE_ENV=$(NODE_ENV) npm run build
+$(WEBPACK_TEST): $(LIB_TEST) $(shell find src/ -type f) package.json webpack.config.js
+	test/download-dist $${DOWNLOAD_DIST_OPTIONS:-} || ($(MAKE) $(NODE_MODULES_TEST) && NODE_ENV=$(NODE_ENV) npm run build)
 
 watch:
 	NODE_ENV=$(NODE_ENV) npm run watch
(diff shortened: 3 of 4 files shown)


martinpitt:
This works with [packit](https://dashboard.packit.dev/results/srpm-builds/19199) and the [initial CI runs which build the webpack themselves](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-35-20210406-041317-a684f93e-debian-stable/log.html) (as they start in parallel with build-dist and thus don't yet have a tarball).

I also tested this with our release scripts:

    RELEASE_SOURCE="_release/source"  ~/upstream/cockpituous/release/release-source

On current main this fails with the previously known
```
download-dist: Downloading dist tarball from https://github.com/cockpit-project/cockpit-machines-dist/raw/main/cc8c4afdc50f483a157c04192fa68a120811a467.tar ...
download-dist: Extracting paths from tarball: dist package-lock.json
/var/home/martin/upstream/cockpit-machines/source.6kKATd/cockpit-machines-242.1.gcc8c4af.tar.gz
ls: cannot access 'cockpit-machines-242.1.tar.gz': No such file or directory
```
which was the reason why we disabled that in commit f660bc09bd5.

On this branch I did a local `git tag -s 243` and ran the same release-source command, which succeeded:
```
> Creating tarball
make: Entering directory '/var/home/martin/upstream/cockpit-machines/source.3ZZiuk'
make: *** No rule to make target 'dist'.  Stop.
make: Leaving directory '/var/home/martin/upstream/cockpit-machines/source.3ZZiuk'
make: Entering directory '/var/home/martin/upstream/cockpit-machines/source.3ZZiuk'
flock Makefile sh -ec '\
    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 240; \
    git checkout --force FETCH_HEAD -- pkg/lib; \
    git reset -- pkg/lib'
sed -e 's/%{VERSION}/243/g' cockpit-machines.spec.in > cockpit-machines.spec
remote: Enumerating objects: 1525, done.
remote: Counting objects: 100% (1525/1525), done.
remote: Compressing objects: 100% (1408/1408), done.
remote: Total 1525 (delta 241), reused 475 (delta 27), pack-reused 0
Receiving objects: 100% (1525/1525), 7.07 MiB | 518.00 KiB/s, done.
Resolving deltas: 100% (241/241), done.
From https://github.com/cockpit-project/cockpit
 * tag               240        -> FETCH_HEAD
mv pkg/lib src/ && rmdir -p pkg
test/download-dist ${DOWNLOAD_DIST_OPTIONS:-} || (make node_modules/.bin/webpack && NODE_ENV=production npm run build)
download-dist: Downloading dist tarball from https://github.com/cockpit-project/cockpit-machines-dist/raw/main/a684f93e1f648bd66b80acb4f6329756a4ef0783.tar ...
download-dist: Extracting paths from tarball: dist package-lock.json
tar czf cockpit-machines-243.tar.gz --transform 's,^,cockpit-machines/,' \
	--exclude cockpit-machines.spec.in \
	$(git ls-files) src/lib/cockpit-po-plugin.js src/lib/patternfly/*.scss package-lock.json cockpit-machines.spec dist/; \

cockpit-machines-243.tar.gz
make: Leaving directory '/var/home/martin/upstream/cockpit-machines/source.3ZZiuk'
```

martinpitt:
I force-retried the previously [successful ubuntu-stable test](https://logs.cockpit-project.org/logs/pull-35-20210406-041317-a684f93e-ubuntu-stable/log.html) to also cover the "download dist/" scenario in CI.

The [retry](https://logs.cockpit-project.org/logs/pull-35-20210406-043820-a684f93e-ubuntu-stable/log.html) successfully used download-dist and did not run webpack. Note that it still does `npm install`, as we need that for running the tests -- at some point we should include these in the downloads, like in cockpit.

packit.yaml
```diff
@@ -6,7 +6,7 @@ downstream_package_name: cockpit-machines
 actions:
   post-upstream-clone: make cockpit-machines.spec
   create-archive:
-    - test/download-dist --wait
+    - make dist-gzip DOWNLOAD_DIST_OPTIONS=--wait
```
KKoukiou, line 9: The env variable here does not need export? It's not clear to me yet, how the env variables are visible to the makefile target processes.

KKoukiou (approved):
I have only one question, but since you verified the release part there it's just for my understanding.

packit.yaml
martinpitt: This is passed as a "make variable", similar to `make V=1`. It has by and large the same effect as

    DOWNLOAD_DIST_OPTIONS=--wait make dist-gzip

or

    export DOWNLOAD_DIST_OPTIONS=--wait
    make dist-gzip

I'm happy to change the syntax if it looks weird to you, of course!

KKoukiou: It's fine of course. Just did not know the syntax.

PR #78, opened 2021-04-06
Virtual networks: Fix address validation

### LINKED ISSUE #36: 'Ipv4 network' field is misleading in virtual networks
In the "Virtual Machine" -> "Virtual Network" -> "Create virtual network", it was asking me for:
"Ipv4 network:", this should be changed to "Ipv4 address", as i did write xx.xx.xx.0 there, and was struggling in many days to figure why the VMs did not have network.
Based on how this is written and the rest of fields, it is easy to think it should be an address like 192.168.1.0.
So I just suggest to change that field to be more clearly written.

skobyda:
Rename `IPvN network` to `IPvN address`

Use [ip package](https://www.npmjs.com/package/ip) to validate addresses

Address inputted by user is both used to figure out network subnet but 
also to assign address to virtual network's bridge.
Network identifier nor brodcast should not be used as address for the 
bridge.
For example, with network 192.168.2.0/24, addresses 192.168.2.0 and 
192.168.2.255 cannot be used as address of bridge.

Fixes #36


KKoukiou (changes requested):
You docstrings need a bit more care, but other than that I am happy with this. The usage of ip although it's actually one more dependency, is really removing a lot of code, so I like it as well .

src/components/networks/createNetworkDialog.jsx
```diff
@@ -55,16 +55,20 @@ function validateParams(dialogValues) {
...
             validationFailed.netmask = _("Mask or prefix length should not be empty");
         else if (!utils.validateNetmask(dialogValues.netmask))
             validationFailed.netmask = _("Invalid IPv4 mask or prefix length");
 
+        if (isEmpty(dialogValues.ipv4.trim()))
+            validationFailed.ipv4 = _("IPv4 network should not be empty");
+        else if (!utils.validateIpv4(dialogValues.ipv4))
+            validationFailed.ipv4 = _("Invalid IPv4 address");
+        else if (utils.ipv4IsBroadcast(dialogValues.ipv4, dialogValues.netmask))
+            validationFailed.ipv4 = _("IPv4 address cannot be same as network's broadcast address");
```
allisonkarlitskaya, line 68: I'd put the, yes.

```suggestion
            validationFailed.ipv4 = _("IPv4 address cannot be same as the network's broadcast address");
```

src/components/networks/createNetworkDialog.jsx
```diff
@@ -55,16 +55,20 @@ function validateParams(dialogValues) {
...
+        else if (utils.ipv4IsNetworkIdentifier(dialogValues.ipv4, dialogValues.netmask))
+            validationFailed.ipv4 = _("IPv4 address cannot be same as network identifier");
```
allisonkarlitskaya, line 70:
```suggestion
            validationFailed.ipv4 = _("IPv4 address cannot be same as the network identifier");
```

src/components/networks/utils.js
```diff
@@ -29,6 +29,46 @@ export function validateIpv4(address) {
     return ip.isV4Format(address);
 }
 
+/**
+ * Valides if user inputted broadcast address
```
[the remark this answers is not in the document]
allisonkarlitskaya: inputted is a weird word.  "gave"?  "used"?  "provided"?

allisonkarlitskaya (commented):
Looked at English grammar only

src/components/networks/utils.js
```diff
@@ -29,6 +29,46 @@ export function validateIpv4(address) {
...
+ * Using broacast address of network space as the address of virtual network's bridge
+ * is forbidden and would disable
```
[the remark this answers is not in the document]
skobyda: Ah the sentence should have continued there. it should say: `would disable the connectivity of the virtual network.`

skobyda:
I addressed all the comments, ammended a test and it's all green now.

src/components/networks/utils.js
```diff
@@ -29,6 +29,46 @@ export function validateIpv4(address) {
     return ip.isV4Format(address);
 }
 
+/**
+ * Valides if the user provided broadcast address
```
KKoukiou, line 33: Validates

src/components/networks/utils.js
```diff
@@ -29,6 +29,46 @@ export function validateIpv4(address) {
...
+ * @param {string} address
+ * @returns {boolean}
+ */
+export function ipv4IsBroadcast(address, netMask) {
+    if (!validateNetmask(netMask))
+        return false;
+    const mask = netmaskConvert(netMask);
+    const subnet = ip.subnet(address, mask);
+
+    console.log(address);
```
KKoukiou, line 46: Do you really want to leave the console logs here?

src/components/networks/utils.js
```diff
@@ -29,6 +29,46 @@ export function validateIpv4(address) {
...
+    console.log(subnet.broadcastAddress);
+    // user provided netmask
+    return address === subnet.broadcastAddress;
+}
+
+/**
+ * Valides if the user provided network identifier as the address
```
KKoukiou, line 53: Validates

src/components/networks/utils.js
```diff
@@ -29,6 +29,46 @@ export function validateIpv4(address) {
...
+ * Using network identifier is however forbidden
+ * During virtual network creation, address is assigned to bridge
+ * But no interface can have the address same as the network identifier,
```
KKoukiou, line 56: So I would rewrite the function docstring to not be so verbose. Also I think this is a helper/util, we don't care about what Is allowed or not in the UI here, we just want to check one thing, if an address is the Network Identifier. 

So:
```
Returns if the provided address is the network's identifier
```  

If you want to mention about the limitations, these comments belong to the place where you use this function.

src/components/networks/utils.js
KKoukiou: See comment bellow: 
Let's rewrite it to: 
```
Returns if the provided address is the network's broadcast address
```

KKoukiou (changes requested):
Your docstrings are still problematic. Lets just simplify them, all the explanatory text did not belong there in the first place.

src/components/networks/utils.js
skobyda: Fixed this and all above

KKoukiou (approved):
Thanks!

KKoukiou:
@martinpitt here packit tests consistently fail with - JavaScript heap out of memory.
I will put the memory webpack adjustment to the packit.yml in another commit, I guess it's fine to merge this as is, right?

martinpitt:
@KKoukiou : packit can't build the webpack, it relies on the pre-built dist cache. I think the reason is that this PR is not rebased, and we landed the reorganization of the dist/ cache in between. Thus this is fine to land, yes.

PR #81, opened 2021-04-07
workflows: Use cockpituous instead of github token for npm/po updates

martinpitt:
The default `GITHUB_TOKEN` does not have the privileges to trigger
further workflows, in particular running unit tests. This is deliberate
behaviour [1].

So follow the recommendation and use our Cockpituous token instead, like
we did in cockpit [2][3]

[1] https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
[2] https://github.com/cockpit-project/cockpit/commit/4950ac6ed3
[2] https://github.com/cockpit-project/cockpit/commit/d65d4c3789

### DIFF
--- a/.github/workflows/npm-update.yml
+++ b/.github/workflows/npm-update.yml
@@ -16,10 +16,13 @@ jobs:
       - name: Set up configuration and secrets
         run: |
           printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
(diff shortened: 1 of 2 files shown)


martinpitt:
This fixes the missing "build-dist" workflow on #80 or #79 and thus the failing packit tests.

marusak (approved):
Thanks!

PR #84, opened 2021-04-09
Rework pre-built dist upload/download

martinpitt:
With test/download-dist now only ever using the pre-built dist/ and
package-lock.json instead of the full tarballs, there is no reason any
more to build and store the actual tarballs.

Simplify build-dist to have a constant artifact name "dist" (which
simplifies downloading/unpacking), and instead store the SHA as a plain
file inside the artifact. Do away with the intermediate tarball
(as there are not so many files in dist/ to make upload-artifact
inefficient).

In publish-dist, directly commit dist/ and package-lock.json to the
-dist repo, and tag the commit with the SHA, so that they are still easy
to look up. Drop the "old tarball" removal. This uses git in a much more
efficient way, as it now only stores the deltas. In particular, for
commits/PRs which don't change src/, that will just add an additional
tag to the previous build and store no extra files at all.

Rework download-dist to get the dist repo in ~/.cache/cockpit-dev/, so
that it persists across project checkouts, and shallow-fetch the desired
SHA build.

Change prune-dist to keep the commits of the last 7 days, and truncate
older ones.

-----

 - [ ] Re-initialize -dist repository before landing this

I tested all of this on my [fork](https://github.com/martinpitt/cockpit-machines/actions), first with a non-existing ~/.cache/cockpit-dev:
```
❱❱❱ git clean -ffdx; time GITHUB_BASE=martinpitt/cockpit-machines test/download-dist 
Removing dist/
Removing package-lock.json
download-dist: Creating dist cache /home/martin/.cache/cockpit-dev/martinpitt/cockpit-machines-dist
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Unpacking objects: 100% (1/1), 195 bytes | 195.00 KiB/s, done.
From https://github.com/martinpitt/cockpit-machines-dist
 * tag               sha-1287884f90826914784bc24c9513c827c2b5b727 -> FETCH_HEAD
download-dist: Extracting dist cache...

real	0m2.776s
```

and after that, with a hot cache:
```
❱❱❱ git clean -ffdx; time GITHUB_BASE=martinpitt/cockpit-machines test/download-dist 
Removing dist/
Removing package-lock.json
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Unpacking objects: 100% (1/1), 195 bytes | 195.00 KiB/s, done.
From https://github.com/martinpitt/cockpit-machines-dist
 * tag               sha-1287884f90826914784bc24c9513c827c2b5b727 -> FETCH_HEAD
download-dist: Extracting dist cache...

real	0m0.995s
```

You can see the commits and contents at https://github.com/martinpitt/cockpit-machines-dist/commits/main

I also tested history truncation with an additional [temporary commit that sets the time to "13 hours" instead of "7 days"](https://github.com/martinpitt/cockpit-machines/commit/1287884f908269147), and it [succeeded](https://github.com/martinpitt/cockpit-machines/actions/runs/731920857). I also did screenshots of the "before" state where it still had a few commits from yesterday, but foolishly deleted them before uploading. D'oh! But I promise that this worked fine.

I also [ran prune-dist again when there is nothing to prune](https://github.com/martinpitt/cockpit-machines/runs/2303141851?check_suite_focus=true), this worked fine as well.

### DIFF
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -12,6 +12,7 @@ jobs:
       - name: Download build-dist artifacts
         uses: dawidd6/action-download-artifact@v2
         with:
+          name: dist
           workflow: build-dist
           run_id: ${{ github.event.workflow_run.id }}
 
@@ -23,40 +24,37 @@ jobs:
           git config --global credential.helper store
           echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' >> ~/.git-credentials
 
-      - name: Commit dist tarball dist repo
+      - name: Commit to dist repo
         run: |
           set -ex
-
-          # we need a fully predictable name/URL, and git introduces these
-          # additional numbers into the version; so wrap it in a predictable named tar
-          sha=$(ls -d dist-* | sed 's/dist-//')
+          SHA=$(cat SHA)
 
           # multiple parallel workflows race against each other, so the final
           # `git push` may fail
           rc=1
           for retry in $(seq 5); do
               git clone https://github.com/${{ github.repository }}-dist.git dist-repo
-              tar -cvf "dist-repo/${sha}.tar" -C "dist-$sha" .
+              rm -rf dist-repo/dist
+              cp -r dist/ package-lock.json dist-repo/
 
               # freshly created empty repo?
               cd dist-repo
               git rev-parse HEAD >/dev/null 2>&1 || git init -b main
-              git add "${sha}.tar"
-              git commit -m "Build for $sha"
 
-              # remove tarballs older than a week
-              now=$(date +%s)
-              for f in *.tar; do
-                  fmtime=$(git log --pretty=%at -n1 -- $f)
-                  [ $(($now - $fmtime)) -lt 604800 ] || git rm $f
-              done
-              [ -z "$(git status --short)" ] || git commit -m 'Drop old builds'
+              # commit the updated files
+              git add .
+              # may have no changes
+              git diff --cached --quiet || git commit -m "Build for $SHA"
+              tag="sha-$SHA"
+              git tag $tag
 
-              if git push; then
+              # push commit and tag
+              if git push origin HEAD $tag; then
                   rc=0
                   break
               else
                   echo "ERROR: conflict? Retrying git commit"
+                  git push origin :$tag || true
                   cd ..
                   rm -rf dist-repo
               fi
--- a/test/download-dist
+++ b/test/download-dist
@@ -54,7 +54,7 @@ def download_dist(wait=False):
         message("download-dist: not a git repository")
         return False
 
-    if subprocess.call(["git", "diff", "--quiet", "--", ":^test", ":^packit.yaml", ":^packaging"]) > 0:
+    if subprocess.call(["git", "diff", "--quiet", "--", ":^test", ":^packit.yaml", ":^packaging", ":^.github"]) > 0:
         message("download-dist: uncommitted local changes, skipping download")
         return False
 
--- a/test/download-dist
+++ b/test/download-dist
@@ -16,14 +16,14 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
-import io
+import argparse
 import os
-import urllib.request
 import subprocess
 import sys
-import tarfile
 import time
-import argparse
+
+REPO = os.getenv("GITHUB_BASE", "cockpit-project/cockpit-machines") + "-dist"
+PROJECT = REPO.split('/')[1]
 
 
 def message(*args):
@@ -31,7 +31,7 @@ def message(*args):
 
 
 def download_dist(wait=False):
-    '''Download and unpack dist/ for current git SHA from GitHub
+    '''Download pre-built dist/ for current git SHA from GitHub
 
     These are produced by .github/workflows/build-dist.yml for every PR and push.
     This is a lot faster than having to npm install and run webpack.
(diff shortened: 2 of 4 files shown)


martinpitt:
Please let me land this myself after reviewing. Right before we push the button, I'd like to reinitialize the current -dist repository, as we don't want to keep the tarballs any more.

martinpitt:
BTW: Eventually I think we want to keep package-lock.json in the main repo, and have a separate workflow which refreshes it every week (together with trying to update package.json). Then the -dist repo will track only dist/. But that needs some more work in bots also.

martinpitt:
The publish-dist workflow for *this* PR [failed](https://github.com/cockpit-project/cockpit-machines/runs/2303130372?check_suite_focus=true), as it already took the build-dist from this PR, and thus there is no dist-* artifact. This is unavoidable for this PR, but also doesn't matter much.

allisonkarlitskaya (commented):
One big question before I dig in here: why are we using artifacts still?  Can't we just have the workflow directly commit to the other git repository for us?

martinpitt:
@allisonkarlitskaya : No, we can't have that, the reason is a bit complicated:

The "build-dist" workflow runs on `pull_request` event,  and processes untrusted code. `pull_request` does not have access to project secrets, which is good because malicious PRs can't exfiltrate secrets; but it also means that build-dist can't commit to any repo or do much else than creating an artifact. The "publish-dist" workflow always runs from the main branch, thus is trusted and does have access to secrets. That must take care to not "inspect" and run anything from the untrusted tarball.

Splitting the workflow that way and using an artifact to exchange data that way is the [recommended best practice](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

.github/workflows/publish-dist.yml
```diff
@@ -23,40 +24,37 @@ jobs:
...
+              # commit the updated files
+              git add .
+              # may have no changes
+              [ -z "$(git diff --cached)" ] || git commit -m "Build for $SHA"
+              tag="sha-$SHA"
+              git tag $tag
 
-              if git push; then
+              # push commit and tag
+              if git push && git push origin $tag; then
```
allisonkarlitskaya, line 52: probably you could use a single git push?

test/download-dist
```diff
@@ -58,66 +59,35 @@ def download_dist(wait=False):
...
-    download_url = f"https://github.com/{ os.getenv('GITHUB_BASE', 'cockpit-project/cockpit-machines') }-dist/raw/main/{sha}.tar"
-    request = urllib.request.Request(download_url)
-    tario = io.BytesIO()
+    dist_git_checkout = os.path.join(os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache")), "cockpit-dev", REPO)
+
+    if not os.path.exists(dist_git_checkout):
+        message(f"download-dist: Creating dist cache {dist_git_checkout}")
+        # we can't directly clone a tag with clone -b, so let's always checkout main
+        # this is usually very close to what we are interested in anyway
+        subprocess.check_call(["git", "clone", "--quiet", "--depth=1", "https://github.com/" + REPO, dist_git_checkout])
```
allisonkarlitskaya, line 68: I think we can do better here.  Two ideas:

  1) Let's rather maintain a raw clone here (ie: `cockpit-machines-dist.git`, no checkout dir) and extract from it directly into the dist/ dir of the working tree that we're inside of

  2) Do the checkout into this directory, but at least take advantage of the files being there, and do reflink-copies. It seems we might not get those by default, possibly related to the recent complaining about sendfile() and /proc.  See https://bugs.python.org/issue37157 for example.

I think I prefer option 1, though.  Having a bunch of checked out files lying around in ~/.cache/ is kinda lame.

allisonkarlitskaya (changes requested):

.github/workflows/publish-dist.yml
martinpitt: Fixed.

test/download-dist
martinpitt: Nice idea, thanks for the tip with `git archive`! Done.

martinpitt:
I [re-tested this on my fork](https://github.com/martinpitt/cockpit-machines/runs/2321944376?check_suite_focus=true) to make sure the combined `git push` works fine. Locally, this works as expected:

```
$ git clean -ffdx; GITHUB_BASE=martinpitt/cockpit-machines TEST_OS=fedora-34 make vm
[...]
From https://github.com/martinpitt/cockpit-machines-dist
 * tag               sha-f4919e6ad526d136336fd39ce5a7eab97d5c3d00 -> FETCH_HEAD
download-dist: Extracting dist cache...
[... (no local build happening) ...]
```

.github/workflows/publish-dist.yml
```diff
@@ -23,40 +24,37 @@ jobs:
...
-
-          # we need a fully predictable name/URL, and git introduces these
-          # additional numbers into the version; so wrap it in a predictable named tar
-          sha=$(ls -d dist-* | sed 's/dist-//')
+          SHA=$(cat SHA)
 
           # multiple parallel workflows race against each other, so the final
           # `git push` may fail
           rc=1
           for retry in $(seq 5); do
```
allisonkarlitskaya, line 35: I don't like this loop, and I think we should be doing this differently:

 - on pull requests, we should push only a tag, as a commit on top of the most recent commit on the main branch.
 - on merge to main, we then push an update also to the main branch here

random PRs ending up on the main branch seems weird, and is also exactly the source of the potential conflict here.

thoughts?

martinpitt: We quickly discussed that on IRC. Alternatives are to just push "bare" tags which are not attached to a branch, or push every SHA build as its own branch. The latter would also make it possible to do `git clone -b ` directly, but I suppose it's somewhat more expensive to store in git (precisely *because* `git clone -b` is possible, so it must publish more metadata). However, these would all lose the nice feature of delta compression against the previous version, so that would not really be an improvement.

test/download-dist
```diff
@@ -54,70 +54,40 @@ def download_dist(wait=False):
...
-
-    # Extract relevant files
-    unpack_paths = ["dist", "package-lock.json"]
-    message("download-dist: Extracting paths from tarball:", ' '.join(unpack_paths))
-    prefixed_unpack_paths = ["cockpit-machines/" + d for d in unpack_paths]
-    subprocess.check_call(["tar", "--touch", "--strip-components=1", "-xf", tar_path] + prefixed_unpack_paths)
-
-    os.remove(tar_path)
+    message("download-dist: Extracting dist cache...")
+    p_git = subprocess.Popen(["git", "-C", dist_git_checkout, "archive", "FETCH_HEAD", "dist", "package-lock.json"],
```
allisonkarlitskaya, line 85: ...or just use the shell?

allisonkarlitskaya (approved):
Great improvement!

test/download-dist
martinpitt: Well, maybe this is over-engineered indeed.. :grin:

.github/workflows/publish-dist.yml
martinpitt: I generally like your idea about pushing to main only for main builds. However, wouldn't we still need the same retry loop? If you land two PRs closely together, one of their publish-dist runs will still lose the race and needs to be retried. It would just be less likely.

PR #86, opened 2021-04-09
Fix hot unplugging a vNIC and add test for interface detaching in general

skobyda:


KKoukiou (commented):
I was really suprised to see this PR, because I was sure nic deletion works for live VMs.

In fact I just tried it now, and the nic was successfully removed after clicking the delete button.
Can you elaborate more on the failure you re trying to solve?

skobyda:
> I was really suprised to see this PR, because I was sure nic deletion works for live VMs.
> 
> In fact I just tried it now, and the nic was successfully removed after clicking the delete button.
> Can you elaborate more on the failure you re trying to solve?

Hmm, strange. Maybe it's just a bug related to libvirt version, or the way I have vNIC which I try to delete configured.
I also try to delete it from a **session** VM, with a following interface configuration, can you check if it fails:

```
    <interface type='user'>
      <mac address='52:54:00:44:15:31'/>
      <model type='virtio'/>
      <alias name='net0'/>
      <address type='pci' domain='0x0000' bus='0x01' slot='0x00' function='0x0'/>
    </interface>
```

If that's not the case, perhaps it's just a bug for my libvirt version, in that case I would create a bugzilla and have this PR as a temporary workaround.

skobyda:
Oh, now I see that there are tests for detaching ifaces, they are just part of testNICAdd and NICAddDialog, so I did not see them. dropping the other commit

skobyda:
In that case I thought the hot unplugging was not covered in tests. But since even hot unplugging is indirectly covered in tests, then my failure to detach vNICs seem like a special case

KKoukiou:
@skobyda tried on session VM with you NIC configuration and it works as well. So if it works also in the test VM I guess it's something on your system indeed.

skobyda:
Alright. So I don't think this PR is relevant anymore, so closing it.

skobyda:
> @skobyda tried on session VM with you NIC configuration and it works as well. So if it works also in the test VM I guess it's something on your system indeed.

Since the tests were there before (just did not see them), it was passing even before. So it's definitely something on my system

PR #91, opened 2021-04-12
Remove css imports which should be autoimported

KKoukiou:
The stylesheets for AccessConsole and DesktopViewer are autoimported
when importing the React components.

This also fixes the missing borders on the console type selector.


martinpitt (approved):
That "fix" feels a bit magic, but indeed I was wondering why we require these manual imports in the first place -- I thought these were only for "manual" HTML that use PF. Thanks!

KKoukiou:
> That "fix" feels a bit magic, but indeed I was wondering why we require these manual imports in the first place -- I thought these were only for "manual" HTML that use PF. Thanks!

it's magic for me as well, I guess some css import order.

PR #92, opened 2021-04-12
workflows: Add timeout to publish-dist

martinpitt:
This ought to be very fast. If it's not, something is seriously wrong
and we want to know about it immediately, rather than stalling a PR for
6 hours.


allisonkarlitskaya (approved):
klar

PR #94, opened 2021-04-12
test: discover also python files which don't contain hash-bang when running pylint

KKoukiou:
And fix all the errors that appear.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ WEBLATE_REPO=tmp/weblate-repo
 WEBLATE_REPO_URL=https://github.com/cockpit-project/cockpit-machines-weblate.git
 WEBLATE_REPO_BRANCH=main
 
-PYEXEFILES=$(shell git grep -lI '^#!.*python')
+PYTHONFILES=$(shell git grep -lI '^#!.*python') $(shell git ls-tree -r --name-only HEAD | grep '\.py')
 
 all: $(WEBPACK_TEST)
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -20,7 +20,7 @@ WEBLATE_REPO=tmp/weblate-repo
 WEBLATE_REPO_URL=https://github.com/cockpit-project/cockpit-machines-weblate.git
 WEBLATE_REPO_BRANCH=main
 
-PYEXEFILES=$(shell git grep -lI '^#!.*python')
+PYEXEFILES=$(shell git grep -lI '^#!.*python') $(shell find test/ -name *.py ! -path 'test/common/*')
```
martinpitt, line 23: Seems a bit weird to call these `PYEXEFILES`. Either call these `PYTHONFILES` and use `git ls-tree -r --name-only HEAD | grep '\.py'` (to really only catch the tracked files, and avoid doing this `! -path` stuff) or call pyflakes/pycodestyle without arguments to let them find all the py files (not 100% sure that this works).

skobyda, line 23: I would also extend the find command to src/ directory. There is for example a file `src/lib/inotify.py` which is not covered by either command

skobyda (changes requested):

martinpitt (changes requested):
Thanks!

Makefile
martinpitt: Indeed, my `ls-tree` proposal from below finds these as well.

skobyda (approved):

martinpitt (approved):
:+1:

PR #97, opened 2021-04-12
When the disk bus type is not in the predefined list show it disabled in the edit dialog

KKoukiou:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1948366

Note: the linked bz uses an IDE disk for the reproducer. IDE
controllers are not supported when used with Q35 machine type, which is the
default machine type in RHEL >= 8. Therefore we don't show IDE in the
predefined list.

### DIFF
--- a/src/components/vm/disks/diskEdit.jsx
+++ b/src/components/vm/disks/diskEdit.jsx
@@ -79,7 +79,9 @@ const CacheRow = ({ onValueChanged, dialogValues, idPrefix, shutoff }) => {
 };
 
 const BusRow = ({ onValueChanged, dialogValues, idPrefix, shutoff }) => {
-    const busTypes = ['sata', 'scsi', 'usb', 'virtio'];
+    const busTypes = ['sata', 'scsi', 'usb', 'virtio'].map(type => ({ value: type }));
+    if (!busTypes.includes(dialogValues.busType))
+        busTypes.push({ value: dialogValues.busType, disabled: true });
 
     return (
         <FormGroup fieldId={`${idPrefix}-bus-type`} label={_("Bus")}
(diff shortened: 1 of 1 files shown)


src/components/vm/disks/diskEdit.jsx
```diff
@@ -79,7 +79,9 @@ const CacheRow = ({ onValueChanged, dialogValues, idPrefix, shutoff }) => {
 };
 
 const BusRow = ({ onValueChanged, dialogValues, idPrefix, shutoff }) => {
-    const busTypes = ['sata', 'scsi', 'usb', 'virtio'];
+    const busTypes = ['sata', 'scsi', 'usb', 'virtio'].map(type => ({ value: type }));
+    if (!busTypes.includes(dialogValues.busType))
+        busTypes.push({ value: dialogValues.busType, disabled: true });
```
martinpitt, line 84: How does that help exactly? This just (sometimes) adds a new bus type as disabled, but never disables any of the ones in the static list. How was "ide" even shown before? I don't see any other code which would extend `busTypes`?

KKoukiou: The 'ide' value never appeared in the dialog actually before, since as you correctly notice the list of bus types was static.

However if the user created somehow (outside of cockpit) a disk with IDE bus, and then tried to edit it, the FormSelect was supplied with value='ide', although none of the options was that. So the rendered component was something like this:

``` js
      <FormSelect value='ide' onChange={this.onChange}>
          <FormSelectOption value='sata' label='sata' />
          <FormSelectOption value='iscsi' label='iscsi' />
          <FormSelectOption value='usb' label='usb' />
          <FormSelectOption value='virtio' label='virtio' />
      </FormSelect>

```

Of course this is invalid, but pf4 won't throw an error. It will display the first value in the UI, misleading the user to think that their selection is sata, (see also bz description) but secretly have value 'ide'. Then if the user will immediately click save, and won't change the bus type value before that, the 'ide' value will just silently remain.

I hope it's clearer now.

martinpitt: Aah! That makes sense, thanks!

martinpitt (approved):

martinpitt:
This introduced a regression, reported as #111

KKoukiou:
Hmm it can't be this PR thought.

PR #100, opened 2021-04-13
workflows: Trigger integration tests from publish-dist

martinpitt:
With that, the tests can use the pre-built dist cache instead of each
test rebuilding it again. This lessens the strain on our CI and
npmjs.com.

Put the tests directly into our AMQP queue. This removes the
pull_request webhook, reacting to it, and calling tests-scan inside our
CI from the webhook queue. This decreases latency and the number of
moving parts.

-----

This is a bit of a trade-off. It's more efficient and reduces the number of moving parts in our infra a bit, but now we will not get the "pending" statuses immediately after pushing again, just after "build-dist" has finished. That may need some getting used to. Perhaps we try it out for a bit, and either we like it and transplant it to cockpit as well,or we just put it to the shelf and declare it a dead end. @marusak , @KKoukiou , opinions?

I tested this on my fork. I sent [this PR](https://github.com/martinpitt/cockpit-machines/pull/4), which triggered [this publish-dist workflow](https://github.com/martinpitt/cockpit-machines/runs/2330640890?check_suite_focus=true#step:5:126), which correctly triggered a debian-stable test in the PR after {build,publish}-dist finished.

For that I used a [temporary bots fork](https://github.com/martinpitt/cockpit-machines/commit/2f7c6f1a3c143) which adjusted the testmap for martinpitt/cockpit-machines.

Note: the actual test run failed to download the dist cache, as it was looking into cockpit-project/cockpit-machines instead of martinpitt/c-m. This is just an artifact of testing on my fork.

### DIFF
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -24,11 +24,17 @@ jobs:
           sudo apt install -y --no-install-recommends npm make gettext sassc
 
       - name: Build
+        run: NODE_ENV=production make
+
+      - name: Save event information
+        env:
+          GITHUB_EVENT: ${{ toJSON(github.event) }}
         run: |
-          NODE_ENV=production make
-           # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions
-           # for pull_requests, use HEAD of the proposed branch, for pushes to origin the current SHA
+          # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions
+          # for pull_requests, use HEAD of the proposed branch, for pushes to origin the current SHA
           echo "${{ github.event.pull_request.head.sha || github.sha }}" > SHA
+          # Save event for follow-up workflows
+          echo "$GITHUB_EVENT" > event.json
 
       - name: Create dist artifact
         uses: actions/upload-artifact@v2
@@ -38,4 +44,5 @@ jobs:
             SHA
             dist/
             package-lock.json
+            event.json
           retention-days: 1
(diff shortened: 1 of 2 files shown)


martinpitt:
Weird, why does this [keep failing](https://github.com/cockpit-project/cockpit-machines/pull/100/checks?check_run_id=2331780265) here, but not against my fork.. Dumping the JSON must have some strange characters for cockpit-project that isn't the case for my fork. I'll try to [explicitly stopping the stream](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#stopping-and-starting-workflow-commands)

martinpitt:
Ah, got it -- the `env:` indirection does the trick, then the spewage does not happen in the `run:` log any more. I pushed the same change to my fork, and re-tested it on https://github.com/martinpitt/cockpit-machines/pull/4 -- [corresponding action](https://github.com/martinpitt/cockpit-machines/runs/2332158422?check_suite_focus=true#step:5:126).

KKoukiou (approved):
So I have nothing against this change, we can just introduce it here and see how it behaves. I can't think of any problems, but I suggest to just go ahead and see it in practice.

martinpitt:
Ack. I'd like to keep the first commit (collecting/storing the github.event info) in any case. I disabled the `pull_request` webhook, let's see how it works in practice. Easy enough to revert if we don't like it.

martinpitt:
BTW, this is also a testing ground for running test nightly on the main branch -- this was actually the reason why I started looking at all this.

PR #115, opened 2021-04-20
package.json: Update @patternfly/react-console

cockpituous:


cockpituous:
Please manually check features related to these files:

```
package.json:    "@patternfly/react-console": "4.5.3",
src/components/vm/consoles/consoles.jsx:import { AccessConsoles } from "@patternfly/react-console";
src/components/vm/consoles/desktopConsole.jsx:import { DesktopViewer } from '@patternfly/react-console';
src/components/vm/consoles/vnc.jsx:import { VncConsole } from '@patternfly/react-console';
```

martinpitt:
This is 4.5.3 now, so the https://github.com/patternfly/patternfly-react/pull/5678#issuecomment-821275444 fix should be included. However, it is not:

image

(I also fixed that blatantly wrong German translation of "Send key" in weblate..)

However, this bug already exists on current main with react-console 4.3.37, so this is not a regression.

martinpitt (approved):
Approving as there is no regression, but maybe @KKoukiou still wants to have a look?

KKoukiou:
@martinpitt This actually has the expected behavior and the bug has been solved. The placeholder option 'Select console type' should have the following behavior:
* Clicking on it resets the choice
* The check icon on the right is missing for this option, since it's not a valid item.
This is the current behavior of the Select, you can also confirm in the Select component documentation: https://www.patternfly.org/v4/components/select/#single

KKoukiou:
The problem previously was the the placeholder did not reset the choice and also had a checkmark on the right when clicken on.
If you think that's an unexpected behavior, we can use a different component than the Select for the console selection, one without placeholder option.

martinpitt:
@KKoukiou : That still feels like a bug. I can *literally* select "Select console type", then the selector closes and I see no console at all. In https://www.patternfly.org/v4/components/select/#single I can *not* do that. That's just the initial default string, but I can never explicitly go back to it.

PR #119, opened 2021-04-21
Stop using branch for -dist cache

martinpitt:
Push every build as an individual tag only, without having them on a
branch.

For the cleanup, just delete all tags which are older than a week, so
that they become unreferenced and can be garbage collected; do that
right in publish-dist.yml, to have fewer moving parts and avoid
duplicating the secrets setup.

This fixes a lot of the previous problems:

 * We don't need any force-pushing and history rewriting any more, which
   fixes the race of publish-dist against prune-dist.

 * There is no more race condition on multiple publish-dist workflows
   pushing to main.

 * prune-dist.yml did not actually clean up anything on GitHub, as it
   did not remove the tags from origin.

Thanks to Allison Karlitskaya for this idea!

---

I tested this on my fork. The [publish-dist run](https://github.com/martinpitt/cockpit-machines/actions/runs/770574221) committed the dist, I can download it with `GITHUB_BASE=martinpitt/cockpit-machines test/download-dist`, it cleaned up the 20-something tags from 8 days ago, and kept the two recent ones. These commits are *not* visible in https://github.com/martinpitt/cockpit-machines-dist now. I didn't reset that repo as I wanted to have some old repo to play with.

But once this is approved, I'll reset the cockpit-machines-dist repo, so that it starts afresh without any branch at all.

### DIFF
--- a/.github/workflows/prune-dist.yml
+++ b/.github/workflows/prune-dist.yml
@@ -10,10 +10,12 @@ jobs:
   run:
     runs-on: ubuntu-latest
     steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
       - name: Set up configuration and secrets
         run: |
           printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
-          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
           # we push to -dist repo via https://github.com, that needs our cockpituous token
           git config --global credential.helper store
           echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' >> ~/.git-credentials
@@ -22,20 +24,22 @@ jobs:
         run: |
           git clone https://github.com/${{ github.repository }}-dist.git dist-repo
 
-      - name: Truncate history of -dist repo
+      - name: Delete old tags
         run: |
           set -ex
+
+          HEAD=$(git rev-parse HEAD)
+
           cd dist-repo
-          # keep one week of builds
-          # oldest commit that applies to any still present tarball
-          REF=$(git log --before '7 days ago' --pretty=format:%H -1)
-          # empty if there are no old commits
-          [ -n "$REF" ] || exit 0
-
-          git checkout --orphan temp $REF
-          git commit -m "Truncated history"
-          git rebase --onto temp $REF main
-          git branch -D temp
-          git reflog expire --expire=now --all
-
-          git push -f
+          now="$(date +%s)"
+          for tag in $(git tag -l); do
+              tag_time="$(git show -s --format=%at $tag)"
+              if [ "$tag" = "sha-${HEAD}" ]; then
+                  echo "$tag refers to current project HEAD, keeping"
+              elif [ $((now - tag_time)) -ge 604800 ]; then
+                  echo "$tag is older than 7 days, deleting..."
+                  git push origin ":$tag"
+              else
+                  echo "$tag is younger than 7 days, keeping"
+              fi
+          done
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -28,39 +28,19 @@ jobs:
       - name: Commit to dist repo
         run: |
           set -ex
-          SHA=$(cat SHA)
+          SHA="$(cat SHA)"
 
-          # multiple parallel workflows race against each other, so the final
-          # `git push` may fail
-          rc=1
-          for retry in $(seq 5); do
-              git clone https://github.com/${{ github.repository }}-dist.git dist-repo
-              rm -rf dist-repo/dist
-              cp -r dist/ package-lock.json dist-repo/
+          mkdir dist-repo
+          cd dist-repo
+          git init
+          cp -r ../dist/ ../package-lock.json .
 
-              # freshly created empty repo?
-              cd dist-repo
-              git rev-parse HEAD >/dev/null 2>&1 || git init -b main
+          git add .
+          git commit -m "Build for $SHA"
+          tag="sha-$SHA"
+          git tag "$tag"
 
-              # commit the updated files
-              git add .
-              # may have no changes
-              git diff --cached --quiet || git commit -m "Build for $SHA"
-              tag="sha-$SHA"
-              git tag $tag
-
-              # push commit and tag
-              if git push origin HEAD $tag; then
-                  rc=0
-                  break
-              else
-                  echo "ERROR: conflict? Retrying git commit"
-                  git push origin :$tag || true
-                  cd ..
-                  rm -rf dist-repo
-              fi
-          done
-          exit $rc
+          git push https://github.com/${{ github.repository }}-dist.git $tag
 
       - name: Trigger integration tests for PRs
         run: |
--- a/test/download-dist
+++ b/test/download-dist
@@ -62,9 +62,9 @@ def download_dist(wait=False):
 
     if not os.path.exists(dist_git_checkout):
         message(f"download-dist: Creating dist cache {dist_git_checkout}")
-        # we can't directly clone a tag with clone -b, so always fetch main initially
-        # this is usually very close to what we are interested in anyway
-        subprocess.check_call(["git", "clone", "--quiet", "--depth=1", "--bare", "https://github.com/" + REPO, dist_git_checkout])
+        os.makedirs(dist_git_checkout)
+        subprocess.check_call(["git", "-C", dist_git_checkout, "init", "--bare", "--quiet"])
+        subprocess.check_call(["git", "-C", dist_git_checkout, "remote", "add", "origin", "https://github.com/" + REPO])
 
     retries = 50 if wait else 1  # 25 minutes, once every 30s
     while retries > 0:


martinpitt:
This was @allisonkarlitskaya 's idea (stolen from https://github.com/cockpit-project/cockpit/pull/15700), and @mvollmer pointed out some bugs in the current workflow this morning -- so asking both of you to have a look. Thanks!

martinpitt:
Meh, [this is annoying](https://logs.cockpit-project.org/logs/pull-119-20210421-135514-b8b23496-rhel-9-0/log.html) -- Let's try and make our tests work offline, without depending on the internal RHEL mirror. It's been on and off for the whole day.

.github/workflows/publish-dist.yml
```diff
@@ -30,37 +30,22 @@ jobs:
...
           SHA=$(cat SHA)
 
-          # multiple parallel workflows race against each other, so the final
-          # `git push` may fail
-          rc=1
-          for retry in $(seq 5); do
-              git clone https://github.com/${{ github.repository }}-dist.git dist-repo
-              rm -rf dist-repo/dist
-              cp -r dist/ package-lock.json dist-repo/
+          git clone https://github.com/${{ github.repository }}-dist.git dist-repo
```
mvollmer, line 33: I think there should be a `git checkout --orphan` here somewhere, to make it super clear that we are producing commits without history.

.github/workflows/publish-dist.yml
```diff
@@ -87,3 +72,18 @@ jobs:
...
+
+      - name: Delete old tags
+        run: |
+          set -ex
+          cd dist-repo
+          now=$(date +%s)
+          for tag in $(git tag -l); do
+              tag_time=$(git show -s --format=%at $tag)
+              if [ $((now - tag_time)) -ge 604800 ]; then
+                  echo "$tag is older than 7 days, deleting..."
```
mvollmer, line 84: This means that the repo will be totally empty after a week when nobody pushes new tags in the mean time, no? What about "keep the N most recent commits" where N is such that we are happy with the size of the repo.

.github/workflows/publish-dist.yml
martinpitt: Sure, I can add a `git checkout --orphan floating-in-space` or so here.

allisonkarlitskaya: In the `node_modules` stuff I just create an empty repo, init, add the files, commit, tag, and push from there.  I think that would be easier here, too.

allisonkarlitskaya (changes requested):

.github/workflows/publish-dist.yml
martinpitt: Good point. I'll change it to at least retain the current cockpit-machines main HEAD, so that this will always stay available.

martinpitt:
@allisonkarlitskaya : There are two reasons that I had in mind:
 - I thought it'd be better to fetch all stuff to be able to re-use existing blobs/commits, so that it does not try to re-push stuff that is already on origin. Like a PR which only changes tests/, or this very PR for workflow changes, neither of which changes tests/. Maybe that's irrelevant and git figures it out already anyway?

 - The workflow later iterates over all tags to clean up. Would it be more efficient to create an empty repo, add origin, and just fetch the tags? (I'm not sure if it would be *easier*, but it *may* transmit less stuff)

.github/workflows/publish-dist.yml
martinpitt: FTR, it will never be *totally* empty -- after all, this publish-dist workflow just pushed a new tag. But that may be for a PR only, and I agree it's kind of nice to keep main HEAD around.

allisonkarlitskaya:
The "reuse existing stuff" doesn't really work like that.  Files are referred to by a hash of their content, so if the files are the same, we are going to be using the same objects anyway, regardless of if they were already there when we made the commit or not.

Put another way: a commit is a complete record of the state of a tree at a given point in time.  It is not a diff.  Having the old commit there or not makes absolutely no difference in how the commit turns out.  Even if you push the duplicate objects, they simply won't be stored on the server side, since they already exist there (unless they're sent as part of a pack, and then you'll get overlap, but it will be cleaned up during GC).  Ditto the delta compression stuff.


The point about cleaning up old tags is a more interesting one.  It would be an awful shame if the only way to do this was to download everything.... I wonder if we can rather do that with a separate pass once a week or so?  Or maybe via some REST API?

.github/workflows/publish-dist.yml
martinpitt: ... or use what @allisonkarlitskaya is proposing below. That also sounds quite interesting, just want to make sure that it works as intended.

martinpitt:
>  Having the old commit there or not makes absolutely no difference in how the commit turns out.

Right, I know that.

> Even if you push the duplicate objects, they simply won't be stored on the server side

That was my concern, as local git can't know which commits are on origin already. It seemed safer to me to fetch them upfront rather than pushing them to origin a second time. But I don't mind trying that.

> I wonder if we can rather do that with a separate pass once a week or so? 

We *could* -- that's what the prune-dist workflow did. I removed it because having to do all this secrets setup twice and yet another workflow to watch seemed too much overhead, when in publish-dist I already have the repo checked out. But indeed it transmits more data. So I'll do a version that keeps prune-dist.

Thanks for the suggestions!

.github/workflows/publish-dist.yml
martinpitt: Moved to Lis'es proposal, which makes this even more clear -- it always starts with an empty repo.

martinpitt:
I reworked this according to your suggestions, it's much simpler now. [publish-dist run](https://github.com/martinpitt/cockpit-machines/actions/runs/773647386) is really fast, "commit to dist repo" takes 3 to 4 seconds, and the whole workflow just 8s. The [prune-dist run](https://github.com/martinpitt/cockpit-machines/runs/2408633850?check_suite_focus=true#step:5:50) recognizes and keeps the current main HEAD now. prune-dist still fetches the whole repo, but that now only happens once a week. I think in the interest of keeping the code as simple as possible, that's probably ok?

.github/workflows/publish-dist.yml
mvollmer: *thumps-up*

mvollmer:
> It would be an awful shame if the only way to do this was to download everything....

A partial clone should be able to only download the tag definitions, but not the commits they point to.

mvollmer (approved):

.github/workflows/prune-dist.yml
```diff
@@ -22,20 +24,22 @@ jobs:
...
-          git checkout --orphan temp $REF
-          git commit -m "Truncated history"
-          git rebase --onto temp $REF main
-          git branch -D temp
-          git reflog expire --expire=now --all
+          HEAD=$(git rev-parse HEAD)
 
-          git push -f
+          cd dist-repo
+          now=$(date +%s)
```
allisonkarlitskaya, line 34: my usual quoting complaints :)
```suggestion
          now="$(date +%s)"
```

.github/workflows/prune-dist.yml
```diff
@@ -22,20 +24,22 @@ jobs:
...
+          for tag in $(git tag -l); do
+              tag_time=$(git show -s --format=%at $tag)
```
allisonkarlitskaya, line 36:
```suggestion
              tag_time="$(git show -s --format=%at $tag)"
```

.github/workflows/prune-dist.yml
```diff
@@ -22,20 +24,22 @@ jobs:
...
+              if [ "$tag" = "sha-${HEAD}" ]; then
+                  echo "$tag refers to current project HEAD, keeping"
+              elif [ $((now - tag_time)) -ge 604800 ]; then
+                  echo "$tag is older than 7 days, deleting..."
+                  git push origin :$tag
```
allisonkarlitskaya, line 41:
```suggestion
                  git push origin ":$tag"
```

.github/workflows/publish-dist.yml
```diff
@@ -30,37 +30,17 @@ jobs:
           set -ex
           SHA=$(cat SHA)
```
allisonkarlitskaya, line 31: can you quote that while you're here?

.github/workflows/publish-dist.yml
```diff
@@ -30,37 +30,17 @@ jobs:
...
+          git init
+          cp -r ../dist/ ../package-lock.json .
 
-              # freshly created empty repo?
-              cd dist-repo
-              git rev-parse HEAD >/dev/null 2>&1 || git init -b main
+          git add .
+          git commit -m "Build for $SHA"
+          tag="sha-$SHA"
+          git tag $tag
```
allisonkarlitskaya, line 41:
```suggestion
          git tag "$tag"
```

.github/workflows/publish-dist.yml
```diff
@@ -30,37 +30,17 @@ jobs:
...
-                  break
-              else
-                  echo "ERROR: conflict? Retrying git commit"
-                  git push origin :$tag || true
-                  cd ..
-                  rm -rf dist-repo
-              fi
-          done
-          exit $rc
+          git push https://github.com/${{ github.repository }}-dist.git $tag
```
allisonkarlitskaya, line 43:
```suggestion
          git push https://github.com/${{ github.repository }}-dist.git "$tag"
```

allisonkarlitskaya (dismissed):
So this looks good.

Next step (but definitely in a separate PR) is to come up with a concrete list of things that impact `dist/` and do a `git log` to find the last commit that changed one of those things, and generate (and download) based on that sha.

martinpitt:
I resolved the quoting issues and re-tested: [publish-dist succeeded](https://github.com/martinpitt/cockpit-machines/runs/2410141079?check_suite_focus=true), `GITHUB_BASE=martinpitt/cockpit-machines test/download-dist` succeeds, and [prune-dist succeeded](https://github.com/martinpitt/cockpit-machines/runs/2410185220?check_suite_focus=true) (and again  kept recent commits and current main HEAD)

martinpitt:
@skobyda: There are now [two](https://logs.cockpit-project.org/logs/pull-119-20210422-123825-7bec56e2-rhel-8-5/log.html) [failures](https://logs.cockpit-project.org/logs/pull-119-20210422-131021-7bec56e2-ubuntu-2004/log.html) of ` TestMachinesNICs.testNICEdit` which are brand new. Fallout  from PR #116?

allisonkarlitskaya (approved):

martinpitt:
I reset https://github.com/cockpit-project/cockpit-machines-dist to make sure this works as intended.

PR #121, opened 2021-04-21
nic edit form should be horizontal like all the others

KKoukiou:


martinpitt:
Meh, we have the [same internal RHEL repo error](https://logs.cockpit-project.org/logs/pull-121-20210421-163301-2476b473-rhel-9-0/log.html) here as in cockpit-machines. I fixed it there in https://github.com/cockpit-project/cockpit-machines/pull/120 , I'll see to applying the same fix to c-podman. Unfortunately we can't make it completely offline due to the container pulls, but at least we can take dnf out of the equation.

martinpitt:
Err, sorry, that was bogus -- I somehow misread that as "podman". As #120 landed, this can just be retried.

garrett:
Here's what it looks like on main for me:
Screenshot_2021-04-22 Virtual Machines - garrett Rain(1)

And here's the PR:
Screenshot_2021-04-22 Virtual Machines - garrett Rain

While the NIC is dependent on the type, it's probably fine to not have it indented like that. It comes after; that should be enough in this case. (And we didn't do indentation before either.)

---

Here's a quick modification of the screenshot to show what it would look like without the indentation:

Screenshot_2021-04-22 Virtual Machines - garrett Rain - mod

(In this case, this is probably what we should go with.)

garrett (changes requested):
Here's what it looks like on main for me:
Screenshot_2021-04-22 Virtual Machines - garrett Rain(1)

And here's the PR:
Screenshot_2021-04-22 Virtual Machines - garrett Rain

While the NIC is dependent on the type, it's probably fine to not have it indented like that. It comes after; that should be enough in this case. (And we didn't do indentation before either.)

---

Here's a quick modification of the screenshot to show what it would look like without the indentation:

Screenshot_2021-04-22 Virtual Machines - garrett Rain - mod

(In this case, this is probably what we should go with.)

KKoukiou:
> While the NIC is dependent on the type, it's probably fine to not have it indented like that. It comes after; that should be enough in this case. (And we didn't do indentation before either.)

We definitely had indentation before, it's implemented in the code, it's just not visible on vertical forms, and this somehow when moving this form from 'ct-form' to Form from PF4 slipped.

But the Nic Add modal has actually this option indented. So I can revert both, if you believe it's wrong.
Screen Shot 2021-04-23 at 09 35 20

garrett:
The indentation looks weird. I understand it's trying to communicate that it will change on the interface type, but it's already common in other places in Cockpit that things above will affect things below, even without indentation.

Usually, we don't indent, unless there are multiple items in a group that are directly affected by the ones above, and when they only appear for a certain selection. For example, selecting something that causes a user/password sub-form to appear.

As source is there for all the interface types and is by itself, I don't think it qualifies for indentation here.

If you think that deserves another PR, I'll approve this one. Otherwise, as it's a simple change and this PR just happens to switch it from not-indented to indented because the top-labelling didn't support indentation, it would be nice in this PR.

KKoukiou:
@garrett nope I am happy to repost, I just wanted to ensure we are aware that at some point (probably me) deliberately made this dialog like that. But I agree with you, let's remove the indentation.

garrett:
I found an old mockup from 2018 that shows it to the side. The indentation is probably a carry-over from that:

vm-nic-edit-old

FWIW, git blame points the finger at me for this mockup. So I'm probably the original one to blame. :rofl:

KKoukiou:
> The indentation looks weird. I understand it's trying to communicate that it will change on the interface type, but it's already common in other places in Cockpit that things above will affect things below, even without indentation.
> 
> Usually, we don't indent, unless there are multiple items in a group that are directly affected by the ones above, and when they only appear for a certain selection. For example, selecting something that causes a user/password sub-form to appear.
> 
> As source is there for all the interface types and is by itself, I don't think it qualifies for indentation here.
> 

It's almost for all interface types :) In session VMs, there is one interfaces type that does not have source. But I would say let's just unindent it anyway.

garrett (approved):
Works for me! Thanks; it looks good. :+1:

PR #122, opened 2021-04-21
fmf: Fix tests on RHEL/CentOS

martinpitt:
Fix typo, `$TestMachines` is not a variable (and not supposed to be
one). Just drop $TESTS on non-Fedora entirely, we want to run all tests
anyway.


marusak (approved):
Thanks

PR #126, opened 2021-04-23
workflows: Run npm-update twice a week only

martinpitt:
This is what we used to do when we still used npm-trigger. Some modules
like react-console are updated very often, and validating these
becomes too expensive/annoying.

Ported from https://github.com/cockpit-project/cockpit/commit/419dc0f19255d


KKoukiou (approved):
Oh yes thank you.

PR #130, opened 2021-04-23
test: Adjust special cases for Ubuntu 21.04

martinpitt:
ubuntu-stable gets updated to 21.04 in
https://github.com/cockpit-project/bots/pull/1952

---

**Blocked**: 
 - [x] Builds on top of #134 to get tests to green
 - [ ] Needs to land in lockstep with the image refresh

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -143,7 +143,7 @@ class TestMachinesCreate(VirtualMachinesCase):
 
         # try to CREATE few machines
         # --cloud-init user-data option exists since virt-install >= 3.0.0
-        if m.image in ['centos-8-stream', 'rhel-8-4', 'rhel-8-5', 'ubuntu-stable', 'ubuntu-2004', 'debian-stable']:
+        if m.image in ['centos-8-stream', 'rhel-8-4', 'rhel-8-5', 'ubuntu-2004', 'debian-stable']:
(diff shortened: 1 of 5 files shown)


martinpitt:
The remaining [failures](https://logs.cockpit-project.org/logs/pull-130-20210423-095742-2725af92-ubuntu-stable-bots-1952/log.html) are flakes -- one qemu, one libvirtd crash, which don't always happen (local single run is fine). TestNICAdd succeeds locally as well (3 times in a row).

martinpitt:
@KKoukiou : Hmm, we are getting qemu and libvirt crashes all over the place. Did you already look at some of them, or shall we have a closer look next week? I suppose most are already reported somewhere, but maybe over the migration to c-machines.git we lost some expected messages filters or so?

martinpitt:
I pulled in the fixes from #134 to hopefully get the tests greener. Let's see how that works.

martinpitt:
The QEMU crashes are https://github.com/cockpit-project/bots/issues/1565, I added the naughty to https://github.com/cockpit-project/bots/pull/1952 . The naughty for rhel-9-0 was wrong, I fixed it there as well. That will sort out the unexpected messages on rhel-9.

KKoukiou (approved):

PR #131, opened 2021-04-23
Fix testNICEdit

KKoukiou:


KKoukiou:
In the first commit I fixed also the nic source, it had some custom CSS resulting in weird alignment issues:

Now:
Screen Shot 2021-04-24 at 12 23 44

Previously:
Screen Shot 2021-04-24 at 13 36 19


The font-size became smaller, but that's the default for inline pf links, let's not override defaults here. What was fixed is that left padding which should not be there.

martinpitt:
[testNICAdd failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-131-20210423-131504-884bd567-debian-testing/log.html#47), not sure if that's related. Please retry if not (but this is suspicious -- normally the tests are reasonably happy, except testNICEdit and the occasional qemu crash)

martinpitt (approved):
Code-wise this looks great, and that test now succeeded everywhere. Thank you!

PR #15, opened 2021-03-24
Live migration

skobyda:
Transport live VM to another host. Configurable options are:
- Destination URI
- Storage: User can either choose to copy all read-only disks, or they
  can choose that they have a shared storage setup where both source and
  destination hosts have access too. In the latter that case, no disks
  are transported.
- Migrated VM's state: This defines if VM at the destination is running
  or not
- Temporary move: This defines if the migrated VM at the destination
  will be only transient or persistent

Release notes
===========

Cockpit Machines: Migration of a virtual machine
-----------------------------------------------------------------

VM migration is now supported and exposing various configuration options. The user has the ability to choose between live and offline migration, can decide if they want to copy the disks or use a shared storage and lastly there is the option to make the migration temporary.

Screen Shot 2021-05-12 at 17 39 29
Screen Shot 2021-05-12 at 17 42 51

PR #132, opened 2021-04-23
tests: set q35 machine type for our test VM

KKoukiou:
This is the default machine type for cirros VMs in all distros we test
against. If we could create the VM with virt-install and supply the os
variant that machine type would have been chosen for us, but since we
define from the XML we need to pass it explicitely.

If not set, we end up with some archaic defaults when attaching nics and
disks, and seeing the future of our tests, these will make operations like migration fail.

For hotplugging multiple PCI devices (which we do in the tests) we need
to add explicitely extra pci controllers.
See https://libvirt.org/pci-hotplug.html

Split out from : https://github.com/cockpit-project/cockpit-machines/pull/15

### DIFF
--- a/test/machinesxmls.py
+++ b/test/machinesxmls.py
@@ -116,7 +116,7 @@ DOMAIN_XML = """
   <name>{name}</name>
   <vcpu>1</vcpu>
   <os>
-    <type arch='x86_64'>hvm</type>
+    <type arch='x86_64' machine='q35'>hvm</type>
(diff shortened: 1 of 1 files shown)


KKoukiou:
There is an issue with: 
* live detaching disks - does not work really after this change
* and one with force reboot - does not seem to get rebooted (plain reboot works)

KKoukiou:
Ok no needed anymore, found another solution in the migration PR.

PR #135, opened 2021-04-26
tests: autogenerate test domain XML with virt-install

KKoukiou:
While working on https://github.com/cockpit-project/cockpit-machines/pull/132 to replace the machine type, I had trouble with setting correct disk parameters for the q35 machine type. Maybe let virt-install give us the XML, and we can control the machine type only by setting the target operating system, which is a more high level option.

DON'T MERGE - I want to see how many tests will fail.


KKoukiou:
Ah cool, this is essentially 1-2 tests failing, i can take a look but first, @martinpitt do you have an opinion? Shall we go with this appoach and delete our XMLs?

martinpitt:
@KKoukiou : Oooh, yes please -- I really like treating the XML as more or less an internal implementation detail, and use virt-install/-xml as our API. I didn't really like that hackery with adding 8 empty PCI slots and such..

martinpitt:
Re-running after #171 to get a clearer view.

martinpitt:
That looks much better now -- so on Fedora 33/firefox/34 it's consistently failing testNICAdd, on [debian-stable](https://logs.cockpit-project.org/logs/pull-135-20210519-105952-198b1cbc-debian-stable/log.html) this approach does not seem to work at all:

    ERROR    Unknown options ['target.type', 'source.path']

and it works everywhere else.

For debian-stable we can see if we can change the invocation to somehow get along -- and if that isn't possible, we could also just drop testing on it. I won't backport it to current Debian stable 10 any more anyway, I'll wait for Debian 11 and backport there.

KKoukiou:
@martinpitt the packit tests on fedora-34 fail all on:
```
bash: line 1: virt-install: command not found
```

This should be installed by c-machines deps, any idea why it's not? Whom can I contact for the packit environment?

martinpitt:
@KKoukiou : We only have it as a `Recommends:`, and presuably tmt does not install recommends by default. 
This has nothing to do with packit as such, it'd affect RHEL/Fedora gating as well.

Just add virt-install to test/browser/main.fmf.

martinpitt (approved):
Yay, nice work! Tests are still running, but this looks good.

PR #137, opened 2021-04-27
test: Add support for pixel testing

mvollmer:


mvollmer:
The new reference images are here: https://github.com/cockpit-project/pixel-test-reference/compare/empty..bf67acf951ff06682d2590d46e9bcd6e99358f60.

(A comment like this would normally be added automatically, but that doesn't work here since this PR also introduices the workflow that is will be responsible for this.)

KKoukiou (commented):
Amazing! So the screenshots still have these memory/storage size dependent strings but for CI should be fine.

There seems to be one flaky new pixel test: https://logs.cockpit-project.org/logs/pull-137-20210427-134053-bb86b6ab-fedora-33/log.html#41

But literally I see no difference in the images. Do you see something? https://logs.cockpit-project.org/logs/pull-137-20210427-134053-bb86b6ab-fedora-33/pixeldiff.html#TestMachinesCreate-testCreateBasicValidation-iso

KKoukiou (approved):

PR #139, opened 2021-04-28
Change the working for the disk image field in the import VM dialog

### LINKED ISSUE #107: Incorrect wording on Import VM page
It says **Installation source** while this should probably be **Existing image** or something like that. Importing is the process when OS usually does not get installed.

image

KKoukiou:
* [x] blocked on #141

When importing an existing disk image there is not any installation
going on. The field label was incorrect and confusing.

Fixes #107


marusak (approved):
Danke :)

PR #140, opened 2021-04-29
package.json: Update redux

cockpituous:
* [x] blocked on https://github.com/cockpit-project/cockpit-machines/pull/141

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "prop-types": "15.7.2",
     "react": "16.14.0",
     "react-dom": "16.14.0",
-    "redux": "4.0.5",
+    "redux": "4.1.0",
     "redux-thunk": "2.3.0",
     "throttle-debounce": "2.3.0"
   }


cockpituous:
Please manually check features related to these files:

```
package.json:    "redux": "4.1.0",
package.json:    "redux-thunk": "2.3.0",
src/reducers.js:import { combineReducers } from 'redux/dist/redux';
src/selectors.js: * Reference: https://redux.js.org/recipes/computing-derived-data/
src/store.js:import { createStore, applyMiddleware, compose } from 'redux';
src/store.js:import thunkMiddleware from 'redux-thunk';
```

martinpitt:
@mvollmer , @KKoukiou : This failed on the pixeldiffs due to the variable free space.
 * We should not retry a failed pixeldiff test, similar to how we don't retry a failure due to an unexpected message
 * Does the "ignore transparent parts in reference images" already work? I.e. could we already manually gimp this in the ref images?

KKoukiou (approved):

PR #142, opened 2021-04-30
Add support for configuring/viewing shared directories between VMs and host

### LINKED ISSUE #103: Request: Shared folder host / virtual machine (libvirtd)
Page: 
Virtual Machine / Libvirtd

I wish it was an feature in cockpit to configure a shared folder between the host and the virtual machine. Currently it is possible to use NFS-share, and I have explored (googled) and figured out it would be possible with some custom configuration.

KKoukiou:
Add support for configuring/viewing shared directories between VMs and host
    
    The shared directories can be mounted using VirtIO-FS [1].
    
    Status of VirtIO-FS
    -------------------
    
    Available in mainline since Linux 5.4, QEMU 5.0, libvirt 6.2.
    
    VirtIO-FS vs VirtIO-9p
    ----------------------
    
    VirtIO-FS is generally recommended over VirtIO-9p, as a faster and with better
    POSIX compliance shared filesystem solution for VMs.
    
    However:
    
    * osinfo-db still does not have support for it. This means that if the
    guest operating system does not support it we can't warn the users
    
    * virt-install does not have support for it (not interesting for us,
    since we allow configuring shared directories after installing the
    guest)
    
    * and neither libvirt exposes an option in domcapabilities for us to detect if the
    virtualization host supports VirtIO-FS
    
    On the othger hand virtIO-9p is not supported at all in RHEL/centos-8
    since the relevant module is not included in the default kernel.
    
    The tests
    ---------
    
    DO:
    * test if adding a filesystem makes it appear correctly in the UI
    
    DON'T:
    * test the actual sharing, meaning that the disk shared from the host
    can be actually written/read from the guest and vise-versa. This is
    because we *don't* have the test tooling at this point to run commands
    inside test VMs.
    
    Notes
    -----
    
    Setting explicitely the xattr is not needed since 5.12 kernel:
    https://lore.kernel.org/selinux/20210113123802.63563-1-omosnace@redhat.com/T/
    
    This is quite unfortunate because it's too much of a technicality to
    document in the UI, but will probably leave some users wondering why
    SELinux is preventing the mount operation they are attempting.
    
    Fixes #103
    
    [1] https://virtio-fs.gitlab.io/

filesystems card:

Screen Shot 2021-06-02 at 16 23 55

dialog if memory backing is already setup or for every new filesystem, apart from the first one.
Screen Shot 2021-06-02 at 16 23 25

dialog if memory backing is not set up:
Screen Shot 2021-06-02 at 16 20 43

### DIFF
--- a/src/components/vm/filesystems/vmFilesystemsCard.jsx
+++ b/src/components/vm/filesystems/vmFilesystemsCard.jsx
@@ -1,0 +1,259 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2021 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+import React, { useState } from 'react';
+import cockpit from 'cockpit';
+import {
+    Button,
+    Checkbox,
+    CodeBlock, CodeBlockCode,
+    ExpandableSection,
+    Form, FormGroup,
+    Grid,
+    List, ListItem,
+    Modal,
+    Popover,
+    Radio, TextInput, Tooltip
+} from "@patternfly/react-core";
+import { HelpIcon } from "@patternfly/react-icons";
+
+import { ListingTable } from "cockpit-components-table.jsx";
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { FileAutoComplete } from "cockpit-components-file-autocomplete.jsx";
+
+import { createFilesystem, setMemoryBacking } from "../../../libvirt-dbus.js";
+import { vmId } from "../../../helpers.js";
+
+import "./vmFilesystemsCard.scss";
+
+const _ = cockpit.gettext;
+
+export const VmFilesystemsCard = ({ vmName, filesystems }) => {
+    const columnTitles = [_("Source path"), _("Mount tag")];
+
+    const rows = filesystems.map(filesystem => {
+        const sourceKey = Object.keys(filesystem.source).find(key => filesystem.source[key]);
+        const filesystemSource = sourceKey ? filesystem.source[sourceKey] : undefined;
+        const filesystemTarget = filesystem.target.dir;
+        const rowId = `${vmId(vmName)}-filesystem-${filesystemSource}-${filesystemTarget}`;
+
+        const columns = [
+            { title: filesystemSource },
+            { title: filesystemTarget },
+        ];
+
+        return {
+            columns,
+            props: { key: rowId, 'data-row-id': rowId }
+        };
+    });
+
+    return (
+        <ListingTable variant='compact'
+                      gridBreakPoint='grid-xl'
+                      emptyCaption={_("No directories shared between the host and this VM")}
+                      columns={columnTitles}
+                      rows={rows} />
+    );
+};
+
+export const VmFilesystemActions = ({ connectionName, dispatch, memory, memoryBacking, objPath, vmName, vmState }) => {
+    const [isOpen, setIsOpen] = useState(false);
+    const idPrefix = `${vmId(vmName)}-filesystems`;
+    const addButton = (
+        <Button id={`${idPrefix}-add`}
+                isAriaDisabled={vmState != 'shut off'}
+                onClick={() => setIsOpen(true)}
+                variant="secondary">
+            {_("Add shared directory")}
+        </Button>
+    );
+
+    return (
+        <>
+            {vmState == 'shut off' ? addButton : <Tooltip content={_("Adding shared directories is possible only when the guest is shut off")}>{addButton}</Tooltip>}
+            {isOpen &&
+            <VmFilesystemAddModal connectionName={connectionName}
+                                  dispatch={dispatch}
+                                  memory={memory}
+                                  memoryBacking={memoryBacking}
+                                  objPath={objPath}
+                                  vmName={vmName}
+                                  vmState={vmState}
+                                  setIsOpen={setIsOpen} />}
+        </>
+    );
+};
+
+const VmFilesystemAddModal = ({ connectionName, dispatch, memory, memoryBacking, objPath, setIsOpen, vmName, vmState }) => {
+    const [additionalOptionsExpanded, setAdditionalOptionsExpanded] = useState(false);
+    const [dialogError, setDialogError] = useState();
+    const [memoryBackingType, setMemoryBackingType] = useState("file");
+    const [mountTag, setMountTag] = useState("");
+    const [source, setSource] = useState("");
+    const [validationFailed, setValidationFailed] = useState({});
+    const [xattr, setXattr] = useState(false);
+    const idPrefix = `${vmId(vmName)}-filesystems`;
+
+    const onAddClicked = () => {
+        const validationFailed = {};
+
+        if (!mountTag)
+            validationFailed.mountTag = _("Mount tag must not be empty");
+        if (!source)
+            validationFailed.source = _("Source must not be empty");
+
+        setValidationFailed(validationFailed);
+
+        if (Object.getOwnPropertyNames(validationFailed).length == 0) {
+            setMemoryBacking({
+                connectionName, objPath,
+                type: memoryBackingType,
+                memory
+            })
+                    .then(() => {
+                        createFilesystem({
+                            connectionName, objPath,
+                            source, target: mountTag,
+                            xattr,
+                        });
+                    })
+                    .then(
+                        () => setIsOpen(false),
+                        exc => setDialogError(exc.message)
+                    );
+        }
+    };
+    return (
+        <Modal position="top" isOpen variant="medium" onClose={() => setIsOpen(false)}
+               title={_("Share a shared host directory with the guest")}
+               footer={
+                   <>
+                       {dialogError && <ModalError dialogError={_("Failed to add shared directory")} dialogErrorDetail={dialogError} />}
+                       <Button id={`${idPrefix}-modal-add`}
+                               variant='primary'
+                               onClick={onAddClicked}>
+                           {_("Share")}
+                       </Button>
+                       <Button id={`${idPrefix}-modal-cancel`}
+                               variant='link'
+                               onClick={() => setIsOpen(false)}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }
+               description={
+                   <>
+                       {_("Shared host directories need to be manually mounted inside the VM")}
+                       <Popover
+                           headerContent={_("You can mount the shared folder using:")}
+                           bodyContent={
+                               <CodeBlock>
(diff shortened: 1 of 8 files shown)


KKoukiou:
@martinpitt if you have time this week to take a quick look to if we even want this feature that would be cool :)

martinpitt:
@KKoukiou sorry, I missed that before. For sure c-machines should at least *show* existing configured shared directories, as that is very interesting information. I find it also nice to create them. AFAICS this does not affect the "create VM" dialog, but it's something you add afterwards? 

My experiences with 9p were a bit disappointing, although I got these many years ago. It had utterly bad performance; reading/writing small files was fine, but for larger files it was at least two magnitudes slower than e.g. NFS. I recently saw some warning like "increase the BLABLA parameter for better performance", does that happen here?  I. e. can you get decent I/O out of 9p now?

Thoughts about the dialog:
 * How about replacing the "[mount tag]" in the info alert with the actually given value?
 * "[source path]" in the info alert is wrong -- that would be the *destination* directory inside the guest. That is unrelated to the source path on the host.

Thanks!

KKoukiou:
@martinpitt I played sortly with dd checking the disk write speed:

First I copied an image:
* from the mounted directory (hostshare) to another directory
* and then did an copy of the same image to another directory which is not shared with the host  

```
// hostshare is the 9p mountpoint:
[root@fedora ~]# dd if=./hostshare/Fedora-Cloud-Base-34-20210305.n.0.x86_64.qcow2 of=./test.img bs=1G count=1 oflag=dsync
0+1 records in
0+1 records out
259961856 bytes (260 MB, 248 MiB) copied, 8.37159 s, 31.1 MB/s

[root@fedora ~]# dd if=./test.img of=./test-bac.img bs=1G count=1 oflag=dsync
0+1 records in
0+1 records out
259961856 bytes (260 MB, 248 MiB) copied, 5.90636 s, 44.0 MB/s
[root@fedora ~]# dd if=./hostshare/Fedora-Cloud-Base-34-20210305.n.0
```

I ran the example a few times, generally copying from the 9p mountpoint, seems ~2times slower than copying to other local directories.

The perfomance can be increased by using 'cache='none'' attribute on the VM's disk. This is however not a solution we can suggest to the users.

VirtioFS has better performance, but it's a pretty new feature for libvirt, so I would rather wait till all our test images distros expose it through libvirt in the same way, and then I can support it.

martinpitt:
@KKoukiou : a 2x slowdown compared to local directories is *really* good performance. Back then I was measuring things in kB/s.. (and I think lots of small files were also much worse than a single big one). So, good to hear that the defaults are much better now!

garrett:
1. The information isn't an alert and should not be styled as one. (We've had similar issues with other dialogs lately, but those are for danger-flavored text about possibly doing something dangerous, and this is for something that's necessary to know, but still done afterward.)
2. Information needs to be available after the mount is created as well, as it will need to be referred to afterward.
3. This kind of information is usually suited for a popover, and the command line should be in a copy block.
4. The same popover with the same exact information could be located in the dialog and also after the mount point exists.

Is there a way to tell if it's in active use in the host or not? That is, if it's mapped in and unmounted versus being mounted? If so, we could show that it's not mounted and then message how to mount it based on that. (It could still be information a popover in that case.)

I suggested text like "Shared host directories need to be manually mounted inside the VM" with a little "more info" popover with the detailed information. (And this same popover would be used from the table, like status of "mounted" vs. "not mounted" and the "not mounted" state would have an info icon with the popover. That is, if we can tell if it's mounted inside the VM or not. Otherwise, it's more difficult to figure out how it is integrated, unless each every row has a redundant "mount details" row or something similar.)

Additional note from IRC: This is _similar_ to Docker's and Podman's mounting a directory as a volume vs. mounting a file, but it's implemented differently — and, unlike Docker and Podman, additional actions need to be taken inside the VM. So, while the two concepts can exist together for containers (as they're basically the same thing over there), they probably need to be separate for VMs like @KKoukiou has them in this PR.

KKoukiou:
@garrett addopted to your change requests. The only divergence I did is that I moved the information text to the modal description area - on the top of the modal. I feel that once users are ready with filling the form the just want to git Enter. Any information which is not a  result of filling the form should probably be on the top.

Screen Shot 2021-05-28 at 13 46 45
Screen Shot 2021-05-28 at 13 46 24

garrett (dismissed):
@KKoukiou: Thanks for fixing up the text by moving it into the popovers. I approve of the design. :+1:

KKoukiou:
There are still some test failures. Updating now.

KKoukiou:
@martinpitt ready for re-review - now with virtiofs - I am really sorry for the confusion with the usage of 9p at start.

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -0,0 +1,254 @@
...
+                variant="secondary">
+            {_("Add shared directory")}
+        </Button>
+    );
+
+    return (
+        <>
+            {vmState == 'shut off' ? addButton : <Tooltip content={_("Adding shared directories is possible only when the guest is shut off")}>{addButton}</Tooltip>}
+            <Modal position="top" isOpen={isOpen} variant="medium" onClose={() => setIsOpen(false)}
+                   title={_("Add a shared host directory with the guest")}
```
KKoukiou, line 138: Maybe 'Share a host directory with the guest', sounds better.

garrett: I agree. "Share a host directory with the guest" does sound better.

Does that mean we should update the action buttons too?

"Share host directory" / "Share directory"

KKoukiou: Sounds indeed better than 'add'. Will update! Thanks

KKoukiou:
> Pixel test references have changed in [9a00bf8](https://github.com/cockpit-project/cockpit-machines/commit/9a00bf808885f68c61c345a409d04dc5e3aec395). You can review them [here](/cockpit-project/pixel-test-reference/compare/1d56dafdc1be755f54ec0bb4e926c27defaaecfb..bf67acf951ff06682d2590d46e9bcd6e99358f60).

Doh? I did not update the test reference, that's weird.

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -0,0 +1,254 @@
...
+                            connectionName, objPath,
+                            source, target: mountTag,
+                            xattr,
+                        });
+                    })
+                    .then(
+                        () => {
+                            setAdditionalOptionsExpanded(false);
+                            setDialogError();
+                            setIsOpen(false);
```
martinpitt, line 114: This seems to reset the dialog back to the defaults. But why does that happen after creating the fs? Shouldn't that be done when clicking the "Add shared directory" button rather? Also, why is this reset necessary in the first place? Wouldn't it be cleaner to completely destroy the component after closing the dialog, and reinstantiate a new one when clicking the button? Then the defaults don't need to be repeated (both here and in `useState()` calls).

martinpitt (changes requested):
This is a really nice feature! I have two code nitpicks below.

I tested this first with a system instance (install Debian from scratch), and sharing worked well (a little too well, I need to investigate something). I forgot to test what happens if you try to add a second share with the same host dir or mount tag as an already existing one -- can you please write a test for that, and make sure that this shows a proper error? Likewise, that you get a proper error when the shared file path does not exist on the host?

I then tested this with a session instance. The "Shared directories" card is available, and the dialog apparently allows me to add a share, but then it goes away without an error message, and the card stays at "No directories shared" empty state. The only indication that something went wrong is the journal:

     libvirtd[1653]: unsupported configuration: virtiofs is not yet supported in session mode

I suggest to disable shared directories for qemu:///session instances for now. It feels really difficult to make this work in a safe manner with virtiofs, as that requires kernel support, and it does not currently behave in an acceptable manner on the host side wrt. rewriting user/group IDs and permissions.

Thanks!

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -0,0 +1,254 @@
...
+
+        const columns = [
+            { title: filesystemSource },
+            { title: filesystemTarget },
+        ];
+        return {
+            columns,
+            props: {
+                key: cockpit.format("$0-filesystem-$1-$2", vmId(vmName), filesystemSource, filesystemTarget),
+                'data-row-id': cockpit.format("$0-filesystem-$1-$2", vmId(vmName), filesystemSource, filesystemTarget),
```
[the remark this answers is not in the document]
KKoukiou: Done

src/components/vm/filesystems/vmFilesystemsCard.jsx
KKoukiou: Moved the modal to a separate component and so I just recreate the component when closed.

KKoukiou:
> Pixel test references have changed in [3e44073](https://github.com/cockpit-project/cockpit-machines/commit/3e44073f6c8f621330994a24acdbe2dda6df5647). You can review them [here](/cockpit-project/pixel-test-reference/compare/a21198bccc708677968b068fdae00086f9ed06b0..1d56dafdc1be755f54ec0bb4e926c27defaaecfb).

This is wrong - the branch is just not rebased. We need to rebase first and then check for changed pixel reference.

martinpitt:
[interesting pixeldiff failure](https://logs.cockpit-project.org/logs/pull-142-20210615-153450-bfff751f-fedora-34/log):
```
qemu-img create -q -f qcow2 -o backing_file=/work/bots/make-checkout-workdir/test/images/fedora-34,backing_fmt=qcow2 /tmp/p4g7utrp.qcow2
  File "/usr/lib64/python3.9/shutil.py", line 806, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/work/bots/make-checkout-workdir/test/common/pixeldiff.html' -> '/var/tmp/attachments._lujhdbk/pixeldiff.html'
```
This happens right at the start. Retrying once, to see if it was a CI flake or some systematic error (the branch is up to date with origin/main, so it's not a missing rebase)

test/check-machines-filesystems
```diff
@@ -0,0 +1,117 @@
...
+        # Try to add a new shared filesystem with non existing source directory
+        b.click("#vm-subVmTest1-filesystems-add")
+        b.set_input_text("#vm-subVmTest1-filesystems-modal-source-group input", "dir3")
+        b.wait_in_text("#vm-subVmTest1-filesystems-modal-source", "No such file or directory")
+        b.click("#vm-subVmTest1-filesystems-modal-cancel")
+
+        # Start VM and ensure that adding filesystem is disabled
+        b.click("#vm-subVmTest1-run")
+        b.wait_visible("#vm-subVmTest1-filesystems-add[aria-disabled=true]")
+
```
martinpitt, line 114: Would be good to check that the card/button don't appear on a session VM. This works right now, but let's make sure it stays that way. :-)

martinpitt (approved):
Thanks! I tested/reviewed this again, and looks good now!

test/check-machines-filesystems
KKoukiou: Will send shortly a follow up. Just want to make sure this is in before the end of the day :) And tests more people can review!

garrett:
There shouldn't be a scrollbar in a dialog, unless the content is quite tall. The content here is not tall whatsoever; it shouldn't overflow like this.

Screen Shot 2021-06-21 at 11 30 46

KKoukiou:
> There shouldn't be a scrollbar in a dialog, unless the content is quite tall. The content here is not tall whatsoever; it shouldn't overflow like this.
> 
> It has already been merged, so we should fix it in a follow-up. If it's easy enough, it would be cool to get it as a hotfix before we make the release.

There is nothing to fix, the screenshot was just hiding by mistake the second additional option. Therefore there was a scrollbar.

Updated the screenshot.

garrett:
Thanks, excellent. I guess it was hidden due to the viewport size, probably related to developer tools?

Glad to see it looks as expected in the screenshot!

PR #143, opened 2021-04-30
Harden publish-dist

martinpitt:
This avoids all recently discussed attack vectors of our build-dist/publish-dist workflow pair.

I tested this on my fork, for both a [push](https://github.com/martinpitt/cockpit-machines/runs/2476801243?check_suite_focus=true) and a [pull_request](https://github.com/martinpitt/cockpit-machines/runs/2476801243?check_suite_focus=true) from https://github.com/martinpitt/cockpit-machines/pull/4.

In both cases publishing succeeded, committed the expected sha-TAG to the -dist repo, and I validated that the cache works with `GITHUB_BASE=martinpitt/cockpit-machines test/download-dist`.

 - [x] After landing this, re-enable `pull_request` webhook

### DIFF
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -42,28 +42,3 @@ jobs:
 
           git push https://github.com/${{ github.repository }}-dist.git "$tag"
 
-      - name: Trigger integration tests for PRs
-        run: |
-          if ! grep -q '"pull_request"' event.json; then
-              echo "build-dist event was not a pull request"
-              exit 0
-          fi
-          # support running this from forks
-          if [ -z '${{ secrets.AMQP_CLIENT_KEY }}' ]; then
-              echo "No AMQP secrets, skipping tests-scan"
-              exit 0
-          fi
-
-          pip install pika
-
-          git clone --depth=1 https://github.com/cockpit-project/bots
-
-          # tests-scan does not currently have a --secrets-dir option
-          sudo mkdir -p /run/secrets/webhook
-          echo '${{ secrets.COCKPIT_CA_PEM }}' | sudo dd status=none of=/run/secrets/webhook/ca.pem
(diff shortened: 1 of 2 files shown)


martinpitt:
Note that publish-dist for *this PR* (and hence packit) **cannot** work -- this PR drops the SHA from the artifact, but publish-dist runs from main and expects this file. Hence the testing on my fork.

martinpitt:
I triggered fedora-33 manually, as that's a required context. As explained above, the others cannot be triggered here.

martinpitt:
I added a commit for eliminating events.json as well. Unfortunately there is no way to get trustworthy information about the original  pull_request event, so we have to give up on this experiment.

martinpitt:
@marusak , @KKoukiou : As Lis is out today again, can you please review this? Should be pretty straightforward, and it's the things that Lis and I already talked about, and her review in https://github.com/cockpit-project/cockpit/pull/15748

.github/workflows/publish-dist.yml
```diff
@@ -28,42 +29,16 @@ jobs:
...
           mkdir dist-repo
           cd dist-repo
           git init
           cp -r ../dist/ ../package-lock.json .
 
           git add .
-          git commit -m "Build for $SHA"
-          tag="sha-$SHA"
+          git commit -m "Build for ${{ github.event.workflow_run.head_sha }}"
+          tag="sha-${{ github.event.workflow_run.head_sha }}"
```
allisonkarlitskaya, line 40: I think you can theoretically switch to single-quotes here.

allisonkarlitskaya (approved):

.github/workflows/publish-dist.yml
martinpitt: Yes, and I already did that in the cockpit PR. There are a few other trivialities which I want to get in sync, I'll do that in a follow-up.

PR #144, opened 2021-05-03
tests: create fake fedora-28 osinfo-db entry also for debian stable

KKoukiou:
debian-stable image has some old osinfo-db version, namely
0.20181120-1+deb10u1, which does not contain the EOL date entry for
fedora-28.

Handle this case in the tests explicitely.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1057,6 +1057,9 @@ class TestMachinesCreate(VirtualMachinesCase):
             root.find('os').find('resources').find('minimum').find('storage').text = '134217728'
             if self.machine.image not in ["debian-stable"]:
                 root.find('os').find('eol-date').text = '9999-12-31'
+            else:
+                eol_date = ET.SubElement(root.find('os'), 'eol-date')
+                eol_date.text = '9999-12-31'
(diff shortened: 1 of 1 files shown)


test/check-machines-create
```diff
@@ -1057,6 +1057,9 @@ class TestMachinesCreate(VirtualMachinesCase):
             root.find('os').find('resources').find('minimum').find('storage').text = '134217728'
             if self.machine.image not in ["debian-stable"]:
                 root.find('os').find('eol-date').text = '9999-12-31'
+            else:
+                eol_date = ET.SubElement(root.find('os'), 'eol-date')
```
martinpitt, line 1061: Do these two queries do something different actually?

martinpitt (approved):
Thanks! Let's fix this quickly to unblock image refreshes.

test/check-machines-create
KKoukiou: In most distro the fedora28 entry in osinfo-db contains the eol-date entry already. In debian-stable the entry is missing completely. That's the only difference, the end result after the modification is the same.

PR #146, opened 2021-05-03
test: Add test/reference to .gitmodules

mvollmer:
This helps Github and others to understand better what's going on, and
it stops "git submodule" invocations from throwing errors.

Also, implementing "pixel-tests pull" in terms of "git submodule"
looks like a very sensible thing anyway.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "test/reference"]
+	path = test/reference
+	url = ../pixel-test-reference


KKoukiou (approved):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "test/reference"]
+	path = test/reference
+	url = ../pixel-test-reference
```
KKoukiou, line 3: Hmm, this is expected actuallY?

KKoukiou (changes requested):

.gitmodules
mvollmer: Yes, it's a feature.  This way, pixel-test-reference clone uses the same access method as the parent cockit-machines clone.

KKoukiou: but does that imply some specific local folder setup for the clones repos?

mvollmer: No, this is only about the URLs that the clones are retrieved from.  The local repositories layout is defined by the `path` entry (which is of course also relative).

KKoukiou (approved):

mvollmer:
`packit/testing-farm-fedora-33-x86_64` has succeeded but is still shown as "pending" here.

PR #151, opened 2021-05-03
Support multiple ways for getting IP address of a virtual machine

### LINKED ISSUE #61: Virtual Machine IP listed as unknown when VM connected via Bridge
Cockpit version: 220
OS:Fedora 32
Page: Machines

I am creating my virtual machines from Cockpit. There is a bridge connection for my machines to reach host network and VM's connect to network via bridge. When I add the VM to bridge machine connected to external DHCP and take IP address but it still shows as Unknown in Network Interfaces tab. In virt-manager I can see the VM's IP.

Steps to reproduce

1. Create a bridge network
2. Create a virtual machine
3. Add this bridge connection from Network Interface 
4. Run VM and see ıf it is getting an IP from DHCP
5. Check network interfaces tab to see IP set to unknown or VM's IP

KKoukiou:
Fixes #61

### DIFF
--- a/src/components/vm/nics/vmNicsCard.jsx
+++ b/src/components/vm/nics/vmNicsCard.jsx
@@ -100,11 +100,13 @@ export class VmNetworkTab extends React.Component {
         super(props);
 
         this.state = {
-            interfaceAddress: [],
+            domifaddressAllSources: [],
             networkDevices: undefined,
+            ips: {},
         };
 
         this.deviceProxyHandler = this.deviceProxyHandler.bind(this);
+        this.getIpAddr = this.getIpAddr.bind(this);
         this.client = cockpit.dbus("org.freedesktop.NetworkManager", {});
         this.hostDevices = this.client.proxies("org.freedesktop.NetworkManager.Device");
         this.hostDevices.addEventListener('changed', this.deviceProxyHandler);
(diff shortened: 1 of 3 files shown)


src/libvirt-dbus.js
```diff
@@ -1568,8 +1570,32 @@ export function storageVolumeDelete(connectionName, poolName, volName) {
...
+                return Promise.resolve(res);
+            })
+            .catch(exc => {
+                error.push(exc.message);
+                return call(connectionName, objPath, 'org.libvirt.Domain', 'InterfaceAddresses', [Enum.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_AGENT, 0], { timeout, type: 'uu' });
+            })
+            .catch(exc => {
+                error.push(exc.message);
+                return Promise.reject(new Error(error.join(", ")));
+            });
```
KKoukiou, line 1598: Hmm I can also use Promise.allSettled() instead. That will make it slightly simpler I feel.

src/components/vm/nics/vmNicsCard.jsx
```diff
@@ -124,9 +145,7 @@ export class VmNetworkTab extends React.Component {
 
         // Load the interface addresses list when the tab mounts
         vmInterfaceAddresses(this.props.vm.connectionName, this.props.vm.id)
-                .then(ifaces => {
-                    this.setState({ interfaceAddress: ifaces[0] });
-                }, ex => {
+                .then(res => this.setState({ domifaddress: res[0] }), ex => {
                     this.props.onAddErrorNotification({
```
skobyda, line 149: What does it mean if it cannot find any IP address? Would that in extent also mean that VM has no network connectivity? If not, then maybe showing the error message like this is too aggressive
Screenshot from 2021-05-04 12-58-26
Since it could make user believe that's some big problem which needs to be fixed. But if it just means that they will not be able to see the IP address in vNIC tab, and the VM will function normally, maybe we could just print the error to `console.error()` 

Again, I suggest this only if this error doesn't cause VM to lose network connectivity. If it does, then it's important and let's keep it like this.

skobyda (changes requested):

src/components/vm/nics/vmNicsCard.jsx
KKoukiou: To be honest I was not able to create this message in any way but to `fake` the method names to make them fail. How did you make it fail?

Basicaly any NIC type I added, ended up working with at least one of these methods.

KKoukiou: Of course for the qemu-guest-agent to work we need the agent inside guest, which is present for cloud images but not generally.
But ARP tables seemed very generic.

skobyda: I don't know why that one in my screenshot fails, but I found out you can get all these errors also if you run our test VM `./bots/vm-run fedora-33 --maintain`, and then thru the UI add some network interface to it (i.e. Userspace SLIRP stack) and reload the page (the fact that you have to reload the page to get IP info is also a problem)

KKoukiou: Good catch, so empty list returned from arp or any of these sources should not be considered an error.
Fixed this, and will repost soon.

src/components/vm/nics/vmNicsCard.jsx
```diff
@@ -115,6 +115,27 @@ export class VmNetworkTab extends React.Component {
...
+                    const type = ifAddress[0] == 0 ? 'inet ' : 'inet6 ';
+                    // 0 cell -> type, 1 cell -> address, 2 cell -> prefix
+                    ipaddr[ifAddress[0]] = type + ifAddress[1] + '/' + ifAddress[2];
+                }
+            });
+        });
+        return ipaddr;
+    }
+
     componentDidMount() {
```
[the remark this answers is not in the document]
KKoukiou: Makes sense, will fix.

src/libvirt-dbus.js
```diff
@@ -1568,8 +1570,32 @@ export function storageVolumeDelete(connectionName, poolName, volName) {
...
+
+                return Promise.resolve(res);
+            })
+            .catch(exc => {
+                error.push(exc.message);
+                return call(connectionName, objPath, 'org.libvirt.Domain', 'InterfaceAddresses', [Enum.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_AGENT, 0], { timeout, type: 'uu' });
+            })
+            .catch(exc => {
+                error.push(exc.message);
+                return Promise.reject(new Error(error.join(", ")));
```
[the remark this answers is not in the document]
KKoukiou: Done

src/components/vm/nics/vmNicsCard.jsx
```diff
@@ -216,17 +235,20 @@ export class VmNetworkTab extends React.Component {
             { name: _("MAC address"), value: 'mac' },
             {
                 name: _("IP address"), value: (network) => {
```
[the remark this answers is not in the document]
KKoukiou: Done

src/libvirt-dbus.js
KKoukiou: Followed this approach indeed

skobyda (approved):
Thank you for the work. LGTM

PR #152, opened 2021-05-04
test: Only ignore #storage-size-helper when it is present

mvollmer:
The "assert_pixels" function used to skip selectors in the "ignore"
list that aren't actually present, but this will soon no longer be the
case.  It is better to be explicit in tests, and it is very easy in
this case, too.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -970,8 +970,10 @@ class TestMachinesCreate(VirtualMachinesCase):
         def assert_pixels(self, subtag=None):
             if self.pixel_test_tag:
                 tag = self.pixel_test_tag + ("-" + subtag if subtag else "")
-                self.browser.assert_pixels("#create-vm-dialog", tag,
-                                           ignore=["#storage-size-helper", "#memory-size-helper"])
+                ignore = ["#memory-size-helper"]
+                if self.storage_pool != "NoStorage":
+                    ignore += ["#storage-size-helper"]
+                self.browser.assert_pixels("#create-vm-dialog", tag, ignore=ignore)
             return self
 
     class CreateVmRunner:


KKoukiou (dismissed):

martinpitt:
@mvollmer: packit tests fail with
```
test/common/pixel-tests pull
Submodule 'test/reference' (https://github.com/mvollmer/pixel-test-reference) registered for path 'test/reference'
Cloning into '/ARTIFACTS/work-allxCCrga/plans/all/discover/default/tests/test/reference'...
fatal: could not read Username for 'https://github.com': No such device or address
fatal: clone of 'https://github.com/mvollmer/pixel-test-reference' into submodule path '/ARTIFACTS/work-allxCCrga/plans/all/discover/default/tests/test/reference' failed
```
which looks systematic, not just a random network flake. This looks like a regression in https://github.com/cockpit-project/cockpit/pull/15479 ?

mvollmer:
> @mvollmer: packit tests fail with
> 
> ```
> test/common/pixel-tests pull
> Submodule 'test/reference' (https://github.com/mvollmer/pixel-test-reference) registered for path 'test/reference'
> Cloning into '/ARTIFACTS/work-allxCCrga/plans/all/discover/default/tests/test/reference'...
> fatal: could not read Username for 'https://github.com': No such device or address
> fatal: clone of 'https://github.com/mvollmer/pixel-test-reference' into submodule path '/ARTIFACTS/work-allxCCrga/plans/all/discover/default/tests/test/reference' failed
> ```
> 
> which looks systematic, not just a random network flake. This looks like a regression in [cockpit-project/cockpit#15479](https://github.com/cockpit-project/cockpit/pull/15479) ?

Yes, this is the "we can't use relatove paths in .gitmodules after all" bug.

KKoukiou (approved):

PR #156, opened 2021-05-07
VM creation script should read qemu.conf to get spice/vnc ports - FIX COMMIT MESSAGE: port → address

### LINKED ISSUE #63: cockpit-machines: GRAPHICS_PARAM should check for qemu.conf listen configuration, but 127.0.0.1 is hardcoded
Cockpit version: 217
OS: Fedora 32
Page: Machines

**Issue Description**

When creating a VM via the cockpit interface, the listening address for external SPICE and VNC viewers will always be 127.0.0.1. Setting the default listen address in /etc/libvirt/qemu.conf has no effect on this, because the 127.0.0.1 address is hardcoded in GRAPHICS_PARAM

**Steps to reproduce**

1. Install cockpit, libvirtd, cockpit-machines, and all other relevant packages.
2. Modify /etc/libvirt/qemu.conf to define vnc_listen and spice_listen as = "0.0.0.0". restart libvirt etc.
3. Create a VM via the web interface.

**Actual results:**
A VM is created and VNC & SPICE are bound and listening to 127.0.0.1

**Expected results:**
VNC & SPICE should be listening to the address as it is defined in qemu.conf, and 127.0.0.1 by default otherwise.

skobyda:
Fixes #63

https://bugzilla.redhat.com/show_bug.cgi?id=1833829

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -370,6 +370,19 @@ class TestMachinesCreate(VirtualMachinesCase):
         b.wait_in_text("#vm-pxe-guest-boot-order", "disk")
         b.wait_not_in_text("#vm-pxe-guest-boot-order", "network")
 
+    def testCreateAndVerifyQemuConf(self):
+        runner = TestMachinesCreate.CreateVmRunner(self)
+        m = self.machine
+
+        self.login_and_go("/machines")
+        self.browser.wait_in_text("body", "Virtual machines")
+
+        spice_listen = "192.168.2.6"
+        m.execute("echo 'spice_listen = \"{0}\"' >> /etc/libvirt/qemu.conf".format(spice_listen))
+
+        runner.createAndVerifyQemuConfParsedTest(TestMachinesCreate.VmDialog(self, sourceType='os', storage_size=256, storage_size_unit='MiB'),
+                                                 spice_listen)
+
     @no_retry_when_changed
     def testCreateThenInstall(self):
         runner = TestMachinesCreate.CreateVmRunner(self)
@@ -803,6 +816,18 @@ class TestMachinesCreate(VirtualMachinesCase):
 
             self.machine.execute(f"virsh destroy {self.name}")
             self.assertIn(f"backing file: {self.location}", self.machine.execute(f"qemu-img info /var/lib/libvirt/images/{self.name}.qcow2"))
+
+        def createAndVerifyVirtInstallArgsQemuConf(self, listen_address):
+            self.browser.click(".pf-c-modal-box__footer button:contains(Create)")
+            self.browser.wait_not_present("#create-vm-dialog")
+
+            # usual tricks: prevent grep from seeing itself in the output of ps
+            virt_install_cmd = "ps aux | grep '[v]irt-install --connect'"
+
+            wait(lambda: self.machine.execute(virt_install_cmd), delay=3)
+            virt_install_cmd_out = self.machine.execute(virt_install_cmd)
+
+            self.assertIn("listen={0}".format(listen_address), virt_install_cmd_out)
 
         def createAndVerifyVirtInstallArgsUnattended(self):
             self.browser.click(".pf-c-modal-box__footer button:contains(Create)")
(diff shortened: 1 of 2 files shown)


KKoukiou (dismissed):
works for me but I don't feel comfortable reviewing shell scripts as I don't know the best practiced. @martinpitt or @allisonkarlitskaya do you want to review the new function?

src/scripts/create_machine.sh
```diff
@@ -22,6 +22,19 @@ vmExists(){
...
 
+readQemuConf(){
+    IFS="="
+    while read -r name value
+    do
+	# removes whitespaces
+	name="$(echo -e "${name}" | tr -d '[:space:]')"
+	if [ "$name" = "$1" ]; then
+		# removes " and whitespaces
+		echo -e "${value//\"/}" | tr -d '[:space:]'
```
martinpitt, line 33: You don't normally want to remove whitespace *inside* the value, especially not when it's quoted. How about something like this:

    awk '/^[[:space:]]*'"$1"'[[:space:]]*=/ { sub(/^.*= "?/, ""); sub(/"[[:space:]]*$/, ""); print }' /etc/libvirt/qemu.conf

I know it's a mouthful, but I believe it's quite a bit more precise. It also avoids the side effect of setting `$IFS`, which will influence the code afterwards (which can  be very surprising and hard to see)

test/check-machines-create
```diff
@@ -778,6 +791,29 @@ class TestMachinesCreate(VirtualMachinesCase):
             virt_install_cmd_out = self.machine.execute(virt_install_cmd)
             self.assertIn("--cloud-init user-data=", virt_install_cmd_out)
 
+        def createAndVerifyVirtInstallArgsQemuConf(self):
+            self.browser.click(".pf-c-modal-box__footer button:contains(Create)")
+            self.browser.wait_not_present("#create-vm-dialog")
+            # Get IP address of 'spice_listen = "IP_ADDRESS"' from qemu.conf, and trim quotes and endline
+            spice_ip = self.machine.execute("cat /etc/libvirt/qemu.conf | grep '^spice_listen' | awk '{print $3}' | sed 's/\"//g' | tr -d '\\n'")
```
martinpitt, line 798: Ugh, five different shell tools... Replace the first four with `awk '/^spice_listen/ {sub(/"/, ""); print $3}', and replace the `tr` with a Python `execute(...).strip()`.

But it would IMHO be even better to just pass the spice_listen and vnc_listen values as parameters to createAndVerifyVirtInstallArgsQemuConf() -- then it can write qemu.conf itself, and does not have to parse them back out of the file that the test just wrote. This is a whole lot simpler and more robust.

martinpitt (changes requested):
Thanks! I couldn't help but bikeshed about the string handling indeed, sorry :grin:

allisonkarlitskaya:
> allisonkarlitskaya

My Friday-afternoon drive-by: rewrite the entire thing in python please.

src/scripts/create_machine.sh
```diff
@@ -22,6 +22,12 @@ vmExists(){
    virsh -c "$CONNECTION_URI" list --all | awk  '{print $2}' | grep -q --line-regexp --fixed-strings "$1"
 }
 
+# Read a parameter value from a qemu.conf
+# i.e.: for a key-value 'spice_listen = "192.168.9.2"', the function will return '192.168.9.2'
+readQemuConf(){
+    awk '/^[[:space:]]*'"$1"'[[:space:]]*=/ { sub(/^.*= "?/, ""); sub(/"[[:space:]]*$/, ""); print }' /etc/libvirt/qemu.conf;
```
allisonkarlitskaya, line 28: -1 from me.  This is really past the realm of "reasonable".

allisonkarlitskaya (changes requested):
I think this script is really past saving.  Another example of the madness going on here: we pass the user and root passwords as commandline arguments, open for everyone on the system to see...

This needs to either be replaced with more logic in the javascript world (for example, the files could easily be constructed there) or be rewritten as a python script which receives its 16(!) arguments as a json blob, securely, on stdin.  virt-install already depends indirectly on python3-libvirt, so we could even avoid the shelling out and use the API directly...

KKoukiou:
Please review as one commit, this needs to be squashed anyway.

martinpitt (approved):
The code and tests look good, thanks! (Much nicer than with shell)

However, let's  fix the commit message on merging -- this is the listening *address* (as the bugzilla correctly says), *not* the port. The port is controlled by `remote_{display,websocket}_port_min`, but assigned dynamically, so that needs to be handled differently.

Danke!

PR #166, opened 2021-05-12
DONTMERGE: Test testlib fix with eliminated dataclass

martinpitt:
This tests the change in https://github.com/cockpit-project/cockpit/pull/15825
but as cockpit itself does not use pixel testing yet, we need to test it
in this project.


martinpitt:
It got far enough, this works. Closing and checking off in https://github.com/cockpit-project/cockpit/pull/15825

PR #171, opened 2021-05-19
Test fixes

martinpitt:
This fixes the mess that we saw in https://github.com/cockpit-project/bots/pull/2020 ([log](https://logs.cockpit-project.org/logs/pull-2020-20210517-150150-d470d5b6-rhel-8-5-cockpit-project-cockpit-machines/log.html)) and in the downstream RHEL 8.5 update.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1086,14 +1086,7 @@ class TestMachinesCreate(VirtualMachinesCase):
 
             # Make sure that the initial state goes away and then try to check what the new state is
             # because we might end up checking the text for the new state the momment it disappears
-            def waitStateWentAway(selector, state):
-                try:
-                    b.wait_text_not(selector, state)
-                    return True
-                except Error:
-                    return False
-
-            wait(lambda: waitStateWentAway(selector, before), delay=3)
+            wait(lambda: b.text(selector) != before, tries=30, delay=3)
             b.wait_in_text(selector, after)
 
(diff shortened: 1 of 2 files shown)


test/check-machines-create
```diff
@@ -1088,12 +1088,13 @@ class TestMachinesCreate(VirtualMachinesCase):
...
                 try:
-                    b.wait_text_not(selector, state)
+                    with b.wait_timeout(10):
+                        b.wait_text_not(selector, state)
                     return True
                 except Error:
                     return False
 
-            wait(lambda: waitStateWentAway(selector, before), delay=3)
+            wait(lambda: waitStateWentAway(selector, before), tries=10, delay=3)
```
KKoukiou, line 1097: We could also do a plain:
wait(lambda: b.text(selector) != state, tries=10, delay=3) and remove waitStateWentAway

KKoukiou (approved):

test/check-machines-create
martinpitt: You are right, this is even better. Admittedly I was not sure what exactly this code is supposed to achieve. I'll wait for this round of test results and push that, it's much easier to read/understand.

martinpitt: Previous testing round looks promising. I applied this simplification.

KKoukiou (approved):

PR #172, opened 2021-05-19
Add git-core to BuildRequires

KKoukiou:


martinpitt:
Can you please add a justification to the commit message? There is nothing in the spec that would require git at runtime. The whole rpm build is effectively just a `cp -r dist/ $DESTDIR` and an `appstream-util validate`.

KKoukiou:
This come from this failure I saw on the CI:

```
+ /usr/bin/make install DESTDIR=/builddir/build/BUILDROOT/cockpit-machines-244-1.el9.noarch 'INSTALL=/usr/bin/install -p'
flock Makefile sh -ec '\
    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 242; \
    git checkout --force FETCH_HEAD -- pkg/lib; \
    git reset -- pkg/lib'
sh: line 2: git: command not found
```

It's because I used the wrong tarball though, I just realized.

PR #174, opened 2021-05-19
Fix libvirt-dbus DELETE_VM

martinpitt:
undefine() previously returned a Promise, which is not valid for a
reducer:

    Actions must be plain objects. Instead, the actual type was: 'Promise'

This was previously hidden by a hack to use cockpit.all(), which the
type checker silenty accepted. But the real cause was that returned
promises need to be dispatched, so do that.

-----

This is the reason why https://github.com/cockpit-project/bots/pull/2029 and packit tests started to fail -- cockpit.all() is gone from cockpit 244. I thought I grepped all our projects, but I missed that one spot, I'm really sorry! (I re-checked -podman, -ostree, sub-man again).

After this lands, we need to do a  new release to Fedora and RHEL.


src/libvirt-dbus.js
```diff
@@ -369,18 +369,16 @@ const LIBVIRT_DBUS_PROVIDER = {
                 }
             }
 
-            // FIXME: use Promise.all() here; but that causes "Error: Actions must be plain objects"
-            // eslint-disable-next-line cockpit/no-cockpit-all
-            return cockpit.all(storageVolPromises)
+            return Promise.all(storageVolPromises)
                     .then(() => {
-                        return call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [flags], { timeout, type: 'u' });
+                        return dispatch(call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [flags], { timeout, type: 'u' }));
```
martinpitt, line 374: I am actually not 100% sure if the dispatch() needs to go here or somewhere else -- I'm a redux n00b. Hence, @KKoukiou , I'd appreciate extra scrutiny here. Thanks!

martinpitt:
I triggered an additional test against the new fedora-34 image, and packit is happy again as well.

KKoukiou (approved):

KKoukiou:
Some f rawhide packit failures: https://dashboard.packit.dev/results/testing-farm/24287

How do I retrigger only the rawhide packit tests? 

Will retrigger for now everything.

/packit build

martinpitt:
@KKoukiou : You can't re-run a single packit release only; but you can use `/packit test` to only re-run the test without rebuilding srpm and copr. And the comment needs to be *only* that command. Done now.

martinpitt:
So at least cloning broke in rawhide -- we need to investigate that separately. It's not related to the "delete VM" action. Landing this now, as this is urgent.

martinpitt:
I kicked off a 244.1 release for this.

PR #176, opened 2021-05-20
List host devices attached to a VM

skobyda:
Release notes:

Machines: list USB, PCI and SCSI host devices that are passed through to the guest
----------------------------------------------------------------------------------------------------------------

Screen Shot 2021-06-22 at 11 43 18

### DIFF
--- a/src/components/vm/hostdevs/hostDevCard.jsx
+++ b/src/components/vm/hostdevs/hostDevCard.jsx
@@ -1,0 +1,212 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2021 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import cockpit from 'cockpit';
+import { vmId, findHostNodeDevice } from "../../../helpers.js";
+import { ListingTable } from "cockpit-components-table.jsx";
+
+const _ = cockpit.gettext;
+
+/* Adds an optional description-value pair to an array which represents multiple values of a table cell
+ *
+ * @param chunks an array into which element is added to
+ * @param value a value of the descr-value pair
+ * @param descr a description of the descr-value pair
+ * @param id
+ */
+function addOptionalToCell(chunks, value, id, descr) {
+    if (value) {
+        chunks.push(
+            <div className='machines-multivalue-column' key={descr}>
+                <strong className='machines-multivalue-column-label'>
+                    {descr}
+                </strong>
+                <span className='machines-multivalue-column-value' id={id}>
+                    {value}
+                </span>
+            </div>
+        );
+    }
+}
+
+export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
+    function getClass(hostDev, hostdevId) {
+        const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
+
+        if (nodeDev && (["usb_device", "pci"].includes(nodeDev.capability.type)))
+            return nodeDev.class;
+    }
+
+    function getProduct(hostDev, hostdevId) {
+        const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
+
+        if (["usb", "pci"].includes(hostDev.type)) {
+            if (nodeDev)
+                return nodeDev.capability.product._value;
+            else if (hostDev.type === "usb")
+                return hostDev.source.product.id;
+        }
+    }
+
+    function getVendor(hostDev, hostdevId) {
+        const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
+
+        if (["usb", "pci"].includes(hostDev.type)) {
+            if (nodeDev)
+                return nodeDev.capability.vendor._value;
+            else if (hostDev.type === "usb")
+                return hostDev.source.vendor.id;
+        }
+    }
+
+    function getSource(hostDev, hostdevId) {
+        const cell = [];
+        const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
+        if (hostDev.type === "usb" && nodeDev) {
+            const device = nodeDev.devnum;
+            const bus = nodeDev.busnum;
+
+            addOptionalToCell(cell, device, `${hostdevId}-device`, _("Device"));
+            addOptionalToCell(cell, bus, `${hostdevId}-bus`, _("Bus"));
+        } else if (hostDev.type === "pci") {
+            let domain = hostDev.source.address.domain.split('x')[1];
+            let bus = hostDev.source.address.bus.split('x')[1];
+            let slot = hostDev.source.address.slot.split('x')[1];
+            let func = hostDev.source.address.func.split('x')[1];
+
+            domain = String(domain).padStart(4, '0');
+            bus = String(bus).padStart(2, '0');
+            slot = String(slot).padStart(2, '0');
+            func = String(func).padStart(1, '0');
+
+            addOptionalToCell(cell, `${domain}:${bus}:${slot}.${func}`, `${hostdevId}-slot`, _("Slot"));
+        } else if (hostDev.type === "scsi") {
+            const bus = hostDev.source.address.bus;
+            const target = hostDev.source.address.target;
+            const unit = hostDev.source.address.lun;
+
+            addOptionalToCell(cell, bus, `${hostdevId}-bus`, _("Bus"));
+            addOptionalToCell(cell, unit, `${hostdevId}-unit`, _("Slot"));
+            addOptionalToCell(cell, target, `${hostdevId}-target`, _("Target"));
+        } else if (hostDev.type === "scsi_host") {
+            const protocol = hostDev.source.protocol;
+            const wwpn = hostDev.source.wwpn;
+
+            addOptionalToCell(cell, protocol, `${hostdevId}-protocol`, _("Protocol"));
+            addOptionalToCell(cell, wwpn, `${hostdevId}-wwpn`, _("WWPN"));
+        } else if (hostDev.type === "mdev") {
+            const uuid = hostDev.source.address.uuid;
+
+            addOptionalToCell(cell, uuid, `${hostdevId}-uuid`, _("UUID"));
+        } else if (hostDev.type === "storage") {
+            const block = hostDev.source.block;
+
+            addOptionalToCell(cell, block, `${hostdevId}-block`, _("Block"));
+        } else if (hostDev.type === "misc") {
+            const ch = hostDev.source.char;
+
+            addOptionalToCell(cell, ch, `${hostdevId}-char`, _("Char"));
+        } else if (hostDev.type === "net") {
+            const iface = hostDev.source.interface;
+
+            addOptionalToCell(cell, iface, `${hostdevId}-interface`, _("Interface"));
+        }
+
+        return cell;
+    }
+
+    const id = vmId(vm.name);
+
+    // Hostdev data mapping to rows
+    const detailMap = [
+        {
+            name: _("Type"), value: (hostdev, hostdevId) => {
+                return (
+                    <span id={`${id}-hostdev-${hostdevId}-type`}>
+                        {hostdev.type}
+                    </span>
+                );
+            }
+        },
+        {
+            name: _("Class"), value: (hostdev, hostdevId) => {
+                return (
+                    <span id={`${id}-hostdev-${hostdevId}-class`}>
+                        {getClass(hostdev, hostdevId)}
+                    </span>
+                );
+            }
+        },
+        {
+            name: _("Model"), value: (hostdev, hostdevId) => {
+                return (
+                    <div id={`${id}-hostdev-${hostdevId}-product`}>
+                        {getProduct(hostdev, hostdevId)}
+                    </div>
+                );
+            }
+        },
+        {
+            name: _("Vendor"), value: (hostdev, hostdevId) => {
+                return (
+                    <div id={`${id}-hostdev-${hostdevId}-vendor`}>
+                        {getVendor(hostdev, hostdevId)}
+                    </div>
+                );
+            }
+        },
+        {
+            name: _("Source"), value: (hostdev, hostdevId) => {
+                return (
+                    <div id={`${id}-hostdev-${hostdevId}-source`}>
+                        {getSource(hostdev, hostdevId)}
+                    </div>
+                );
+            }
+        },
+    ];
+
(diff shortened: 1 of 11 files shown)


src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,175 @@
...
+                </strong>
+                <span className='machines-multivalue-column-value' id={id}>
+                    {value}
+                </span>
+            </div>
+        );
+    }
+}
+
+addOptionalToCell.propTypes = {
```
KKoukiou, line 50: Ehm I always thought that props is an object, above it's not however.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,175 @@
...
+                column = { title: d.value(target, hostdevId, vm.connectionName) };
+            }
+            return column;
+        });
+        hostdevId++;
+        return { columns, props: { key: hostdevId } };
+    });
+
+    return (
+        <ListingTable aria-label={`VM ${vm.name} Host Devices`}
```
KKoukiou, line 162: This needs to be translated.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,175 @@
...
+            gridBreakPoint='grid-xl'
```
KKoukiou, line 163: Please ensure that mobile mode works as expected and post some screenshots.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,175 @@
...
+            variant='compact'
+            emptyCaption={_("No host devices assigned to this VM")}
+            rows={rows}
+            columns={columnTitles} />
+    );
+};
+
+VmHostDevCard.propTypes = {
+    vm: PropTypes.object.isRequired,
+    config: PropTypes.object.isRequired,
```
KKoukiou, line 173: Missing nodeDevices

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,175 @@
...
+const _ = cockpit.gettext;
+
+/* Adds an optional description-value pair to an array which represents multiple values of a table cell
+ *
+ * @param chunks an array into which element is added to
+ * @param value a value of the descr-value pair
+ * @param descr a description of the descr-value pair
+ * @param id
+ */
+function addOptionalToCell(chunks, value, id, descr) {
```
KKoukiou, line 35: This function is copied from somewhere (disks?) right?
Let's move it to helpers or some common file and reuse it instead.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,175 @@
...
+    ];
+
+    let hostdevId = 1;
+    detailMap = detailMap.filter(d => !d.hidden);
+
+    const columnTitles = detailMap.map(target => target.name);
+    const rows = vm.hostDevices.sort().map(target => {
+        const columns = detailMap.map(d => {
+            let column = null;
+            if (typeof d.value === 'string') {
```
KKoukiou, line 147: This never exists above. All are functions.

test/check-machines-hostdevs
```diff
@@ -0,0 +1,86 @@
...
+
+from machineslib import VirtualMachinesCase  # noqa
+from testlib import nondestructive, test_main  # noqa
+from machinesxmls import USB_HOSTDEV, PCI_HOSTDEV  # noqa
+
+
+@nondestructive
+class TestMachinesHostDevs(VirtualMachinesCase):
+
+    def testVmHostDevs(self):
```
KKoukiou, line 36: testHostDevicesLists is better maybe? So differentiate later with creating/removing. etc.

KKoukiou (commented):
I did not try it locally yet, some minor comments :)

src/components/vm/hostdevs/hostDevCard.jsx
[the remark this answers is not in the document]
skobyda: I'm not sure. It does not render any component. It just a helper function which ADDS a component to a list. That list is rendered somewhere else, no?
So I think calling such function which only adds to a list but doesn't render anything with a react functional component interface (i.e. `<AddOptionalCell chunks={chunks} ... />`) would be confusing.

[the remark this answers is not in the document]
KKoukiou: A React component does not have to be a whole part of the page, it can definitely be a row in a table, or a row in some other structure like here. 

Basically we are just discussing if returning React Node types makes sense to do in normal functions or we should write them as React functional components. See https://stackoverflow.com/questions/64453528/in-react-what-is-the-difference-between-using-functional-component-and-just-a-f

In any case, you need to decide, either you use a normal function and don't use propTypes, or a functional component with proptypes.

garrett (dismissed):
The host device table for VMs is inconsistent with the host device table @ `/system/hwinfo` page (https://localhost:9090/system/hwinfo as a local URL). 

Here's a screenshot excerpt from my system:

Screenshot_2021-05-25 Overview - garrett Rain(1)

The order of the columns is human-friendly, where it has the class of device, the device name, who makes it, and then the additional slot information. Additionally, the class is a human-friendly name instead of a (lowercased) acronym. (I'm not sure if you can use the same source of information for that though.)

It was based on super-old mockups @ https://garrett.github.io/cockpit-mockweb/hardware/ — this is my laptop's data: https://garrett.github.io/cockpit-mockweb/hardware/t460s ... we currently don't list USB on the information page, it seems. But we could/should. (But that's out of scope of this PR, obviously. I'm only mentioning it because it's useful to show all the devices attached to a VM, even if we currently don't show them in the hwinfo page at the moment.)

**TL;DR**: Please reorder and make it like the hwinfo page.

skobyda:
@garrett So I tried to make it like hwinfo page.
Screenshot from 2021-06-02 14-02-54

But as this table shows many types of devices (PCI, USB, mdev, block...), It will be a bit different.
The differences are: The **class** and **type** properties are not the same, and **type** property needs to be shown nevertheless, as it is quite a crucial piece of information what type of device it is.
**Class** is available for almost all the PCI devices, but only for some USB devices.
I moved the **slot** into a shared **Source** column, as each devices, wherever it is pci, USB, block, mdev, etc... has different source identifiers. For USB, it's not slot, but a combination of device/bus numbers. So in order to not have a dozens of columns, maybe we can show those identifier numbers in one column?

garrett:
> But as this table shows many types of devices (PCI, USB, mdev, block...), It will be a bit different.

Sure; I didn't mean that it should be 1:1, as it's not exactly the same thing, but that the design should feel similar.

How is the table sorted? It feels like it would be best to sort it by type first, then source. Those are the only two bits of data I See that could determine the order. Everything else is arbitrary (including alphanumeric sort on the model and vendor).

Source is odd, things don't line up and it shouldn't have to be vertical.

This is what I mean by things not lining up. They're 8px apart.

Screenshot from 2021-06-02 16-12-08

But it's looking good to me aside from the above!

KKoukiou (changes requested):
inserting: 
```
  <hostdev mode='subsystem' type='usb'>
    <source startupPolicy='optional'>
      <vendor id='0x1234'/>
      <product id='0xbeef'/>
    </source>
  </hostdev>
```

Crashes the page.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,215 @@
...
+        const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
+
+        if (nodeDev && (["usb_device", "pci"].includes(nodeDev.capability.type)))
+            return nodeDev.capability.vendor._value;
+    }
+
+    function getSource(hostDev, hostdevId) {
+        const cell = [];
+        if (hostDev.type === "usb") {
+            const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
```
KKoukiou, line 82: Please add the if nodeDev check here as well. See the example I posted which crashes it.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,215 @@
...
+
+            addOptionalToCell(cell, iface, `${hostdevId}-interface`, _("Interface"));
+        }
+
+        return cell;
+    }
+
+    const id = vmId(vm.name);
+
+    // Network data mapping to rows
```
KKoukiou, line 130: Fix comment

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,215 @@
...
+                    <div id={`${id}-hostdev-${hostdevId}-source`}>
+                        {getSource(hostdev, hostdevId)}
+                    </div>
+                );
+            }
+        },
+    ];
+
+    let hostdevId = 1;
+    detailMap = detailMap.filter(d => !d.hidden);
```
KKoukiou, line 180: I don't think this is applicable here.

KKoukiou (changes requested):
In src/components/vm/vmDetailsPage.jsx:

Add also the new card here:

<CardBody className={["disks-card", "networks-card", "snapshots-card"].includes(card.className) ? "contains-list" : ""}>

KKoukiou (changes requested):
Same it crashes with:

```
    <hostdev mode='subsystem' type='pci' managed='yes'>
      <source>
        <address domain='0x0000' bus='0x06' slot='0x02' function='0x0'/>
      </source>
      <rom bar='on' file='/etc/fake/boot.bin'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x0b' function='0x0'/>
    </hostdev>

```

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,215 @@
...
+            const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
+            const slot = nodeDev.pciSlotName;
+
+            addOptionalToCell(cell, slot, `${hostdevId}-slot`, _("Slot"));
+        } else if (hostDev.type === "scsi") {
+            const bus = hostDev.source.address.bus;
+            const target = hostDev.source.address.target;
+            const unit = hostDev.source.address.lun;
+
+            addOptionalToCell(cell, bus, `${hostdevId}-bus`, _("Bus"));
```
KKoukiou, line 98: I noticed in virt-manager that they reprecent the hos devices like: PCI 0000:06:02.0

domain:\bus:slot:function

Shall we do it also like this to save some space - and not have so many obscure labels?

KKoukiou (changes requested):

src/components/vm/hostdevs/hostDevCard.jsx
skobyda: I try to make it closest as possible with what is the result of very popular commands "lspci" and "lsusb". 
- While "lspci" returns a list of pci devices where multiple identifiers are put together into a "slotname" with an annotation like PCI 0000:06:02.0
- the command "lsusb" returns a list of usb devices in a format "Bus 002 Device 002: ID 0bda:0316 Realtek Semiconductor Corp. Card Reader", where it states bus and device explicitely.

Also same for other command, `udevadm`:
- For pci devices it exposes one impotant property: PCI_SLOT_NAME=0000:00:1f.6
- For usb devices however it exposes two properties:
E: BUSNUM=002
E: DEVNUM=002

So I don't feel like trying to invent something new. If it's common to show one property a `slot_name` in a format 0000:00:1f.6 for pci devices, but to show 2 different properties for usb devices, I think we should stick with that

skobyda:
Screenshot for mobile:

Screenshot from 2021-06-10 13-01-00

skobyda:
> Same it crashes with:
> 
> ```
>     <hostdev mode='subsystem' type='pci' managed='yes'>
>       <source>
>         <address domain='0x0000' bus='0x06' slot='0x02' function='0x0'/>
>       </source>
>       <rom bar='on' file='/etc/fake/boot.bin'/>
>       <address type='pci' domain='0x0000' bus='0x00' slot='0x0b' function='0x0'/>
>     </hostdev>
> ```

Fixed both crashes

KKoukiou (changes requested):
@skobyda I am using the first first examples from https://libvirt.org/formatdomain.html#host-device-assignment and I get:

Screen Shot 2021-06-12 at 10 15 52


Why are most of the details missing?

src/components/vm/vmDetailsPage.jsx
```diff
@@ -145,6 +146,12 @@ export const VmDetailsPage = ({
                                 networks={networks}
                                 onAddErrorNotification={onAddErrorNotification} />,
         },
+        {
+            id: `${vmId(vm.name)}-hostdevs`,
+            className: "hostdevs-card",
+            title: _("Host devices"),
+            body: <VmHostDevCard vm={vm} nodeDevices={nodeDevices} config={config} onAddErrorNotification={onAddErrorNotification} />,
+        },
```
KKoukiou, line 154: onAddErrorNotification is not used.

KKoukiou (changes requested):

KKoukiou:
@skobyda would be great if you could add one pixel test :) I can show you how!

skobyda:
> @skobyda I am using the first first examples from https://libvirt.org/formatdomain.html#host-device-assignment and I get:
> 
> Screen Shot 2021-06-12 at 10 15 52
> 
> Why are most of the details missing?

The Source not being shown for pci was indeed a bug.
For others, they are missing because those details are not present anywhere (neither vm xml).

Screenshot from 2021-06-21 14-31-27

src/libvirt-dbus.js
```diff
@@ -599,13 +602,31 @@ const LIBVIRT_DBUS_PROVIDER = {
...
-                        const deviceXmlObject = parseNodeDeviceDumpxml(deviceXml);
+                        deviceXmlObject = parseNodeDeviceDumpxml(deviceXml);
                         deviceXmlObject.connectionName = connectionName;
 
-                        dispatch(updateOrAddNodeDevice(deviceXmlObject));
+                        if (deviceXmlObject.path && ["pci", "usb_device"].includes(deviceXmlObject.capability.type)) {
+                            return cockpit.spawn(["udevadm", "info", "--path", deviceXmlObject.path], { err: "message" })
+                                    .then(output => {
+                                        const nodeDev = parseUdevDB(output);
+                                        if (nodeDev.SUBSYSTEM === "pci" && nodeDev) {
```
KKoukiou, line 616: if nodeDev does not exist the first will crash. Same for the else part.

KKoukiou (changes requested):

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -0,0 +1,216 @@
...
+        },
+    ];
+
+    let hostdevId = 1;
+
+    const columnTitles = detailMap.map(target => target.name);
+    const rows = vm.hostDevices.sort((a, b) => a.type < b.type && -1).map(target => {
+        const columns = detailMap.map(d => {
+            let column = null;
+            if (typeof d.value === 'function') {
```
KKoukiou, line 193: There is nothing else than function here. You can remove the check and add const column = ...

KKoukiou (changes requested):

src/libvirt-dbus.js
skobyda: parseUdevDB always returns an empty object if it cannot find anything, `nodeDev.SUBSYSTEM` would not lead to a crash, just to an `undefined === "pic"`

src/components/vm/hostdevs/hostDevCard.jsx
skobyda: Fixed

KKoukiou (approved):

PR #177, opened 2021-05-20
test: Add workaround for RHEL 8.5 libvirt/qemu firmware regression

martinpitt:
This works around "internal error: unknown feature amd-sev-es" until the
proper fix lands, and thus unblocks image refreshes.

----

See https://bugzilla.redhat.com/show_bug.cgi?id=1961562#c16

Triggering extra run against https://github.com/cockpit-project/bots/pull/2020

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -59,3 +59,9 @@ printf "[WebService]\\nAllowUnencrypted=true\\n" > /etc/cockpit/cockpit.conf
 if type firewall-cmd >/dev/null 2>&1; then
     firewall-cmd --add-service=cockpit --permanent
 fi
+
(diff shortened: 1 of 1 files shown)


martinpitt:
That does not [quite fix the new image yet](https://logs.cockpit-project.org/logs/pull-177-20210520-121038-544f75a7-rhel-8-5-bots-2020/log.html), the VM somehow got messed up. I'll retry once to compare. But indeed that workaround seems to generally work.

martinpitt:
Retry is similarly messy, so this isn't good enough yet.

martinpitt:
We won't be needing this any more, the refresh in https://github.com/cockpit-project/bots/pull/2085 includes the fix.

PR #180, opened 2021-05-24
test: add some checks about boot order list details

yunmingyang:


test/check-machines-settings
```diff
@@ -272,6 +272,11 @@ class TestMachinesSettings(VirtualMachinesCase):
         # Open dialog
         b.click("#vm-subVmTest1-boot-order button")
         b.wait_visible("#vm-subVmTest1-order-modal-window")
+        # Check some details of boot options
+        b.wait_in_text("#vm-subVmTest1-order-modal-device-row-0 div.list-view-pf-main-info span",
```
KKoukiou, line 276: Can you add in the comment what kind of 'details' are these? That or add an id to the JS code that makes clear what that selector represents and replace the selector here with the id.

KKoukiou (changes requested):

test/check-machines-settings
yunmingyang: OK, I add comment there, and I will sperate

KKoukiou (approved):

martinpitt:
@KKoukiou : The failed [pixel tests](https://logs.cockpit-project.org/logs/pull-180-20210607-145438-40fabac1-fedora-34/log.html#7-2) seem old and the ones that are recently fixed, right? I'll retry that one.

PR #182, opened 2021-05-25
Replace Redux library with the useReducer Hook

KKoukiou:


KKoukiou:
Doh the new dispatch does not return a promise

KKoukiou:
This is not as straightforward as I thought, will reopen the PR when I have something more concrete.

PR #183, opened 2021-05-26
fmf: Drop pillow test dependency

martinpitt:
It has been uninstallable in RHEL for a while, will eventually disappear
[1], and we don't really need pixel tests in downstream gating.

[1] https://github.com/minimization/content-resolver-input/blob/master/configs/sst_cs_apps-unwanted-python3.yaml

### DIFF
--- a/test/browser/main.fmf
+++ b/test/browser/main.fmf
@@ -11,7 +11,6 @@ require:
   - make
   - npm
   - python3
-  - python3-pillow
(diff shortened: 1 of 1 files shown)


mvollmer:
> and we don't really need pixel tests in downstream gating.

Also, we couldn't look at the failures in a convenient way.

mvollmer (approved):

PR #186, opened 2021-05-27
Fix editing the MAC address of NICs

### LINKED ISSUE #185: [Machines] VM’s Network interface disappeared after editing the MAC address to a wrong character
Cockpit version: 
cockpit-244-1.fc33.x86_64
cockpit-machines-244-1.fc33.noarch
OS: Fedora 33
Page: Virtual Machines -- Network interface
[Machines] VM’s Network interface disappeared after editing the MAC address to a wrong character 

Steps to reproduce

1. Prepare a running vm
2.Enter the vm details page, click edit after the network interface, change the mac address to a wrong character like:
52:54:00:29:af:*&
3. Click save
There will be an error which is expected：
Danger alert:Network interface settings could not be saved
Error message: XML error: unable to parse mac address '52:54:00:29:af:*&'

But at the same time, the network interface disappeared on the page, which need to be fixed.
screenshot-10 66 92 39_9090-2021 05 27-14_40_40

KKoukiou:
If attaching the NIC after changing the MAC address failed we previously
ended up with the original NIC being removed as well.

Let's fix it in the following way:

* First attach the new interface (new MAC address) and only if the
operation is successful remove the obsoleted NIC
* In order that the attachment and removal of the obsoleted NIC is
guaranteed let's allow this operation for offline VMs only.
Hotplugging and hot-unplugging NICs is not guaranteed to succeed,
as it relies on the OS being responsive, enough PCI slots being
available etc.

Fixes #185

### DIFF
--- a/src/components/vm/nics/nicEdit.jsx
+++ b/src/components/vm/nics/nicEdit.jsx
@@ -19,7 +19,7 @@
 import React from 'react';
 import cockpit from 'cockpit';
 import PropTypes from 'prop-types';
-import { Button, Alert, Form, FormGroup, Modal, TextInput } from '@patternfly/react-core';
+import { Button, Alert, Form, FormGroup, Modal, TextInput, Tooltip } from '@patternfly/react-core';
 
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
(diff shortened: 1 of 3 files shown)


skobyda (approved):
Thanks. LGTM

PR #187, opened 2021-06-01
[no-test] Update translations from Fedora Weblate

cockpituous:


martinpitt:
@mvollmer , @KKoukiou : The pixel tests [failed](https://logs.cockpit-project.org/logs/pull-187-20210601-203022-5808beb4-fedora-34/log.html#7-2) because the "Memory" input now seems to have a different color in its focus frame. The translation update certainly did not change this, and other current PRs (#180 and #142) have the same problem. I wonder, did that break by itself through some `devDependencies` update? That feels like the kind of change which ought to be a runtime dependency, like PF..

KKoukiou:
> @mvollmer , @KKoukiou : The pixel tests [failed](https://logs.cockpit-project.org/logs/pull-187-20210601-203022-5808beb4-fedora-34/log.html#7-2) because the "Memory" input now seems to have a different color in its focus frame. The translation update certainly did not change this, and other current PRs (#180 and #142) have the same problem. I wonder, did that break by itself through some `devDependencies` update? That feels like the kind of change which ought to be a runtime dependency, like PF..

https://github.com/cockpit-project/cockpit-machines/pull/190

martinpitt (approved):

PR #190, opened 2021-06-03
Update reference pixels for changes to :focus-visible rendering

mvollmer:
Patternfly controls the appearance of the `:focus` pseudo class, but not `:focus-visible`. The Browser has some defaults for `:focus-visible` which seem to have changed.


mvollmer:
Here are some screenshots of a text input in various states, to try to sort this out.

Nothing
image

`:hover`
image

`:focus`
image

`:focus-visible`
image (Firefox)
image (Chrome)

`:focus` and `:focus-visible`
image (Firefox)
image (Chrome)

mvollmer:
Oky, I think I understand now what `:focus-visible` is about. It's a fix for the annoying dashed-line focus ring that was left behind after clicking on a button with the mouse.

Now, when clicking on a button with the mouse, the element gets :focus but not :focus-visible. When the user tabs to a button, it gets both :focus and :focus-visible.  This way one can write the CSS so that the focus ring only appears when the user has brought the button into focus with the keyboard. [I would have made it so that clicking a button would not give it focus at all, but there might be compatibility concerns that make this change impossible.]

This works nicely for us with Patternfly, for buttons.

For a text input, the rules seem to be such that it always gets both :focus and :focus-visible no matter how it got focused (mouse or keyboard). [This doesn't immediately make sense to me, I would treat inputs the same as buttons, but again, what do I know.]

I think the extra ring produced by :focus-visible is not necessary for text inputs. Patternfly should probably override it.

KKoukiou (approved):
It's not clear to me why the initial screenshots had yellow ring, but this change still makes sense.

mvollmer:
I have asked for clarification from the Patternflies: https://github.com/patternfly/patternfly/issues/4121

garrett:
Firefox screenshot is wrong; it has the focus ring clipped. I fixed that via z-index and it landed in PF a while back. I think this is related to using an older version of PatternFly and not updating yet.

In other words, I'm letting you know that this version of PF is wrong and it will will definitely change soon whenever we upgrade PF.

mvollmer:
> Firefox screenshot is wrong; it has the focus ring clipped.

Which one exactly? The one with just `:focus-visible`?

That state will never happen, as I have learned: a element can only have `:focus-visible` when it also has `:focus`. So I don't think we need to worry about that.

garrett:
Right, `:focus-visible`.

See https://github.com/patternfly/patternfly/pull/3991 which should be in PF >= 4.99.1... and [it applies to :focus and :focus-within](https://github.com/patternfly/patternfly/pull/3991/files#diff-d322235a2efa27dfb9c6f5bde685bdd84fcc6774014393ccea44e1c148d54c0cR36-R39).

PR #191, opened 2021-06-03
Port bootorder dialog to PF4

KKoukiou:
Make use of Draggable list.

See https://github.com/patternfly/patternfly/issues/3712

Blocked on:
* [x] https://github.com/patternfly/patternfly/issues/4138
* [x] https://github.com/patternfly/patternfly-react/issues/5900
* [ ] https://github.com/patternfly/patternfly-react/issues/6317

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,8 +54,8 @@
   "dependencies": {
     "@patternfly/patternfly": "4.135.2",
     "@patternfly/react-console": "4.13.5",
-    "@patternfly/react-core": "4.157.3",
-    "@patternfly/react-styles": "4.11.16",
+    "@patternfly/react-core": "4.160.3",
+    "@patternfly/react-styles": "4.12.2",
     "@patternfly/react-table": "4.31.0",
     "date-fns": "2.24.0",
     "patternfly": "3.59.5",
--- a/src/components/vm/overview/bootOrder.css
+++ b/src/components/vm/overview/bootOrder.css
@@ -1,64 +1,20 @@
-.boot-order .list-group {
-    --machines-boot-list-offset: 15rem;
-    max-height: calc(100vh - var(--machines-boot-list-offset));
+.boot-order-modal-cell {
+    display: flex;
 }
 
-.boot-order-list-view {
-    margin-top: 0;
-    margin-bottom: 0;
-}
-
-/* Overriding Patternfly's ListView */
-.boot-order .list-view-pf-body {
-    align-items: baseline;
-    min-height: 26px;
-}
-
-.boot-order .list-group-item {
-    padding: 1rem 1.5rem;
-}
-
-.boot-order .list-view-pf-main-info,
-.boot-order .list-view-pf-checkbox,
-.boot-order .list-view-pf-actions {
-    align-self: baseline;
-    margin-top: 0;
-    margin-bottom: 0;
-    padding-top: 0;
-    padding-bottom: 0;
-}
-
-.boot-order .list-view-pf-checkbox {
-    position: relative;
-    top: 3px;
-}
-
-.boot-order .list-view-pf-description {
+.boot-order-description {
     flex: 1 0 30%;
     flex-basis: 30%;
+    font-weight: var(--pf-global--FontWeight--bold);
 }
 
-.boot-order .list-view-pf-additional-info {
+.boot-order-additional-info {
     flex-basis: 70%;
 }
 
-.boot-order .list-group-item:not(.is-checked) .list-view-pf-actions {
-    display: none;
-}
-
-.boot-order .ct-form + .ct-form {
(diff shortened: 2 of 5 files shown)


mvollmer:
This PR seems to have accidentally reverted the changes to the reference pixels made in #190.

mvollmer (dismissed):
The changes to test/reference should be removed from this PR.

garrett (changes requested):
I like the idea in general, but this is currently unusable.

https://user-images.githubusercontent.com/10246/121147678-60f48880-c841-11eb-8369-fae17c78a25e.mp4

When dragging rows, over anything more than 1 row high, there's an obscene amount of flicker as the component does not know which row should be above.

I recorded this video in 30 FPS; I probably should've recorded it at 60 or 120 FPS, but even at a lower framerate it captures the issue well enough. (It flickers even more than you see here, which is already a nightmare.)

This is not just bad from a UX perspective; it can also trigger motion sickness, migraines, and seizures.

KKoukiou:
> I like the idea in general, but this is currently unusable.
> drag-flicker.mp4
> 
> When dragging rows, over anything more than 1 row high, there's an obscene amount of flicker as the component does not know which row should be above.
> 
> I recorded this video in 30 FPS; I probably should've recorded it at 60 or 120 FPS, but even at a lower framerate it captures the issue well enough. (It flickers even more than you see here, which is already a nightmare.)
> 
> This is not just bad from a UX perspective; it can also trigger motion sickness, migraines, and seizures.

I see what you mean. I can't see the problem upstream so it's something wrong with this specific PR. I will try to identify.

garrett:
> it's something wrong with this specific PR

It seems to happen with the tall lines. I wonder if there's some kind of positioning causing them to be calculated differently for the height? Or if the JS is hacky and makes assumptions that break down?

KKoukiou:
> > it's something wrong with this specific PR
> 
> It seems to happen with the tall lines. I wonder if there's some kind of positioning causing them to be calculated differently for the height? Or if the JS is hacky and makes assumptions that break down?

Hmm the JS is stolen from the website. Same code. So I lean towards that it's something with the content of our rows.

garrett:
If I make the height the same for each row, then it works fine. There's something weird with different sized items. I wonder if this happens when dragging a larger item versus smaller and same sized ones, or if it's only a problem with smaller versus larger?

garrett:
I tested this in-browser just now and found that dragging larger items is fine. So it's when you drag a smaller item around with larger items present. (Others being same or smaller is fine.)

KKoukiou:
> I tested this in-browser just now and found that dragging larger items is fine. So it's when you drag a smaller item around with larger items present. (Others being same or smaller is fine.)

Oh thanks for identidying this. It would  have taken me some hours I bet. Shall we set fixed height?

KKoukiou:
@garrett I don't tihnk inlining the information can be an option here. If you look at the code, https://github.com/cockpit-project/cockpit-machines/blob/fe945e3aa14eff9ba5646818c9090022f170fd10/src/components/vm/overview/bootOrder.jsx#L99 some devices can have really more informtion that three items, and for some these info are not obvious at all, so that we hide the label.

garrett:
I've modified the PF draggable list component example to make one element larger than the others, then tried dragging and hit the same issue we're having here. So this bug is an upstream one.

KKoukiou:
@garrett created upstream issue: https://github.com/patternfly/patternfly/issues/4138

KKoukiou:
This would also be a nice alternative component for this feature: https://github.com/patternfly/patternfly-react/issues/6265

(Since the draggable data list is broken)

garrett:
We'd need to either drop the additional information or have it under each item. Since there's drag and drop here too, it might also suffer from the same issue with variable heights.

I guess the left side would be "Available disks" and the right side would be "Boot disks"? And we'd use a simplified version without filtering? Or the labels could be "Excluded disks" and "Boot disks", which is a bit more contextual and explains that the disks are not used for booting?

KKoukiou:
> We'd need to either drop the additional information or have it under each item. Since there's drag and drop here too, it might also suffer from the same issue with variable heights.
> 
> I guess the left side would be "Available disks" and the right side would be "Boot disks"? And we'd use a simplified version without filtering? Or the labels could be "Excluded disks" and "Boot disks", which is a bit more contextual and explains that the disks are not used for booting?

Generally I agree, just s/disks/devices/ as not only disks are used for booting from but also nics.

garrett:
> Generally I agree, just s/disks/devices/ as not only disks are used for booting from but also nics.

Good point. You're right.

"Excluded devices" "Boot devices", I guess?

KKoukiou:
In favor of https://github.com/cockpit-project/cockpit-machines/pull/512

PR #192, opened 2021-06-04
Fix localization on VncConsole

serboupal:
Small fix to localize some strings in VncConsole tab.


KKoukiou (approved):
Thanks for your contribution!

martinpitt:
I approved the build-dist workflow, this was the reason for the packit build failure. Nevertheless, packit is down now due to the Westford power outage, but at least the srpm build should work.

For the same reason the RHEL tests can't run right now, but for this PR we can ignore them, I think.

PR #197, opened 2021-06-09
Refactor libvirt-dbus.js

skobyda:
- Currently, some APIs can be imported from libvirt-dbus.js directly and some have to use wrappers in actions/provider-actions.js. Unify this approach by dropping actions/provider-actions.js and exporting all methods directly from libvirt-dbus.js or libvirt-common.js
- Do not import store.dispatch at the src/index.js and then pass it all the way down to libvirt-dbus.js  libvirt-common.js methods. Instead, import store.dispatch directly to libvirt-dbus.js and libvirt-common.js.
- Remove redundant dispatches and use dispatch only when updating redux store directly


martinpitt:
packit tests fail a lot on 

    RuntimeError: TypeError: Cannot read property 'indexOf' of undefined

KKoukiou:
Simon you need:
``` diff
diff --git src/components/libvirtSlate.jsx src/components/libvirtSlate.jsx
index 03a08e1..820f850 100644
--- src/components/libvirtSlate.jsx
+++ src/components/libvirtSlate.jsx
@@ -57,7 +57,7 @@ class LibvirtSlate extends React.Component {
     checkStatus() {
         const service = this.props.libvirtService;
 
-        checkLibvirtStatus(service.name);
+        checkLibvirtStatus({ serviceName: service.name });
     }
 
     startService() {

```

KKoukiou:
@skobyda still a few failures ;(

KKoukiou:
@skobyda 

``` diff
        [kkoukiou@sequioa cockpit-machines]$ git diff
        diff --git src/libvirt-common.js src/libvirt-common.js
        index 5e41d36a..2095ab4b 100644
        --- src/libvirt-common.js
        +++ src/libvirt-common.js
        @@ -57,7 +57,7 @@ export function buildConsoleVVFile(consoleDetail) {
         }
         
         export function buildScriptTimeoutFailHandler(handler, delay) {
        -    return () => window.setTimeout(handler, delay);
        +    return window.setTimeout(handler, delay);
         }
         
         export function canLoggedUserConnectSession (connectionName, loggedUser) {

```

skobyda:
> @skobyda
> 
> ```diff
> [kkoukiou@sequioa cockpit-machines]$ git diff
> diff --git src/libvirt-common.js src/libvirt-common.js
> index 5e41d36a..2095ab4b 100644
> --- src/libvirt-common.js
> +++ src/libvirt-common.js
> @@ -57,7 +57,7 @@ export function buildConsoleVVFile(consoleDetail) {
>  }
>  
>  export function buildScriptTimeoutFailHandler(handler, delay) {
> -    return () => window.setTimeout(handler, delay);
> +    return window.setTimeout(handler, delay);
>  }
>  
>  export function canLoggedUserConnectSession (connectionName, loggedUser) {
> ```

With that fix, it's a bit strange. The test is still failing, but now the error message is present on different page. It should be here:
Screenshot from 2021-06-24 16-33-21

But instead it's here:
Screenshot from 2021-06-24 16-33-02

KKoukiou (approved):
Thank you!!

PR #88, opened 2021-04-09
Move to webpack 5

martinpitt:
Cherry-picked from starter-kit commit c52136c0514.

----

 - [x] Builds on top of PR #87 
 - [x] Consider if the Table warnings are important enough to block: https://github.com/patternfly/patternfly-react/issues/5644
 - [x] Fix/work around build failures with `ip` npm package
 
This currently spits out 23 warnings of this form:
```
WARNING in ./node_modules/patternfly-react/dist/esm/components/Table/Table.js 21:0-31
export 'actionHeaderCellFormatter' (imported as 'Table') was not found in 'reactabular-table' (possible exports: Body, BodyRow, Header, Provider, __esModule, columnsAreEqual, evaluateFormatters, evaluateTransforms, mergeProps, resolveRowKey)
```

This feels like a PatternFly bug. It still seems to work anyway, let's see what CI says.

PR #198, opened 2021-06-09
tests: When creating the test VMs use VNC console as default graphics

KKoukiou:
RHEL-9 does not support spice so spice graphics which we used before
as default are not an options for the RHEL 9 test images

Ideally, we would let virt-install pick up the right graphics for us by not specifying it
at all. However virt-install also chooses spice for the RHEL-9-0 image
which is obviously wrong.

See https://bugzilla.redhat.com/show_bug.cgi?id=1969952


martinpitt:
These failures on [rawhide](http://artifacts.dev.testing-farm.io/9fa2505d-ac76-463e-bea9-18fa9a243ec3/) look worrying, but they look unrelated.
```
# testSnapshots (__main__.TestMachinesSnapshots)
error: Failed to start network default
error: Requested operation is not valid: network is already active
error: Failed to start domain 'subVmTest1'
error: can't connect to virtlogd: Unable to open system token /run/libvirt/common/system.token: Permission denied
```
Test LGTM, but I'd still like to trigger the full set of tests, just in case.

martinpitt:
I already retried, but on [ubuntu-2004](https://logs.cockpit-project.org/logs/pull-198-20210610-053656-efe25769-ubuntu-2004/log.html#20)  this reliably causes tests to fail with an unexpected "couldn't shutdown fd: Transport endpoint is not connected" message.

martinpitt:
The [journal](https://logs.cockpit-project.org/logs/pull-198-20210610-053656-efe25769-ubuntu-2004/TestMachinesLifecycle-testBasic-ubuntu-2004-127.0.0.2-2301-FAIL.log.gz) has a little better message:

    cockpit-bridge[6914]: 127.0.0.1:5900: couldn't shutdown fd: Transport endpoint is not connected

@KKoukiou : Port 5900 is spice, isn't it? Could it be that this  change now tries to enable spice on ubuntu-2004, and it's not yet supported there?

martinpitt:
@KKoukiou , @mvollmer : Argh, [pixel test failures](https://logs.cockpit-project.org/logs/pull-198-20210610-083929-c91259f2-fedora-34/log.html#2-2) again. Katerina's branch is 4 commits behind origin:main, but nothing there should touch pixels (and I expect that this gets rebased anyway). Any idea about that one?

KKoukiou:
> @KKoukiou , @mvollmer : Argh, [pixel test failures](https://logs.cockpit-project.org/logs/pull-198-20210610-083929-c91259f2-fedora-34/log.html#2-2) again. Katerina's branch is 4 commits behind origin:main, but nothing there should touch pixels (and I expect that this gets rebased anyway). Any idea about that one?

Looks like a race condition - the OS input field is disabled when the OS list is being loaded from osinfo db. That's the greyed out screenshot. I suspect it's really a timing issue.

martinpitt:
Ah, so I suppose we need to wait for the field to not be disabled before we do the screenshot? Thanks! (Retrying, as that's blocking other stuff, and the race isn't new)

KKoukiou:
> Ah, so I suppose we need to wait for the field to not be disabled before we do the screenshot? Thanks! (Retrying, as that's blocking other stuff, and the race isn't new)

Correct I have a commit about it here: https://github.com/cockpit-project/cockpit-machines/pull/88/commits/64025d7b81d29e1e305f65719040837325524e69

I will send it in seperate pR - did not realize it's so often breaking.

martinpitt:
I failed again, so I stole your fix from https://github.com/cockpit-project/cockpit-machines/commit/64025d7b81d29e1e305f65719040837325524e69 and added it here. Let's see how it fares!

martinpitt:
Dang, I missed the separate PR #202, and that is failing. Let's sort it out in #202, I unpushed the separate commit here.

KKoukiou:
> Dang, I missed the separate PR #202, and that is failing. Let's sort it out in #202, I unpushed the separate commit here.

Fixed and merged.

martinpitt (approved):
Cheers! I'm glad it went green at the third attempt or so :)

PR #199, opened 2021-06-09
Fix comparison when building the available disk bus type list

KKoukiou:
Previously we were comparing the current disk bus type value (string)
with the list of objects, therefore the current disk bus type was always
added as disabled option in the end of the list.

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1948366


marusak (approved):
Thanks

martinpitt (approved):
Oops! Thanks for fixing!

PR #204, opened 2021-06-16
tests: add check that 'Add shared filesystem' is not available for session connection guests

KKoukiou:
createVm helper function had to be extended to support session connection VMs.

### DIFF
--- a/test/check-machines-filesystems
+++ b/test/check-machines-filesystems
@@ -31,18 +31,23 @@ from testlib import skipImage, nondestructive, test_main  # noqa
 
 @nondestructive
 class TestMachinesFilesystems(VirtualMachinesCase):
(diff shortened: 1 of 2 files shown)


martinpitt:
[this test failure](https://logs.cockpit-project.org/logs/pull-204-20210616-125714-80158cce-ubuntu-2004/log.html#60) smells like some test helper is writing into the static file name `/tmp/xml` without cleaning it up -- root runs first, then session can't overwrite the existing root-owned file

martinpitt (commented):
Nice! Packit failures are because this branch is out of date. Do you mind pushing a rebase? This makes structural changes to machineslib.py, let's please validate that they work in FMF as well.

martinpitt (approved):
Danke!

PR #206, opened 2021-06-17
index.html: Add missing opening `<html>` tag

oakaigh:
The `html` tag needs to be closed too.


martinpitt (approved):
Thank you! The fix is right, but the commit message needs to be fixed. I suggest "index.html: Add missing opening `<html>` tag".

KKoukiou (approved):

PR #210, opened 2021-06-21
Fix reducers updating interfaces and node devices

KKoukiou:
Previously the resources would be updated to an undefined object,
because of the misspelled named parameter, resulting in JS exceptions.

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1974228

### DIFF
--- a/src/reducers.js
+++ b/src/reducers.js
@@ -109,7 +109,7 @@ function interfaces(state, action) {
         }
 
         const updatedIface = Object.assign({}, state[index], iface);
-        return replaceResource({ state, updatedIface, index });
+        return replaceResource({ state, updatedResource: updatedIface, index });
     }
     default:
         return state;
@@ -174,7 +174,7 @@ function nodeDevices(state, action) {
         }
 
         const updatedNodedev = Object.assign({}, state[index], nodedev);
-        return replaceResource({ state, updatedNodedev, index });
+        return replaceResource({ state, updatedResource: updatedNodedev, index });
     }
     default:
         return state;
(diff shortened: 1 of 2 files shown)


KKoukiou:
ah I want to add a test as well

src/reducers.js
```diff
@@ -109,7 +109,7 @@ function interfaces(state, action) {
         }
 
         const updatedIface = Object.assign({}, state[index], iface);
-        return replaceResource({ state, updatedIface, index });
+        return replaceResource({ state, updatedResource: updatedIface, index });
```
skobyda, line 112: How does this exactly affect a bug which only happens after shutting down the libvirtd.service? Are these function not also called initially when the page is first loaded? Just trying to understand the relation to the bug.

KKoukiou: They are also called when you stop libvirtd just with delay polling:
https://github.com/cockpit-project/cockpit-machines/blob/main/src/libvirt-dbus.js#L1138

skobyda: Oh I see, so then they would be replaced by incorrect object/list?
The code itself seems good, so giving.

KKoukiou: Correct, that's what happened resulting in a JS exception.

skobyda (approved):
Thanks for fix

KKoukiou:
centos failed with unrelated: FileNotFoundError: [Errno 2] No such file or directory: '/work/bots/make-checkout-workdir/test/common/pixeldiff.html'

PR #211, opened 2021-06-21
create-vm-dialog:  Don't use default nodeMaxMemory

KKoukiou:


skobyda (changes requested):
I believe this introduces a bug which I fixed some time ago with Math.Floor(). The reason why there was a Math.Floor, which rounds number **down**, was that the number of available memory is never larger by a rounding error.

Imagine you have 15.4603516GiB of free space. The `toReadableNumber()` would round it **up** to 15.5GiB. If then user actually inputs 15.5GiB, they would assign more memory than is actually available.

KKoukiou:
> I believe this introduces a bug which I fixed some time ago with Math.Floor(). The reason why there was a Math.Floor, which rounds number **down**, was that the number of available memory is never larger by a rounding error.
> 
> Imagine you have 15.4603516GiB of free space. The `toReadableNumber()` would round it **up** to 15.5GiB. If then user actually inputs 15.5GiB, they would assign more memory than is actually available.

toReadableNumber will not round it but trim everything after 4. So it will be 15.4.

skobyda:
> toReadableNumber will not round it but trim everything after 4. So it will be 15.4.

Are you sure? I just copied the function to nodeJs, tried to run it with `toReadableNumber(15.4603546)` and it returned string "15.5".

skobyda:
> > I believe this introduces a bug which I fixed some time ago with Math.Floor(). The reason why there was a Math.Floor, which rounds number **down**, was that the number of available memory is never larger by a rounding error.
> > Imagine you have 15.4603516GiB of free space. The `toReadableNumber()` would round it **up** to 15.5GiB. If then user actually inputs 15.5GiB, they would assign more memory than is actually available.
> 
> toReadableNumber will not round it but trim everything after 4. So it will be 15.4.

The function is basically just a wrapper for .toFixed(), which seems to round numbers up/down. You can try by running: 15.4603516.toFixed(1), it will round it up to 15.5

src/helpers.js
```diff
@@ -30,9 +30,9 @@ export function dummyVmsFilter(vms, uiVms) {
 
 export function toReadableNumber(number) {
     if (number < 1) {
-        return number.toFixed(2);
+        return Math.floor(number * 100) / 100;
     } else {
-        const fixed1 = number.toFixed(1);
+        const fixed1 = Math.floor((number * 10 ) / 10);
```
skobyda, line 35: Breaks make check:

```suggestion
        const fixed1 = Math.floor((number * 10) / 10);
```

skobyda (changes requested):

src/helpers.js
KKoukiou: Oops

src/helpers.js
```diff
@@ -30,9 +30,9 @@ export function dummyVmsFilter(vms, uiVms) {
 
 export function toReadableNumber(number) {
     if (number < 1) {
-        return number.toFixed(2);
+        return Math.floor(number * 100) / 100;
     } else {
-        const fixed1 = number.toFixed(1);
+        const fixed1 = Math.floor((number * 10) / 10);
         return (number - fixed1 === 0) ? number.toFixed(0) : fixed1;
```
skobyda, line 36: I believe this has to be also adjusted?
```suggestion
        return (number - fixed1 === 0) ? Math.floor(number) : fixed1;
```

skobyda (approved):
Thanks, LGTM

PR #212, opened 2021-06-22
Fix overflow of long description with single word in snapshots list

KKoukiou:


mvollmer (approved):
Approved, including the snapshost typo!

PR #213, opened 2021-06-22
fix validation of snapshot create dialog

KKoukiou:


skobyda:
2 test failures which seem related to snapshots.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -114,6 +112,8 @@ export class CreateSnapshotModal extends React.Component {
                         this.setState({ inProgress: false });
                         this.dialogErrorSet(_("Snapshot failed to be created"), exc.message);
                     });
+        } else {
+            this.setState({ validationError });
         }
     }
```
skobyda, line 119: I would akeep validationError in state, because then when the error is shown, once user adjust the error in UI, the error is immediately hidden. 
With current approach, even after user corrects incorrect snapshot name, the error message is still present until they click on "Create" again, which could make them confused.

I would recommend the approach which is done in VM creation dialog: https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/create-vm-dialog/createVmDialog.jsx#L956

skobyda (changes requested):

skobyda (approved):
Thanks, looks great

PR #214, opened 2021-06-22
Delete VM snapshots before deleting the VM to remove unusable overlays from image

KKoukiou:
Consider the following scenario:

* The user deletes the VM (let's call it testVm) but not the disk image
(testVm.qcow2), which has snapshot overlays and he want to reuse the
aforementioned image
* The user imports testVm.qcow2 into a new VM which of course can't parse the
snapshots of the previous VM, however the overlays still exist on the
testVm.qcow2 image

Since these snapshots are of not use anymore, let's delete these when
the first domain is deleted.

Relevant to: https://bugzilla.redhat.com/show_bug.cgi?id=1974216


martinpitt (approved):
Looks nice, thanks!

PR #216, opened 2021-06-23
Reset disk edit modal state when closing it

KKoukiou:
This is achieved by keeping in a different component than the action and
recreating every time it's opened.

This commit also replaces `samp` with a `div` in disk Edit and network
Create dialogs. samp should only be used when quoting a program generated output,
and it's not the case in none of the two usages in this repository.

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1947241


martinpitt (approved):
Looks nice, thanks!

PR #218, opened 2021-06-24
tests: make sure package.json is ignored when installing test deps

KKoukiou:


martinpitt (approved):
Sure, let's try that. This will bite us if we ever need to actually pin down a specific version of these two modules, but let's hope "never" :grin:

PR #220, opened 2021-06-24
It's enough to run the fail handler once in the install VM operation

KKoukiou:
Previously the notification was appearing immediately, so we can safely
remove the setTimeout handler without affecting the current behavior.

This caused some confusion @ https://github.com/cockpit-project/cockpit-machines/pull/197 

Let's fix it seperately .


skobyda (approved):
Tests are passing, the fix seems good. Thank you

PR #221, opened 2021-06-25
Bump up pkg/lib

KKoukiou:
We can't just use tag 247 (latest) because the needed lib changes for
muting some sass related deprecation warnings are contained in
4e553647a81b6ec49e9fa13e2ed72d472c52965b.

This now indeed enables us to use dart sass.


allisonkarlitskaya (changes requested):
The content looks awesome.  The summary on the commit message is terrible :rofl:

allisonkarlitskaya (approved):
approving and rewriting the commit message via squash landing, as requested by @KKoukiou

PR #222, opened 2021-06-28
When creating VMs use the XML file virt-install prints for adding metadata

KKoukiou:
Previously we were using an XML we fetched once the virt-install
command finished. This was wrong since it was using the live XML
which can differ from the offline XML.

example: When the CPU type is chosen as 'host-model' libvirt will copy the host CPU
definition from capabilities XML into domain XML once the domain is started.
In that case the offline XML is different than the live XML.

This was reported by @yunmingyang, he plans to create a BZ for it. Also we need to add a test for this ideally as well.

### DIFF
--- a/src/scripts/create_machine.sh
+++ b/src/scripts/create_machine.sh
@@ -191,7 +191,7 @@ fi
 
 if [ "$START_VM" = "true" ]; then
     vmExists "$VM_NAME" || handleFailure $?
-    virsh -c "$CONNECTION_URI" -q dumpxml "$VM_NAME" > "$XMLS_FILE"
+    virsh -c "$CONNECTION_URI" -q dumpxml --inactive "$VM_NAME" > "$XMLS_FILE"
 fi
 
 # LAST STEP ONLY - virt-install can output 1 or 2 steps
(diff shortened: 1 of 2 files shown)


KKoukiou:
Converting to draft - I would like to add a test first.

src/scripts/create_machine.sh
```diff
@@ -191,7 +191,7 @@ fi
 
 if [ "$START_VM" = "true" ]; then
     vmExists "$VM_NAME" || handleFailure $?
-    virsh -c "$CONNECTION_URI" -q dumpxml "$VM_NAME" > "$XMLS_FILE"
+    virsh -c "$CONNECTION_URI" -q dumpxml --inactive "$VM_NAME" > "$XMLS_FILE"
```
martinpitt, line 194: Whee, I wasn't aware of `--inactive` either -- this totally makes sense (and it even feels like that would be a better default..)

martinpitt (approved):
Danke!

KKoukiou:
@martinpitt the packit rawhide failures are not overlapping between the first run and this current run. Are you fine with ignoring them, they seem to be rather timeouts because of slow env or similar (at least the current run)

martinpitt:
@KKoukiou : Yes, we've been ignoring them for a while already. We need to track them down (they also block downstream landing, and [waiving is broken](https://github.com/fedora-infra/bodhi/issues/4237)), but not here.

PR #224, opened 2021-06-28
Add test for deleting an interface of a running VM

KKoukiou:
This has regressed on RHEL-9- an unfortunately we did not have a test for this.

https://bugzilla.redhat.com/show_bug.cgi?id=1976945

Will add the naughty as well.

### DIFF
--- a/src/libvirt-dbus.js
+++ b/src/libvirt-dbus.js
@@ -1473,7 +1473,7 @@ export function detachIface(mac, connectionName, id, live, persistent) {
                 if (ifaceInactiveXML && persistent)
                     detachFlags |= Enum.VIR_DOMAIN_AFFECT_CONFIG;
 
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Nice, thanks! Due to the regression that will fail on current rhel-9-0 already, or "just" in the pending refresh? I.e. will this already need a naughty?

Good if this goes green.

KKoukiou:
> Nice, thanks! Due to the regression that will fail on current rhel-9-0 already, or "just" in the pending refresh? I.e. will this already need a naughty?
> 
> Good if this goes green.

This will break on https://github.com/cockpit-project/bots/pull/2144 which brings libvirt 7.4.0.

martinpitt (approved):
Thanks! I'm glad it was a bug fix on our side.

PR #226, opened 2021-06-29
Adding and removing VM's host devices

skobyda:
## Machines: Support adding and removing host devices

Host devices are physical devices of the host. This includes PCI (for example NICs, GPUs), and USB (for example mice, cameras, and keyboards) types. Users can now assign host USB and PCI devices to a VM and also detach them again. When a device is assigned to a VM it can on longer be used by the host.

Screen Shot 2021-09-24 at 09 14 34
Screen Shot 2021-09-24 at 09 13 48

### DIFF
--- a/src/components/vm/hostdevs/hostDevAdd.jsx
+++ b/src/components/vm/hostdevs/hostDevAdd.jsx
@@ -1,0 +1,238 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2021 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+import React, { useState, useEffect } from "react";
+import cockpit from "cockpit";
+import {
+    Table,
+    TableBody,
+    TableHeader,
+    TableVariant,
+} from '@patternfly/react-table';
+import {
+    Button,
+    DescriptionList,
+    Form,
+    FormGroup,
+    Modal,
+    Radio
+} from "@patternfly/react-core";
+
+import { ModalError } from "cockpit-components-inline-notification.jsx";
+import { domainAttachHostDevice, domainGet } from "../../../libvirtApi/domain.js";
+import { findHostNodeDevice } from "../../../helpers.js";
+import { getOptionalValue } from "./hostDevCard.jsx";
+import "./hostDevAdd.scss";
+
+const _ = cockpit.gettext;
+
+const TypeRow = ({ idPrefix, type, setType }) => {
+    return (
+        <FormGroup fieldId="usb_device"
+                   label={_("Type")}
+                   isInline
+                   hasNoPaddingTop>
+            <Radio id="usb_device"
+                   name="type"
+                   label={_("USB")}
+                   isChecked={type === "usb_device"}
+                   onChange={() => setType("usb_device")} />
+            <Radio id="pci"
+                   name="type"
+                   label={_("PCI")}
+                   isChecked={type === "pci"}
+                   onChange={() => setType("pci")} />
+        </FormGroup>
+    );
+};
+
+function devicesHaveAChild(selectableDevices) {
+    const all = {};
+
+    selectableDevices.forEach(item => {
+        all[item.name] = { ...item };
+        all[item.name].hasChildren = false;
+    });
+
+    Object.values(all).forEach(item => {
+        if (item.parent && item.parent !== "computer" && (item.capability.type === "usb_device" || item.capability.type === "pci")) {
+            all[item.parent].hasChildren = true;
+        }
+    });
+
+    return Object.values(all).sort(a => a.hasChildren ? 1 : -1);
+}
+
+const DevRow = ({ idPrefix, type, selectableDevices, setSelectableDevices }) => {
+    function getSource(nodeDev, id) {
+        const cells = [];
+        if (nodeDev.capability.type === "usb_device") {
+            const device = nodeDev.devnum;
+            const bus = nodeDev.busnum;
+
+            cells.push(getOptionalValue(device, `${id}-device`, _("Device")));
+            cells.push(getOptionalValue(bus, `${id}-bus`, _("Bus")));
+        } else if (nodeDev.capability.type === "pci") {
+            let domain = nodeDev.capability.domain;
+            let bus = nodeDev.capability.bus;
+            let slot = nodeDev.capability.slot;
+            let func = nodeDev.capability.function;
+
+            domain = domain.toString(16).padStart(4, '0');
+            bus = bus.toString(16).padStart(2, '0');
+            slot = slot.toString(16).padStart(2, '0');
+            func = func.toString(16).padStart(1, '0');
+
+            cells.push(getOptionalValue(`${domain}:${bus}:${slot}.${func}`, `${id}-slot`, _("Slot")));
+        }
+
+        return cells;
+    }
+
+    function onSelect(event, isSelected, rowId) {
+        let newDevs;
+        if (rowId === -1) {
+            newDevs = selectableDevices.map(oneRow => {
+                oneRow.selected = isSelected;
+                return oneRow;
+            });
+        } else {
+            newDevs = [...selectableDevices];
+            newDevs[rowId].selected = isSelected;
+        }
+
+        setSelectableDevices(newDevs);
+    }
+
+    return (
+        <FormGroup fieldId={`${idPrefix}-dev`} label={_("Device")} hasNoPaddingTop isInline>
+            <Table onSelect={onSelect}
+                   variant={TableVariant.compact}
+                   canSelectAll={false}
+                   className="vm-device-table"
+                   aria-label={_("Table of selectable host devices")}
+                   cells={[_("Product"), _("Vendor"), _("Location")]}
+                   rows={selectableDevices.map((dev, idx) => {
+                       return {
+                           selected: dev.selected, disableSelection: dev.nodeDev.hasChildren, cells: [
+                               dev.nodeDev.capability.product._value || "(" + _("Undefined") + ")",
+                               dev.nodeDev.capability.vendor._value,
+                               { title: <DescriptionList key='source' isHorizontal>{getSource(dev.nodeDev, idx)}</DescriptionList> }
+                           ]
+                       };
+                   })}>
+                <TableHeader />
+                <TableBody />
+            </Table>
+        </FormGroup>
+    );
+};
+
+function getSelectableDevices(nodeDevices, vm, type) {
+    const devicesNotAlreadyAttached = [];
+
+    const devicesWithCorrectType = nodeDevices.filter(dev => dev.capability && dev.capability.type === type);
+    devicesWithCorrectType.filter(dev => dev.capability && dev.capability.type === type).forEach(nodeDev => {
+        let deviceIsAlreadyAttached = false;
+
+        vm.hostDevices.forEach(hostDev => {
+            const foundNodeDevice = findHostNodeDevice(hostDev, nodeDevices);
+
+            if (foundNodeDevice && nodeDev.path === foundNodeDevice.path)
+                deviceIsAlreadyAttached = true;
+        });
+
+        if (!deviceIsAlreadyAttached)
+            devicesNotAlreadyAttached.push({ selected: false, nodeDev });
+    });
+
+    return devicesNotAlreadyAttached;
+}
+
+const AddHostDev = ({ idPrefix, vm, nodeDevices, close }) => {
+    const [type, setType] = useState("usb_device");
+    const [selectableDevices, setSelectableDevices] = useState([]);
+    const [dialogError, setDialogError] = useState("");
+    const [dialogErrorDetail, setDialogErrorDetail] = useState("");
+
+    const allDevices = devicesHaveAChild([...nodeDevices]);
+
+    useEffect(() => {
+        setSelectableDevices(getSelectableDevices(allDevices, vm, type));
+    }, []);
+
+    const setTypeWrapper = (newType) => {
+        setSelectableDevices(getSelectableDevices(allDevices, vm, newType));
+        setType(newType);
+    };
+
+    const attach = () => {
+        Promise.all(selectableDevices.map(device => {
+            if (device.selected)
+                return domainAttachHostDevice({ connectionName: vm.connectionName, vmName: vm.name, live: vm.state !== "shut off", dev: device.nodeDev });
+        }))
+                .then(() => {
+                    domainGet({ connectionName: vm.connectionName, id: vm.id });
+                    close();
+                })
+                .catch(exc => {
+                    setDialogError(_("Host device could not be attached"));
+                    setDialogErrorDetail(exc.message);
+                });
+    };
+
+    const body = (
+        <Form isHorizontal>
+            <TypeRow idPrefix={idPrefix} type={type} setType={setTypeWrapper} />
+            <DevRow idPrefix={idPrefix} type={type} selectableDevices={selectableDevices} setSelectableDevices={setSelectableDevices} />
+        </Form>
+    );
+
+    const footer = (
+        <>
+            {dialogError &&
+                <ModalError dialogError={dialogError}
+                            dialogErrorDetail={dialogErrorDetail} />}
+            <Button id={`${idPrefix}-attach`}
+                    variant="primary"
+                    onClick={attach}>
+                {_("Add")}
+            </Button>
+            <Button id={`${idPrefix}-cancel`}
+                    variant="link"
+                    className="btn-cancel"
+                    onClick={close}>
+                {_("Cancel")}
+            </Button>
+        </>
+    );
+
+    return (
+        <Modal position="top"
+               variant="medium"
+               id={`${idPrefix}-dialog`}
+               onClose={close}
+               title={_("Add host device")}
+               footer={footer}
+               isOpen>
+            {body}
+        </Modal>
+    );
+};
+
+export default AddHostDev;
--- a/src/components/vm/hostdevs/hostDevAdd.scss
+++ b/src/components/vm/hostdevs/hostDevAdd.scss
@@ -1,0 +1,12 @@
+@import "../../../lib/_global-variables.scss";
+
+.pf-c-table.vm-device-table {
+  /* Set overflow with a limiter on the list, so it scrolls instead of the dialog */
+  max-height: min(50vh, 50rem);
+  overflow: auto;
+
+  /* The list is used within a table; remove excess padding */
+  th {
+    --pf-c-table--m-compact-th--PaddingTop: 0;
+  }
+}
--- a/src/components/vm/hostdevs/hostDevCard.jsx
+++ b/src/components/vm/hostdevs/hostDevCard.jsx
@@ -34,24 +34,21 @@ const _ = cockpit.gettext;
 
 /* Adds an optional description-value pair to an array which represents multiple values of a table cell
  *
- * @param chunks an array into which element is added to
  * @param value a value of the descr-value pair
  * @param descr a description of the descr-value pair
  * @param id
  */
-function addOptionalToCell(chunks, value, id, descr) {
-    if (value) {
-        chunks.push(
-            <DescriptionListGroup key={descr}>
-                <DescriptionListTerm>
-                    {descr}
-                </DescriptionListTerm>
-                <DescriptionListDescription id={id}>
-                    {value}
-                </DescriptionListDescription>
-            </DescriptionListGroup>
-        );
-    }
+export function getOptionalValue(value, id, descr) {
+    return (
+        <DescriptionListGroup key={descr}>
+            <DescriptionListTerm>
+                {descr}
+            </DescriptionListTerm>
+            <DescriptionListDescription id={id}>
+                {value}
+            </DescriptionListDescription>
+        </DescriptionListGroup>
+    );
 }
 
 export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
@@ -85,14 +82,14 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
     }
 
     function getSource(hostDev, hostdevId) {
-        const cell = [];
+        const cells = [];
         const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
         if (hostDev.type === "usb" && nodeDev) {
             const device = nodeDev.devnum;
             const bus = nodeDev.busnum;
 
-            addOptionalToCell(cell, device, `${hostdevId}-device`, _("Device"));
-            addOptionalToCell(cell, bus, `${hostdevId}-bus`, _("Bus"));
+            cells.push(getOptionalValue(device, `${hostdevId}-device`, _("Device")));
+            cells.push(getOptionalValue(bus, `${hostdevId}-bus`, _("Bus")));
         } else if (hostDev.type === "pci") {
             let domain = hostDev.source.address.domain.split('x')[1];
             let bus = hostDev.source.address.bus.split('x')[1];
@@ -104,40 +101,40 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
             slot = String(slot).padStart(2, '0');
             func = String(func).padStart(1, '0');
 
-            addOptionalToCell(cell, `${domain}:${bus}:${slot}.${func}`, `${hostdevId}-slot`, _("Slot"));
+            cells.push(getOptionalValue(`${domain}:${bus}:${slot}.${func}`, `${hostdevId}-slot`, _("Slot")));
         } else if (hostDev.type === "scsi") {
             const bus = hostDev.source.address.bus;
             const target = hostDev.source.address.target;
             const unit = hostDev.source.address.lun;
 
-            addOptionalToCell(cell, bus, `${hostdevId}-bus`, _("Bus"));
-            addOptionalToCell(cell, unit, `${hostdevId}-unit`, _("Slot"));
-            addOptionalToCell(cell, target, `${hostdevId}-target`, _("Target"));
+            cells.push(getOptionalValue(bus, `${hostdevId}-bus`, _("Bus")));
+            cells.push(getOptionalValue(unit, `${hostdevId}-unit`, _("Slot")));
+            cells.push(getOptionalValue(target, `${hostdevId}-target`, _("Target")));
         } else if (hostDev.type === "scsi_host") {
             const protocol = hostDev.source.protocol;
             const wwpn = hostDev.source.wwpn;
 
-            addOptionalToCell(cell, protocol, `${hostdevId}-protocol`, _("Protocol"));
-            addOptionalToCell(cell, wwpn, `${hostdevId}-wwpn`, _("WWPN"));
+            cells.push(getOptionalValue(protocol, `${hostdevId}-protocol`, _("Protocol")));
+            cells.push(getOptionalValue(wwpn, `${hostdevId}-wwpn`, _("WWPN")));
         } else if (hostDev.type === "mdev") {
             const uuid = hostDev.source.address.uuid;
 
-            addOptionalToCell(cell, uuid, `${hostdevId}-uuid`, _("UUID"));
+            cells.push(getOptionalValue(uuid, `${hostdevId}-uuid`, _("UUID")));
         } else if (hostDev.type === "storage") {
             const block = hostDev.source.block;
 
-            addOptionalToCell(cell, block, `${hostdevId}-block`, _("Path"));
+            cells.push(getOptionalValue(block, `${hostdevId}-block`, _("Path")));
         } else if (hostDev.type === "misc") {
             const ch = hostDev.source.char;
 
-            addOptionalToCell(cell, ch, `${hostdevId}-char`, _("Path"));
+            cells.push(getOptionalValue(ch, `${hostdevId}-char`, _("Path")));
         } else if (hostDev.type === "net") {
             const iface = hostDev.source.interface;
 
-            addOptionalToCell(cell, iface, `${hostdevId}-interface`, _("Interface"));
+            cells.push(getOptionalValue(iface, `${hostdevId}-interface`, _("Interface")));
         }
 
-        return <DescriptionList isHorizontal>{cell}</DescriptionList>;
+        return <DescriptionList isHorizontal>{cells}</DescriptionList>;
     }
 
     const id = vmId(vm.name);
--- a/src/components/vm/hostdevs/hostDevCard.jsx
+++ b/src/components/vm/hostdevs/hostDevCard.jsx
@@ -16,10 +16,11 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 
 import {
+    Button,
     DescriptionList,
     DescriptionListTerm,
     DescriptionListGroup,
@@ -29,8 +30,29 @@ import {
 import cockpit from 'cockpit';
 import { vmId, findHostNodeDevice } from "../../../helpers.js";
 import { ListingTable } from "cockpit-components-table.jsx";
+import AddHostDev from "./hostDevAdd.jsx";
 
 const _ = cockpit.gettext;
+
+export const VmHostDevActions = ({ vm, nodeDevices }) => {
+    const [showAttachModal, setShowAttachModal] = useState(false);
+
+    const idPrefix = `${vmId(vm.name)}-hostdevs`;
+
+    return (
+        <>
+            <Button id={`${idPrefix}-add`} variant='secondary' onClick={() => setShowAttachModal(true)}>
+                {_("Add host device")}
+            </Button>
+            {showAttachModal &&
+                <AddHostDev close={() => setShowAttachModal(false)}
+                            idPrefix={idPrefix}
+                            vm={vm}
+                            nodeDevices={nodeDevices} />
+            }
+        </>
+    );
+};
 
 /* Adds an optional description-value pair to an array which represents multiple values of a table cell
  *
--- a/src/components/vm/hostdevs/hostDevCard.jsx
+++ b/src/components/vm/hostdevs/hostDevCard.jsx
@@ -28,9 +28,11 @@ import {
 } from '@patternfly/react-core';
 
 import cockpit from 'cockpit';
-import { vmId, findHostNodeDevice } from "../../../helpers.js";
+import { vmId, findHostNodeDevice, getNodeDevSource } from "../../../helpers.js";
 import { ListingTable } from "cockpit-components-table.jsx";
 import AddHostDev from "./hostDevAdd.jsx";
+import { domainDetachHostDevice } from '../../../libvirtApi/domain.js';
+import { DeleteResourceButton, DeleteResourceModal } from '../../common/deleteResource.jsx';
 
 const _ = cockpit.gettext;
 
@@ -74,6 +76,8 @@ export function getOptionalValue(value, id, descr) {
 }
 
 export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
+    const [deleteDialogProps, setDeleteDialogProps] = useState(undefined);
+
     function getClass(hostDev, hostdevId) {
         const nodeDev = findHostNodeDevice(hostDev, nodeDevices);
 
@@ -208,6 +212,28 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
                 );
             }
         },
+        {
+            name: "", value: (hostdev, hostdevId) => {
+                const nodeDev = findHostNodeDevice(hostdev, nodeDevices);
+
+                const deleteNICAction = (
+                    <DeleteResourceButton objectId={`${id}-hostdev-${hostdevId}`}
+                                          actionName={_("Remove")}
+                                          showDialog={() => setDeleteDialogProps({
+                                              objectType: hostdev.type + " host device",
+                                              objectName: getNodeDevSource(nodeDev),
+                                              onClose: () => setDeleteDialogProps(undefined),
+                                              deleteHandler: () => domainDetachHostDevice({ connectionName: vm.connectionName, vmName: vm.name, live: vm.state !== 'shut off', dev: nodeDev }),
+                                          })} />
+                );
+
+                return (
+                    <div className='machines-listing-actions'>
+                        {nodeDev && deleteNICAction}
+                    </div>
+                );
+            }
+        },
     ];
 
     let hostdevId = 1;
@@ -222,12 +248,16 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
     });
 
     return (
-        <ListingTable aria-label={_(`VM ${vm.name} Host Devices`)}
-            gridBreakPoint='grid-xl'
-            variant='compact'
-            emptyCaption={_("No host devices assigned to this VM")}
-            rows={rows}
-            columns={columnTitles} />
+        <>
+            {deleteDialogProps &&
+                <DeleteResourceModal {...deleteDialogProps} actionName={_("Remove")} />}
+            <ListingTable aria-label={_(`VM ${vm.name} Host Devices`)}
+                gridBreakPoint='grid-xl'
+                variant='compact'
+                emptyCaption={_("No host devices assigned to this VM")}
+                rows={rows}
+                columns={columnTitles} />
+        </>
     );
 };
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -855,3 +855,28 @@ export function getNextAvailableTarget(existingTargets, busType) {
         i++;
     }
 }
+
+export function getNodeDevSource(dev) {
+    let source;
+
+    if (dev.capability.type === "pci") {
+        let domain = Number(dev.capability.domain);
+        let bus = Number(dev.capability.bus);
+        let slot = Number(dev.capability.slot);
+        let func = Number(dev.capability.function);
+
+        domain = domain.toString(16).padStart(4, '0');
+        bus = bus.toString(16).padStart(2, '0');
+        slot = slot.toString(16).padStart(2, '0');
+        func = func.toString(16).padStart(1, '0');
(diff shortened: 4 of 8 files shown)


skobyda:
The main thing I want to discuss before finishing this PR is the design.
So the use case is straightforward: User wants to attach or detach a device to a VM. They open a "Add host device" dialog, where they can choose type of device (only usb and pci implemented yet) and then choose a specific device from a datalist. (Theoretically the dropdown for type doesn't need to be there, but if all devices in a single data list, I think it would be too large).

I'm not sure if data list is the best decision. And also I think selecting a specific item in data list does not make it distinct enough (maybe we could change styles to color the selected row?)

Screenshot from 2021-06-29 13-59-01

Screenshot from 2021-06-29 13-59-49

martinpitt:
I agree that cramming this into a standard selector is not very friendly. We should display the slots/bus, as that's the only way to disambiguate identical devices.

My current gut feeling is to use a [data list with checkboxes](https://www.patternfly.org/v4/components/data-list#checkboxes-actions-and-additional-cells), and allow attaching multiple devices at once. This would be a bit more friendly than having to open/close/go through the dialog multiple times.

The "Delete Host Device usb" dialog looks scary. I would think that's the dialog when you click "Remove", i.e. it disconnects the device from the VM, but certainly does not "delete" it from the host POV. Also, it should say the name of the device again.

Thanks!

garrett:
Data lists do not align properly (they're hardcoded and there's no relationship between items), so we might want something like a [selectable table](https://www.patternfly.org/v4/components/table#selectable) instead.

How many types are there (PCI/USB/etc.?). I only see two: PCI and USB. We should consider some kind of switcher for filtering, probably a set of radios, if the number of items is appropriate (2 - 5). Generally, only use a dropdown for something like this is the number is > 5, do not use a select widget for just two items.

The list doesn't expand to the edge and alignment is all sorts of strange. (Perhaps a table would help with some of that instead?) Device and bus are still on their own lines and aren't aligned. They should probably be on one line, not two. It's hard to see which is which in this screenshot. Are they supposed to be "007" instead of "7"?

image

Scrolling is wrong. The dialog itself scrolls instead of the list:

Screenshot 2021-07-26 at 15-42-47 Virtual Machines - garrett Thunder

I played around in Firefox's devtools DOM and applied this additional CSS (which should _not_ be applied directly to our codebase, as we need to change things and part of it needs to be in a class instead of various selectors) to demonstrate what I mean:

```css
.pf-c-data-list {
  /* Stretch the list to fill the container */
  width: 100%;
  /* Set overflow with a limiter on the list, so it scrolls instead of the dialog */
  max-height: min(50vh,50rem);
  overflow: auto;
}

@media screen and (min-width: 768px) {
  /* This should be a class on the element and then the class should be used here... */
  .pf-c-data-list.pf-m-grid-md .pf-c-data-list__cell + .pf-c-data-list__cell:last {
    display: flex;
    gap: 1rem;
  }
}

.machines-multivalue-column-label {
  /* Change label margin to be a space, not a character */
  /* Not sure why this is set in the CSS instead of being in the HTML though. It's not accessible when there's just a presentational setting in CSS instead of using actual HTML elements... */
  margin-right: var(--pf-global--spacer--sm);
}
```

After the above CSS, it looks like this, which is improved, but still has shortcomings due to not being a table (alignment) and the issue with the dropdown instead of radios still persists, of course:

Screenshot 2021-07-26 at 16-02-31 Virtual Machines - garrett Thunder

Screen Shot 2021-07-26 at 16 02 19

Oh, and here's what I mean with alignment issues for the data list (look at device and bus — _especially bus_ — labels and values):

Screenshot from 2021-07-26 16-14-48


The only two fixes that are possible:
1. Use a table element
2. Use a subgrid (which only works in Firefox... even still, a few years after it was introduced)

Workarounds include:
1. Hardcode the width (although this breaks with various text strings and various translations)
2. Alignment (such as right-alignment in the last column) and monospaced fonts for the values, in addition to hardcoding the width (with similar issues as just hardcoding the width)
3. Wacky usage of a CSS grid at multiple levels, with some inheriting the grid sizing and others not

Really, the only proper way to fix this is to use an HTML table element and the PatternFly table component provides a table with an HTML table element.

Additionally, I see a LOT of things that probably shouldn't be mapped to a VM. What happens when I map a USB hub to a VM and there are devices attached? Shouldn't the list filter out the hubs and only show actual devices?

skobyda:
So currently I'm stuck with tests. I tried to create a dummy nodedevices, but it seems that archiving it in tar returns an error: https://pastebin.com/dnmkchCm . There are two types of error when tar'ing:
```
tar: ./msr/power/autosuspend_delay_ms: Read error at byte 0, while reading 4096 bytes: Input/output error
tar: ./LNXSYSTM\:00/LNXSYBUS\:00/PNP0A03\:00/device\:03/eject: Cannot open: Permission denied
```
Working around them by excluding the file from tar doesn't help, as then nodedevices cannot be found by libvirt.
My approach to faking nodedevices is [here](https://github.com/cockpit-project/cockpit-machines/pull/226/files#diff-345bc36ca7a25e3e5dfaa3233a3f2cc077a2ffa8b0b3cffe8153d1c6b7f164faR101)

martinpitt:
@skobyda : /sys is indeed a rather brittle matter. Not all the files in there are meant for plain reading, it's full of symlinks (which you need to preserve, instead of copy as file), and some attributes behave rather weird (being write-only, giving EIO and whatnot). Creating this tarball dynamically is both too hard, and also besides the point: You want to create a static one locally, then hand-edit it to get rid of most of the noise and just keep one or two devices, and commit the tar into test/, so that your tests are fully predictable.

Also, you need to grab the full /sys  -- often libraries enumerate devices through the /sys/bus/pci/devices/ or /sys/class/net/ symlink hierarchies. If these are from the host, but /sys/devices is different, both udev and libvirt will get seriously confused.

martinpitt:
Here's my attempt. First, capture a PCI and an USB device on my hos:

 *  `sudo dnf install umockdev`
 * From `lspci` I picked my graphics card:

    00:02.0 VGA compatible controller: Intel Corporation WhiskeyLake-U GT2 [UHD Graphics 620] (rev 02)

 * record that: `umockdev-record /sys/bus/pci/devices/0000:00:02.0 > /tmp/vga.umockdev`. That file is plaintext, reasonably small, and easy to edit. So if you want to change names to something more funny/obvious that it's a test device, that's the place to do it.

 * From `lsusb` I picked my scanner:
 
        Bus 005 Device 004: ID 04a9:221c Canon, Inc. CanoScan LiDE 60

 * Again, record that: `umockdev-record /dev/bus/usb/005/004 > /tmp/scanner.umockdev` (give it the /dev/bus/usb/ device node, it's much easier to map to the lsusb bus/device output than trying to find it on /sys/bus/usb -- umockdev DTRT).

 * Quick consistency check:
 
```
$ umockdev-run -d /tmp/vga.umockdev -d /tmp/scanner.umockdev  lspci
02.0 VGA compatible controller: Intel Corporation WhiskeyLake-U GT2 [UHD Graphics 620] (rev 02)
00:1d.4 PCI bridge: Intel Corporation Cannon Point-LP PCI Express Root Port #13 (rev f1)
05:00.0 PCI bridge: Intel Corporation JHL6540 Thunderbolt 3 Bridge (C step) [Alpine Ridge 4C 2016] (rev 02)
06:01.0 PCI bridge: Intel Corporation JHL6540 Thunderbolt 3 Bridge (C step) [Alpine Ridge 4C 2016] (rev 02)
08:00.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge DD 2018] (rev 06)
09:02.0 PCI bridge: Intel Corporation JHL7540 Thunderbolt 3 Bridge [Titan Ridge DD 2018] (rev 06)
0a:00.0 USB controller: Intel Corporation JHL7540 Thunderbolt 3 USB Controller [Titan Ridge DD 2018] (rev 06)

$ umockdev-run -d /tmp/vga.umockdev -d /tmp/scanner.umockdev  lsusb
Bus 005 Device 004: ID 04a9:221c Canon, Inc. CanoScan LiDE 60
Bus 005 Device 002: ID 17ef:3080 Lenovo USB2.0 Hub             
Bus 005 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
```
(umockdev records/restores all parent devices, due to how sysfs works; libraries and udev expect this)

 * Create a tarball of the mocked /sys. This requires a little trick, as umockdev is not good enough to simulate everything that `tar` does (this is a bug...):

        umockdev-run -d /tmp/vga.umockdev -d /tmp/scanner.umockdev -- sh -c 'tar -C $UMOCKDEV_DIR/sys -cvf /tmp/sys.tar .'

 * Copy the tarball to a `./vm-run fedora-34` machine with `scp /tmp/sys.tar c:/tmp/` or similar.

 * Install tools into the VM: `dnf install -y /usr/sbin/lspci /usr/bin/lsusb`

 * Unpack the tarball in the VM into a temporary /sys: `mount -t tmpfs none /sys; tar -C /sys -xvf /tmp/sys.tar`

 * Validate in the VM that `lsusb` and `lspci` look as expected.

@skobyda: Is there any `virsh` command that I could run to validate that the simulation is sufficient? I hope that this is just looking at /sys for the detection -- the moment it tries to open a /dev device node, this is doomed of course.

skobyda:
> @skobyda: Is there any `virsh` command that I could run to validate that the simulation is sufficient? I hope that this is just looking at /sys for the detection -- the moment it tries to open a /dev device node, this is doomed of course.

you can attach those devices to a Vm with commands:

`virt-xml -c qemu:///system subVmTest1 --add-device --hostdev 00:02.0`
... for your pci device

`virt-xml -c qemu:///system subVmTest1 --add-device --hostdev 005.004`
... for your usb device

The VM needs to be shut off. Otherwise use --update flag to update a running VM, but as we discussed earlier hotplug of a dummy device will probably not work.

If commands pass without error, you can also check with `virsh dumpxml subVmTest1` to see if xml contains <hostdevice> element which fits the device you just attached.

martinpitt:
So I created a simple VM, stopped it, *then* unpacked the tarball, and tried to attach:
```
# virt-install --memory 50 --pxe --virt-type qemu --os-variant alpinelinux3.8 --disk none --wait 0 --name test1
# virsh destroy test1
# mount -t tmpfs none /sys; tar -C /sys -xvf /tmp/sys.tar
# virt-xml -c qemu:///system test1 --add-device --hostdev 00:02.0
Domain 'test1' defined successfully.
# virt-xml -c qemu:///system test1 --add-device --hostdev 005:004
ERROR    Error: --hostdev 005:004: Did not find a matching node device for '005:004'
```

Now `virsh dumpxml test1` contains this:
```xml
    <hostdev mode='subsystem' type='pci' managed='yes'>
      <source>
        <address domain='0x0000' bus='0x00' slot='0x02' function='0x0'/>
      </source>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x08' function='0x0'/>
    </hostdev>
```
It may actually look at the /dev node instead of/in addition to /sys for USB devices at least? (PCI devices don't generally have a /dev node, but USB does)

martinpitt:
@skobyda : As that emulation thing is getting rather out of hand, how about this: write the test with assumptions about the PCI devices that we get in our upstream QEMU testbeds, and skip that particular test in downstream gating (where we can't make these assumptions). That should significantly simplify the test?

skobyda:
So I made changed Dropdown for Radio buttons, and moved the Datalist component for Table, which fixed alignment issues. I try to apply the fix for scrolling, but can't seem to get it right. Could you try to apply it for current table @garrett ?

Screenshot from 2021-07-29 01-09-47
Screenshot from 2021-07-29 01-08-27

src/components/vm/hostdevs/hostDevAdd.jsx
```diff
@@ -0,0 +1,212 @@
...
+            </Button>
+        </>
+    );
+
+    return (
+        <Modal position="top"
+               variant="medium"
+               id={`${idPrefix}-dialog`}
+               onClose={close}
+               title={_("Add host device to VM")}
```
KKoukiou, line 204: Please let's be consistent, how we use 'Add disk', we probably don't need 'to VM' here.

src/components/vm/hostdevs/hostDevAdd.scss
```diff
@@ -0,0 +1,5 @@
+.pf-c-table {
```
KKoukiou, line 1: You selectors needs to be more specific. Otherwise this will apply to all tables in the page.

src/libvirt-dbus.js
```diff
@@ -1305,6 +1306,15 @@ export function attachIface({ connectionName, vmId, mac, permanent, hotplug, sou
     return attachDevice({ connectionName, vmId, permanent, hotplug, xmlDesc });
 }
 
+export function attachHostDevice({ connectionName, vmName, live, dev }) {
+    const source = getNodeDevSource(dev);
+
+    return cockpit.script(
+        `virt-xml -c qemu:///${connectionName} ${vmName} --add-device --hostdev ${source}` + (live ? " --update" : ""),
+        { err: "message" }
```
KKoukiou, line 1314: Don't we need a conditional superuser try here?

KKoukiou (changes requested):

src/libvirt-dbus.js
```diff
@@ -1315,6 +1315,15 @@ export function attachHostDevice({ connectionName, vmName, live, dev }) {
     );
 }
 
+export function detachHostDevice({ connectionName, vmName, live, dev }) {
+    const source = getNodeDevSource(dev);
+
+    return cockpit.script(
+        `virt-xml -c qemu:///${connectionName} ${vmName} --remove-device --hostdev ${source}` + (live ? "--update" : ""),
+        { err: "message" }
```
KKoukiou, line 1323: Same here the comment about superuser.

KKoukiou (changes requested):

src/libvirt-dbus.js
```diff
@@ -1305,6 +1306,15 @@ export function attachIface({ connectionName, vmId, mac, permanent, hotplug, sou
     return attachDevice({ connectionName, vmId, permanent, hotplug, xmlDesc });
 }
 
+export function attachHostDevice({ connectionName, vmName, live, dev }) {
+    const source = getNodeDevSource(dev);
+
+    return cockpit.script(
+        `virt-xml -c qemu:///${connectionName} ${vmName} --add-device --hostdev ${source}` + (live ? " --update" : ""),
```
martinpitt, line 1313: As you already use a format string, please use it for the `--update` part as well. Don't mix \` and +.

src/libvirt-dbus.js
```diff
@@ -1305,6 +1306,15 @@ export function attachIface({ connectionName, vmId, mac, permanent, hotplug, sou
     return attachDevice({ connectionName, vmId, permanent, hotplug, xmlDesc });
 }
 
+export function attachHostDevice({ connectionName, vmName, live, dev }) {
+    const source = getNodeDevSource(dev);
```
martinpitt, line 1310: This is only being used once, and attachHostDevice() is trivial. I suggest to just fold it in here, that makes the code easier to understand.

src/components/vm/hostdevs/hostDevAdd.jsx
```diff
@@ -0,0 +1,212 @@
+/*
```
martinpitt, line 1: Commit message "Enable an addition of host devices to VM" sounds strange. How about "Add host devices to VM"?

test/check-machines-hostdevs
```diff
@@ -85,6 +85,84 @@ class TestMachinesHostDevs(VirtualMachinesCase):
...
+        b.wait_in_text("body", "Virtual machines")
+
+        polkit = """
+polkit.addRule(function(action, subject) {
+    if (subject.user == "admin") {
+        return polkit.Result.YES;
+    }
+});
+"""
+        m.execute("echo '{0}' >> /etc/polkit-1/rules.d/80-libvirt.rules".format(polkit))
```
martinpitt, line 104: That is a gross test hack that must disappear. I suppose this hides the bug with the missing `{ superuser: "try" }` that @KKoukiou already spotted above.

test/check-machines-hostdevs
```diff
@@ -85,6 +85,84 @@ class TestMachinesHostDevs(VirtualMachinesCase):
...
+
+        self.goToVmPage("subVmTest1")
+        b.wait_visible("#vm-subVmTest1-hostdevs")
+
+        # Shut off domain
+        self.performAction("subVmTest1", "forceOff")
```
martinpitt, line 110: That should wait for the VM to actually be off, so that you don't try and add a device to a still running one.

test/check-machines-hostdevs
```diff
@@ -85,6 +85,84 @@ class TestMachinesHostDevs(VirtualMachinesCase):
...
+
+        class HostDevAddDialog(object):
+            def __init__(
+                self, test_obj, dev_type="usb_device", dev_id=0, vm_dev_id=1
+            ):
```
martinpitt: Please don't provide a default value for `dev_type`,  let's always make this explicit in the callers (and they already do). Also avoid the line breaks in the function arguments, this is very unpythonic.

test/check-machines-hostdevs
```diff
@@ -85,6 +85,84 @@ class TestMachinesHostDevs(VirtualMachinesCase):
...
+                b.click("button#vm-subVmTest1-hostdevs-add")
+                b.wait_in_text(".pf-c-modal-box .pf-c-modal-box__header .pf-c-modal-box__title", "Add host device to VM")
+
+            def fill(self):
+                b.click("input#{0}".format(self.dev_type))
+                b.set_checked(".pf-c-table input[name='checkrow{0}']".format(self.dev_id), True)
+                self._model = b.text("#vm-subVmTest1-hostdevs-dialog table tbody tr:nth-child({0}) td:nth-child(2)".format(self.dev_id + 1))
+                self._vendor = b.text("#vm-subVmTest1-hostdevs-dialog table tbody tr:nth-child({0}) td:nth-child(3)".format(self.dev_id + 1))
+
+            def cancel(self):
```
martinpitt, line 142: This isn't being exercised from anything.

test/check-machines-hostdevs
```diff
@@ -85,6 +85,84 @@ class TestMachinesHostDevs(VirtualMachinesCase):
...
+                b.wait_not_present("#vm-subVmTest1-hostdevs-dialog")
+
+            def add(self):
+                b.click(".pf-c-modal-box__footer button:contains(Add)")
+                b.wait_not_present("#vm-subVmTest1-hostdevs-dialog")
+
+            def verify(self):
+                b.wait_visible("#vm-subVmTest1-hostdev-{0}-product".format(self.vm_dev_id))
+                b.wait_in_text("#vm-subVmTest1-hostdev-{0}-product".format(self.vm_dev_id), self._model)
+                b.wait_in_text("#vm-subVmTest1-hostdev-{0}-vendor".format(self.vm_dev_id), self._vendor)
```
martinpitt, line 153: I know it's hard to detect that the devices are really visible inside the VM. But please at least make sure that (1) the dumped XML now contains the device, and (2) the VM still boots (to guard against syntax/XML structure errors and such)

martinpitt (changes requested):
Thank you! I leave the design questions to Garrett and Katerina, I concentrated more on the backend and tests.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -97,40 +95,40 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
...
 
-            addOptionalToCell(cell, protocol, `${hostdevId}-protocol`, _("Protocol"));
-            addOptionalToCell(cell, wwpn, `${hostdevId}-wwpn`, _("WWPN"));
+            cells.push(getOptionalValue(protocol, `${hostdevId}-protocol`, _("Protocol")));
+            cells.push(getOptionalValue(wwpn, `${hostdevId}-wwpn`, _("WWPN")));
         } else if (hostDev.type === "mdev") {
             const uuid = hostDev.source.address.uuid;
 
-            addOptionalToCell(cell, uuid, `${hostdevId}-uuid`, _("UUID"));
+            cells.push(getOptionalValue(uuid, `${hostdevId}-uuid`, _("UUID")));
```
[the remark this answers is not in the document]
skobyda: This code is a place where we list host devices already attached to VM. According to libvirt docs, UUID is mandatory field in XML for `source` of `mdev`: https://libvirt.org/formatdomain.html#usb-pci-scsi-devices
Also libvirt has a XML schema where you can find what values are optional and what values are not optional.  "Char" value for "misc" device is not optional: https://gitlab.com/libvirt/libvirt/-/blob/master/docs/schemas
/domaincommon.rng#L5835 . Same for uuid of mdev device.
Therefore for libvirt API to allow attachment, an UUID needs to be specified when attaching it to a VM, so I think we can rely that a "mdev" host device already attached to VM is present.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -97,40 +95,40 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
...
         } else if (hostDev.type === "storage") {
             const block = hostDev.source.block;
 
-            addOptionalToCell(cell, block, `${hostdevId}-block`, _("Block"));
+            cells.push(getOptionalValue(block, `${hostdevId}-block`, _("Block")));
         } else if (hostDev.type === "misc") {
             const ch = hostDev.source.char;
 
-            addOptionalToCell(cell, ch, `${hostdevId}-char`, _("Char"));
+            cells.push(getOptionalValue(ch, `${hostdevId}-char`, _("Char")));
```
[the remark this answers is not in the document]
skobyda: Same as in case above. Libvirt docs specify it's a mandatory field https://libvirt.org/formatdomain.html#block-character-devices, just as libvirt XML schema, so it should always be present.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -97,40 +95,40 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
...
         } else if (hostDev.type === "net") {
             const iface = hostDev.source.interface;
 
-            addOptionalToCell(cell, iface, `${hostdevId}-interface`, _("Interface"));
+            cells.push(getOptionalValue(iface, `${hostdevId}-interface`, _("Interface")));
```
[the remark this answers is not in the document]
skobyda: Same, guaranteed in libvirt [docs](https://libvirt.org/formatdomain.html#block-character-devices)
and also libvirt XML [schema](https://gitlab.com/libvirt/libvirt/-/blob/master/docs/schemas/domaincommon.rng#L5846)

src/libvirt-dbus.js
skobyda: It's being later used also for detachHostDevice in another commit

src/components/vm/hostdevs/hostDevCard.jsx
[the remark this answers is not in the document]
martinpitt: Thanks! Resolving this one, then.

garrett:
- I can't click on a row or _even a label_ to select the item. That's a usability _and_ accessibility issue.

- There are several devices with no name whatsoever
 
  Screen Shot 2021-08-17 at 14 25 14

- Small screen modes have a funky layout, where checks are in unexpected locations

  Screen Shot 2021-08-17 at 14 24 36
  Screen Shot 2021-08-17 at 14 25 03

Additionally, the whole content of the dialog scrolls instead of just the list. However, it's kind of a feature instead of a bug, I guess, as it maximizes the vertical space, as one doesn't switch between PCI and USB except just once at the beginning.

src/components/vm/hostdevs/hostDevAdd.scss
garrett: :100: Agreed.

I just ran across this issue with the disks table in mobile mode and spent a few minutes debugging it and found this problem.

Here's a screenshot to show that this is more than just theoretical. It's introducing bugs:

Screen Shot 2021-08-17 at 15 36 12

(Notice the two scrollbars, which wouldn't be visible by default on a real mobile device. Plus, the scrolling within scrolling is a very bad pattern.)

garrett:
I'm seeing duplicate CSS on the page again. Are we importing CSS twice?

It makes debugging things very difficult.

Screenshot from 2021-08-17 15-39-41

garrett:
`lspci`:

```
00:00.0 Host bridge: Intel Corporation Comet Lake-U v1 4c Host Bridge/DRAM Controller (rev 0c)
00:02.0 VGA compatible controller: Intel Corporation CometLake-U GT2 [UHD Graphics] (rev 02)
00:04.0 Signal processing controller: Intel Corporation Xeon E3-1200 v5/E3-1500 v5/6th Gen Core Processor Thermal Subsystem (rev 0c)
00:08.0 System peripheral: Intel Corporation Xeon E3-1200 v5/v6 / E3-1500 v5 / 6th/7th/8th Gen Core Processor Gaussian Mixture Model
00:12.0 Signal processing controller: Intel Corporation Comet Lake Thermal Subsytem
00:14.0 USB controller: Intel Corporation Comet Lake PCH-LP USB 3.1 xHCI Host Controller
00:14.2 RAM memory: Intel Corporation Comet Lake PCH-LP Shared SRAM
00:14.3 Network controller: Intel Corporation Comet Lake PCH-LP CNVi WiFi
00:16.0 Communication controller: Intel Corporation Comet Lake Management Engine Interface
00:1c.0 PCI bridge: Intel Corporation Comet Lake PCI Express Root Port #1 (rev f0)
00:1c.4 PCI bridge: Intel Corporation Comet Lake PCI Express Root Port #5 (rev f0)
00:1d.0 PCI bridge: Intel Corporation Comet Lake PCI Express Root Port #9 (rev f0)
00:1d.4 PCI bridge: Intel Corporation Comet Lake PCI Express Root Port #13 (rev f0)
00:1f.0 ISA bridge: Intel Corporation Comet Lake PCH-LP LPC Premium Controller/eSPI Controller
00:1f.3 Audio device: Intel Corporation Comet Lake PCH-LP cAVS
00:1f.4 SMBus: Intel Corporation Comet Lake PCH-LP SMBus Host Controller
00:1f.5 Serial bus controller [0c80]: Intel Corporation Comet Lake SPI (flash) Controller
00:1f.6 Ethernet controller: Intel Corporation Ethernet Connection (10) I219-LM
02:00.0 SD Host controller: Genesys Logic, Inc GL9750 SD Host Controller (rev 01)
03:00.0 PCI bridge: Intel Corporation JHL6240 Thunderbolt 3 Bridge (Low Power) [Alpine Ridge LP 2016] (rev 01)
04:00.0 PCI bridge: Intel Corporation JHL6240 Thunderbolt 3 Bridge (Low Power) [Alpine Ridge LP 2016] (rev 01)
04:01.0 PCI bridge: Intel Corporation JHL6240 Thunderbolt 3 Bridge (Low Power) [Alpine Ridge LP 2016] (rev 01)
04:02.0 PCI bridge: Intel Corporation JHL6240 Thunderbolt 3 Bridge (Low Power) [Alpine Ridge LP 2016] (rev 01)
05:00.0 System peripheral: Intel Corporation JHL6240 Thunderbolt 3 NHI (Low Power) [Alpine Ridge LP 2016] (rev 01)
2b:00.0 USB controller: Intel Corporation JHL6240 Thunderbolt 3 USB 3.1 Controller (Low Power) [Alpine Ridge LP 2016] (rev 01)
2e:00.0 Non-Volatile memory controller: Micron/Crucial Technology Device 5412
```

`lspci -vt`:

```
-[0000:00]-+-00.0  Intel Corporation Comet Lake-U v1 4c Host Bridge/DRAM Controller
           +-02.0  Intel Corporation CometLake-U GT2 [UHD Graphics]
           +-04.0  Intel Corporation Xeon E3-1200 v5/E3-1500 v5/6th Gen Core Processor Thermal Subsystem
           +-08.0  Intel Corporation Xeon E3-1200 v5/v6 / E3-1500 v5 / 6th/7th/8th Gen Core Processor Gaussian Mixture Model
           +-12.0  Intel Corporation Comet Lake Thermal Subsytem
           +-14.0  Intel Corporation Comet Lake PCH-LP USB 3.1 xHCI Host Controller
           +-14.2  Intel Corporation Comet Lake PCH-LP Shared SRAM
           +-14.3  Intel Corporation Comet Lake PCH-LP CNVi WiFi
           +-16.0  Intel Corporation Comet Lake Management Engine Interface
           +-1c.0-[02]----00.0  Genesys Logic, Inc GL9750 SD Host Controller
           +-1c.4-[03-2b]----00.0-[04-2b]--+-00.0-[05]----00.0  Intel Corporation JHL6240 Thunderbolt 3 NHI (Low Power) [Alpine Ridge LP 2016]
           |                               +-01.0-[06-2a]--
           |                               \-02.0-[2b]----00.0  Intel Corporation JHL6240 Thunderbolt 3 USB 3.1 Controller (Low Power) [Alpine Ridge LP 2016]
           +-1d.0-[2d]--
           +-1d.4-[2e]----00.0  Micron/Crucial Technology Device 5412
           +-1f.0  Intel Corporation Comet Lake PCH-LP LPC Premium Controller/eSPI Controller
           +-1f.3  Intel Corporation Comet Lake PCH-LP cAVS
           +-1f.4  Intel Corporation Comet Lake PCH-LP SMBus Host Controller
           +-1f.5  Intel Corporation Comet Lake SPI (flash) Controller
           \-1f.6  Intel Corporation Ethernet Connection (10) I219-LM
```

`lsusb`:

```
Bus 004 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 003 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 002 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 001 Device 003: ID 04f2:b6d0 Chicony Electronics Co., Ltd Integrated Camera
Bus 001 Device 002: ID 1050:0407 Yubico.com Yubikey 4/5 OTP+U2F+CCID
Bus 001 Device 007: ID 0451:82ff Texas Instruments, Inc. 
Bus 001 Device 019: ID 046d:082d Logitech, Inc. HD Pro Webcam C920
Bus 001 Device 016: ID 0bda:0307 Realtek Semiconductor Corp. Card Reader
Bus 001 Device 014: ID 0451:8043 Texas Instruments, Inc. Hub
Bus 001 Device 011: ID 0451:8043 Texas Instruments, Inc. Hub
Bus 001 Device 009: ID 262a:100c SAVITECH Corp. FiiO USB DAC K1
Bus 001 Device 015: ID 3297:1969 ZSA Moonlander Mark I
Bus 001 Device 012: ID 047d:2041 Kensington SlimBlade Trackball
Bus 001 Device 010: ID 05ac:0265 Apple, Inc. Magic Trackpad 2
Bus 001 Device 008: ID 2109:2813 VIA Labs, Inc. VL813 Hub
Bus 001 Device 006: ID 05e3:0610 Genesys Logic, Inc. Hub
Bus 001 Device 005: ID 413c:8442 Dell Computer Corp. 
Bus 001 Device 004: ID 8087:0026 Intel Corp. AX201 Bluetooth
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
```

`lsusb -t`:

```
/:  Bus 04.Port 1: Dev 1, Class=root_hub, Driver=xhci_hcd/2p, 10000M
/:  Bus 03.Port 1: Dev 1, Class=root_hub, Driver=xhci_hcd/2p, 480M
/:  Bus 02.Port 1: Dev 1, Class=root_hub, Driver=xhci_hcd/6p, 10000M
/:  Bus 01.Port 1: Dev 1, Class=root_hub, Driver=xhci_hcd/12p, 480M
    |__ Port 2: Dev 5, If 0, Class=Hub, Driver=hub/6p, 480M
        |__ Port 4: Dev 6, If 0, Class=Hub, Driver=hub/4p, 480M
            |__ Port 3: Dev 11, If 0, Class=Hub, Driver=hub/3p, 480M
                |__ Port 3: Dev 14, If 0, Class=Hub, Driver=hub/3p, 480M
                    |__ Port 3: Dev 16, If 0, Class=Mass Storage, Driver=usb-storage, 480M
            |__ Port 1: Dev 8, If 0, Class=Hub, Driver=hub/4p, 480M
                |__ Port 3: Dev 12, If 0, Class=Human Interface Device, Driver=usbhid, 1.5M
                |__ Port 1: Dev 10, If 3, Class=Human Interface Device, Driver=usbhid, 12M
                |__ Port 1: Dev 10, If 1, Class=Human Interface Device, Driver=usbhid, 12M
                |__ Port 1: Dev 10, If 2, Class=Human Interface Device, Driver=usbhid, 12M
                |__ Port 1: Dev 10, If 0, Class=Human Interface Device, Driver=usbhid, 12M
                |__ Port 4: Dev 15, If 2, Class=Human Interface Device, Driver=usbhid, 12M
                |__ Port 4: Dev 15, If 0, Class=Human Interface Device, Driver=usbhid, 12M
                |__ Port 4: Dev 15, If 3, Class=Vendor Specific Class, Driver=, 12M
                |__ Port 4: Dev 15, If 1, Class=Human Interface Device, Driver=usbhid, 12M
            |__ Port 4: Dev 19, If 2, Class=Audio, Driver=snd-usb-audio, 480M
            |__ Port 4: Dev 19, If 0, Class=Video, Driver=uvcvideo, 480M
            |__ Port 4: Dev 19, If 3, Class=Audio, Driver=snd-usb-audio, 480M
            |__ Port 4: Dev 19, If 1, Class=Video, Driver=uvcvideo, 480M
            |__ Port 2: Dev 9, If 3, Class=Audio, Driver=snd-usb-audio, 12M
            |__ Port 2: Dev 9, If 1, Class=Audio, Driver=snd-usb-audio, 12M
            |__ Port 2: Dev 9, If 0, Class=Human Interface Device, Driver=usbhid, 12M
        |__ Port 5: Dev 7, If 0, Class=Human Interface Device, Driver=usbhid, 480M
    |__ Port 3: Dev 2, If 2, Class=Chip/SmartCard, Driver=usbfs, 12M
    |__ Port 3: Dev 2, If 0, Class=Human Interface Device, Driver=usbhid, 12M
    |__ Port 3: Dev 2, If 1, Class=Human Interface Device, Driver=usbhid, 12M
    |__ Port 8: Dev 3, If 3, Class=Video, Driver=uvcvideo, 480M
    |__ Port 8: Dev 3, If 1, Class=Video, Driver=uvcvideo, 480M
    |__ Port 8: Dev 3, If 4, Class=Application Specific Interface, Driver=, 480M
    |__ Port 8: Dev 3, If 2, Class=Video, Driver=uvcvideo, 480M
    |__ Port 8: Dev 3, If 0, Class=Video, Driver=uvcvideo, 480M
    |__ Port 10: Dev 4, If 0, Class=Wireless, Driver=btusb, 12M
    |__ Port 10: Dev 4, If 1, Class=Wireless, Driver=btusb, 12M
```

src/components/vm/hostdevs/hostDevAdd.jsx
```diff
@@ -110,8 +110,23 @@ const DevRow = ({ idPrefix, type, selectableDevices, setSelectableDevices }) =>
...
-                       return { selected: dev.selected, cells: [dev.nodeDev.capability.product._value, dev.nodeDev.capability.vendor._value, ...getSource(dev.nodeDev, idx)] };
-                   })}>
+                       // Filtered matchers should be USB hub, PCI bridge;
+                       // FIXME: both should filtered by class, not string...
+                       // but USB isn't exposing class at the moment (only PCI does?)
+                       if (dev.nodeDev.capability.product._value &&
+                        !dev.nodeDev.capability.product._value.match(/^hub$| hub$|^$/i) &&
+                        !(dev.nodeDev.class || "").match(/bridge$/i)) {
+                           console.info(dev.nodeDev.class || "UNDEFINED");
+                           console.table(dev.nodeDev);
```
garrett, line 120: Obviously, we should remove the console calls before merging. I think it's still useful for debugging during development right now.

garrett:
I've rebased (because I originally merged in a bunch of changes from main... whoops!) and added two commits. One fixes the table scrolling by using a specific class and the other is an attempt at filtering the USB and PCI device lists (currently in a bit of a hacky way).

garrett:
I haven't been able to figure out the checkbox odd layout nor have I been able to figure out having a label for the item and associating it with the checkbox (which would use a `for=""` attribute in the label and an `id=""` with the same unique value in the checkbox input element).

KKoukiou (commented):
This resolves https://bugzilla.redhat.com/show_bug.cgi?id=1994898 

Let's add it in the commit message.

KKoukiou:
Forced pushed with a rebase. Many things has changes - including API files split and getOptionalOption function change.
The layout is broken now - especially for small screens.

src/components/vm/hostdevs/hostDevAdd.scss
```diff
@@ -1,3 +1,9 @@
+@import "../../../lib/_global-variables.scss";
+@import "@patternfly/patternfly/components/Table/table.scss";
+@import "@patternfly/patternfly/components/Table/table-grid.scss";
+
+
+
```
skobyda, line 6: @KKoukiou is there any specific reason (in scss) for 3 empty lines after imports, or just a mistake?

KKoukiou: Just mistake.

src/libvirtApi/domain.js
```diff
@@ -138,6 +139,16 @@ export function domainAttachDisk({
     return domainAttachDevice({ connectionName, vmId, permanent, hotplug, xmlDesc });
 }
 
+export function domainAttachHostDevice({ connectionName, vmName, live, dev }) {
+    const source = getNodeDevSource(dev);
+    const update = live ? " --update" : "";
+
+    return cockpit.script(
+        `virt-xml -c qemu:///${connectionName} ${vmName} --add-device --hostdev ${source} ${update}`,
```
KKoukiou, line 147: See Martins comment in this other PR https://github.com/cockpit-project/cockpit-machines/pull/374#discussion_r709541242 
Same applies here.

KKoukiou (changes requested):

src/components/vm/hostdevs/hostDevAdd.jsx
```diff
@@ -0,0 +1,236 @@
...
+const _ = cockpit.gettext;
+
+const TypeRow = ({ idPrefix, type, setType }) => {
+    return (
+        <FormGroup fieldId="usb_device"
+                   label={_("Type")}
+                   isInline
+                   hasNoPaddingTop>
+            <Radio id="usb_device"
+                   name="usb_device"
```
KKoukiou, line 51: `name` in a group should be the same, in this case `type` for example.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#defining_a_radio_group

src/components/vm/hostdevs/hostDevAdd.jsx
```diff
@@ -0,0 +1,236 @@
...
+        setSelectableDevices(newDevs);
+    }
+
+    return (
+        <FormGroup fieldId={`${idPrefix}-dev`} label={_("Device")} hasNoPaddingTop isInline>
+            <Table onSelect={onSelect}
+                   variant={TableVariant.compact}
+                   canSelectAll={false}
+                   className="vm-device-table"
+                   aria-label="Selectable Table"
```
KKoukiou, line 128: Aria-labels should be more descriptive and also should be translated.

src/helpers.js
```diff
@@ -855,3 +855,51 @@ export function getNextAvailableTarget(existingTargets, busType) {
...
 }
+
+/**
+ * Parses output of "udevadm info --path [devicePath]" into a node device object
+ * Inspired by parseUdevDB() from cockpit's pkg/lib/machine-info.js
+ *
+ * @param {string} text output of "udevadm" commdn
+ * @returns {object}
+ */
+export function parseUdevDB(text) {
```
KKoukiou, line 866: This is not used anywhere. The function now exists in src/libvirtApi/nodeDevice.js

KKoukiou (changes requested):

KKoukiou:
@skobyda also mobile is completely broken. I fine however if we fix it as a followup. 
Screen Shot 2021-09-22 at 11 28 45

KKoukiou:
@skobyda still some failing tests. https://logs.cockpit-project.org/logs/pull-226-20210922-121409-bcb2b095-rhel-9-0/log.html#66-1 

Also can you please rebase to let packit tests run? Thanks.

garrett:
I was trying all sorts of ways to make this work with the definition list (because it's using a definition list for some odd reason still ... see my comments way above) and couldn't get it working in a pure PF4 way.

Meanwhile, I discovered that the SCSS was importing files, causing lots of duplicated CSS, as the same imports were already included in the Machines page, elsewhere. _Splitting CSS into separate files does not make it compartmentalized!_

Additionally, I discovered that headings were missing from the table. PatternFly, for better or worse, depends on headings when decomposing the table elements to render as a grid in mobile mode. The lack of headings and PatternFly's assumptions were causing the issue we were seeing in this PR. (Ideally, there would be an option to turn off the display of headings in PatternFly, where it would still use them for screen readers, but not display them, even in mobile mode... and mobile mode would then handle not having headers too. But this isn't what we have, so we still need headers... or we would have to do some custom stuff.)

I did want the location presented in a much better way (dropping the definition list, as I mentioned above and showed off mockups), but I guess that'll be in a follow-up? (I can't get it working in PF4/React; I can only do it with CSS... so please fix. Again, that's fine in a follow-up, but please do it.)

garrett (approved):
I was trying all sorts of ways to make this work with the definition list (because it's using a definition list for some odd reason still ... see my comments way above) and couldn't get it working in a pure PF4 way.

Meanwhile, I discovered that the SCSS was importing files, causing lots of duplicated CSS, as the same imports were already included in the Machines page, elsewhere. _Splitting CSS into separate files does not make it compartmentalized!_

Additionally, I discovered that headings were missing from the table. PatternFly, for better or worse, depends on headings when decomposing the table elements to render as a grid in mobile mode. The lack of headings and PatternFly's assumptions were causing the issue we were seeing in this PR. (Ideally, there would be an option to turn off the display of headings in PatternFly, where it would still use them for screen readers, but not display them, even in mobile mode... and mobile mode would then handle not having headers too. But this isn't what we have, so we still need headers... or we would have to do some custom stuff.)

I did want the location presented in a much better way (dropping the definition list, as I mentioned above and showed off mockups), but I guess that'll be in a follow-up? (I can't get it working in PF4/React; I can only do it with CSS... so please fix. Again, that's fine in a follow-up, but please do it.)

---

Oh, and I'm certain I broke pixel tests. But it's not done building, I'm already working later than intended, and it's the weekend. Sorry! :wink: 

 Once @KKoukiou is happy with this, the PR passes tests, and there are new PNGs for the visual tests, it's fine to merge, IMO. We can solve the rest in a follow-up.

KKoukiou:
Just repushed with the mobile pixel test update which was failing.

KKoukiou (approved):

KKoukiou (approved):

test/check-machines-hostdevs
martinpitt: I think this is resolved now -- verify_backend() does that.

skobyda:
All green

garrett:
Updated release note text

martinpitt:
@MatterSec : Yes, it was released in cockpit-machines 253. Ubuntu [has version 256](https://launchpad.net/ubuntu/+source/cockpit-machines) in backports. Which verrsion do you have? And which Ubuntu version?

MatterSec:
> 

Thanks - Sorry, got it now.  Didn't realise I wasn't updating the focal-backports when updating.   Great product!

PR #228, opened 2021-06-30
tests: add test for NIC switching from direct to virtual network type

KKoukiou:
The bugzilla reported issue was most likely fixed in 6fa84cf757.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1977669

### DIFF
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -198,6 +198,14 @@ class TestMachinesNICs(VirtualMachinesCase):
     def testNICEdit(self):
         b = self.browser
 
+        def open(iface):
+            b.click("#vm-subVmTest1-network-{0}-edit-dialog".format(iface))
+            b.wait_visible("#vm-subVmTest1-network-{0}-edit-dialog-modal-window".format(iface))
+
+        def save(iface):
+            b.click("#vm-subVmTest1-network-{0}-edit-dialog-save".format(iface))
+            b.wait_not_present("#vm-subVmTest1-network-{0}-edit-dialog-modal-window".format(iface))
+
         self.add_veth("eth42")
         self.add_veth("eth43")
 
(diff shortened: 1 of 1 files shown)


test/check-machines-nics
```diff
@@ -198,6 +198,14 @@ class TestMachinesNICs(VirtualMachinesCase):
     def testNICEdit(self):
         b = self.browser
 
+        def open(iface):
+            b.click("#vm-subVmTest1-network-{0}-edit-dialog".format(iface))
```
martinpitt, line 202: One of these days we'll teach you f-strings :grin:

test/check-machines-nics
```diff
@@ -242,16 +248,20 @@ class TestMachinesNICs(VirtualMachinesCase):
...
-        b.wait_not_present("#vm-subVmTest1-network-2-edit-dialog-modal-window")
+        save(2)
 
         # The NICs table should display the permanent configuration which is different than the live which we just set
         b.wait_in_text("#vm-subVmTest1-network-2-source", "eth43")
 
+        # Change source type from direct to virtual network - https://bugzilla.redhat.com/show_bug.cgi?id=1977669
+        open(2)
+        b.select_from_dropdown("#vm-subVmTest1-network-2-select-type", "network")
+        save(2)
```
martinpitt, line 263: This is a pretty shallow test, as it only checks that the UI does not crash. Ideally it should test the actual effect in the virt XML, or in what the "attached networks" UI shows after that.

martinpitt (approved):
Thanks!

test/check-machines-nics
KKoukiou: Previously it generated an error in the dialog making the dialog to actually not close, now the dialog closes. I can indeed check that the change took effect.

KKoukiou:
Ok merging this and I am sending now follow up with extension of the testEditNIC, in order that it always verifies with the XML.

PR #230, opened 2021-06-30
tests: introduce NICEditDialog class for usage within testNICEdit

KKoukiou:
This class now contains a more extended testing of the NIC Edit dialog,
including verification that the selected options were applied in the
libvirt XML and that the changes are visible in the UI.

Some identifiers were adjusted in order to better reflect their purpose.


KKoukiou:
@martinpitt as you needed XML verification for the NIC edit dialog and usage of f string (which indeed is the first time I use :))

martinpitt (approved):
Nice generalization! A bit hard to follow when reading, but it does fit into the general style of the tests, and makes it easier to write/cover similar scenarios.

PR #231, opened 2021-06-30
Fix [x] button not closing in disk edit dialog

KKoukiou:


KKoukiou:
There are many tiny issue like this which appear and are not tested. I am considering that we need a class in our test suite which will be the test base class for all dialogs, and it will have checks like this, so that we don't need to repeat them for each dialog separately.

martinpitt (approved):

PR #233, opened 2021-07-01
Migration: Move Popover from FormGroup's label to labelIcon

### LINKED ISSUE #160: migration: Clicking on the "Duration:" label or the circled question mark (popover trigger) toggles the checkbox
This is because the 'Duration' is literally the label for the checkbox. Labels behave like, clicking on them selects their content, this so it's not so unexpected. But not desired in this case.

Reported on https://github.com/cockpit-project/cockpit-machines/pull/15#issuecomment-838755703

skobyda:
Fixes #160


skobyda:
Setting `pf-c-form__group-label-help` className for question mark icon moved it a bit, but I think only for better:
Screenshot from 2021-07-01 13-54-23

KKoukiou (approved):

PR #234, opened 2021-07-01
Center align the spinner which appears when openning the disk add modal

KKoukiou:


martinpitt (approved):
Haha, what a great component name!

PR #235, opened 2021-07-01
Fix false submission in dialog's form

### LINKED ISSUE #161: migration: Click "Migrate", type "qemu://foo", press Enter, the dialog goes away without an error message
The second time pressing Enter does nothing (which is fine). Clicking on "Migrate" gives me an appropriate error message.

Reported by @martinpitt on https://github.com/cockpit-project/cockpit-machines/pull/15#issuecomment-838755703

skobyda:
If you open a modal, select a text input and then press enter, a form
submission is trigerred, which leads to a page reloading.

Fixes #161

### DIFF
--- a/src/components/vm/vmCloneDialog.jsx
+++ b/src/components/vm/vmCloneDialog.jsx
@@ -76,7 +76,11 @@ export const CloneDialog = ({ name, connectionName, toggleModal }) => {
                    </Button>
                </>
            }>
-            <Form isHorizontal>
+            <Form onSubmit={e => {
+                e.preventDefault();
+                onClone();
+            }}
+            isHorizontal>
                 <FormGroup label={_("Name")} fieldId="vm-name"
                            id="vm-name-group"
                            helperTextInvalid={validationFailed.name}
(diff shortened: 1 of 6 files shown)


src/components/vm/vmCloneDialog.jsx
```diff
@@ -76,7 +76,7 @@ export const CloneDialog = ({ name, connectionName, toggleModal }) => {
                    </Button>
                </>
            }>
-            <Form isHorizontal>
+            <Form onSubmit={e => e.preventDefault()} isHorizontal>
```
KKoukiou, line 79: Only one tiny comment. if the modal has a single value, then the primary button operation should run on submit. https://github.com/patternfly/patternfly-react/issues/5680#issuecomment-831538214

KKoukiou (changes requested):

KKoukiou:
This also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1934323, to be added in the commit message.

KKoukiou (approved):

PR #236, opened 2021-07-01
Fix bus selection when attaching iso images

KKoukiou:
Previously the dialog was setting the bus type to undefined, when an iso
path was selected.

In order to test the previously problematic behavior, adjust
AddDiskDialog class to that it fills in only the options supplied.

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1977810


martinpitt:
Thank you! Looks sensible to me, but I'll let @skobyda do the detailed review.

skobyda (approved):
LGTM. Thanks for adding a test

PR #237, opened 2021-07-01
Creating a VM based on a cloud image should not force users to pass cloud init options

KKoukiou:
Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1978206

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -562,19 +562,20 @@ const UsersConfigurationRow = ({
 };
 
 const CloudInitOptionsRow = ({
-    useCloudInit,
     onValueChanged,
     rootPassword,
     userLogin, userPassword,
     validationFailed,
 }) => {
+    const [showOptions, setShowOptions] = useState(false);
+
     return (
         <FormGroup fieldId="cloud-init-checkbox">
             <Checkbox id="cloud-init-checkbox"
-                      isChecked={useCloudInit}
-                      onChange={checked => onValueChanged('useCloudInit', checked)}
-                      label={_("Cloud init parameters")} />
-            {useCloudInit && <UsersConfigurationRow rootPassword={rootPassword}
+                      isChecked={showOptions}
+                      onChange={setShowOptions}
+                      label={_("Set cloud init parameters")} />
+            {showOptions && <UsersConfigurationRow rootPassword={rootPassword}
                                                     rootPasswordLabelInfo={_("Leave the password blank if you do not wish to set a root password")}
                                                     showUserFields
                                                     userLogin={userLogin}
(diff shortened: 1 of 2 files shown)


martinpitt:
I'm not sure that this is really what we want? See https://bugzilla.redhat.com/show_bug.cgi?id=1978206#c2

KKoukiou:
The unbootable disk
---------------------------
So the problem with the unbootable disk was that before if 'clout init options' checkbox was left unset, we did dot use the base image as backing store, instead the generated virt-install command was like that:
```
virt-install --connect qemu:///system --name test --os-variant cirros0.5.1 --memory 1024 --quiet --disk size=1,format=qcow2 --noautoconsole --wait -1 --cdrom /home/kkoukiou/Downloads/cirros-0.5.1-x86_64-disk.img --graphics spice,listen=127.0.0.1 --graphics vnc,listen=127.0.0.1
```

This is obviously wrong :/ 

What does the command generate when checking the 'set cloud parameters' checkbox:
--------------------------------------------------------------------------------------------------------------------

When the user selected the checkbox but not set the parameters one would get this cmd to run:
```
virt-install --connect qemu:///system --name test --os-variant cirros0.5.1 --memory 1024 --quiet --disk size=0.048828125,format=qcow2,backing_store=/home/kkoukiou/Downloads/cirros-0.5.1-x86_64-disk.img --noautoconsole --wait -1 --graphics spice,listen=127.0.0.1 --graphics vnc,listen=127.0.0.1 --cloud-init 
```

Passing the `cloud-init` option will tell virt-install to generated an iso file for the VM to boot from. 

What scenarios we should support:
-----------------------------------------------

When using an existing qcow2 image there are three scenarios I can think of that we need to support.

1) Import it and use it directly as the VM's disk (supported from import dialog)
2) Use a cloud image as base image and pass cloud init options to it (supported from the 'Cloud base image' in create dialog)
3) Use a plain qcow2 image as base image and create the overlay image for the VM to use (not supported now)

I am thinking that I should move the 'Cloud base image' option to the 'import' dialog, and then have the three options above in that modal. Even with the base images, these options are more like importing an existing image than creating a new one.

KKoukiou:
> I'm not sure that this is really what we want? See https://bugzilla.redhat.com/show_bug.cgi?id=1978206#c2

I gave this some more research today, and it seems we should keep supporting the option to *not* pass cloud init parameters when creating VM out for cloud base images, since there are sometimes 'test' images which do have preconfigured root password login. See https://docs.openstack.org/image-guide/obtain-images.html <- cirros

So I would keep this PR as is, with the test fixes, and rethink the big picture with the extra feature discussed above in the future.

martinpitt (approved):
Alright, thanks! I agree that fixing the boot failure aspect is much more important.

PR #242, opened 2021-07-05
migration: Redirect browser to main page after permanent migration

skobyda:
After permanent migration, the VM is undefined and destroyed on origin host. However, it browser would be left hanging at VM's page. Fix this by redirecting to a page with a list of VMs

### DIFF
--- a/src/components/vm/vmMigrateDialog.jsx
+++ b/src/components/vm/vmMigrateDialog.jsx
@@ -34,7 +34,7 @@ import {
 } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from "@patternfly/react-icons";
 
-import { migrateToUri } from '../../libvirt-dbus.js';
+import { getAllVms, migrateToUri } from '../../libvirt-dbus.js';
 import { isEmpty, isObjectEmpty } from '../../helpers.js';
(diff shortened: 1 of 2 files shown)


src/components/vm/vmMigrateDialog.jsx
```diff
@@ -149,7 +149,15 @@ export const MigrateDialog = ({ vm, connectionName, toggleModal }) => {
...
         setInProgress(true);
         return migrateToUri(connectionName, vm.id, destUri, storage, temporary)
-                .then(toggleModal, exc => {
+                .then(() => {
+                    toggleModal();
+                    if (!temporary)
+                        cockpit.location.go(["vms"]);
+                })
+                .then(() => {
+                    return getAllVms({ connectionName });
```
KKoukiou, line 158: Why you need this here? When the VM will get undefined we receive a event and we  update the list.

skobyda: Oh you are right, it should fire an event. I put it there because after migration finished, but the VM stayed in state, assuming I have to call getVm manually. Looks like a bug where no "undefined" event is fired when migration finished and undefined VM on source. I reported it: https://gitlab.com/libvirt/libvirt/-/issues/186

KKoukiou (approved):

PR #252, opened 2021-07-14
Don't round or floor memory and storage size values unless they are not consumed by virt-install

KKoukiou:
Let's use toReadableNumber function only when the number is presented
in a helper text, and not consumed by our scripts.

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1979152


martinpitt (approved):
Danke!

PR #256, opened 2021-07-15
Use cockpit's proxy API for monitoring libvirt service changes

### LINKED ISSUE #37: Wrong memory limit in create VM when libvirt is started for fist time via cockpit in a fresh install
Cockpit version: cockpit-240-1.fc33.x86_64
OS: F33 with latest updates

Hey guys,

I just installed completely fresh F33, updated, rebooted, enabled Cockpit and when creating new VM the form does not allow me to put more than 1 GiB of memory while my host has 32 GB of RAM.

I have found this: I started libvirt from cockpit itself. Just to experience how it would be to install and manage services through cockpit, it was nice sure, but it looks like when the libvirt daemon is started through cockpit, some kind of maximum memory limit is applied (by systemd or namespaces?).

Anyway, workaround for the problem is to restart the libvirtd from a terminal and then it works.

Cheers!

KKoukiou:
Fixes #37
Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1974223

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -56,6 +56,10 @@ export const App = ({ name }) => {
         store.dispatch(updateLibvirtState({
             activeState: libvirtService.state,
         }));
+        if (libvirtService.state == 'failed') {
+            console.warn("Libvirtd has failed");
+            setLoadingResources(false);
+        }
         if (libvirtService.state == 'running' && prevLibvirtServiceRef.current !== 'running' && prevLibvirtServiceRef.current !== null) {
             setLoadingResources(true);
             // If promises rejected here show an toast alert in the main page
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -16,8 +16,7 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { useState, useEffect, useRef } from 'react';
 import { AlertGroup, Button } from '@patternfly/react-core';
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import { superuser } from "superuser.js";
@@ -30,18 +29,65 @@ import { NetworkList } from "./components/networks/networkList.jsx";
 import { VmDetailsPage } from './components/vm/vmDetailsPage.jsx';
 import { CreateVmAction } from "./components/create-vm-dialog/createVmDialog.jsx";
 import LibvirtSlate from "./components/libvirtSlate.jsx";
-import { isObjectEmpty, dummyVmsFilter, vmId } from "./helpers.js";
+import { dummyVmsFilter, vmId } from "./helpers.js";
 import { InlineNotification } from 'cockpit-components-inline-notification.jsx';
 import {
+    getApiData,
     usageStartPolling,
     usageStopPolling,
 } from "./libvirt-dbus.js";
+import { getOsInfoList, getLoggedInUser } from './libvirt-common.js';
+import { useObject, useEvent } from "hooks";
+import * as service from 'service.js';
+import store from './store.js';
 
+import { updateLibvirtState } from './actions/store-actions.js';
 const _ = cockpit.gettext;
 
 superuser.reload_page_on_change();
 
-class App extends React.Component {
+export const App = ({ name }) => {
+    const libvirtService = useObject(() => service.proxy(name), null, []);
+    const prevLibvirtServiceRef = useRef();
+    const [loadingResources, setLoadingResources] = useState(false);
+
+    useEvent(libvirtService, "changed", () => {
+        store.dispatch(updateLibvirtState({
+            activeState: libvirtService.state,
+        }));
+        if (libvirtService.state == 'running' && prevLibvirtServiceRef.current !== 'running' && prevLibvirtServiceRef.current !== null) {
+            setLoadingResources(true);
+            getApiData({ connectionName: 'system' })
+                    .catch(exc => console.warn("getApiData failed:", exc.message))
+                    .finally(() => setLoadingResources(false));
+        }
+        prevLibvirtServiceRef.current = libvirtService.state;
+    });
+    useEvent(superuser, "changed");
+    useEffect(() => {
+        getOsInfoList();
+        getLoggedInUser();
+        prevLibvirtServiceRef.current = libvirtService.state;
+
+        setLoadingResources(true);
+        getApiData({ connectionName: null })
+                .catch(exc => console.warn("getApiData failed: ", exc.message))
+                .finally(() => setLoadingResources(false));
+    }, []);
+
+    // Show libvirtSlate component if libvirtd is not running only to users that are allowed to start the service.
+    if (((!libvirtService.exists || libvirtService.state !== 'running') && superuser.allowed) ||
+        loadingResources) {
+        return (
+            <LibvirtSlate libvirtService={libvirtService}
+                          loadingResources={loadingResources} />
+        );
+    } else return (
+        <AppActive />
+    );
+};
+
+class AppActive extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
@@ -54,21 +100,14 @@ class App extends React.Component {
         this.onAddErrorNotification = this.onAddErrorNotification.bind(this);
         this.onDismissErrorNotification = this.onDismissErrorNotification.bind(this);
         this.onNavigate = () => this.setState({ path: cockpit.location.path });
-        this.onSuperuserChanged = this.onSuperuserChanged.bind(this);
     }
 
     componentDidMount() {
         cockpit.addEventListener("locationchanged", this.onNavigate);
-        superuser.addEventListener("changed", this.onSuperuserChanged);
     }
 
     componentWillUnmount() {
         cockpit.removeEventListener("locationchanged", this.onNavigate);
-        superuser.removeEventListener("changed", this.onSuperuserChanged);
-    }
-
-    onSuperuserChanged() {
-        this.setState({ allowed: !!superuser.allowed });
     }
 
     /*
@@ -112,7 +151,7 @@ class App extends React.Component {
     }
 
     render() {
-        const { vms, config, storagePools, systemInfo, ui, networks, nodeDevices, interfaces } = this.props.store.getState();
+        const { vms, config, storagePools, systemInfo, ui, networks, nodeDevices, interfaces } = store.getState();
         const path = this.state.path;
         const combinedVms = [...vms, ...dummyVmsFilter(vms, ui.vms)];
         const properties = {
@@ -123,16 +162,6 @@ class App extends React.Component {
         const createVmAction = <CreateVmAction {...properties} mode='create' />;
         const importDiskAction = <CreateVmAction {...properties} mode='import' />;
         const vmActions = <> {createVmAction} {importDiskAction} </>;
-        const resources = [...storagePools, ...networks, ...nodeDevices, ...interfaces, ...vms];
-        const loadingResources = resources.some(resource => isObjectEmpty(resource));
-
-        // Show libvirtSlate component if libvirtd is not running only to users that are allowed to start the service.
-        if ((systemInfo.libvirtService.activeState !== 'running' && (this.state.allowed === undefined || this.state.allowed)) ||
-            loadingResources) {
-            return (<LibvirtSlate libvirtService={systemInfo.libvirtService}
-                        loadingResources={loadingResources} />);
-        }
-
         const pathVms = path.length == 0 || (path.length > 0 && path[0] == 'vms');
 
         let vmContent;
@@ -231,8 +260,5 @@ class App extends React.Component {
         );
     }
 }
-App.propTypes = {
-    store: PropTypes.object.isRequired,
-};
 
 export default App;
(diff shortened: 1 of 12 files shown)


KKoukiou:
ups I broke unprivileged logins, need to repost.

src/app.jsx
```diff
@@ -36,12 +36,43 @@ import {
...
 
 superuser.reload_page_on_change();
 
-class App extends React.Component {
+export const App = ({ store, name }) => {
+    const libvirtService = useObject(() => service.proxy(name), null, []);
+
+    useEvent(libvirtService, "changed", () => {
+        store.dispatch(updateLibvirtState({
+            activeState: libvirtService.exists ? libvirtService.state : 'stopped',
```
martinpitt, line 56: is this special case necessary? `libvirtService.state === undefined` for a non-existing service, which may work just as well. There is a lot of states, so I suppose the code can't rely on an exhaustive comparison anyway? The code below only checks for `!= running` anyway.

src/app.jsx
```diff
@@ -36,12 +36,43 @@ import {
...
+        }));
+    });
+    useEvent(superuser, "changed");
+
+    const { interfaces, networks, nodeDevices, storagePools, vms } = store.getState();
+    const resources = [...interfaces, ...networks, ...nodeDevices, ...storagePools, ...vms];
+    const loadingResources = resources.some(resource => isObjectEmpty(resource));
+
+    // Show libvirtSlate component if libvirtd is not running only to users that are allowed to start the service.
+    if (((!libvirtService.exists || libvirtService.state !== 'running') && superuser.allowed) ||
```
martinpitt, line 79: So that means that now privileged users don't see their user libvirt domains if system libvirt is not running. Was that on purpose? On podman we treat them independently, and show the "system podman is also available... [start]" alert/notification/empty state" in this case.

martinpitt (changes requested):
Some initial review. Please feel free to split out the earlier commits to one or two separate PRs, they seem rather easy to land. Also, I suppose the "make eslint happy" commit needs to be squashed somewhere?

KKoukiou:
> Some initial review. Please feel free to split out the earlier commits to one or two separate PRs, they seem rather easy to land. Also, I suppose the "make eslint happy" commit needs to be squashed somewhere?

the eslint commit is needed also on master. No idea how it slipped in. I can split the earlier commits sure!

src/app.jsx
KKoukiou: Not at all, no change of functionality done here. We just show the empty state if libvirtd is not running only if the connected user has privilege to do something about it. That behavior did not change though.

KKoukiou: Podmans behavior is better now that you mention it, do you mind if I change this in a follow up?

src/app.jsx
KKoukiou: We indeed don't need this special case, will remove.

src/app.jsx
martinpitt: Sure, it's fine as a follow-up. I confirm that the current behaviour on main is the same, sorry for the noise!

KKoukiou:
debian-testing crash fixed in https://github.com/cockpit-project/bots/pull/2221/

src/libvirt-common.js
```diff
@@ -44,7 +44,6 @@ import {
 import store from './store.js';
 import VMS_CONFIG from './config.js';
 
-const _ = cockpit.gettext;
```
martinpitt, line 47: I don't understand how it can build in CI? That uses exactly the same eslint version.. do you have a local one which is prefered over the one from `npm install`?

src/components/libvirtSlate.jsx
```diff
@@ -60,10 +61,19 @@ class LibvirtSlate extends React.Component {
...
         const service = this.props.libvirtService;
 
-        enableLibvirt({ enable: this.state.libvirtEnabled, serviceName: service.name });
-        startLibvirt({ serviceName: service.name });
+        enableLibvirt({
+            enable: this.state.libvirtEnabled,
+            serviceName: service.name
+        })
+                .then(() => {
+                    startLibvirt({
```
martinpitt: Big +1 on this serialization -- enable() does a daemon-reload, which tends to lock up with pending start/stop transactions. (I've spent a lot of time debugging this with pcp)

src/components/libvirtSlate.jsx
```diff
@@ -60,10 +61,19 @@ class LibvirtSlate extends React.Component {
...
+                        serviceName: service.name
+                    }).catch(ex => this.setState({ error: _("Failed to start virtualization service"), errorDetail: ex.message }));
+                }, ex => this.setState({ error: _("Failed to enable virtualization service"), errorDetail: ex.message }))
```
martinpitt: It might be a bit simpler to read by chaining the two promises (i.e. drop the `{}` block around startLibvirt) and using a single exception handler for both. But this is not a strong opinion, if you like this better, by all means keep it.

src/app.jsx
```diff
@@ -30,35 +29,51 @@ import { NetworkList } from "./components/networks/networkList.jsx";
...
-export const App = ({ store, name }) => {
+export const App = ({ name }) => {
     const libvirtService = useObject(() => service.proxy(name), null, []);
+    const prevLibvirtServiceRef = useRef();
+    const [loadingResources, setLoadingResources] = useState(false);
 
     useEvent(libvirtService, "changed", () => {
         store.dispatch(updateLibvirtState({
-            activeState: libvirtService.exists ? libvirtService.state : 'stopped',
+            activeState: libvirtService.state,
```
martinpitt: If you need to touch this again, please squash this into the previous commit. (But no need to repush just for this)

src/app.jsx
```diff
@@ -30,35 +29,51 @@ import { NetworkList } from "./components/networks/networkList.jsx";
...
         }));
+        if (libvirtService.exists && libvirtService.state == 'running' && prevLibvirtServiceRef.current !== 'running' && prevLibvirtServiceRef.current !== null) {
```
martinpitt, line 58: .exists test is redundant; .state === undefined if !.exists

src/app.jsx
```diff
@@ -30,35 +29,51 @@ import { NetworkList } from "./components/networks/networkList.jsx";
...
+            setLoadingResources(true);
+            getApiData({ connectionName: 'system' })
+                    .catch(exc => console.warn("getApiData failed: ", exc.message))
```
martinpitt, line 61: stray space at end of string.

Also, we didn't even catch failures from getApiData() before, so supposedly if libvirtd.service runs, this is a Should Not Happen™, and thus a mere console.warn() is enough?

src/app.jsx
```diff
@@ -30,35 +29,51 @@ import { NetworkList } from "./components/networks/networkList.jsx";
...
+                    .finally(() => setLoadingResources(false));
+        }
+        prevLibvirtServiceRef.current = libvirtService.state;
     });
     useEvent(superuser, "changed");
+    useEffect(() => {
+        getOsInfoList();
```
martinpitt, line 68: That looks rather expensive -- useEffect() runs on every state change, no? It's fine to do this once libvirtd.service moves to "running", but e.g. not if it stops, or even if unrelated states like `loadingResources` change. You can filter for events in the second argument, but I'm not very familiar with this. But ISTM that the OS info list sohuld rather be moved into line 59, i.e. the code that runs exactly when libvirtd.service moves to "running"?

test/check-machines-lifecycle
```diff
@@ -83,8 +83,9 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         self.login_and_go("/machines", user=user, superuser=superuser)
 
         if expect_empty_list:
-            b.wait_in_text("#virtual-machines-listing .pf-c-empty-state", "No VM is running")
-            return
+            with b.wait_timeout(20):
```
martinpitt, line 86: Not sure if we really need that (CI machines can be quite busy), but as long as it works, fine for me.

test/check-machines-lifecycle
```diff
@@ -254,73 +255,101 @@ class TestMachinesLifecycle(VirtualMachinesCase):
...
 
-            self.browser.switch_to_top()
-            self.browser.wait_not_visible("#navbar-oops")
+        b.switch_to_top()
+        b.wait_not_visible("#navbar-oops")
 
-            self.browser.reload()
-            self.browser.enter_page('/machines')
-            self.browser.wait_in_text("body", "Virtual machines")
+        b.reload()
```
martinpitt, line 309: why the reload? IMHO every reload in the tests should have a justifying comment at least.

martinpitt (changes requested):
Nice cleanup! I have a bunch of complaints as usual, but hopefully nothing earthshaking.

src/libvirt-common.js
KKoukiou: I thought so as well, but now I just double checked and this is not the case. Seems some issue with webpack5/eslint probably.

Running make on master now emits no warnigns/errros.
However, $ ./node_modules/eslint/bin/eslint.js /home/kkoukiou/repos/cockpit-machines/src/components/vm/snapshots/vmSnapshotsCreateModal.jsx

Shows one warning.

I will check it

src/components/libvirtSlate.jsx
KKoukiou: Right, I mainly did it to have the titles in errors Failed to start / Failed to enable. No strong opinion here either.

src/app.jsx
KKoukiou: This is definitely a `should not happen`, but also I can't imagine a scenario where this specific would happen. anyway, I added a separate commit (Show errors if resource fetching failed after enabling/starting libvirtd) showing errors from this in the UI.

src/app.jsx
KKoukiou: This useEffect code runs one time when the component mounts. see [] as the dependency list. https://medium.com/@felippenardi/how-to-do-componentdidmount-with-react-hooks-553ba39d1571

getOsInfoList is not libvirt API, but libosinfo, and running that on app mount is correct I feel. It also does not need superuser to I feel there is not need to recall it on libvirtd changes.

test/check-machines-lifecycle
KKoukiou: I noticed 1 times a test timing out without this at exactly this point. Let's leave it for now, and I might try to clean up more of these wait_timeout calls in a bulk cleanup.

test/check-machines-lifecycle
KKoukiou: reload basically ensures that the app does not have a state already, ensuring that the next part of the test in (in matters of app state) independent from the previous parts.

It's especially important here, because we:
* have loaded the VMs list in the previous steps (subVmTest1 is present in the UI)
* I want to check in the next steps that starting from the empty state (stopped libvirtd) and then start the service from the UI will still fetch resources, thus make subVmTest1 appear in the UI. This was buggy before.

https://bugzilla.redhat.com/show_bug.cgi?id=1974223#c3

Will add the comment in the test as well.

src/app.jsx
KKoukiou: Done

src/app.jsx
KKoukiou: Done

KKoukiou:
Just did also a clean rebase to origin/main because of pixel tests

src/app.jsx
martinpitt: Sorry about that -- I misremembered what the second `[]` argument does. Thanks!

KKoukiou:
Adding release blocker as it fixes a bug.

src/app.jsx
```diff
@@ -50,15 +50,22 @@ export const App = ({ name }) => {
...
         }));
         if (libvirtService.state == 'running' && prevLibvirtServiceRef.current !== 'running' && prevLibvirtServiceRef.current !== null) {
             setLoadingResources(true);
+            // If promises rejected here show an toast alert in the main page
             getApiData({ connectionName: 'system' })
-                    .catch(exc => console.warn("getApiData failed:", exc.message))
+                    .then(promises => {
+                        promises.forEach(promise => {
+                            if (promise.status == 'rejected' && promise.reason)
+                                setError(promise.reason.message + '\n' + error);
```
martinpitt, line 66: I think you want a `findFirst()` or so here, otherwise you call setState() multiple times in quick succession? (My initial reaction was `break`, but of course this is not a proper loop)

martinpitt (changes requested):
Thanks! One little issue, and I suppose the bots PR for adjusting the naughties needs to land first. Let's rebase after that to not make our test stats less useful.

src/app.jsx
KKoukiou: Fixed.

src/app.jsx
```diff
@@ -30,18 +29,78 @@ import { NetworkList } from "./components/networks/networkList.jsx";
...
+    const prevLibvirtServiceRef = useRef();
+    const [loadingResources, setLoadingResources] = useState(false);
+    const [error, setError] = useState('');
+
+    useEvent(libvirtService, "changed", () => {
+        store.dispatch(updateLibvirtState({
+            activeState: libvirtService.state,
+        }));
+        if (libvirtService.state == 'failed') {
+            console.warn("Libvirtd has failed");
```
martinpitt, line 60: Thanks! If that helps to write a more robust naughty pattern, then by all means let's do so. I.e. keep the current one and additionally match on this line.

KKoukiou: i am waiting for test results to check :)

martinpitt (approved):
(would still like a rebase once the naughty lands, but then someone else can trivially approve this)

PR #257, opened 2021-07-16
Convert from moment to date-fns

martinpitt:
Steal and slightly adjust localize_datetime() from c-podman, and replace
the custom prettyTime() that used the deprecated moment. This fixes
translated times, and avoids our project specific translations of
"Today" and "Yesterday".

Change snapshot creation default name from moment to Date.toISOFormat()
(with stripping off seconds). This is reasonably similar.

-----

Create snapshot dialog and list on main (note the wrong time format -- it's in English, but locale is German):
main-create-snapshot
main-list-snapshot

The same with this PR:
pr-create-snapshot
pr-list-snapshot-de

Also looks correct in English:
pr-list-snapshot-en

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -16,6 +16,10 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
+
+import * as dfnlocales from 'date-fns/locale';
+import { formatRelative } from 'date-fns';
+
 import cockpit from 'cockpit';
 import VMS_CONFIG from './config.js';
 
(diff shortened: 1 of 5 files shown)


martinpitt:
@KKoukiou , @skobyda : Does that green icon thingy for the "current snapshot" work for you? That might rely on a local font which I don't have on my rather minimal system. But supposedly we should ship that font? It's easy to put a handful of glyphs into the project CSS as inline base64 format.

martinpitt:
I sent #258 to fix the icon.

martinpitt:
I just landed https://github.com/cockpit-project/bots/pull/2219 . Re-running packit test, rawhide should hopefully get green now.

src/helpers.js
```diff
@@ -37,6 +41,11 @@ export function toReadableNumber(number) {
     }
 }
 
+export function localize_datetime(time) {
```
KKoukiou, line 44: Not very important but we could also consume timeformat from src/lib.

KKoukiou (approved):

src/helpers.js
martinpitt: Eventually, yes. But timeformat does not yet wrap `formatRelative()`, so we need to add that first. Indeed I replaced a few moment.calendar() with absolute times, I think it'd be nice to bring these back.

PR #258, opened 2021-07-16
Fix "current snapshot" icon

martinpitt:
The icon was broken at least in some cases. Replace it with CheckIcon,
and wrap it in Flex to get a reasonable spacing.

----

 - [x] Rebase after #257 (collision due to code proximity)

On main it looks like that for me:

main-list-snapshot

With this fix:

current-snapshot-pr

(ignore the "today at", that is the change from #257 -- the PRs will collide, but it's a separate change)

### DIFF
--- a/src/components/vm/snapshots/vmSnapshotsCard.jsx
+++ b/src/components/vm/snapshots/vmSnapshotsCard.jsx
@@ -22,8 +22,8 @@ import cockpit from 'cockpit';
 import { vmId, localize_datetime } from "../../../helpers.js";
 import { CreateSnapshotModal } from "./vmSnapshotsCreateModal.jsx";
 import { ListingTable } from "cockpit-components-table.jsx";
-import { Button, Tooltip } from '@patternfly/react-core';
-import { InfoAltIcon } from '@patternfly/react-icons';
+import { Button, Tooltip, Flex, FlexItem } from '@patternfly/react-core';
+import { CheckIcon, InfoAltIcon } from '@patternfly/react-icons';
 import { DeleteResourceButton, DeleteResourceModal } from '../../common/deleteResource.jsx';
 import { RevertSnapshotModal } from './vmSnapshotsRevertModal.jsx';
 import { deleteSnapshot, getVmSnapshots } from '../../../libvirt-dbus.js';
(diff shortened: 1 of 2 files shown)


src/components/vm/snapshots/vmSnapshotsCard.jsx
```diff
@@ -99,14 +99,14 @@ export class VmSnapshotsCard extends React.Component {
...
+                        <FlexItem id={`${id}-snapshot-${snapId}-date`} spacer={{ default: 'spacerSm' }}>
                             {date}
-                        </div>
-                        { snap.isCurrent && <Tooltip content={_("Current")}>
-                            <i id={`${id}-snapshot-${snapId}-current`} className="pficon pficon-ok" />
-                        </Tooltip> }
-                    </div>);
+                        </FlexItem>
+                        { snap.isCurrent && <FlexItem><Tooltip content={_("Current")}>
+                            <CheckIcon id={`${id}-snapshot-${snapId}-current`} color="#6EC664" />
```
KKoukiou, line 107: How did you find this color? I see here the success color is #3E8635 (alert icon background) https://www.patternfly.org/v4/guidelines/colors ?

marusak: Why not just doing `<CheckIcon color="green" />`?

KKoukiou: That would work!

martinpitt: It's "green-300" on that page. "green" is too dark, I tried "lightgreen" and "pf-color-green-300" but that didn't change anything. I'd be much happier about a symbolic name, of course.

martinpitt:
Rebased to current main to resolve conflict. If anyone has a better idea about the green color, I'll gladly improve it!

src/components/vm/snapshots/vmSnapshotsCard.jsx
KKoukiou: The only other way to change the color is to pass a class with className and that class can set:
color: pf-color-green-300

I really don't like color code in the JSX but I am happy to send a followup for this.

KKoukiou (approved):

martinpitt:
I re-pushed this with using the approach you suggested. Looks the same:

image

The extra `@import` in scss is a bit annoying, I copied that from cockpit's pkg/shell/listing.scss (shell seems to be the only page which uses an explicit PF color)

KKoukiou (dismissed):

martinpitt:
@KKoukiou : Sorry for the extra push -- I cleaned up the sass import again, and used the `var()` approach. Works fine.

KKoukiou (approved):

PR #259, opened 2021-07-16
Revert "fmf: Only run a subset of tests on RHEL 9"

KKoukiou:
This reverts commit 4bcb3ec32fb3bb16c4dc78c5d0090e1b1782e8cd.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -25,10 +25,17 @@ export TEST_AUDIT_NO_SELINUX=1
 
 EXCLUDES=""
 
+if [ "$ID" = "rhel" ]; then
+    EXCLUDES="TestMachinesDisks.testDetachDisk
+              TestMachinesDisks.testDiskEdit
+              TestMachinesDisks.testAddDiskDirPool
+              TestMachinesNetworks.testNetworkSettings
+    "
+fi
+
 # Fedora gating tests are running on infra without /dev/kvm; Machines tests are too darn slow there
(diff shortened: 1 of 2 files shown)


martinpitt:
I just landed https://github.com/cockpit-project/bots/pull/2219 . Re-running packit test, rawhide should hopefully get green now.

martinpitt:
https://github.com/cockpit-project/bots/pull/2213 just got green for c-machines, and is about to land.

@KKoukiou : Can you please rebase this, bump the `ref:` on the downstream PR, and try again? It should look a lot happier now.

martinpitt (dismissed):
Danke! LGTM, please run tests.

martinpitt (approved):

test/browser/run-test.sh
```diff
@@ -25,10 +25,17 @@ export TEST_AUDIT_NO_SELINUX=1
 
 EXCLUDES=""
 
+if [ "$ID" = "rhel" ]; then
```
martinpitt, line 28: Actually, we should not exclude these tests for RHEL 8, as they already work there. Can you please add `-a "${VERSION_ID#9}" != "$VERSION_ID` to only exclude them for RHEL 9?

martinpitt (changes requested):
Sorry, didn't spot this before!

test/browser/run-test.sh
martinpitt: I just checked that we don't call *this* script in downstream RHEL 8 testing, so ok for now.

martinpitt (approved):

PR #262, opened 2021-07-19
test: libvirt 7.5.0 detects duplicate shared mount IDs now

martinpitt:
https://bugzilla.redhat.com/show_bug.cgi?id=1972218 got fixed in libvirt
7.5.0. Ensure that the "Add shared directory" shows an error about the
already existing ID instead of silently adding it twice.

----

I tested this locally on the [refreshed rhel-9-0 image](https://github.com/cockpit-project/bots/pull/2213), but I won't trigger it here as that test still fails later on (see that very image refresh).

### DIFF
--- a/test/check-machines-filesystems
+++ b/test/check-machines-filesystems
@@ -93,14 +93,18 @@ class TestMachinesFilesystems(VirtualMachinesCase):
         self.assertEqual('', self.machine.execute(xmllint_element.format(prop='devices/filesystem[2]/binary/@xattr')).strip())
 
         # Try to add a new shared filesystem with the same mount tag
-        # Due to a libvirt bug this will be actually added without an error
-        # Keep the test here and add the expected behavior when the BZ is fixed
-        # https://bugzilla.redhat.com/show_bug.cgi?id=1972218
         b.click("#vm-subVmTest1-filesystems-add")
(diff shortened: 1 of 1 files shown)


test/check-machines-filesystems
```diff
@@ -114,6 +114,7 @@ class TestMachinesFilesystems(VirtualMachinesCase):
 
         # Start VM and ensure that adding filesystem is disabled
         b.click("#vm-subVmTest1-run")
+        b.wait_in_text("#vm-subVmTest1-state", "Running")
```
KKoukiou, line 117: That is not really needed. The next line (disabled button) is an implicit check for this, as the add button is disabled for running VMs.

This is failing on RHEL-9 https://bugzilla.redhat.com/show_bug.cgi?id=1966580 so if you decide to keep it you need to adjust the naughty

martinpitt: Oh, heh, thanks -- I was about to file that bug, I thought it was a recent regression. Adding this seems to make the error more clear (as it then fails on a plausible thing, not "huh, did we break the disabling of the button?"). But as that already exists, I just back it out.

martinpitt:
Updated, and additionally triggered test against the new rhel-9-0 image.

KKoukiou (approved):

PR #272, opened 2021-07-23
Disable non-shared storage migration for some distros

### LINKED ISSUE #247: Don't offer live migration on non-shared storage
According to the libvirt team, this is the support status:

Supported: Copy a disk from a shutdown VM (simple cp/scp)

Grey area: Live migrate a VM with shared storage

Not supported: Live Migrate a VM with non-shared storage

I am not sure, but I believe we currently offer the third scenario. Let's disable this then. That's the one which I was already wondering about in https://github.com/cockpit-project/cockpit-machines/pull/15#issuecomment-828969408 (most of the discussion happened on IRC, sorry), as that's not robust by design -- trying to copy a moving target.

I realize that we have to trust the user's input for "shared storage", as that's not something which we can detect. But we already have an option for it, and we do know whether the VM is running or paused/shut down.

skobyda:
When copy-storage option is disabled for certain distro, storage row is hidden and shared-storage is assumed. This is not perfect, as user might not realize they need to have shared storage set up, so perhaps some warning or change of design could be done.

Screenshot from 2021-07-23 18-10-21

Fixes #247

### DIFF
--- a/src/components/vm/vmMigrateDialog.jsx
+++ b/src/components/vm/vmMigrateDialog.jsx
@@ -132,6 +132,21 @@ export const MigrateDialog = ({ vm, connectionName, toggleModal }) => {
     const [storage, setStorage] = useState("nocopy");
     const [temporary, setTemporary] = useState(false);
     const [validationFailed, setValidationFailed] = useState(false);
+    const [copyStorageHidden, setCopyStorageHidden] = useState(true);
+
+    cockpit.file("/etc/os-release").read()
+            .then(data => {
+                let isRhel = false;
+
+                data.split('\n').forEach(line => {
+                    const parts = line.split('=');
+
+                    if (parts.length === 2 && parts[0] === "ID" && parts[1].replace(/^"(.*)"$/, '$1') === "rhel")
+                        isRhel = true;
+                });
+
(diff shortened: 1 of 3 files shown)


garrett (approved):
For a simple removal of live migration copying, this works, provided the help text is still accurate (and I think it is).

UI approved; someone else should review the React. :wink:

KKoukiou:
> For a simple removal of live migration copying, this works, provided the help text is still accurate (and I think it is).
> 
> UI approved; someone else should review the React. wink

I don't think there is a help text at this point. We need to add in the modal modal description a hint that the storage is supposed to be shared.

KKoukiou (changes requested):
@garrett can you please suggest a text to be added in the modal description that suggest that the storage can only be shared? (applicable for RHEL)

garrett:
Sorry; I didn't understand what was needed. Thanks for pointing it out.

What about this?

"Storage volumes must be shared between this host and the destination host."

machines-migrate-dialog

KKoukiou:
@skobyda friendly ping for update :)

skobyda:
Updated:

Screenshot from 2021-07-29 01-27-22

KKoukiou (approved):

PR #274, opened 2021-07-26
Rewrite create/install VM scripts to python

KKoukiou:
* [x] Consider hiding traceback or adding under an expandable section in the alert


src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+def virsh(connectionName, args):
+    cmd = ["virsh", "-c", f"qemu:///{connectionName}"]
+    for arg in args:
+        cmd.append(arg)
+
+    logging.debug("Running virsh command :" + ' '.join(cmd))
+
+    return subprocess.run(cmd, check=False, capture_output=True)
```
allisonkarlitskaya, line 20: Here's a bit more Pythonic and more convenient to use (you'll need to update all call sites to remove the array around the paramters, and just pass them as directly to parameters of the function), and also just read stdout directly.

```suggestion
def virsh(connection, *args, check=True):
    cmd = ("virsh", "-c", f"qemu:///{connection}") + args
    logging.debug("Running virsh command :" + ' '.join(cmd))
    return subprocess.run(cmd, check=check, capture_output=True).stdout
```

or just use `check_output()` and catch expected exceptions.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+def assertVmExists(args):
+    cmd = f"virsh -c qemu:///{args.get('connectionName')} list --all | awk  " + "'{print $2}'" + f" | grep -q --line-regexp --fixed-strings {args.get('vmName')}"
+
+    subprocess.run(cmd, shell=True, check=True, capture_output=True)
```
allisonkarlitskaya, line 26: Probably needs more than a simple substring check, but you get the idea.

```suggestion
def assert_vm_exists(connection, name):
    assert name in virsh(connection, 'list', '--all')
```

src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+def getGraphicsCapabilies():
+    result = subprocess.run(["virsh", "domcapabilities"], check=True, capture_output=True)
```
allisonkarlitskaya, line 30: You have this nice `virsh()` function at the top.  Use it!

src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+
+    root = ET.ElementTree(ET.fromstring(result.stdout))
+    graphics = root.find('devices') \
+                   .find('graphics')
```
allisonkarlitskaya, line 34: I'd skip the \ and just do one line

src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+def cleanup(files):
+    for file in files:
+        os.unlink(file)
+
+
+def createVm(args, files):
+    params = prepareVirtInstallParams(args, files)
+
+    result = subprocess.run(params, check=True, capture_output=True)
+    xml = result.stdout
```
allisonkarlitskaya, line 252:
```suggestion
    xml = subprocess.check_output(params)
```

allisonkarlitskaya (changes requested):

src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+def installVm(args, files):
+    # If the VM already exists and we want to run the installer we need to remove it first
+    result = virsh(args.get('connectionName'), ["destroy", args.get('vmName')])
+
+    result = virsh(args.get('connectionName'), ["dumpxml", args.get('vmName')])
+    prevXML = result.stdout
+
+    result = virsh(args.get('connectionName'), ["undefine", args.get('vmName'), "--managed-save"])
```
allisonkarlitskaya, line 277: no point in assigning this variable just to ignore it

src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+        params.append("-1")
+
+    if args.get('type') == 'install' or (args.get('startVm') and args.get('sourceType') != 'disk_image'):
+        params.append("--noautoconsole")
+
+    # Disks
+    if args.get('type') == 'install':
+        hasParam = False
+        for disks in args.get('disks'):
+            params.append("--disk")
```
allisonkarlitskaya, line 160: For all of the cases like this you can write `params.extend(['--disk', disks])`

src/scripts/install_machine.py
```diff
@@ -0,0 +1,362 @@
...
+        params.append("--wait")
+        params.append("-1")
+
+    if args.get('type') == 'install' or (args.get('startVm') and args.get('sourceType') != 'disk_image'):
+        params.append("--noautoconsole")
+
+    # Disks
+    if args.get('type') == 'install':
+        hasParam = False
+        for disks in args.get('disks'):
```
allisonkarlitskaya, line 159:
```suggestion
        for disk in args.get('disks'):
```

allisonkarlitskaya (changes requested):

src/scripts/install_machine.py
allisonkarlitskaya: Or `params += ['--disk', disks]` maybe even nicer.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,334 @@
...
+    supported = graphics.get('supported')
+
+    consoles = []
+    if supported == 'yes':
+        for value in graphics.find('enum').findall('value'):
+            consoles.append(value.text)
+
+    logging.debug('get_graphics_capabilies: ' + ', '.join(consoles))
+
+    return list(filter(lambda console: console in ['vnc', 'spice'], consoles))
```
allisonkarlitskaya, line 37:
```suggestion
    return [c in consoles if c in ['vnc', 'spice']]
```

src/scripts/install_machine.py
```diff
@@ -0,0 +1,334 @@
...
+
+    # NICs
+    if args.get('type') == "install":
+        has_param = False
+        for network in args.get('vnics'):
+            params += ['--network', network]
+            has_param = True
+
+        if not has_param:
+            params += ['--network', "none"]
```
allisonkarlitskaya, line 175: stylistic suggestion
```suggestion
        vnics = args['vnics'] or ['none']
        for network in vnics:
            params += ['--network', network]
```

...and similar for the graphics stuff below

src/scripts/install_machine.py
```diff
@@ -0,0 +1,334 @@
...
+        has_param = False
+        for graphics in get_graphics_capabilies(args.get('connectionName')):
+            params += ['--graphics', f"{graphics},listen=127.0.0.1"]
+            has_param = True
+
+        if not has_param:
+            params += ['--graphics', "none"]
+
+    # Installation media
+    if not args.get('sourceType') == "cloud":
```
allisonkarlitskaya, line 196: that's just `!=`

src/scripts/install_machine.py
```diff
@@ -0,0 +1,334 @@
...
+            sys.exit(0)
+
+    xml_last_phase = xml.strip().split(b'\n\n')[-1]
+    # Get last step only - virt-install can output 1 or 2 steps
+    inject_metadata(xml_last_phase.decode())
+
+
+def install_vm(args, files):
+    # If the VM already exists and we want to run the installer we need to remove it first
+    virsh(args.get('connectionName'), "destroy", args.get('vmName'), check=False)
```
allisonkarlitskaya, line 246: This seems to be the only place you do `check=False`.  At this point I'd lean towards unconditional use of `subprocess.check_output()` and catch (and ignore) the exception in this one place.

allisonkarlitskaya (changes requested):

src/scripts/install_machine.py
```diff
@@ -0,0 +1,334 @@
...
+        for value in graphics.find('enum').findall('value'):
+            consoles.append(value.text)
+
+    logging.debug('get_graphics_capabilies: ' + ', '.join(consoles))
+
+    return list(filter(lambda console: console in ['vnc', 'spice'], consoles))
+
+
+def prepare_unattended(params, args, files):
+    if args.get('unattended'):
```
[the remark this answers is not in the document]
KKoukiou: That's not the same. I want to check if it's truthy, and in the case it is add the option. The key does not always exist.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,334 @@
...
+        params.append("--unattended")
+
+        unattended_params = []
+        if args.get('profile'):
+            unattended_params.append(f"profile={args.get('profile')}")
+
+        if args.get('rootPassword'):
+            with tempfile.NamedTemporaryFile(
```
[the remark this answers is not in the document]
KKoukiou: This option can be used either as "root-password-file=/path/to/password-file" or as "root-password-file=/dev/fd/n", being n the file descriptor of the password-file.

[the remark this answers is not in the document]
KKoukiou: I have to open the file descriptor from python as well, so I don't see the advantage.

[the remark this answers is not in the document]
KKoukiou: @allisonkarlitskaya I am really not sure how I can nicely use python's context manager in this scenario.
 I could open all files at start, but it's really not better than the current approach I feel.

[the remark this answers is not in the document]
allisonkarlitskaya: Something like this

```python
from tempfile import NamedTemporaryFile
from contextlib import contextmanager
import subprocess

@contextmanager
def make_args():
    pw_file = NamedTemporaryFile(mode='w+')
    pw_file.write('abc\n')
    pw_file.flush()
    
    root_pw_file = NamedTemporaryFile(mode='w+')
    root_pw_file.write('xyz\n')
    root_pw_file.flush()
    
    yield [pw_file.name, root_pw_file.name]
    
def main():
    with make_args() as args:
        subprocess.check_call(['cat'] + args)
        
main()  
```

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -927,7 +927,6 @@ class CreateVmModal extends React.Component {
                 userPassword: this.state.userPassword,
                 rootPassword: this.state.rootPassword,
                 userLogin: this.state.userLogin,
-                useCloudInit: this.state.sourceType === CLOUD_IMAGE,
```
martinpitt, line 929: What happened to this? It looks like nothing actually looks at this property any more anywhere, was this just some forgotten cleanup? (again → separate commit)

test/check-machines-create
```diff
@@ -1267,7 +1267,7 @@ class TestMachinesCreate(VirtualMachinesCase):
                 # In this case don't use runuser -l admin because we get errors 'authentication unavailable'
                 listCmd = "virsh -c qemu:///system dumpxml --inactive {0}".format(name)
 
-            wait(lambda: "6" in self.machine.execute(listCmd + " | grep cockpit_machines | wc -l"))
+            wait(lambda: "7" in self.machine.execute(listCmd + " | grep cockpit_machines | wc -l"))
```
martinpitt, line 1270: Why would the running counter change here? I don't see any changes to the test which would start a new machine?

src/scripts/install_machine.py
```diff
@@ -0,0 +1,326 @@
...
+
+def virsh(connection, *args):
+    cmd = ("virsh", "-c", f"qemu:///{connection}") + args
+    logging.debug("Running virsh command :" + ' '.join(cmd))
+    return subprocess.check_output(cmd)
+
+
+def assert_vm_exists(connection, name):
+    logging.debug(virsh(connection, 'list', '--all', '--name').strip().split())
+    assert name.encode() in virsh(connection, 'list', '--all', '--name').strip().split()
```
martinpitt, line 21: Don't call the same command twice, save its output into a variable. Also, this kind of substring match is brittle -- if you ask for "test1" it would be true if only "xxtest12" exists. How about calling `virsh domuuid name` and checking for 0 vs. non-zero exit code? (You'd need to handle `CalledProcessError`, but (1) that's a lot more reliable, and (2) you can certainly create another helper function for calls which are legit to fail)

Also, eventually we should stop calling `virsh` from Python -- we can count on python3-libvirt (as it's a dependency of virt-manager), so using the proper API should be a lot nicer than this shell parsing stuff. However, as you are already that far, let's do this some other time.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,326 @@
...
+
+def prepare_unattended(params, args, files):
+    if args.get('unattended'):
+        params.append("--unattended")
+
+        unattended_params = []
+        unattended_params.append(f"profile={args['profile']}")
+
+        if 'rootPassword' in args:
+            with tempfile.NamedTemporaryFile(
```
martinpitt, line 49: Right now this can't work, as the temporary file will get cleaned up before calling virt-install (i.e. as soon as the `with` block ends). (but you are already discussing this in the thread with Lis)

src/scripts/install_machine.py
```diff
@@ -0,0 +1,326 @@
...
+    except subprocess.CalledProcessError as e:
+        # If virt-install returned non-zero return code, redefine
+        # the VM show that we get back the metadata which enable the 'Install'
+        # button, so that the user can re-attempt installation
+        with tempfile.NamedTemporaryFile() as file:
+            logging.debug("virt-install failed, redefining to renable the 'Install' button")
+            file.write(prevXML)
+            file.flush()
+            virsh(args['connectionName'], "define", file.name)
+            raise e
```
martinpitt, line 256: Did you mean to outdent this raise? Raising the exception inside the temp file context is rather confusing.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,326 @@
...
+    logging.debug(args)
+
+    if args['type'] == 'create':
+        create_vm(args, files)
+    else:
+        install_vm(args, files)
+
+    cleanup(files)
+except Exception as ex:
+    cleanup(files)
```
martinpitt, line 321: Do this in a separate `finally:` block, so that (1) you don't have to call cleanup twice, and (2) you can probably drop the entire exception handling and just rely on the default behaviour.

martinpitt (changes requested):
Thanks! I tried to not interfere too much with @allisonkarlitskaya 's review, I looked at some other places.

src/scripts/install_machine.py
allisonkarlitskaya: It works because of the `delete=False`, but it's very much missing the point of using it as a context manager.

src/scripts/install_machine.py
allisonkarlitskaya: All of this `files` array handling stuff will disappear once we're using `tempfile` correctly.

KKoukiou: Done

src/scripts/install_machine.py
KKoukiou: Yes

src/scripts/install_machine.py
KKoukiou: Fixed with the contextmanager approach. It worked because of the delete=False.

src/scripts/install_machine.py
KKoukiou: Applied the domuuid fix now, will change to python API, and maybe we can remove completely the libvirt-client  dependency like that.

test/check-machines-create
KKoukiou: So this I explained it in the commit message. I don't get why, by building now the XML from python, I don't seem to have control to which element has the xmlns tags.

Now it looks like this:
```
  <metadata xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0" xmlns:cockpit_machines="https://github.com/cockpit-project/cockpit-machines">
    <libosinfo:libosinfo>
      <libosinfo:os id="http://centos.org/centos-stream/8"/>
    </libosinfo:libosinfo>
    <cockpit_machines:data>
      <cockpit_machines:has_install_phase>false</cockpit_machines:has_install_phase>
      <cockpit_machines:install_source_type>os</cockpit_machines:install_source_type>
      <cockpit_machines:install_source/>
      <cockpit_machines:os_variant>centos-stream8</cockpit_machines:os_variant>
    </cockpit_machines:data>
  </metadata>
```

Previously the xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0" and xmlns:cockpit_machines="https://github.com/cockpit-project/cockpit-machines" tags where part of the libosinfo and cockpit_macines:data elements respectively

I thought this is some XML standard that ElementTree has, to add the xmlnfs to the parents.

src/components/create-vm-dialog/createVmDialog.jsx
KKoukiou: Moved to separate commit.

src/app.jsx
```diff
@@ -36,7 +36,7 @@ import {
     usageStartPolling,
     usageStopPolling,
 } from "./libvirt-dbus.js";
-import { getOsInfoList, getLoggedInUser } from './libvirt-common.js';
+import { initState } from './libvirt-common.js';
```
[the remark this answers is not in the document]
KKoukiou: Done.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -761,6 +761,7 @@ class CreateVmModal extends React.Component {
             minimumStorage: 0,
 
             // Unattended installation or cloud init options for cloud images
+            profile: 'jeos',
```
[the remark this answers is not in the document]
KKoukiou: Initialized to '', added an explanation and moved to seperate commit.

test/check-machines-create
martinpitt: Ooh! I was confused. I thought this was `virsh list`, but it is `virsh dumpxml`. Then the numerical change makes sense. (It's a bit odd to expect number of lines instead of doing a substring match for the actual things you expect, but that's follow-up matter). Maybe rename `listCmd` to `dumpCmd`?

src/scripts/install_machine.py
```diff
@@ -0,0 +1,324 @@
...
+        if args['type'] == 'install':
+            disks = args['disks'] or ['none']
+            for disk in disks:
+                params += ['--disk', disk]
+        else:
+            params.append("--disk")
+            if args['sourceType'] == "cloud":
+                backing_store = f",backing_store={args['source']},"
+            else:
+                backing_store = ""
```
allisonkarlitskaya, line 168: It feels weird that this gets set when it might not be used, depending on which case is picked below.  Maybe delete it from here:

```suggestion
```

and later...

src/scripts/install_machine.py
```diff
@@ -0,0 +1,324 @@
...
+
+            if args['sourceType'] == 'disk_image':
+                params.append(f"{args['source']},device=disk")
+            elif args['storagePool'] == 'NoStorage':
+                params.append("none")
+            elif args['storagePool'] != 'NewVolume':
+                params.append(f"vol={args['storagePool']}/{args['storageVolume']}{backing_store}")
+            else:
+                params.append(f"size={args['storageSize']},format=qcow2{backing_store}")
```
allisonkarlitskaya, line 177: ... do something like this

```suggestion
            else:
                if args['storagePool'] != 'NewVolume':
                    params.append(f"vol={args['storagePool']}/{args['storageVolume']}")
                else:
                    params.append(f"size={args['storageSize']},format=qcow2")

                if args['sourceType'] == "cloud":
                    params.append(",backing_store={args['source']}")
```

?

src/scripts/install_machine.py
```diff
@@ -0,0 +1,324 @@
...
+    prevXML = virsh(args['connectionName'], "dumpxml", args['vmName'])
+
+    virsh(args['connectionName'], "undefine", args['vmName'], "--managed-save")
+
+    with prepare_virt_install_params(args) as params:
+        try:
+            subprocess.check_output(params)
+        except subprocess.CalledProcessError as e:
+            # If virt-install returned non-zero return code, redefine
+            # the VM show that we get back the metadata which enable the 'Install'
```
allisonkarlitskaya, line 258:
```suggestion
            # the VM so that we get back the metadata which enable the 'Install'
```

src/scripts/install_machine.py
```diff
@@ -0,0 +1,324 @@
...
+        logging.debug(params)
+
+        yield params
+
+
+def create_vm(args):
+    with prepare_virt_install_params(args) as params:
+        xml = subprocess.check_output(params)
+
+        if args['startVm']:
```
allisonkarlitskaya, line 227: I guess everything starting here can be unintended to be outside of the install params context: you only need the temporary files to survive until the subprocess call returns, right?

allisonkarlitskaya (dismissed):
OK.  This is a very close reading.  If we manage to resolve all of these, I guess there won't be a lot more.

One big issue outstanding is the pipes question, but I get the feeling that you don't feel like solving that.  It could also come later, or not at all. :)

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+            try:
+                assert_vm_exists(args['connectionName'], args['vmName'])
+
+                xml = virsh(args['connectionName'], "dumpxml", "--inactive", args['vmName'])
+            except subprocess.CalledProcessError:
+                logging.info("The VM got deleted while being installed")
+                logging.info(traceback.format_exc())
```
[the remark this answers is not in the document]
KKoukiou: We should not fail here. If the case where the VM was being created and the user deleted it. This should not show a notification in the UI.

src/scripts/install_machine.py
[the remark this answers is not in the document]
KKoukiou: if type==create it's always present, adjusted the condition that way

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+                mode='w+'
+            )
+            user_data_file.write(content)
+            user_data_file.flush()
+            params.append(f"user-data={user_data_file.name}")
+
+    yield params
+
+
+def prepare_installation_source(params, args):
```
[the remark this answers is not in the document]
KKoukiou: Fixed.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+        params += ['--pxe', '--network', args['source']]
+    elif args['sourceType'] == "os":
+        params += ['--install', f"os={args['os']}"]
+    elif args['sourceType'] in ['disk_image', 'cloud']:
+        params.append("--import")
+    elif ((args['source'][0] == '/' and os.path.isfile(args['source'])) or
+            (args['sourceType'] == 'url' and args['source'].endswith(".iso"))):
+        if only_define:
+            return
+        params += ['--cdrom', args['source']]
```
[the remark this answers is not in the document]
KKoukiou: Fixed.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+
+            if not has_param:
+                params += ['--graphics', "none"]
+
+        # Installation media
+        if args['sourceType'] != "cloud":
+            prepare_installation_source(params, args)
+
+        # Autostart
+        if args.get('autostart'):
```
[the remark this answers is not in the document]
KKoukiou: autostart is only present for type==install, adjusted the condition accodingly

src/scripts/install_machine.py
KKoukiou: This needs to be in a variable together, so that subprocess can process is properly. So I applied this fix with the change that I prestore it in a variable

src/scripts/install_machine.py
KKoukiou: Done

src/scripts/install_machine.py
KKoukiou: Indeed, just virt-install needs the files.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+
+@contextmanager
+def prepare_cloud_init(args):
+    params = []
+    if args['sourceType'] == 'cloud':
+        params.append("--cloud-init")
+
+        content = "#cloud-config\n"
+        if 'userPassword' in args and 'userLogin' in args:
+            content += "users:\n"
```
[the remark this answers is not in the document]
KKoukiou: Done

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+            content += f"  - name: {args['userLogin']}:{args['userPassword']}\n"
+
+        if 'rootPassword' in args:
+            content += "chpasswd:\n"
+            content += "  list: |\n"
+            content += f"    root:{args['rootPassword']}\n"
+            content += "  expire: False\n"
+
+            user_data_file = tempfile.NamedTemporaryFile(
```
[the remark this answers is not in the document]
KKoukiou: I just indended it wrong. the file should be created if any of the parameters is passed. userLogin and userPassword are provided together.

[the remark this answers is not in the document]
KKoukiou: Thanks for catching this, hard to notice :)

src/scripts/install_machine.py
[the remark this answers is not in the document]
KKoukiou: This was cool, thanks :)

src/scripts/install_machine.py
KKoukiou: Fixed.

src/libvirt-common.js
```diff
@@ -1154,14 +1158,18 @@ export function enableLibvirt({ enable, serviceName }) {
...
+function getLoggedInUser() {
     logDebug(`GET_LOGGED_IN_USER:`);
     return cockpit.user().then(loggedUser => {
         store.dispatch(setLoggedInUser({ loggedUser }));
     });
 }
 
-export function getOsInfoList () {
+function getPythonPath() {
+    return cockpit.spawn(["/bin/sh", "-c", "which /usr/libexec/platform-python 2>/dev/null || which python3 2>/dev/null || which python"]).then(pyexe => { pythonPath = pyexe.trim() });
```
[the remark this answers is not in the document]
KKoukiou: but we are not having python3 it as dependency, one apparently can remove it without problem on RHEL

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+            only_define = True
+            params.append("--print-xml")
+
+    if args['sourceType'] == "pxe":
+        params += ['--pxe', '--network', args['source']]
+    elif args['sourceType'] == "os":
+        params += ['--install', f"os={args['os']}"]
+    elif args['sourceType'] in ['disk_image', 'cloud']:
+        params.append("--import")
+    elif ((args['source'][0] == '/' and os.path.isfile(args['source'])) or
```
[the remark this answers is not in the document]
KKoukiou: To be honest I just tried to translate the condition from the bash script here,  https://github.com/cockpit-project/cockpit-machines/blob/35c24110395eea8217c85f34eb5bc32d47dd536b/src/scripts/create_machine.sh#L132

This script can be definitely improved, but I tried to avoid actual functional changes and really stick to the rewrite to python, without affecting the functionality.

src/libvirt-common.js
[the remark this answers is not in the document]
martinpitt: @allisonkarlitskaya : Long story, we discussed it at length on IRC. I'm afraid we have to keep this.

(TL/DR: /usr/bin/python3 is for "user-visible/3rd party dependencies on RHEL, and shipped in AppStream; distro packages are supposed to use platform-python on RHEL)

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+
+    yield params
+
+
+def prepare_installation_source(params, args):
+    only_define = False
+    if args['type'] == 'create':
+        if not args['startVm']:
+            only_define = True
+            params.append("--print-xml")
```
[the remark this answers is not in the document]
KKoukiou: Changed

src/scripts/install_machine.py
```diff
@@ -0,0 +1,323 @@
...
+    if args.get('unattended'):
+        params.append("--unattended")
+
+        unattended_params = []
+        unattended_params.append(f"profile={args['profile']}")
+
+        if 'rootPassword' in args:
+            root_pass_file = tempfile.NamedTemporaryFile(
+                prefix="cockpit-machines-",
+                suffix=("-admin-password"),
```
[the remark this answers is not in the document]
KKoukiou: Fixed the parenthesis

test/check-machines-create
martinpitt: FTR: The previous structure was a bug -- you can't declare a namespace *inside* of a tag that already uses it. So what ET produces here is correct, and this is a bug fix. 

Still keeping this thread open for the confusing `listCmd` variable

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -944,7 +944,7 @@ class CreateVmModal extends React.Component {
                 (exception) => {
                     onAddErrorNotification({
                         text: cockpit.format(_("Creation of VM $0 failed"), vmParams.vmName),
-                        detail: exception.message,
+                        detail: exception.message.split(/Traceback(.+)/)[0],
```
martinpitt, line 947: This is rather "meh", but we can clean this up in a follow-up.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+@contextmanager
+def prepare_unattended(args):
+    params = []
+    if args['type'] == 'create' and args['unattended']:
+        params.append("--unattended")
+
+        unattended_params = []
+        unattended_params.append(f"profile={args['profile']}")
```
martinpitt, line 65: Style nitpick -- just do that right away on initialization.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,330 @@
...
+    # - The script is called from the 'Import' dialog
+    if args['type'] == 'install' or args['startVm'] or args['sourceType'] == 'disk_image':
+        has_install_phase = "false"
+
+    # Register used namespaces
+    ET.register_namespace("cockpit_machines", "https://github.com/cockpit-project/cockpit-machines")
+    ET.register_namespace("libosinfo", "http://libosinfo.org/xmlns/libvirt/domain/1.0")
+
+    METADATA = f'''
+<cockpit_machines:data xmlns:cockpit_machines="https://github.com/cockpit-project/cockpit-machines"> \
```
martinpitt, line 301: This still looks weird -- it's declaring the namespace inside a tag which uses it already. I'm not sure if that's legit (but then again libvirt doesn't seem to care)

src/scripts/install_machine.py
```diff
@@ -0,0 +1,330 @@
...
+logging.basicConfig(level=logging.ERROR, format='%(message)s')
+logging.debug(sys.argv[1])
+
+args = json.loads(sys.argv[1], strict=False)
+
+logging.debug(args)
+
+if args['type'] == 'create':
+    create_vm(args)
+else:
```
martinpitt, line 329: I'd add an "elif type == 'install' here, and do a proper error if the script gets called with an unexpected mode. Just for robustness..

martinpitt (approved):
Thanks, amazing work! I still have three small suggestions, but I'd be happy enough with landing this as-is.

test/check-machines-create
KKoukiou: FIxed.

src/scripts/install_machine.py
```diff
@@ -0,0 +1,329 @@
...
+logging.debug(sys.argv[1])
+
+args = json.loads(sys.argv[1], strict=False)
+
+logging.debug(args)
+
+if args['type'] == 'create':
+    create_vm(args)
+elif args['type'] == 'install':
+    install_vm(args)
```
martinpitt, line 329: Now the script silently does nothing for unknown `type`s, instead of installing. That's slightly better, but I think it'd be even better to do something like `raise NotImplementedError("unknown type " + args['type'])`

martinpitt (dismissed):
Danke!

src/scripts/install_machine.py
KKoukiou: Fixed

src/scripts/install_machine.py
KKoukiou: I removed it but then tests failed, so I put it back. I know it's not optimal, but can we make a follow up to clean this up?

martinpitt (approved):

PR #276, opened 2021-07-27
Show only supported pool types in pool create dialog

KKoukiou:
Information is fetched by pool capabilities API.

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1985947

### DIFF
--- a/src/components/storagePools/createStoragePoolDialog.jsx
+++ b/src/components/storagePools/createStoragePoolDialog.jsx
@@ -31,7 +31,7 @@ import { LIBVIRT_SYSTEM_CONNECTION } from '../../helpers.js';
 import { MachinesConnectionSelector } from '../common/machinesConnectionSelector.jsx';
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
 import { FileAutoComplete } from 'cockpit-components-file-autocomplete.jsx';
-import { createStoragePool } from '../../libvirt-dbus.js';
+import { createStoragePool, getPoolCapabilities } from '../../libvirt-dbus.js';
 import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
@@ -52,7 +52,7 @@ const StoragePoolNameRow = ({ onValueChanged, dialogValues }) => {
     );
 };
 
-const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion }) => {
+const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion, poolCapabilities }) => {
     const poolTypes = [
         { type: 'dir', detail: _("Filesystem directory") },
         { type: 'netfs', detail:_("Network file system") },
@@ -63,6 +63,8 @@ const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion }) =>
     // iscsi-direct exists since 4.7.0
     if (libvirtVersion && libvirtVersion >= 4007000)
         poolTypes.push({ type: 'iscsi-direct', detail: _("iSCSI direct target") });
+
+    const supportedPoolTypes = poolTypes.filter(pool => poolCapabilities[pool.type] ? poolCapabilities[pool.type].supported : true);
 
     /* TODO
         { type: 'fs', detail _("Pre-formatted Block Device") },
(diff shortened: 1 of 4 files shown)


martinpitt:
This [Debian stable error](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-276-20210727-115745-5b23a7ae-debian-stable/log.html#61) seems relevant?
```
> warning: virsh pool-capabilities failed: error: unknown command: 'pool-capabilities'
> error: 
RuntimeError: TypeError: Cannot read property 'dir' of undefined
```

src/libvirt-dbus.js
```diff
@@ -1483,6 +1484,17 @@ export function getDomainCapabilities(connectionName, arch, model) {
...
 }
 
+export function getPoolCapabilities({ connectionName }) {
+    // TODO: replace with D-Bus API onec available https://bugzilla.redhat.com/show_bug.cgi?id=1986321
+    const opts = { err: "message", environ: ['LC_ALL=C'] };
+    if (connectionName === 'system')
+        opts.superuser = 'try';
+
+    return cockpit.spawn(["virsh", "-c", "qemu:///" + connectionName, "pool-capabilities"], opts).then(poolCapabilities => {
+        return parsePoolCapabilities(poolCapabilities);
```
martinpitt, line 1494: Please drop the `{ return }` wrapping.

src/libvirt-common.js
```diff
@@ -135,6 +135,18 @@ export function getElem(xml) {
     return xmlDoc.firstElementChild;
 }
 
+export function parsePoolCapabilities(capsXML) {
```
martinpitt, line 138: If `virsh` does not support the pool-capabilities command, then capsXML would be empty, right? For some reason it seems to get past this function without crashing (it crashes later, see above), but I figure it couldn't hurt to add an explicit check if `poolCapsElem` is defined.

src/components/storagePools/createStoragePoolDialog.jsx
```diff
@@ -64,6 +64,8 @@ const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion }) =>
     if (libvirtVersion && libvirtVersion >= 4007000)
         poolTypes.push({ type: 'iscsi-direct', detail: _("iSCSI direct target") });
 
+    const supportedPoolTypes = poolTypes.filter(pool => poolCapabilities[pool.type].supported);
```
martinpitt, line 67: This is where it crashes on debian-stable, `poolCapabilities.dir` is empty.  I suggest this:

    poolCapabilities[pool.type] ? poolCapabilities[pool.type].supported : true

i.e. it falls back to the old behaviour of offering all pool types. On these old distros without runtime detection we'd continue to get the dialog inline error message, which is not a regression (and also the best we can do anyway).

martinpitt (changes requested):
Danke! This is a nice and clean approach (other than having to call a CLI, of course, but thanks for filing the libvirt-dbus bug!)

src/libvirt-common.js
KKoukiou: If virsh does not support pool-capabilities this code will not be called, because it's the first parameter of then.

martinpitt: AH, right -- the promise would fail there, and `poolCapabilities` would simply be uninitialized. Which reminds me -- please fix the constructor to initialize `poolCapabilities` to `null`. It's nice if the constructor shows a complete picture what `this.state` is.

src/libvirt-dbus.js
```diff
@@ -1483,6 +1484,21 @@ export function getDomainCapabilities(connectionName, arch, model) {
...
+    const opts = { err: "message", environ: ['LC_ALL=C'] };
+    if (connectionName === 'system')
+        opts.superuser = 'try';
+
+    return cockpit.spawn(
+        ["virsh", "-c", "qemu:///" + connectionName, "pool-capabilities"],
+        opts
+    ).then(poolCapabilities => parsePoolCapabilities(poolCapabilities), ex => {
+        console.warn('virsh pool-capabilities failed:', ex.toString());
+        return cockpit.reject(ex);
```
martinpitt, line 1498: Let's not turn JS Promises into the obsolete cockpit jQuery promises. However, I see that we already do this in another place, I'll just send a follow-up after this.

martinpitt (approved):
This looks great now, thank you!

martinpitt:
[unrelated tuned flake](https://logs.cockpit-project.org/logs/pull-276-20210727-140032-eaaa7ab0-ubuntu-2004/log.html#1), retrying

PR #279, opened 2021-07-28
Promise API cleanup

martinpitt:


martinpitt:
Some regressions. I'm being lazy and pair-program with CI. I'll send the last (big) commit separately, changed this to only the fairly obvious stuff.

KKoukiou (commented):
There a persistent failure on debian-stable with RuntimeError: te

martinpitt:
The [debian failure](https://logs.cockpit-project.org/logs/pull-279-20210728-073549-c6fdeec1-debian-stable/log.html#14-2) happens because the "Replace deprecated cockpit.defer() with standard Promise" interferes with the error passing through the old cockpit.defer/\.fail API. I need to fix and land that other commit first before I can bring this back. I backed out that commit, should work now. Baby steps!

KKoukiou (approved):

PR #280, opened 2021-07-28
Disallow adjustment of NIC settings for transient NICs

KKoukiou:
Possibly this can be supported but let's dissalow it until we have the
capabity to propertly implement and test this scenario.

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1977669


martinpitt (approved):
Thank you! Looks great, it's nice to give some extra hints why something is not working.

PR #281, opened 2021-07-28
jQuery → Promise handler API

martinpitt:
Replace the old jQuery .done/.fail/.always aliases with their standard
Promise .then/.catch/.finally names. Just keep three .fail() handlers
which rely on some cockpit.defer() specific behaviour; these are more
tricky, and will be ported separately.

Drop the second `data` argument from some exception handlers. This is
not a thing on either the old cockpit.defer nor the Promise API.

 - [x] builds on top of PR #279

### DIFF
--- a/src/libvirt-common.js
+++ b/src/libvirt-common.js
@@ -1120,13 +1120,13 @@ export function createVm({
         profile,
         useCloudInit,
     ], opts)
-            .done(() => {
+            .then(() => {
                 finishVmCreateInProgress(vmName, connectionName);
                 clearVmUiState(vmName, connectionName);
             })
-            .fail((exception, data) => {
+            .fail(ex => {
                 clearVmUiState(vmName, connectionName); // inProgress cleanup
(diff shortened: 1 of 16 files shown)


src/libvirt-common.js
```diff
@@ -1120,13 +1120,13 @@ export function createVm({
         profile,
         useCloudInit,
     ], opts)
-            .done(() => {
+            .then(() => {
                 finishVmCreateInProgress(vmName, connectionName);
                 clearVmUiState(vmName, connectionName);
             })
-            .fail((exception, data) => {
+            .fail(ex => {
```
martinpitt, line 1127: This is the one place which breaks [a single test on Debian stable](https://logs.cockpit-project.org/logs/pull-279-20210728-073549-c6fdeec1-debian-stable/log.html#14-2), presumably because of some buggy cockpit.js promise API in cockpit 239. This needs at least a comment, and what fixing it depends on.

marusak (approved):
Danke!

PR #284, opened 2021-07-29
Disable clone button in 'Clone' dialog when the operation is in progress

KKoukiou:


src/components/create-vm-dialog/uiState.js
```diff
@@ -31,6 +31,18 @@ const CREATE_TIMEOUT = 'CREATE_TIMEOUT';
...
+export function setVmCloneInProgress(name, connectionName) {
+    const vm = {
+        name,
+        connectionName,
+        isUi: true,
+        cloneInProgress: true,
+    };
+
+    store.dispatch(addUiVm(vm));
+    setupCleanupTimeout(name, connectionName, CREATE_TIMEOUT);
```
martinpitt, line 43: Is that really desirable? Creating a new VM is somewhat predictable, but cloning a VM might take arbitrarily long if there is lots of data to copy?

martinpitt (commented):
I mostly wonder about the timeout -- I don't think there is any single static timeout duration which is "correct". But at the same time, with this "do this in the background" it's too much of a "fire and forget".

It might be better to show some "in progress" stub VM which the user can cancel. Honestly, leaving the dialog open with a spinner/progress and an explicit cancel button (i.e. what we have in main) feels a bit safer to me. The bug is just about cancelling not working properly, maybe let's just fix that?

KKoukiou:
> I mostly wonder about the timeout -- I don't think there is any single static timeout duration which is "correct". But at the same time, with this "do this in the background" it's too much of a "fire and forget".
> 
> It might be better to show some "in progress" stub VM which the user can cancel. Honestly, leaving the dialog open with a spinner/progress and an explicit cancel button (i.e. what we have in main) feels a bit safer to me. The bug is just about cancelling not working properly, maybe let's just fix that?

At start I also was leaning to that direction, but cancelling the virt-install process in the middle of a clone, will leave some rubbish behind and a VM with a probably partially copied disk. So I guess it should be cancel and delete the new VM.

martinpitt:
Right, cancel == interrupt + delete the clone sounds good (doesn't make sense to leave anything behind, there is no precious data)

KKoukiou:
The second commit is useless, since we decided to go the other way. So removed it.

martinpitt:
[packit keeps failing](http://artifacts.dev.testing-farm.io/979577e8-db71-42af-8e34-13ecadb2c9d3/), possibly because the branch is 15 commits behind main (and packit doesn't rebase). I pushed a rebase.

martinpitt:
[this packit failure](https://dashboard.packit.dev/results/srpm-builds/31002) is now entirely different -- the PVC of sandcastle is full, quota is exceeded @TomasTomecek

martinpitt (approved):

TomasTomecek:
you are working during weekend! :) :O

sorry about the failures: we are planning to address the issues which resulted in the outage, so they would not happen in future again

PR #288, opened 2021-07-29
Fix unit/value conversion when memory is auto-filled after OS selection

KKoukiou:
Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=1987120


martinpitt (approved):
Thank you!

PR #289, opened 2021-07-29
Allow adding disks with custom path even when no storage pools exist

KKoukiou:
The above condition should remain present for 'Create new' and 'Use
existing' modes, since these rely on storage pools being present and the
user to choose a volume on these.

https://bugzilla.redhat.com/show_bug.cgi?id=1985228


skobyda (approved):
Seems like all tests are passing. Thanks for additional test

PR #294, opened 2021-07-30
Only respect getBestUnit if it returns unit value that is supported by the dialog

KKoukiou:
getBestUnit can return also units > GiB like TiB etc but the create VM dialog has
only MiB and GiB.


KKoukiou:
Ah wait I need to make sure it's also >= MiB

martinpitt (approved):
Danke!

PR #296, opened 2021-07-30
Some `Add disk` dialog fixes

KKoukiou:


KKoukiou:
This makes some pixel tests crazy - and therefore they need some adjustment - but other than that the code is ready for review.

martinpitt:
The [testAddDiskPool](https://logs.cockpit-project.org/logs/pull-296-20210730-173407-f63bf8dc-fedora-34/log.html#59) changed pixels are really evil -- not only does the "Add button" sometimes get a bit larger, but it also captures the text cursor in the "Size" field. Sometimes it's visible, sometimes not. We first need to move the focus to a field without the cursor to pin this down.

martinpitt:
The [testAddDiskPool](https://logs.cockpit-project.org/logs/pull-296-20210730-173407-f63bf8dc-fedora-34/log.html#59) changed pixels are really evil -- not only does the "Add button" sometimes get a bit larger, but it also captures the text cursor in the "Size" field. Sometimes it's visible, sometimes not. We first need to move the focus to a field without the cursor to pin this down.

marusak:
>  We first need to move the focus to a field without the cursor to pin this down.

`b.blur("<selector for the 'Size' field>")` should do, shouldn't it?

martinpitt:
I updated the pixel tests to avoid the cursor in #302 ,  so let's try this again here.

martinpitt:
OK, seems we *have* to get down to the bottom of this weird "Add" button size change.. it's not predictable right now.

martinpitt:
I sent PR #303 to land the two simple fixes. I rebased this PR. The first commit is the one that introduces the dialog instability, so keeping as draft until this gets debugged.

KKoukiou:
Sorry for the ignore selector, I really was not able to figure out how to stabilize them :(

martinpitt:
@KKoukiou : Pragmatic solution, let's not block this fix indeed. Thanks!

marusak:
Needs rebase. I can take a look in the evening.

marusak (approved):
Thanks!

PR #302, opened 2021-08-02
test: Avoid text cursor in "Add disk" dialog pixel tests

martinpitt:
These are a bit unstable, sometimes the blinking cursor is not visible
and causes failed tests. To avoid that, focus the cancel button, which
is know to be stable.

### DIFF
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -297,6 +297,8 @@ class TestMachinesDisks(VirtualMachinesCase):
             self.fill()
 
             if self.pixel_test_tag:
+                # focus away from any input field, to avoid the text cursor in screenshots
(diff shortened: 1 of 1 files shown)


mvollmer:
The "Cancel" button does not actually have the focus ring in all the new reference images, only in two. Is that expected?

martinpitt:
@mvollmer : that's indeed a little weird and smells like a bug in our code. I could work around by selecting the primary button instead, but maybe it's actually a good thing to update the screenshots once we fix the focus ring?

marusak:
Interesting, when I manually try it out, focus ring works as expected. But if tests are happy and it is not flaky again, then :man_shrugging:

marusak (approved):
Thanks!

mvollmer (approved):

PR #303, opened 2021-08-02
Two storage pool fixes

martinpitt:
Broken out from PR #296 as that one needs more work.


skobyda (approved):
Both fixes seem good to me. If tests pass, it's okay to merge

PR #305, opened 2021-08-04
Two test fixes that break gating

martinpitt:
Noted in current round of trying to update RHEL 8 to cockpit-machines 249.


martinpitt:
Ah, [some screenshots](https://logs.cockpit-project.org/logs/pull-305-20210804-122112-baf51e98-fedora-34/log.html#55) show 256MB and need to be updated.

skobyda (approved):

PR #307, opened 2021-08-05
Fix parsing of pool capabilities

skobyda:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1985947

### DIFF
--- a/test/check-machines-storage-pools
+++ b/test/check-machines-storage-pools
@@ -286,7 +286,11 @@ class TestMachinesStoragePools(VirtualMachinesCase):
             def fill(self):
                 b.set_input_text("#storage-pool-dialog-name", self.name)
 
+                if "rhel-9" in m.image:
+                    b.wait_not_present("#storage-pool-dialog-type option[value='iscsi-direct']")
+
                 if self.pool_type:
+                    b.wait_visible("#storage-pool-dialog-type")
                     b.set_val("#storage-pool-dialog-type", self.pool_type)
 
                 if self.target:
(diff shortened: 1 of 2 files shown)


martinpitt:
Ouch, thanks Simon! As we botched this twice, it feels like we should cover this with a test? I.e. on rhel-9-0 we expect no iscsi-direct, on the other OSes we do, right?

skobyda:
> Ouch, thanks Simon! As we botched this twice, it feels like we should cover this with a test? I.e. on rhel-9-0 we expect no iscsi-direct, on the other OSes we do, right?

We should, I thought no OS yet have any pool type unsupported, so if we can test it on rhel-9-0, that's great. I'm gonna update the PR with some test

martinpitt:
@skobyda: Yes, the [image refresh from 8 days ago](https://github.com/cockpit-project/bots/pull/2247) removed iscsi support, which [made a test fail](https://logs.cockpit-project.org/logs/pull-2247-20210727-041814-bf079c16-rhel-9-0-cockpit-project-cockpit-machines/log.html#61-2). https://github.com/cockpit-project/cockpit-machines/pull/276 fixed it, but apparently these two bugs cancelled each other out or something such? It already tests that iscsi-direct is not shown on rhel-9-0, so something is wrong with that test as well maybe?

test/check-machines-storage-pools
```diff
@@ -286,7 +286,11 @@ class TestMachinesStoragePools(VirtualMachinesCase):
             def fill(self):
                 b.set_input_text("#storage-pool-dialog-name", self.name)
 
+                if "rhel-9" in m.image:
+                    b.wait_not_present("#storage-pool-dialog-type option[value='iscsi-direct']")
+
                 if self.pool_type:
+                    b.wait_visible("#storage-pool-dialog-type")
                     b.set_val("#storage-pool-dialog-type", self.pool_type)
```
martinpitt, line 294: Makes me wonder whether we should better use `b.select_PF4()` for this? This is a more realistic way than just hard-poking the value into the DOM.

martinpitt (approved):
Thanks for fixing that! How embarrassing..

PR #315, opened 2021-08-12
Some fixes for mobile mode

KKoukiou:
New:
Screen Shot 2021-08-12 at 11 19 04
Old:
Screen Shot 2021-08-12 at 11 14 18

### DIFF
--- a/src/components/common/stateIcon.scss
+++ b/src/components/common/stateIcon.scss
@@ -17,8 +17,8 @@
 }
 
 .resource-state-text-cell {
-  // Fix labels to not fully stretch when in responsive mode
-  justify-items: start;
+  // Add some spacer if multiple labels exist
+  row-gap: var(--pf-global--spacer--sm);
 }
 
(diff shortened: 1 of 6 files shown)


marusak:
I still see it like this:
Screenshot from 2021-08-12 15-03-27hmm

Are you sure you did not forget to commit something? :)

src/components/common/stateIcon.scss
```diff
@@ -19,6 +19,8 @@
 .resource-state-text-cell {
```
marusak, line 19: Those badges are not under `.resource-state-text-cell`

marusak:
Oh wait, I was looking at VM details, this was on overview. But still, can we also fix it on details? :)

src/components/common/stateIcon.scss
KKoukiou: Good catch, fixed.

marusak (approved):
Thanks!

PR #317, opened 2021-08-12
Fail form validation of 'Add disk' dialog when the preselected pool does not support volume creation

KKoukiou:
Only applicable for when the mode is 'Create new'.

Explanation:
Some pool types, including "iscsi" do not support volume creation. Instead
of hiding these pools from the dialog, which would create a mystery
around the missing entries in the dropdown, we have decided in the past
to show all pools. The ones which don't support the volume create API
are disabled in the dropdown, preventing the user to select them.

However, the alphabetical sort of pools can bring to the top a pool
of that blacklist. In this case we still have to prohibit the dialog submission with
inline validation.

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1992966


marusak (approved):
Thanks!

PR #318, opened 2021-08-12
Fix creation of cloud-init user-data file

KKoukiou:
Now if only root passwd was set, and the user field were left empty we
would get:

users:
  - name:
chpasswd:
  list: |
    root:strongpassword
    :
  expire: False

Which is obviously wrong. We don't need to check if the keys exist, they
always exist if we reach that part of the code, we need to check that
they are not empty.

Note: we do it correctly for other parameters, like unattended installs

This broke in commit 1c286554f3b7f05441588822aefcdc0a92bed0e3

### DIFF
--- a/src/scripts/install_machine.py
+++ b/src/scripts/install_machine.py
@@ -129,16 +129,16 @@ def prepare_cloud_init(args):
             mode='w+'
         )
         user_data_file.write("#cloud-config\n")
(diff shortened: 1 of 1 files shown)


KKoukiou:
This fixes a nasty bug - thus release blocker.

allisonkarlitskaya (approved):
dankedanke

PR #320, opened 2021-08-15
test: add checks of disk and network interface removing button when the VM is paused

yunmingyang:


test/check-machines-lifecycle
```diff
@@ -126,6 +126,9 @@ class TestMachinesLifecycle(VirtualMachinesCase):
 
         # pause
         self.performAction("subVmTest1", "pause")
+        # check removing button of disk and network interface when the VM is paused
+        b.wait_visible("#delete-subVmTest1-disk-vda:disabled")
```
marusak, line 130: I think this is covered already by https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-disks#L990

But it is fine. It is very cheap to check this, so we can surely do it.

Running tests now, when they are green I'll approve

yunmingyang: Many thanks :D

marusak (approved):

PR #326, opened 2021-08-20
Fix error messages not showing

skobyda:
Weeks ago we replaced all the `.fail()` promise handlers with `.catch()`.
Handler `.fail()` returns an original promise, while `.catch()` returns a new promise or a value. This causes a problem, since `.failt().then()` was a non-problematic chain combination, since `.then()` would in case of error never be entered. However chaining `.catch().then()` causes a problem when both `.catch()` and `.then()` are entered, leading to a situation where an error is shown by `.catch()`, but modal is imediatelly closed by .then() without a chance to see the error.

Fix it by switching the two handlers.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1691662


marusak (approved):
Thanks!

PR #329, opened 2021-08-24
test: Lower memory allocation for created VMs

marusak:
In RHEL 8 gating machines these often hit the roof of the gating test
VM:

  qemu-kvm: cannot set up guest memory 'pc.ram': Cannot allocate memory

We don't expect these VMs to actually do much, so lower the RAM
requirement from 300 to 130 MiB.

Also see https://github.com/cockpit-project/cockpit-machines/commit/33272e4b91ca8a2


skobyda (approved):
Looks good. Let's wait until tests pass

PR #337, opened 2021-08-30
Forbid configurations where users would attempt to use cdroms in virtio bus

KKoukiou:
Libvirt would generate an error for that anyway, but let's hide this
combination from the UI.

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1993434

### DIFF
--- a/src/components/vm/disks/diskEdit.jsx
+++ b/src/components/vm/disks/diskEdit.jsx
@@ -29,7 +29,7 @@ import { InfoAltIcon } from '@patternfly/react-icons';
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
 
 import { domainUpdateDiskAttributes } from '../../../libvirtApi/domain.js';
-import { diskCacheModes, getDiskPrettyName, getDiskFullName } from '../../../helpers.js';
+import { diskBusTypes, diskCacheModes, getDiskPrettyName, getDiskFullName } from '../../../helpers.js';
 
 import 'form-layout.scss';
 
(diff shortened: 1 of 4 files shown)


src/components/vm/disks/diskEdit.jsx
```diff
@@ -78,8 +78,8 @@ const CacheRow = ({ onValueChanged, dialogValues, idPrefix, shutoff }) => {
     );
 };
 
-const BusRow = ({ onValueChanged, dialogValues, idPrefix, shutoff }) => {
-    const busTypes = ['sata', 'scsi', 'usb', 'virtio'].map(type => ({ value: type }));
+const BusRow = ({ onValueChanged, dialogValues, diskDevice, idPrefix, shutoff }) => {
+    const busTypes = diskBusTypes[diskDevice].map(type => ({ value: type }));
```
martinpitt, line 82: This assumes that `diskDevice` only ever has the values `disk` or `cdrom`. Is that safe? Could it have any other value or be undefined? Then this would crash.

test/check-machines-disks
```diff
@@ -173,6 +173,16 @@ class TestMachinesDisks(VirtualMachinesCase):
...
 
+        # Virtio bus type should not be shown for CDROM devices
+        m.execute("touch /var/lib/libvirt/novell.iso")
+        m.execute("virsh attach-disk --domain subVmTest1 --source /var/lib/libvirt/novell.iso --target hda --type cdrom --mode readonly --persistent")
+        # virsh attach-disk want send an event for offline VM changes
+        b.reload()
+        b.enter_page('/machines')
+        open("hda")
+        b.wait_not_present("#vm-subVmTest1-disks-hda-edit-bus-type option[value=virtio]")
+        b._wait_present("#vm-subVmTest1-disks-hda-edit-bus-type option[value=sata]")
```
martinpitt, line 184: Please swap these two around -- the wait_present then makes sure that the page is sufficiently initialized, and then `assertFalse(b.is_present()` or `wait_not_present()` (for convenience, not functionality). Otherwise this would be a race.

martinpitt (changes requested):
Thanks! One question and one thing to improve.

src/components/vm/disks/diskEdit.jsx
KKoukiou: Indeed. Added the missing device types and disabled the button just for the case a type is something that's not documented.

martinpitt (approved):
Epharisto! :+1:

PR #338, opened 2021-08-30
Fix crash when adding disk with custom path while no storage pools exist

KKoukiou:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1985228


KKoukiou:
Adding release blocker as it's a nice bug fix.

martinpitt (approved):
Thanks!

PR #339, opened 2021-08-30
Use --location virt-install option also for local media and ISOs

KKoukiou:
Fix unattendded installation usage with ISO files

This commit enables unattended installations with 'Local install media' source option in the UI.

This is because --unattended virt-install option does not work with --cdrom but
it works with --location for linux distros. --cdrom still needs to be used for
windows media however.

note: These was found by trial and error since the docs are very vague about
it. See https://github.com/virt-manager/virt-manager/issues/296

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1868594

* [ ] waiting for clarification https://github.com/virt-manager/virt-manager/issues/296

### DIFF
--- a/src/scripts/install_machine.py
+++ b/src/scripts/install_machine.py
@@ -6,7 +6,6 @@ import traceback
 import json
 import subprocess
 import sys
-import os
 import xml.etree.ElementTree as ET
 import tempfile
 import logging
@@ -162,8 +161,7 @@ def prepare_installation_source(args):
         params += ['--install', f"os={args['os']}"]
     elif args['sourceType'] in ['disk_image', 'cloud']:
         params.append("--import")
-    elif ((args['source'][0] == '/' and os.path.isfile(args['source'])) or
-            (args['sourceType'] == 'url' and args['source'].endswith(".iso"))):
+    elif args['source'].endswith(".iso") and args['os'].startswith('win'):
         if not only_define:
             params += ['--cdrom', args['source']]
     else:
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -410,6 +410,7 @@ class TestMachinesCreate(VirtualMachinesCase):
         # Test create VM with disk of type "block"
         # Check choosing existing volume as destination storage
         runner.createThenInstallTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                                 os_name=config.WINDOWS_SERVER_10,
                                                                  location=config.NOVELL_MOCKUP_ISO_PATH,
                                                                  memory_size=128, memory_size_unit='MiB',
                                                                  storage_pool="poolDisk",
@@ -434,6 +435,7 @@ class TestMachinesCreate(VirtualMachinesCase):
 
             # Check choosing existing volume as destination storage
             runner.createThenInstallTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                                     os_name=config.WINDOWS_SERVER_10,
                                                                      location=config.NOVELL_MOCKUP_ISO_PATH,
                                                                      memory_size=128, memory_size_unit='MiB',
                                                                      storage_pool="iscsi-pool",
@@ -441,12 +443,14 @@ class TestMachinesCreate(VirtualMachinesCase):
 
         # Click the install button through the VM details
         runner.createThenInstallTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                                 os_name=config.WINDOWS_SERVER_10,
                                                                  location=config.NOVELL_MOCKUP_ISO_PATH,
                                                                  memory_size=128, memory_size_unit='MiB',
                                                                  storage_pool="NoStorage"), True)
 
         # Try performing an installation which will fail and ensure that the 'Install' button is still available
         runner.createThenInstallTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                                 os_name=config.WINDOWS_SERVER_10,
                                                                  location=config.NOVELL_MOCKUP_ISO_PATH,
                                                                  memory_size=128, memory_size_unit='MiB',
                                                                  storage_pool="NoStorage"), True, True)
@@ -463,6 +467,7 @@ class TestMachinesCreate(VirtualMachinesCase):
 
         # Check that when there is no storage pool defined a VM can still be created
         runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                      os_name=config.WINDOWS_SERVER_10,
                                                       location=config.NOVELL_MOCKUP_ISO_PATH,
                                                       storage_pool="NoStorage",
                                                       start_vm=True))
@@ -483,6 +488,7 @@ class TestMachinesCreate(VirtualMachinesCase):
         self.browser.wait_in_text("body", "Virtual machines")
 
         runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                      os_name=config.WINDOWS_SERVER_10,
                                                       location=config.NOVELL_MOCKUP_ISO_PATH,
                                                       memory_size=128, memory_size_unit='MiB',
                                                       storage_pool="NoStorage",
@@ -503,6 +509,7 @@ class TestMachinesCreate(VirtualMachinesCase):
 
         # Check choosing existing volume as destination storage
         runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                      os_name=config.WINDOWS_SERVER_10,
                                                       location=config.NOVELL_MOCKUP_ISO_PATH,
                                                       memory_size=128, memory_size_unit='MiB',
                                                       storage_pool="tmp pool",
@@ -511,6 +518,7 @@ class TestMachinesCreate(VirtualMachinesCase):
 
         # Check "NoStorage" option (only define VM)
         runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                      os_name=config.WINDOWS_SERVER_10,
                                                       location=config.NOVELL_MOCKUP_ISO_PATH,
                                                       memory_size=128, memory_size_unit='MiB',
                                                       storage_pool="NoStorage",
@@ -519,6 +527,7 @@ class TestMachinesCreate(VirtualMachinesCase):
         self.machine.execute("touch '{0}'".format(config.PATH_WITH_SPACE))
         # Check file with empty space
         runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                      os_name=config.WINDOWS_SERVER_10,
                                                       location=config.PATH_WITH_SPACE,
                                                       memory_size=128, memory_size_unit='MiB',
                                                       storage_pool="NoStorage",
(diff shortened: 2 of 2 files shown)


fidencio:
@KKoukiou, have you checked whether this works with Windows ISOs?  IIRC Windows ISOs would explicitly require the usage of `--cdrom` instead of `--location`.

KKoukiou:
> @KKoukiou, have you checked whether this works with Windows ISOs? IIRC Windows ISOs would explicitly require the usage of `--cdrom` instead of `--location`.

Oh thanks for the hint! and super thanks for following the notifications here! That's not mentioned in the man pages - the difference between cdrom and location in the docs is very vague. I will check and add a special case if true.

KKoukiou:
> > @KKoukiou, have you checked whether this works with Windows ISOs? IIRC Windows ISOs would explicitly require the usage of `--cdrom` instead of `--location`.
> 
> Oh thanks for the hint! and super thanks for following the notifications here! That's not mentioned in the man pages - the difference between cdrom and location in the docs is very vague. I will check and add a special case if true.

So virt-manager indeed uses the cdrom option to create VMs from ISOs. As cdrom is not compatible with --unattended I fear this will create an annoying switch case code. @fidencio are unattended installations supported for Windows as well?

fidencio:
> > > @KKoukiou, have you checked whether this works with Windows ISOs? IIRC Windows ISOs would explicitly require the usage of `--cdrom` instead of `--location`.
> > 
> > 
> > Oh thanks for the hint! and super thanks for following the notifications here! That's not mentioned in the man pages - the difference between cdrom and location in the docs is very vague. I will check and add a special case if true.
> 
> So virt-manager indeed uses the cdrom option to create VMs from ISOs. As cdrom is not compatible with --unattended I fear this will create an annoying switch case code. @fidencio are unattended installations supported for Windows as well?

Unattended is supposed to work with Windows.

KKoukiou:
ubuntu-2004 test failures.

martinpitt:
> Ideally we should use a real ISO where validation os isinfo command would succeed, but having a real ISO file in the repo would cost us a few MB

We can/should  just maintain one in bots, not commit it to the git repo IMHO. If it literally just needs to be a valid .iso file, you could use bots' cloud-init.iso. If it has to be bootable, we can add/track the [16 MB TinyCore linux](http://www.tinycorelinux.net/downloads.html) or something similarly compact.

src/scripts/install_machine.py
```diff
@@ -162,8 +161,7 @@ def prepare_installation_source(args):
         params += ['--install', f"os={args['os']}"]
     elif args['sourceType'] in ['disk_image', 'cloud']:
         params.append("--import")
-    elif ((args['source'][0] == '/' and os.path.isfile(args['source'])) or
-            (args['sourceType'] == 'url' and args['source'].endswith(".iso"))):
+    elif args['source'].endswith(".iso") and args['os'].startswith('win'):
```
martinpitt, line 164: Hardcoding image names that way is really ugly -- could that have a `# HACK` comment pointing to https://github.com/virt-manager/virt-manager/issues/296 or a similar bz?

test/check-machines-create
```diff
@@ -410,6 +410,7 @@ class TestMachinesCreate(VirtualMachinesCase):
         # Test create VM with disk of type "block"
         # Check choosing existing volume as destination storage
         runner.createThenInstallTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                                 os_name=config.WINDOWS_SERVER_10,
```
martinpitt, line 413: Do we now have any test case which checks an iso with Linux?

KKoukiou: No that would fail now anyway on iso validation - see commit message.

src/scripts/install_machine.py
KKoukiou: I am checking here the prefix of the operating system short name, this is a list that osinfo-db has and is not random. So the only Oses that start with win are versions windows.
 
virt-manager does not support unattended installs and it always uses --cdrom for iso files exactly I we had it before.

I needed to make the change, because as Fabiano pointed out in the BZ here https://bugzilla.redhat.com/show_bug.cgi?id=1868594 in order for unattended installs to work we need to use --location for linux, however for WIndows we need to keeo using --cdrom. That's really confusing and badly documented that's why I opened -https://github.com/virt-manager/virt-manager/issues/296

If you think that's too vaguely explained I can wait till they answer in the issue here https://github.com/virt-manager/virt-manager/issues/296.

KKoukiou: I can add an inline comment with this issue of course, I just wonder if that's enough even.

martinpitt: I'm mostly worried that in a few months we'll look at this and can't make sense of "why are we using --cdrom and --location like that"? Some pointer to the details would be useful there. Maybe point to that bugzilla and virt-manager#296?

test/check-machines-create
martinpitt: I'm confused -- the commit message still talks about "use --location for linux", and that's what the code does AFAIUI, in the [else branch](https://github.com/cockpit-project/cockpit-machines/pull/339/files#diff-1ca34eaf4fc88fdb7fb68685e74a6c1701b7bc2ebb92541110b1b6da8433eb16L171). This should be covered somehow, to make sure that --location works with Linux? (That may require having an iso first, of course -- what you already started to discuss above, like we need to add one to our bots?)

KKoukiou:
This is not ready for merging. This '--location' arg does not work with all ISOs. 

I just downloaded http://archive.ubuntu.com/ubuntu/dists/bionic-updates/main/installer-amd64/current/images/netboot/mini.iso

Running with --location fails: 
```
$ virt-install --connect qemu:///system --quiet --name test --os-variant ubuntu20.10 --memory 2048 --wait -1 --noautoconsole --disk size=5,format=qcow2 --graphics vnc,listen=127.0.0.1 --graphics spice,listen=127.0.0.1 --location /home/kkoukiou/Downloads/mini.iso
ERROR    Error validating install location: Could not find an installable distribution at URL '/home/kkoukiou/Downloads/mini.iso'
```

Whilst this work fine with cdrom.

fidencio:
> This is not ready for merging. This '--location' arg does not work with all ISOs.
> 
> I just downloaded http://archive.ubuntu.com/ubuntu/dists/bionic-updates/main/installer-amd64/current/images/netboot/mini.iso
> 
> Running with --location fails:
> 
> ```
> $ virt-install --connect qemu:///system --quiet --name test --os-variant ubuntu20.10 --memory 2048 --wait -1 --noautoconsole --disk size=5,format=qcow2 --graphics vnc,listen=127.0.0.1 --graphics spice,listen=127.0.0.1 --location /home/kkoukiou/Downloads/mini.iso
> ERROR    Error validating install location: Could not find an installable distribution at URL '/home/kkoukiou/Downloads/mini.iso'
> ```
> 
> Whilst this work fine with cdrom.

I think that fails because osinfo can't detect that ISO.
Taking a look at the data we use to detect the ISO, I can see:
```
fidencio@quino ~/src/upstream/osinfo-db $ isoinfo -d -i ~/Descargas/mini.iso 
CD-ROM is in ISO 9660 format
System id: LINUX
Volume id: CDROM
Volume set id: 
Publisher id: 
Data preparer id: 
Application id: GENISOIMAGE ISO 9660/HFS FILESYSTEM CREATOR (C) 1993 E.YOUNGDALE (C) 1997-2006 J.PEARSON/J.SCHILLING (C) 2006-2007 CDRKIT TEAM
Copyright File id: 
Abstract File id: 
Bibliographic File id: 
Volume set size is: 1
Volume set sequence number is: 1
Logical block size is: 2048
Volume size is: 35659
El Torito VD version 1 found, boot catalog is in sector 63
Joliet with UCS level 3 found
Rock Ridge signatures version 1 found
Eltorito validation header:
    Hid 1
    Arch 0 (x86)
    ID ''
    Key 55 AA
    Eltorito defaultboot header:
        Bootid 88 (bootable)
        Boot media 0 (No Emulation Boot)
        Load segment 0
        Sys type 0
        Nsect 4
        Bootoff 40 64
```

That gives us basically **zero** information to identify the ISO and then properly know what to do about it.
My understanding is that `--location` would work well with ISOs that we'd need to extract the kernel & initrd to boot the machine up, whice `--cdrom` would work with everything else.  I'm not a virt-manager / virt-install maintainer, I'm merely a contributor there (and not even a good one), but I recall @crobinso raising some concerns about mixing both methods.

I'd say the safest call would be to wait till @crobinso is back (beginning of October) to have this sorted out.

KKoukiou:
> > This is not ready for merging. This '--location' arg does not work with all ISOs.
> > I just downloaded http://archive.ubuntu.com/ubuntu/dists/bionic-updates/main/installer-amd64/current/images/netboot/mini.iso
> > Running with --location fails:
> > ```
> > $ virt-install --connect qemu:///system --quiet --name test --os-variant ubuntu20.10 --memory 2048 --wait -1 --noautoconsole --disk size=5,format=qcow2 --graphics vnc,listen=127.0.0.1 --graphics spice,listen=127.0.0.1 --location /home/kkoukiou/Downloads/mini.iso
> > ERROR    Error validating install location: Could not find an installable distribution at URL '/home/kkoukiou/Downloads/mini.iso'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Whilst this work fine with cdrom.
> 
> I think that fails because osinfo can't detect that ISO.
> Taking a look at the data we use to detect the ISO, I can see:
> 
> ```
> fidencio@quino ~/src/upstream/osinfo-db $ isoinfo -d -i ~/Descargas/mini.iso 
> CD-ROM is in ISO 9660 format
> System id: LINUX
> Volume id: CDROM
> Volume set id: 
> Publisher id: 
> Data preparer id: 
> Application id: GENISOIMAGE ISO 9660/HFS FILESYSTEM CREATOR (C) 1993 E.YOUNGDALE (C) 1997-2006 J.PEARSON/J.SCHILLING (C) 2006-2007 CDRKIT TEAM
> Copyright File id: 
> Abstract File id: 
> Bibliographic File id: 
> Volume set size is: 1
> Volume set sequence number is: 1
> Logical block size is: 2048
> Volume size is: 35659
> El Torito VD version 1 found, boot catalog is in sector 63
> Joliet with UCS level 3 found
> Rock Ridge signatures version 1 found
> Eltorito validation header:
>     Hid 1
>     Arch 0 (x86)
>     ID ''
>     Key 55 AA
>     Eltorito defaultboot header:
>         Bootid 88 (bootable)
>         Boot media 0 (No Emulation Boot)
>         Load segment 0
>         Sys type 0
>         Nsect 4
>         Bootoff 40 64
> ```
> 
> That gives us basically **zero** information to identify the ISO and then properly know what to do about it.
> My understanding is that `--location` would work well with ISOs that we'd need to extract the kernel & initrd to boot the machine up, whice `--cdrom` would work with everything else. I'm not a virt-manager / virt-install maintainer, I'm merely a contributor there (and not even a good one), but I recall @crobinso raising some concerns about mixing both methods.
> 
> I'd say the safest call would be to wait till @crobinso is back (beginning of October) to have this sorted out.

Thanks a lot for the reply (again) :) . Out of interest, since you know about the --unattended option, why doesn't it work with --cdrom?

fidencio:
> Thanks a lot for the reply (again) :) . Out of interest, since you know about the --unattended option, why doesn't it work with --cdrom?

When we do unattended installations we extract the kernel / initrd from the ISO, then we insert the unattended script (as in, kickstart file, preseed file, autoyast file) into the initrd and pass the kernel (and a specific kernel command line) to virt-install.  So, we need to use `--location` for that as `--cdrom` code path isn't aware of how to extract & deal with the kernel / initrd location inside the ISO.

Please, take this with a grain of salt as my memory is not the best and there was a reasonable rework on that  code base since I added the feature, but I think the idea was that.

Does this shed some light to the discussion?  Or just make things more confusing? /o\

PR #350, opened 2021-09-01
Stop providing options for shared memory when creating shared directories

### LINKED ISSUE #343: machines: Wrong size when creating 'hugepages' shared directories
Versions:
cockpit-250-1.fc34.x86_64

Steps:
1. Prepare a shutoff VM,
2. Create a shared directory with 'hugepages' for the VM
3. Check the VM XML, and 'hugepages' size is set to equal with memory(That should be 2M usually， also it may be 1G in some server, 'virsh freepages --all' can check that)
4. Run the VM, there will be an error, which is "Memory backend 'pc.ram' not found"

### LINKED ISSUE #344: machines: There should be an hint before creating shared directory with "hugepages"
Versions:
cockpit-250-1.fc34.x86_64

Steps:
1. Prepare a shutoff VM
2. Create a shared directory with 'hugepages' for the VM
3. Check hugepages by "virsh freepages --all", that will show how many 4k and 2M pages on different node, if we don't create in advance, it probably be 0 for 2M pages. So, for here, it is 0 for 2M pages
4. Click run, the VM will be failed to start with an error, which is "Cannot allocate memory"
5. Add hugepages by "virsh allocpages 2M 1024"
6. Re-run the VM, it will be successful

So, maybe it is better to remind user to check and create "hugepages" manually before use shared directories with "hugepages"

KKoukiou:
Only relevant when using hugepage-backed memory.

When creating the memoryBacking XML for hugepages, use 2M page size, which seems
so be the default. Previously we were using falsely the memory size of the guest.

Following the instructions from: https://libvirt.org/kbase/virtiofs.html#host-setup

Fixes #344
Fixes #343


KKoukiou:
@yunmingyang do you want to check this PR locally so that you can verify that the issue you reported is solved?

yunmingyang:
Yeah, I will do it tomorrow

yunmingyang:
@KKoukiou After some checks locally, I think there still some little questions:
1. If we want to change "Memory backing" type and already have created shared directories, there is no way except editing VM XML. So maybe we can add a hint about that if we don't want to delete <memoryBacking> by the page.
2. After deleting the shared directory, hugepages memory will not be deleted
3. If we set a VM memory the same as hosts', there will be probably an error like "Unable to allocate 3500 pages. Allocated only 3064", then that shared directory will be never created successfully, but at host, the hugepages are created. So maybe there should be an hint
4. If we create two VMs and they both have a shared directory, firstly, run one of them, then run another one, it will fail since "virsh allocpages" will be always seems to set new pool size(which means "virsh allocpages 2M 512" will set 1G size hugepage pool no matter how many times this command is executed). So, maybe we should add "--add" for "virsh allocpages", and we also need to consider add a hint for reminding user created will be failed when the sum of memory of all VM with hugepages is close to host memory  
5. "virsh allocpages" can set hugepages in different node of a NUMA host, and maybe we should point a cellno(by --cellno) for this command. But I don't know whether it is OK without setting it since I don't have a host with NUMA, maybe I can borrow one to test later, but I think we need to consider it.
6. In some server, there will be 1G hugepages, so should we add a selection for that
7. In this [link](https://virtio-fs.gitlab.io/howto-qemu.html), it shows how to use qemu with virtiofs, also with DAX, and it seems there will be "cache-size=" after "-device vhost-user-fs-pci" checkinig by "ps -ef | grep qemu". But there is no "cache-size=" after creating VM and shared directory by the page. So I want to make sure it that since it maybe affect "Shared directory" performance.

KKoukiou:
@yunmingyang So I decided to abandon using hugepages and file based backing for shared memory and stick to memfd. Seems the best option and that's what libivrt devs confirmed. 

Wanna check this again?

yunmingyang:
Sure, let me check it tomorrow, thanks

yunmingyang:
After checking, I think it is good to me. And I think that there is still one thing that we need to pay attention, in https://libvirt.org/kbase/virtiofs.html#other-options-for-vhost-user-memory-setup, it seems that guest need to specify the NUMA topology. I am not sure whether it will affect our function. But I think it doesn't matter, I can borrow a host with NUMA to check that later.

KKoukiou:
Added release blocker as it fixes an issue.

KKoukiou:
> After checking, I think it is good to me. And I think that there is still one thing that we need to pay attention, in https://libvirt.org/kbase/virtiofs.html#other-options-for-vhost-user-memory-setup, it seems that guest need to specify the NUMA topology. I am not sure whether it will affect our function. But I think it doesn't matter, I can borrow a host with NUMA to check that later.

Let's consider this for future feature.

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -63,6 +63,10 @@ export const VmFilesystemsCard = ({ connectionName, vmName, vmState, objPath, fi
                                           actionName: _("Remove"),
                                           objectName: filesystemTarget,
                                           onClose: () => setDeleteDialogProps(undefined),
+                                          actionDescription: <TextContent>
+                                              <Text component={TextVariants.p}>{_("The current memory backing configuration will not be affected.")}</Text>
+                                              <Text component={TextVariants.p}>{_("Confirm this action.")}</Text>
```
martinpitt, line 68: What does "confirm this action" mean in this context? I don't see any additional confirmation step that the user has to take here?

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -248,21 +247,9 @@ const VmFilesystemAddModal = ({ connectionName, memory, memoryBacking, objPath,
...
-                                   name="memoryBackingType"
-                                   onChange={(_, event) => setMemoryBackingType(event.currentTarget.value)}
-                                   value="file" />
-                            <Radio id={`${idPrefix}-modal-hugepages`}
-                                   isChecked={memoryBackingType == "hugepages" }
-                                   label={_("Hugepages")}
-                                   name="memoryBackingType"
-                                   onChange={(_, event) => setMemoryBackingType(event.currentTarget.value)}
-                                   value="hugepages" />
+                            <div id={`${idPrefix}-modal-memfd-backed`}>{_("Using virtiofs requires shared memory. memfd memory backing will be automatically configured for this guest")}</div>
```
martinpitt, line 252: This is quite a technicality, and feels like an internal implementation detail. Does this really need to be exposed in the UI? (this is a long and complicated translatable string, after all)

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -28,7 +28,7 @@ import {
     List, ListItem,
     Modal,
     Popover,
-    Radio, Text, TextContent, TextInput, TextVariants, Tooltip
```
martinpitt: This "Stop providing options for shaed memory when creating shared directories" commit has a typo (not important), but refers to #350 which is *this* PR. I think it wants to refer to #343 and #344 instead, and having correct issue links in changelogs is fairly useful/important IMHO.

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -146,7 +145,7 @@ const VmFilesystemAddModal = ({ connectionName, memory, memoryBacking, objPath,
         if (Object.getOwnPropertyNames(validationFailed).length == 0) {
             domainSetMemoryBacking({
                 connectionName, objPath,
-                type: memoryBackingType,
+                type: "memfd",
```
martinpitt, line 148: Will that memfd contain the actual file system, or just some metadata for communciation? It seems fine for the latter, but squeezing possibly giga- or terabytes into a memfd is an utter abuse (and may not even work). If this contains the file system, what happens if you try to create a non-trivial one? Do our tests try to create a memfd-backed FS which exceeds the RAM size?

martinpitt (changes requested):
Thanks!

src/components/vm/filesystems/vmFilesystemsCard.jsx
KKoukiou: When deleting resources we always show the 'Confirm this action' text. We do exactly the same for disks, nics etc.

src/components/vm/filesystems/vmFilesystemsCard.jsx
KKoukiou: The reason I exposed it - it because there are 3 available memory backing options from libvirt. After discussions with the virt team - I went with this one (memfd). I thought the users might be as well informed about their guess config changing to enable memory backing once they set up a shared filesystem.

In addition - memory backing configuration is added - if not yet present in the guest XML - but it's not removed when removing the shared folders. For this reason - I find it important that the user get's notified about it both on addind and also on removal.

src/components/vm/filesystems/vmFilesystemsCard.jsx
KKoukiou: The shared memory is used only for metadata according to https://lwn.net/Articles/774495/

Quoting the file is you 're not interested in all the doc:
```

- For data access inside guest, mmap portion of file in QEMU address
  space and guest accesses this memory using dax. That way guest page
  cache is bypassed and there is only one copy of data (on host). This
  will also enable mmap(MAP_SHARED) between guests.

- For metadata coherency, there is a shared memory region which contains
  version number associated with metadata and any guest changing metadata
  updates version number and other guests refresh metadata on next
  access. This is still experimental and implementation is not complete.
```

Does this cover you @martinpitt ? honestly I don't know the implementation details of this technology, but I read the libvirt/virtiofs docs and they did not specify any memory constraints of the host when using memfs for shared memory.

martinpitt: Thanks! That sounds fine then, memfd is a rather nice technology for such stuff (and similar to classic shared memory, anyway).

martinpitt:
@KKoukiou : Can you please rebase this? The pixel tests are messy, and I figure your branch just has an outdated reference (they still contain the "unattended installation" bits)

src/components/vm/filesystems/vmFilesystemsCard.jsx
KKoukiou: Fixed.

src/components/vm/filesystems/vmFilesystemsCard.jsx
KKoukiou: As we talked over IRC _ removed the string.

martinpitt (approved):
Thanks! Sorry for me being so picky :grin:

KKoukiou:
> Thanks! Sorry for me being so picky grin

Please do :)

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -122,7 +122,6 @@ export const VmFilesystemActions = ({ connectionName, memory, memoryBacking, obj
 const VmFilesystemAddModal = ({ connectionName, memory, memoryBacking, objPath, setIsOpen, vmName, vmState }) => {
```
skobyda, line 122: "memoryBacking" is not not used anywhere in the whole file

skobyda (changes requested):

src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -242,24 +241,6 @@ const VmFilesystemAddModal = ({ connectionName, memory, memoryBacking, objPath,
                                    onToggle={() => setAdditionalOptionsExpanded(!additionalOptionsExpanded)}
                                    isExpanded={additionalOptionsExpanded}>
                     <Grid hasGutter>
```
skobyda, line 243: This Grid with gutter is now unecessary since there is only one element inside it

skobyda (changes requested):

src/components/vm/filesystems/vmFilesystemsCard.jsx
KKoukiou: Removed - thanks!

skobyda (approved):
Thanks

PR #351, opened 2021-09-01
Fix validation check for the source field in 'Add shared directory' dialog

### LINKED ISSUE #348: machines: There is an error missed when only inputting "Mount tag" in "Add shared directory" dialog
Versions:
cockpit-250-1.fc34.x86_64

Steps:
1 Prepare a shutoff VM
2 Enter the VM detail page
3 Input to "Mount tag" like "/mnt", then clicking "Share". There will be no error shown, but there should be an error shown below "Source path" input. It probably because these two <FormGroup> use the same validated in vmFilesystemCard.jsx

KKoukiou:
The test already exists, but since the 'tag' was also invalid, both
fields got invalidated.

Fixes #348


martinpitt (approved):
Thanks! Retrying a [flake](https://logs.cockpit-project.org/logs/pull-351-20210901-143915-1ad1a84b-rhel-9-0/log.html)

PR #355, opened 2021-09-02
Introduce virtqemud unit and check status of both .service and .socket

skobyda:
- Since fedora 35, virtqemud service should be preferred instead of libvirtd service
- Because socket activation is used on all distros now, we should never turn on the service manually, but instead let socket turn it on when it needs it. It's also acceptable for a service to be turned of after few minutes of inactivity. Therefore it should be virtqemud/libvirtd.socket's state which we should look if it's running at all times.


src/README.md
```diff
@@ -10,7 +10,7 @@ In the vagrant:
     $ sudo virt-host-validate
     $ sudo su
     # systemctl start virtlogd.socket
-    # systemctl start libvirtd.service
+    # systemctl start libvirtd.socket or # systemctl start virtqemud.socket
```
KKoukiou, line 13: If you want to comment out the alternative to allow copy pasting into terminal put the # before the or.

src/actions/store-actions.js
```diff
@@ -116,9 +117,16 @@ export function undefineVm({ connectionName, name, id, transientOnly }) {
...
-export function updateLibvirtState(state) {
+export function updateLibvirtService(state) {
     return {
-        type: UPDATE_LIBVIRT_STATE,
+        type: UPDATE_LIBVIRT_SERVICE,
+        state,
+    };
+}
+
+export function updateLibvirtSocket(state) {
```
KKoukiou, line 127: You don't use this anywhere at this point - just put into the store data which you plan to use in other places in the code.

src/components/libvirtSlate.jsx
```diff
@@ -43,25 +43,29 @@ const LibvirtSlate = ({ loadingResources, libvirtService }) => {
 
         enableLibvirt({
             enable: libvirtEnabled,
-            serviceName: libvirtService.unit.Id
+            serviceName: libvirtUnit.unit.Id
```
KKoukiou, line 46: Since this is not always the .service can we enable the property to unitName?

KKoukiou (changes requested):

KKoukiou:
Closing in favor of #387

PR #356, opened 2021-09-06
Replace adding/removing vNICs API's with virt-xml

skobyda:
I'm just trying out the waters with the pull request to see that we agree with the approach used for using virt-xml.
Once we agree on a correct approach, APIs for adding/removing disks, filesystems, memory-blocking can be replaced with virt-xml too.

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -40,7 +40,6 @@ import {
 } from '../actions/store-actions.js';
 import {
     getDiskXML,
-    getIfaceXML,
     getFilesystemXML,
     getMemoryBackingXML,
 } from '../libvirt-xml-create.js';
@@ -61,7 +60,6 @@ import {
     getDiskElemByTarget,
     getDoc,
     getElem,
-    getIfaceElemByMac,
     getSingleOptionalElem,
     parseDumpxml,
 } from '../libvirt-xml-parse.js';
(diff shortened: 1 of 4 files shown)


src/libvirtApi/domain.js
```diff
@@ -137,10 +135,22 @@ export function domainAttachDisk({
...
-    return domainAttachDevice({ connectionName, vmId, permanent, hotplug, xmlDesc });
+    let update = "";
+    if (hotplug)
+        update = "--update";
+    let define = "--define";
+    if (hotplug && !permanent)
+        define = "--no-define";
+
+    return cockpit.script(
+        `virt-xml -c qemu:///${connectionName} ${vmName} --add-device --network mac=${mac},type=${sourceType},source=${source},model=${model} ${define} ${update}`,
```
KKoukiou, line 151: When 'generate automatically' is chosen for the MAC the mac variable is undefined and clicking add would result in an error:
```  ERROR XML error: unable to parse mac address 'undefined'  ```

KKoukiou (changes requested):

KKoukiou:
Reg the virt-xml approach I am happy with the change.

KKoukiou:
@skobyda I updated the PR _ it was just a tiny tweak :)

KKoukiou (commented):
testNICAdd consistently fails on some distros - with:
> warning: ERROR    Error: --network mac=52:54:00:a5:f8:c0,type=direct,source=eth1,model=virtio: The MAC address '52:54:00:a5:f8:c0' is in use by another virtual machine.

skobyda:
Fixed the failing test and updated pixel reference (the MAC address filled in NIC Add dialog changed)

KKoukiou (changes requested):
Previously one could choose 'Generate automatically' and the API would be fed with an undefined value for the mac address, resulting in an auto generation of a new / non conflicting mac address. Now this commit - stops testing this functionality of 'Generate automatically'  if I am seeing correctly. 

Is virt-xml not able to auto generate mac addresses?

KKoukiou (commented):
Ok I just realized that autogeneration of macs works still fine - but then why did you set the mac in all the tests now?

KKoukiou (commented):
OK I see now - we had a preset mac address in the dialog class. How about removing it and let the dialog generate for us? (So that we tests that functionality as well)

KKoukiou (commented):
Alright - let's get this in - the tests rely on knowing the mac address for doing the xmllint checks. And we need to add a tests for mac address autogeneration as a followup.

KKoukiou (approved):

KKoukiou (approved):

PR #359, opened 2021-09-07
Perform virt-install feature detection in the parent most component

KKoukiou:
This will prevent the storm of virt-install processes getting spawned
and eating CPU resources.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -108,6 +108,12 @@ class AppActive extends React.Component {
             resourceHasError: {},
             notificationIdCnt: 0,
             path: cockpit.location.path,
+            /* virt-install feature support checks */
+            cloudInitSupported: undefined,
+            downloadOSSupported: undefined,
+            unattendedSupported: undefined,
+            unattendedUserLogin: undefined,
+            virtInstallAvailable: undefined,
         };
         this.onAddErrorNotification = this.onAddErrorNotification.bind(this);
         this.onDismissErrorNotification = this.onDismissErrorNotification.bind(this);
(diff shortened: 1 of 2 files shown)


martinpitt (dismissed):
Ooh, good one!

mvollmer (dismissed):

src/app.jsx
```diff
@@ -108,6 +108,12 @@ class AppActive extends React.Component {
             resourceHasError: {},
             notificationIdCnt: 0,
             path: cockpit.location.path,
+            /* virt-install feature support checks */
+            cloudInitSupported: undefined,
```
mvollmer, line 112: We could put all virt-install related booleans into a structure, then passing them around is more generic.  Like
```
virtInstallFeatures: {
   available: undefined,
   cloudInitSupport: undefined,
   ...
}
```

KKoukiou: ah sure - let me repost then.

KKoukiou: Created the struct when passing to the child component. For having the struct like this already in the state I feel it makes the code uglier, as each of these properties are updates independently.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1191,10 +1167,10 @@ export class CreateVmAction extends React.Component {
                     vms={this.props.vms}
                     osInfoList={this.props.systemInfo.osInfoList}
                     onAddErrorNotification={this.props.onAddErrorNotification}
-                    cloudInitSupported={this.state.cloudInitSupported}
-                    downloadOSSupported={this.state.downloadOSSupported}
-                    unattendedSupported={this.state.unattendedSupported}
-                    unattendedUserLogin={this.state.unattendedUserLogin}
+                    cloudInitSupported={this.props.virtInstallFeatures.cloudInitSupported}
```
mvollmer, line 1170: The idea was that this could just be `virtInstallFeatures={this.props.virtInstallFeatures}`, otherwise wrapping the individual props into a structure is no win at all, no?

src/app.jsx
mvollmer: Actually, it is possible to run `virt-install --install=? --cloud-init=? --unattended=?`, but that makes the parsing more difficult, I guess.

KKoukiou: This last command won't work because it early fails if ones of these is missing.
That means that I need to update the state like this for each of the calls:
 this.setState({ virtInstallFeatures: { ...this.state.virtInstallFeatures, downloadOSSupported: true } }), ..

This seems actually uglier than the initial state - don't you think?

mvollmer: Yes, let's not bother with this. :-)

mvollmer (approved):

PR #361, opened 2021-09-07
Add popover with documentation about the network options in NIC add dialog

### LINKED ISSUE #285: Clarify exact meaning & limitations of 'direct attachment' network interface type
When adding a network interface, there are three options for Interface Type.

* Virtual network: pretty clear what this means - attaches to a libvirt-defined network
* Bridge to LAN: pretty clear what this means - attaches to an existing bridge
* Direct Attachment: the mystery option!

I've done some rudimentary user research and when googling for [cockpit "direct attachment"](https://www.google.com/search?hl=en&q=cockpit%20%22direct%20attachment%22) there are not so many hits explaining what "Direct Attachment" means\*, but several from users puzzled as to why their newly created virtual machine is able to reach other hosts on their lan, but not the host running the VM, or other VMs on the same host.

* https://forum.openmediavault.org/index.php?thread/33829-virtual-machine-with-cockpit-with-networking-set-to-direct-attachment-issue/
* https://serverfault.com/questions/1013111/guest-vms-and-host-cant-ping-each-other
* https://github.com/cockpit-project/cockpit/issues/14676 😉

My first thought was changing the option to "Direct Attachment (macvtap)". But `macvtap` probably won't mean much to the users who asked the above questions (I happily used it myself for quite a while before I ever tried to create a second VM on the same host, and discovered the limitations of `macvtap` the hard way...

So my suggestion is to display some helpful text when "Direct Attachment" is chosen, explaining the following:

1. "Direct Attachment" requires `eth0` to be attached to a switch operating in Virtual Ethernet Port Aggregator (VEPA) mode for full network functionality.
2. Without a switch operating in VEPA-mode, the network interface will not be be to communicate with other virtual machines running on the same host, nor with the host itself.
3. If this communication is desired then try using 'Virtual network' or 'Bridge to LAN' instead.

I guess the text _could_ mention that the interface can be run in bridge/private/passthrough mode as well, but I don't think there's a way to change the mode in the `cockpit-machines` UI so probably best to omit this information.

Alternative text for the first paragraph above:

1. "Direct Attachment" attaches the guest to `eth0` via a `macvtap` interface. ~~As such,~~ for full network functionality, `eth0` must be attached to a switch operating in Virtual Ethernet Port Aggregator (VEPA) mode.

This is better in that it has the magic word `macvtap`, so people who already knows what that implies will instantly understand what it implies. But it's worse in that it front loads the explanation with more jargon.

----

\* unless you know enough about the software stack to dig into [RHEL's virtualization manual](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_virtualization/configuring-virtual-machine-network-connections_configuring-and-managing-virtualization#direct-attachment-of-the-virtual-network-device_types-of-virtual-machine-network-connections), or the 101-page long [libvirt Domain XML Format](https://www.libvirt.org/formatdomain.html#elementsNICSDirect) page.

KKoukiou:
* [x] update pixel tests

Fixes #285

The popover does not work nicely in mobile mode. See https://github.com/patternfly/patternfly-react/issues/5993

The documentation was copied from
https://libvirt.org/formatdomain.html#network-interfaces.

### DIFF
--- a/src/components/vm/nics/nicBody.jsx
+++ b/src/components/vm/nics/nicBody.jsx
@@ -20,9 +20,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
+    Flex,
     FormGroup,
     FormSelect, FormSelectOption,
+    Popover, PopoverPosition,
+    Text, TextContent, TextVariants,
 } from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 
 import cockpit from 'cockpit';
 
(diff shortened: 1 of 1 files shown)


src/components/vm/nics/nicBody.jsx
```diff
@@ -75,16 +79,35 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
         availableNetworkTypes = [
-            { name: 'network', desc: 'Virtual network' },
-            { name: 'bridge', desc: 'Bridge to LAN' },
-            { name: 'direct', desc: 'Direct attachment' },
+            ...virtualNetwork,
+            {
+                name: 'bridge',
+                desc: 'Bridge to LAN',
+                detailHeadline: _("This is the recommended config for general guest connectivity on hosts with static wired networking configs."),
+                detailParagraph: _("Provides a bridge from the guest virtual machine directly onto the LAN. This assumes there is a bridge device on the host physical machine which has one or more of the host physical machines physical NICs enslaved.")
```
martinpitt, line 95: This is very long and hard to read due to the repetitions. How about

```suggestion
                detailParagraph: _("Provides a bridge from the guest virtual machine directly onto the LAN. This needs a bridge device on the host with one or more physical NICs.")
```

martinpitt (approved):
Danke!

martinpitt:
Please squash this when landing.

KKoukiou:
Repushed to add the mobile pixel tests we just introduced.

martinpitt (approved):
Unfortunately the [diff view](https://github.com/cockpit-project/pixel-test-reference/compare/8f6f3efee309342cee1a2bea1f485c67833b67c0...67d8c720f5c19093c8be00ab2c37d3e1e74a8f60?short_path=50afbca#diff-50afbca5eb9709efacd1e3ae3b4cb7416def97a422921c7595a4b0389587341d) is broken for all mobile images, but "View file" works at least. (This was already like that in the original PR that introduced mobile tests, I just downloaded the branch)

Thanks!

PR #363, opened 2021-09-08
Stop exposing unattended installation option for ISO installation media

KKoukiou:
* [x] update pixel tests

Unattended is not working for linux ISO media which we pass with
--cdrom. See https://bugzilla.redhat.com/show_bug.cgi?id=1868594 for
details.

Till we get a response here https://github.com/virt-manager/virt-manager/issues/296
let's disable this option that's very much broken.


martinpitt (approved):
Danke!

PR #364, opened 2021-09-08
Fix CPU editing for session connection guests

### LINKED ISSUE #44: machines: editing cpu model does not work for session connection
It's because we call virt-xml script with superuser try, and session connection does not exist for root of course.

KKoukiou:
Fixes #44


martinpitt (approved):
Thanks! A bit sad to not cover this with tests, but at least the system one does have them.

PR #366, opened 2021-09-09
Add popover with documentation about the Direct attachment network type - DON'T GET IN 252 release (changes translations)

### LINKED ISSUE #285: Clarify exact meaning & limitations of 'direct attachment' network interface type

yrro:
Fixes #285

### DIFF
--- a/src/components/vm/nics/nicBody.jsx
+++ b/src/components/vm/nics/nicBody.jsx
@@ -20,13 +20,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
+    Button,
     Flex,
     FormGroup,
     FormSelect, FormSelectOption,
     Popover, PopoverPosition,
     Text, TextContent, TextVariants,
 } from '@patternfly/react-core';
-import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import { ExternalLinkSquareAltIcon, OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 
 import cockpit from 'cockpit';
 
@@ -97,7 +98,20 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
             {
                 name: 'direct',
                 desc: 'Direct attachment',
-                detailParagraph: _("Uses the macvtap driver to attach the guest's NIC directly to a specified physical interface of the host machine.")
+                detailHeadline: _("This is the recommended config for high performance or enhanced security."),
+                detailParagraph: _("In the default 'vepa' mode, switching is offloaded to the external switch. If the switch is not VEPA-capable, communication between guest virtual machines, or between a guests and the host is not possible."),
+                externalDocs: (
+                    <Button isInline
+                            className='ct-external-docs-link'
+                            variant="link"
+                            component="a"
+                            icon={<ExternalLinkSquareAltIcon />}
+                            iconPosition="right"
+                            target="__blank"
+                            href="https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface">
+                        {_("more info")}
+                    </Button>
+                )
             },
         ];
     } else {
@@ -149,11 +163,15 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
                        labelIcon={
                            <Popover aria-label={_("Interface type help")}
                                     position={PopoverPosition.bottom}
+                                    enableFlip={false}
                                     bodyContent={<Flex direction={{ default: 'column' }}>
                                         {availableNetworkTypes.map(type => <TextContent key={type.name}>
                                             <Text component={TextVariants.h4}>{type.desc}</Text>
                                             <strong>{type.detailHeadline}</strong>
-                                            <p>{type.detailParagraph}</p>
+                                            <p>
+                                                {type.detailParagraph}
+                                                {type.externalDocs}
+                                            </p>
                                         </TextContent>)}
                                     </Flex>}>
                                <button onClick={e => e.preventDefault()} className="pf-c-form__group-label-help">
--- a/src/machines.scss
+++ b/src/machines.scss
@@ -83,3 +83,7 @@
 .ct-machines-select-unit, .ct-machines-size-input {
     max-width: 5rem;
 }
+
+.ct-external-docs-link.pf-c-button.pf-m-link.pf-m-inline {
+  padding-left: var(--pf-global--spacer--sm);
+}


src/components/vm/nics/nicBody.jsx
```diff
@@ -97,7 +97,8 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
             {
                 name: 'direct',
                 desc: 'Direct attachment',
-                detailParagraph: _("Uses the macvtap driver to attach the guest's NIC directly to a specified physical interface of the host machine.")
+                detailParagraph: _("Uses the macvtap driver to attach the guest's NIC directly to a specified physical interface of the host machine."),
+                detailHeadline: _("This config is suitable where high performance or enhanced security is required. Varying degrees of isolation between guests, or between the guest and the host are possible, depending on the mode and the capabilities of the network to which the physical interface is connected.")
```
KKoukiou, line 101: Can you bring the headline above the paragraph as they are presented in the UI? 

Shall we start with 'This is the recommended config' like the two options above? 

Also where did you find this information? (high performance ... enhanced security ...)

Shouldn't we mention in the detailParagraph about the VEPA which is the mode it will use by default if chosen?

yrro, line 101: I've avoided using the magic jargon of VEPA/Reflective Relay/Hairpinning because cockpit-machines doesn't let you choose the `mode` of a Direct attachment network type. If it did then I think the tooltip for that option (`vepa`/`bridge`/`private`/`passthrough`) would be the appropriate place to explain the details.

KKoukiou: The default mode is VEPA. Do you think it needs explanation in this case?

yrro: I'm trying to sum up all the use cases of the four modes in a couple of words. I think adding extra detail it's too helpful since cockpit-machines doesn't let you change the mode.

The background info comes from https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface from which I take away:

`vepa`: used to offload switching from the host to the network (performance)  
`bridge`: I think this avoids some overhead for VM-to-VM traffic, but VM-to-host communication is not possible (maybe unless you have a VEPA-capable switch, not sure?) (performance)  
`private`: used when you want to explicitly prevent VM-to-VM and VM-to-host traffic (security)
`passthrough`: for SR-IOV (performance)

But since the default mode is `vepa`, I will add some more details about this use case & the requirement of a VEPA-capable switch.

yrro: I'll move the headline above the paragraph for all the options, in a separate commit.

KKoukiou: But cockpit-machines will automatically use VEPA mode - why should be summarize the rest?

If then the user goes and adjusts the XML they probably are fully aware of what they are doing.

src/components/vm/nics/nicBody.jsx
```diff
@@ -97,6 +97,7 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
             {
                 name: 'direct',
                 desc: 'Direct attachment',
+                detailHeadline: _("This is the recommended config for high performance or enhanced security. In the default 'vepa' mode, a guest virtual machine's packets are always sent via the physical interface. Communication between guest virtual machines, or between a guest virtual machine and the host is not possible unless the physical interface is connected to a VEPA-capable network switch. Other modes are available; consult the libvirt documentation for further information."),
```
KKoukiou, line 100: This is rather  wordy but it definitely explains the option. @martinpitt do you want to have a second word here?

yrro: Is it possible to use markup to render "vepa" in monospace, or to include a nice link directly to https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface?

KKoukiou: Should we follow libvirt docs and say ' all VMs' packets are sent to the external bridge' instead of the current wording?
I also just checked it live in the popover and it's incredibly long, we need to shorten it.

yrro: How about:

> In the default 'vepa' mode, switching is offloaded to the ethernet switch. If the switch is not VEPA-capable, communication between guest virtual machines, or between a guests and the host is not possible. See the linked documentation for information about the other modes.

KKoukiou: I like it! I am only hesitant about being so verbose about existanse of 'other modes' since we offer the option to change the mode. 

All in all - I would keep the whole suggested text - but for the last sentence. 
Instead of that I would rather use a plain 'more info' link as I had in the example branch I sent above.

yrro: Sounds good

KKoukiou (approved):
The popover works terribly on mobile or laptop screen sizes - but it's not your fault - will send a followup.

PR #367, opened 2021-09-10
packit: Force dist download

martinpitt:
The sandcastle env has too little RAM to do a webpack build, so it
inevitably fails. Let's fail in a faster and more obvious way instead.


martinpitt:
This avoids failures to look like https://dashboard.packit.dev/results/srpm-builds/34763

KKoukiou (approved):

PR #369, opened 2021-09-10
Parse supported disk bus types from domcapabities and filter the displayed select options

KKoukiou:
* [ ] tested against a ppc machine

Move fetching of the domain domcapabities one component up so as to avoid
doing the API call twice, as it's now needed in two places.

This affects the Add and Edit disk dialogs.

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1862779

### DIFF
--- a/src/components/vm/disks/diskAdd.jsx
+++ b/src/components/vm/disks/diskAdd.jsx
@@ -141,7 +141,9 @@ class AdditionalOptions extends React.Component {
     }
 
     render() {
-        const displayBusTypes = diskBusTypes[this.props.device].map(type => ({ value: type }));
+        const displayBusTypes = diskBusTypes[this.props.device]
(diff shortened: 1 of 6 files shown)


KKoukiou:
@yunmingyang could you please pre-verify this? I don't have access to any ppc machine to test this.

KKoukiou (commented):
@martinpitt this needs to be tested against ppc theoretically for preverification - but honestly the beaker story was not so much straightforward.
This is pretty straightforward change, are you fine with reviewing it as is, and the QEs can deal with the bugzilla verification after it's in.

martinpitt (approved):
Thanks! Agreed that the change here is not platform specific, and looks good -- if there was something wrong with parsing the disk types, it would fail on x86 also.

PR #371, opened 2021-09-13
test: add checks for NICs pluging and unpluging

yunmingyang:


test/check-machines-nics
```diff
@@ -85,6 +82,34 @@ class TestMachinesNICs(VirtualMachinesCase):
         b.wait_in_text("#vm-subVmTest1-network-3-source", "virbr0")
         b.wait_in_text("#vm-subVmTest1-network-3-ipv4-address", "192.168.122.")
 
+        # Check Plug and unplug NICs
+        def checkPlugAndUnplugNICs():
```
KKoukiou, line 86: Can you create a new test for it instead of putting in it in the existing one?

test/check-machines-nics
```diff
@@ -85,6 +82,34 @@ class TestMachinesNICs(VirtualMachinesCase):
...
+            mac = b.text("#vm-subVmTest1-network-1-mac")
+
+            b.wait_in_text("#vm-subVmTest1-network-1-state", "up")
+            b.wait_in_text("#delete-vm-subVmTest1-iface-1 + button", "Unplug")
+
+            # Click "Unplug"
+            b.click("#delete-vm-subVmTest1-iface-1 + button")
```
KKoukiou, line 93: That's really weird way to select the unplug button. You essentially say - take the one after the delete button. 
Please add in the JS an id for the unplug/plug buttons and use that instead.

KKoukiou (changes requested):

test/check-machines-nics
yunmingyang: Sure, re-push it

test/check-machines-nics
yunmingyang: OK, let me add the id in another PR

KKoukiou (approved):
This looks ok, I am a bit afraid that the live hotunplugging/plugging will not be stable, as it relies on the OS being responsive etc.

But let's see what tests say.

PR #374, opened 2021-09-14
Use virt-xml to change vcpu settings - this is an attempt to unblock https://github.com/cockpit-project/bots/pull/2401

KKoukiou:


src/libvirtApi/domain.js
```diff
@@ -761,11 +759,13 @@ export function domainSetVCPUSettings ({
     threads,
     isRunning
 }) {
-    return call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' })
-            .then(domXml => {
-                const updatedXML = updateVCPUSettings(domXml[0], count, max, sockets, cores, threads);
-                return call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
-            });
+    const opts = { err: "message", environ: ['LC_ALL=C'] };
```
martinpitt, line 762: Not necessary for this PR, but in general: C.UTF-8 is a thing everywhere since RHEL 8, and preferable in general.

src/libvirtApi/domain.js
```diff
@@ -761,11 +759,13 @@ export function domainSetVCPUSettings ({
...
+    if (connectionName === 'system')
+        opts.superuser = 'try';
+
+    return cockpit.script(
+        `virt-xml -c qemu:///${connectionName} --vcpu ${max},vcpu.current=${count},sockets=${sockets},cores=${cores},threads=${threads} ${name} --edit`, opts
```
martinpitt, line 767: Can we please not introduce shell again? We just got rid of this with converting the huge shell scripts to Python. If the name contains spaces, quotes, or any funny characters, this is going to blow up in the user's face. Let's please use cockpit.spawn() and assemble a proper array. This is not significantly more difficult, but dramatically more robust.

test/check-machines-settings
```diff
@@ -390,24 +390,24 @@ class TestMachinesSettings(VirtualMachinesCase):
         b.click("#cpu-config-dialog-apply")
         b.wait_visible("#cpu-tooltip")
 
+        dom_xml = "virsh -c qemu:///system dumpxml --inactive subVmTest1"
+
+        xmllint_elem = "{0} | xmllint --xpath 'string(//domain/cpu/@mode)' - 2>&1 || true".format(dom_xml)
+        wait(lambda: "host-passthrough" in m.execute(xmllint_elem).strip())
```
martinpitt, line 398: Was moving that check from "after shutting off" to "still running" the magic trick to pacify the test? :grin: This feels unrelated to virt-xml, unless I'm missing something? Not a biggie, but might be cleaner to split it into a separate commit.

test/check-machines-settings
```diff
@@ -384,6 +384,8 @@ class TestMachinesSettings(VirtualMachinesCase):
         b.wait_in_text("#vm-subVmTest1-state", "Running")
         self.goToVmPage("subVmTest1")
 
+        wait(lambda: "login as 'cirros' user" in m.execute("cat {0}".format(args["logfile"])), delay=3)
```
martinpitt, line 387: Oh -- changing CPUs while the kernel is still booting might screw up too much in the guest? (although I didn't expect that we actually care what's running inside). But if it just slows down the test enough, that's fine :grin:

martinpitt (changes requested):
Nice job tracking this down! I also like the cleanups, and getting away from cobbling together our own XML. I'm afraid I am just really insisting on not building shell strings any more.

test/check-machines-settings
KKoukiou: Actually not - I can just revert this.

src/libvirtApi/domain.js
KKoukiou: Of course! Sorry for not realizing this!

test/check-machines-settings
KKoukiou: this is actually the commit that stabilized the test. the others are just test improvents and cleanups on the code. The 'why' part I can't answer :/

test/check-machines-settings
KKoukiou: This was not needed, reverted.

KKoukiou:
Also did clean rebase to avoid that weird 'pixel tests changed' comment.

test/check-machines-settings
martinpitt: Sure, let's leave some mysteries :grin: I'm fine with it, I was just thinking aloud what could happen here. But I suppose it's just a glorified sleep.

KKoukiou: What was happening in the test, is that the API call which was changing the CPU settings was just not taking effect - or IOW leaving the previous CPU settings values, and that without some error message.

So indeed there was some race when changing the XML of the just created domain. 

I think this scenario for the users to hit is impossible so I am also ok with this change.

martinpitt (approved):
Thanks! Very nice cleanup all around!

PR #377, opened 2021-09-15
fmf: Go back to chromium on Fedora, enable CentOS 8 in packit

martinpitt:
Firefox has not been installable in rawhide for a long time, which
breaks our tests. Go back to chromium for Fedora, but keep Firefox on
RHEL/CentOS as that's its official browser.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -12,11 +12,17 @@ LOGS="$(pwd)/logs"
 mkdir -p "$LOGS"
 chmod a+w "$LOGS"
 
-# install browser; on RHEL, use firefox-nightly as chromium broke CDP on rhel-9
-# Install firefox to pull in all the deps
-dnf -y install firefox
-curl --location 'https://download.mozilla.org/?product=firefox-nightly-latest-ssl&os=linux64&lang=en-US' | tar -C /usr/local/lib/ -xj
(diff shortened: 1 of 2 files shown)


martinpitt:
Lo and behold, green! This likely means that we can enable fedora-35 (and disable 33) as well now. But first @marusak , WDYT?

test/browser/browser.sh
```diff
@@ -12,11 +12,17 @@ LOGS="$(pwd)/logs"
 mkdir -p "$LOGS"
 chmod a+w "$LOGS"
 
-# install browser; on RHEL, use firefox-nightly as chromium broke CDP on rhel-9
-# Install firefox to pull in all the deps
-dnf -y install firefox
-curl --location 'https://download.mozilla.org/?product=firefox-nightly-latest-ssl&os=linux64&lang=en-US' | tar -C /usr/local/lib/ -xj
-ln -s /usr/local/lib/firefox/firefox /usr/local/bin/
+# install browser; on RHEL, use firefox
+if grep -q 'ID=.*rhel' /etc/os-release; then
```
martinpitt, line 16: Let's do the same on CentOS 8, and enable that in packit.

martinpitt: Done.

martinpitt:
Oops -- I'll send that as a new PR then.

PR #387, opened 2021-09-20
Stop trying to guess the name of the libvirt service/socket (fedora-35 necessary changes)

KKoukiou:
And leave the service activation to libvirt itself.

In all our supported distros the libvirt service is socket activated.
The socket is running by default, from the very first moment the user
installs the package. A stopped libvirt socket means that the user
manually changed the default state, and should be left alone to do so.

In addition to that, the libvirt daemon changed recently to a modular
achitecture, so instead of a single monolithic libvirtd daemon we now
get one for each driver (domain, storage, network, interfaces etc). Some distros already
opted in the new modular daemons, (see https://www.mail-archive.com/libvir-list@redhat.com/msg219931.html)
but many keep using the old monolithic libvirtd daemon.

As now we will have many sockets to check for their state, we decided to get
rid of any libvirt service/socket name detection logic from
cockpit-machines.
This logic was previously used to detect if the service was running
and if not we showed the user an empty state screen.
That also helped us to listen to socket/service changes and update
the UI accordingly.

This commit replaces this logic will a simple API call in the initial
render of this app which allows us to check if a connection type is
available; if that fails for any of the system/session connections, that means that the libvirt service is not available for the connection.

For the same reason of the name detection logic being problematic, let's
stop providing to the cockpit-machines users the ability to start/enable
the libvirt service/socket from the cockpit-machines UI directly.


test/machineslib.py
```diff
@@ -198,6 +198,9 @@ def setUp(self):
             # https://bugs.launchpad.net/ubuntu/+source/libvirt-dbus/+bug/1892757
             m.execute("usermod -a -G libvirt libvirtdbus")
 
+        # we don't have configuration to open the firewall for local libvirt machines, so just stop firewalld
+        m.execute("systemctl stop firewalld")
+
         self.startLibvirt()
```
martinpitt, line 204: So that means even *starting* libvirt fails with firewalld running? This is a release-critical bug,  can you please report it?

(Also: We should really figure this out properly, and fix c-machines with firewalld running)

martinpitt (changes requested):
Thanks! That's a nice cleanup as well. One question  (doesn't affect the code). Can you please rebase this, to run this through packit? That's quite important in this case IMHO.

martinpitt:
Still [lots of failures on fedora-35](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-387-20210922-070305-7e4c94e8-fedora-35/log.html#6-2), capturing log URL. Looks like you need to start some libvirt-network service as well perhaps?

test/machineslib.py
KKoukiou: ok removed the commit that does that - and will make a naughty for it once I figure this out what exactly made the test fail

martinpitt: Sorry for the misunderstanding -- my "requested changes" was just for the rebase. But sure, we can land this PR without a green f35, and fix that separately.

test/machineslib.py
```diff
@@ -85,11 +85,19 @@ def waitNetworkRow(self, networkName, connectionName="system", present="true"):
         else:
             b.wait_not_present(network_row)
 
+    def getLibvirtServiceName(self):
+        m = self.machine
+
+        if m.image in ["fedora-35"]:
+            return "virtqemud"
+        else:
+            return "libvirtd"
```
martinpitt, line 94: We should invert the condition here. AFAIUI this wouldn't fail on fedora-36, as the monolithic daemon is still a thing, isn't it? If it will fail on F36/RHEL 10/debian-testing etc., then it's ok, and we can invert it once we get more distros with the split daemon.

martinpitt (approved):
Thanks! Still rather ad-hoc, but we'll clean this up over time when more OSes require this. Great work!

test/machineslib.py
KKoukiou: It will fail for sure exactly how fedora-35 tests failed before this PR. 
For the record, the monolithic daemon will be removed much later (fedora 38 they assume)

martinpitt (changes requested):
[fails all over](http://artifacts.dev.testing-farm.io/e2684e8c-5932-456b-8a7f-9f5719e588c1/) on the testing farm: :cry: 
```
error: Failed to connect socket to '/var/run/libvirt/virtnetworkd-sock': No such file or directory
error: failed to get network 'default'
```

As packit/f35 currently succeeds, this is a regression I'm afraid.

test/check-machines-migrate
```diff
@@ -45,6 +45,11 @@ class TestMachinesMigration(VirtualMachinesCase):
         machine2.execute("hostnamectl set-hostname machine2")
         machine2.execute(f"systemctl restart {self.getLibvirtServiceName()}")
 
+        if machine2.image in ["fedora-35"]:
+            # HACK for missing deps on virtlockd/virtlogd sockets - needed till f-35 uses libvirt v7.7.0
+            # https://github.com/libvirt/libvirt/commit/88c5b9f827779ae6fe5a6f08100a4b6184492a1c
+            machine2.execute("systemctl start virtlockd.socket && systemctl start virtlogd.socket")
```
martinpitt, line 51: I had to read this three times :grinning:

martinpitt (approved):
Thanks! Looks great, good to land on green. I also commented on the libvirt issue, thanks for filing!

test/browser/browser.sh
```diff
@@ -62,6 +62,12 @@ systemctl start firewalld
 firewall-cmd --add-service=cockpit --permanent
 firewall-cmd --add-service=cockpit
 
+if grep -Eq 'PLATFORM_ID==.*(f35)' /etc/os-release; then
```
martinpitt, line 65: This [did not work](http://artifacts.dev.testing-farm.io/1554aa3c-c62e-41b1-89b3-940d9a2cf4c3/). os-release has no double `==`, but it should work with a single one.

martinpitt (changes requested):

test/browser/browser.sh
martinpitt: Fixe (FPed your branch)

martinpitt:
At least the [failure](http://artifacts.dev.testing-farm.io/82aebb2d-b4c9-4fd1-9482-bfa922c5a711/) looks different now..
```
error: Failed to define network from /etc/libvirt/qemu/networks/default.xml
error: operation failed: network 'default' already exists with uuid 5e0fbeda-3731-45be-8fd2-7a8098f92da6
error: Failed to start network default
error: Requested operation is not valid: network is already active
```

martinpitt (approved):
Niiice! (But no cookies for the Fedora package policy.. this will hit users in real-life as well..)

PR #393, opened 2021-09-24
Fix few issues for NIC APIs

skobyda:


skobyda:
Just a few fixes which I forgot to implement in https://github.com/cockpit-project/cockpit-machines/pull/356

skobyda:
Ah failing tests, I see where I broke the virt-xml command

KKoukiou (approved):

PR #394, opened 2021-09-24
Introduce virt-xml into disk APIs

skobyda:


KKoukiou:
Fixed the filesystem, disks still broken: https://logs.cockpit-project.org/logs/pull-394-20210924-130035-1e3d9861-fedora-34/log.html

skobyda:
This is currently blocked on https://github.com/virt-manager/virt-manager/pull/311

martinpitt:
The fix landed in upstream virt-manager 26 days ago, but not released yet AFAICS.

skobyda:
There is already a release in fedora 36 and ubuntu 2204. I can get going with this PR :)

skobyda:
This now passes on all distros where virt-xml (virt-manager package) is >= 4.0 . I wonder if I should just keep this PR blocked until it's released in the other distros too, or just write an exception/antipattern for it

src/libvirtApi/domain.js
```diff
@@ -138,9 +123,32 @@ export function domainAttachDisk({
...
+        source = `,source.pool=${poolName},source.volume=${volumeName}`;
+    let driverType = "";
+    if (format && ['qcow2', 'raw'].includes(format))
+        driverType = `,driver.type=${format}`;
+    let serialNum = "";
+    if (serial)
+        serialNum = `,serial=${serial}`;
+    const shareableOption = shareable ? "yes" : "no";
+
+    return cockpit.script(
```
martinpitt, line 148: Please use cockpit.spawn() for this, pass this as a proper array, and avoid the shell.

src/libvirt-xml-parse.js
```diff
@@ -427,6 +428,17 @@ export function parseDumpxmlForDisks(devicesElem) {
                 readonly: !!readonlyElem,
                 shareable: !!shareableElem,
                 removable: targetElem.getAttribute('removable'),
+                address: { // optional
```
martinpitt, line 431: This commit should not stand on its own. It adds this new field which is not used anywhere, nor does the commit message justify it. I suppose it should be squashed into some future commit that uses it?

src/libvirt-xml-parse.js
```diff
@@ -427,6 +428,17 @@ export function parseDumpxmlForDisks(devicesElem) {
...
+                    type: addressElem ? driverElem.getAttribute('type') : undefined,
```
martinpitt, line 432: Repeating the `addressElem` here is a bit nasty -- how about `address: addresElem ? { bus: ... } : {}`

martinpitt (changes requested):
I see that using `virt-xml` is probably nicer in the long run, as it's a bit more abstract than directly modifying the XML. But as long as this still breaks on so many supported OSes, I feel that we shouldn't land this -- this would appear as a regression. So either we need to support both code paths for now (ugh!), or just shelve this for a while?

KKoukiou:
@skobyda this looks still blocked at least for rhel-8-8, as we still update that, but seems like virt-xml is not updated there.

skobyda:
> @skobyda this looks still blocked at least for rhel-8-8, as we still update that, but seems like virt-xml is not updated there.

Yeah I updated this PR last week to see where it still continues to fail. I'd not try to push this until every distro gets new version of virt-xml

KKoukiou:
This is probably unblocked by now?

skobyda:
> This is probably unblocked by now?

No idea, let's re-trigger

src/libvirtApi/domain.js
```diff
@@ -1101,10 +1108,45 @@ export function domainStart({ connectionName, id: objPath }) {
...
+    let newTarget = target;
+    let addressBus;
+    let addressType;
+    if (busType !== oldBusType) {
+        newTarget = getNextAvailableTarget(existingTargets, busType);
+        // Workaround for https://github.com/virt-manager/virt-manager/issues/430
+        // Until that issue is fixed, we have to change address type and address bus manually
+        if (busType === "virtio")
+            addressType = "pci";
+        if (busType === "usb" || busType === "scsi" || busType === "sata") {
```
cockpituous, line 1127: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-394-20231013-100800-8c396984-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1101,10 +1108,45 @@ export function domainStart({ connectionName, id: objPath }) {
...
+            // The only allowed bus value is '0' as defined in function qemuValidateDomainDeviceDefAddressDrive at
+            // https://gitlab.com/libvirt/libvirt/-/blob/master/src/qemu/qemu_validate.c
+            addressBus = 0;
+            if (busType === "usb")
+                addressType = "usb";
```
cockpituous, line 1132: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-394-20231013-100800-8c396984-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

skobyda:
> This is probably unblocked by now?

Nope, old version of virt-xml is still present on centos-8-stream and rhel-8-*, so this PR will not get merged until we stop supporting those distros

mvollmer:
@martinpitt, @KKoukiou, what's the point of using `virt-xml`? Does it augment the XML with its own knowledge of best defaults? Otherwise this looks like a step back to me, in the JSON > XML > CSV hierachy.  (I think we need to worry about quoting when constructing the virt-xml invocations, for example.)

mvollmer:
Please reopen if this is still wanted.

mvollmer:
Obsoleted by #2581.

PR #398, opened 2021-09-26
test: add check about OS auto detecting

yunmingyang:
When there is an OS already input, new "Installation source" could change the OS by detecting automatically

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -73,12 +73,15 @@ class TestMachinesCreate(VirtualMachinesCase):
         runner.checkOsInputTest(TestMachinesCreate.VmDialog(self, sourceType="disk_image"))
 
         # Test OS autodetection from URL tree media
+        # Auto auto-detection can change "Operating system" field without Ooops
+        # https://bugzilla.redhat.com/show_bug.cgi?id=1715409
         runner.cancelDialogTest(TestMachinesCreate.VmDialog(self, sourceType='url',
                                                             location=config.TREE_URL,
                                                             memory_size=128, memory_size_unit='MiB',
                                                             storage_pool="NoStorage",
-                                                            os_name=config.FEDORA_28,
+                                                            os_name=config.CENTOS_7,
                                                             start_vm=False,
+                                                            expected_os_name=config.FEDORA_28,
                                                             pixel_test_tag="auto"))
 
         # check if older os are filtered
@@ -644,6 +647,8 @@ class TestMachinesCreate(VirtualMachinesCase):
         FEDORA_28 = 'Fedora 28'
         FEDORA_28_SHORTID = 'fedora28'
 
+        CENTOS_7 = 'CentOS 7'
+
         MANDRIVA_2011_FILTERED_OS = 'Mandriva Linux 2011'
 
         MAGEIA_3_FILTERED_OS = 'Mageia 3'
@@ -663,6 +668,7 @@ class TestMachinesCreate(VirtualMachinesCase):
                      os_name="Fedora 28",
                      os_search_name=None,
                      os_short_id="fedora28",
+                     expected_os_name=None,
                      is_unattended=None,
                      profile=None,
                      root_password=None,
@@ -703,6 +709,7 @@ class TestMachinesCreate(VirtualMachinesCase):
             self.os_name = os_name
             self.os_search_name = os_search_name
             self.os_short_id = os_short_id
+            self.expected_os_name = expected_os_name
             self.is_unattended = is_unattended
             self.profile = profile
             self.root_password = root_password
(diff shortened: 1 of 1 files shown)


KKoukiou (changes requested):
There is already a test - which at some point exactly what you 're trying to do here, but we broke it. 

Search for: # Test OS autodetection from URL tree media
https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-create#L75

So instead of introducing this new one - let's fix that one.

You can fix that we removing os_name=config.FEDORA_28, and adding a new expected_os_name which will just wait for the field to get filled.

Then you don't need that checkOsDetect function.

yunmingyang:
OK, let me check and modify it later.

test/check-machines-create
```diff
@@ -911,6 +928,17 @@ class TestMachinesCreate(VirtualMachinesCase):
             elif self.os_name:
                 b.click("#os-select-group > div button")
                 b.click("#os-select li:contains('{0}') button".format(self.os_name))
+                b.wait_attr("#os-select-group input", "value", self.os_name)
+                if (self.expected_os_name and
+                    self.expected_location and
+                    self.sourceType == 'url'):
+                    b.set_input_text("#source-url", self.expected_location)
```
KKoukiou, line 935: Why don't you use self.location. expected_ prefix is used for when the dialog autofills that input field, in this case self.expected_os_name. The fields that we set in the test manually should not use expected.

test/check-machines-create
```diff
@@ -911,6 +928,17 @@ class TestMachinesCreate(VirtualMachinesCase):
...
+                    b.wait_attr("#os-select-group", "data-loading", "false")
+                    b.wait_attr("#os-select-group input", "value", self.expected_os_name)
+
+                    b.switch_to_top()
```
KKoukiou, line 939: This does not belong here. If you want to check that the page did not crash please introduce a function and run it after the fill method.

KKoukiou (changes requested):

test/check-machines-create
yunmingyang: OK, then change it from expected_location to locationSecondInput

test/check-machines-create
```diff
@@ -81,6 +81,17 @@ class TestMachinesCreate(VirtualMachinesCase):
                                                             start_vm=False,
                                                             pixel_test_tag="auto"))
 
+        # Test OS auto detection can change "Operatin system" field without Ooops
+        runner.cancelDialogTest(TestMachinesCreate.VmDialog(self, sourceType='url',
+                                                            location=config.VALID_URL,
```
KKoukiou, line 86: I don't get why you want to use locationSecondInput new variable. Can't you just set directly the location to TREE_URL and let the dialog autodetect the OS?

test/check-machines-create
```diff
@@ -81,6 +81,17 @@ class TestMachinesCreate(VirtualMachinesCase):
                                                             start_vm=False,
                                                             pixel_test_tag="auto"))
 
+        # Test OS auto detection can change "Operatin system" field without Ooops
```
KKoukiou, line 84: s/Operatin/Operating

KKoukiou: Do you have some bugzilla to attach in the test as comment? No needed just wondering.

test/check-machines-create
yunmingyang: Yeah, make sense, I just follow the step of BZ and forget that

test/check-machines-create
yunmingyang: Yeah, I missed it, this case is related with https://bugzilla.redhat.com/show_bug.cgi?id=1715409

test/check-machines-create
```diff
@@ -908,9 +923,16 @@ class TestMachinesCreate(VirtualMachinesCase):
...
-            elif self.os_name:
-                b.click("#os-select-group > div button")
+
+            if self.os_name:
+                # After OS input already has a value, the expand button need selector with more details
+                b.click("#os-select-group > div button.pf-c-select__toggle-button")
                 b.click("#os-select li:contains('{0}') button".format(self.os_name))
+                b.wait_attr("#os-select-group input", "value", self.os_name)
+                if self.expected_os_name and self.sourceType == 'url':
+                    b.set_input_text("#source-url", self.location)
```
KKoukiou, line 933: Sorry for the iteration again.
This set_input_text("#source-url", self.location) is not needed.

It's already done in https://github.com/cockpit-project/cockpit-machines/pull/398/commits/d608cceef38a94702ac55fec87f4bec1fc81ec3d#diff-b284ac42752f40597c4dd12d74e16f6ed82ad5e12d68f0437f242f5ae20bbaa5R916 a few lines above.

Same for the condition, you can drop the self.sourceType == 'url'.

The checks for expected_os_name is enough.

test/check-machines-create
```diff
@@ -655,14 +667,16 @@ class TestMachinesCreate(VirtualMachinesCase):
         vmId = 0
 
         def __init__(self, test_obj, name=None,
-                     sourceType='file', sourceTypeSecondChoice=None, location='',
+                     sourceType='file', sourceTypeSecondChoice=None,
+                     location='',
```
KKoukiou, line 671: Unrelated newline change

KKoukiou (changes requested):

KKoukiou (approved):

PR #399, opened 2021-09-27
Support configuring static MAC -> IP mappings in a libvirt network

KKoukiou:


src/libvirtApi/network.js
```diff
@@ -33,6 +33,21 @@ export function networkActivate({ connectionName, objPath }) {
...
+        opts.superuser = 'try';
+
+    const cmd = [
+        'virsh', '-c', `qemu:///${connectionName}`,
+        'net-update', networkName,
+        'add-last', 'ip-dhcp-host', `<host mac='${macAddress}' ip='${ipAddress}' />`,
+        '--parent-index', parentIndex,
+        '--config'
+    ];
+    return cockpit.spawn(cmd, opts);
```
skobyda, line 48: Why not instead use the proper libvirt-dbus API which has implementation for: https://libvirt.org/html/libvirt-libvirt-network.html#virNetworkUpdate
I think they are equivalent with the virsh net-update.

src/components/networks/networkOverviewTab.jsx
```diff
@@ -134,12 +158,40 @@ export const NetworkOverviewTab = ({ network }) => {
...
+                            </List>
+                        </DescriptionListDescription>
+                    </DescriptionListGroup>
+
+                    <List isPlain>
+                        { ip[1].dhcp.hosts.length
+                            ? ip[1].dhcp.hosts.map((host, index) => DHCPHost(host, index, ip[1].family, idPrefix))
+                            : <ListItem>{_("none")}</ListItem>
+                        }
+                    </List>
```
skobyda, line 194: I think there is, by accident, the same code twice, which causes to render "none" twice in IPv6's Static host entries section
Screenshot from 2021-09-30 01-03-37

test/check-machines-networks
```diff
@@ -786,6 +786,50 @@ class TestMachinesNetworks(VirtualMachinesCase):
...
+        b.set_input_text("#add-new-static-entry-mac-address", self.getDomainMacAddress("subVmTest1"))
+        b.set_input_text("#add-new-static-entry-ip-address", "192.168.122.222")
+        b.click("#add-new-static-entry-save")
+        b.wait_in_text(".pf-c-modal-box .pf-c-alert.pf-m-danger", "there is an existing dhcp host entry")
+        b.click("#add-new-static-entry-cancel")
+
+        # Check that the static IP is correctly assigned to the VM
+        m.execute("virsh net-start default && virsh start subVmTest1")
+        wait(lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.122.222 | wc -l"), delay=3)
+
```
skobyda, line 832: We should also test the same for ipv6, as that has slightly different XML format for <host> section

src/libvirtApi/network.js
```diff
@@ -33,6 +33,21 @@ export function networkActivate({ connectionName, objPath }) {
...
 
+export function networkAddStaticHostEntries({ connectionName, networkName, macAddress, ipAddress, parentIndex }) {
+    const opts = { err: "message", environ: ['LC_ALL=C.UTF-8'] };
+    if (connectionName === 'system')
+        opts.superuser = 'try';
+
+    const cmd = [
+        'virsh', '-c', `qemu:///${connectionName}`,
+        'net-update', networkName,
+        'add-last', 'ip-dhcp-host', `<host mac='${macAddress}' ip='${ipAddress}' />`,
```
skobyda, line 44: This will not work for IPv6, as with ipv6 there is no such thing as mac address. Instead, we should use "id" or "name" property (you can find examples of such XML [here](https://libvirt.org/formatnetwork.html#examplesPrivate6))
Citing the libvirt documentation:
`The IPv6 host element differs slightly from that for IPv4: there is no mac attribute since a MAC address has no defined meaning in IPv6. Instead, the name attribute is used to identify the host to be assigned the IPv6 address. For DHCPv6, the name is the plain name of the client host sent by the client to the server. Note that this method of assigning a specific IP address can also be used for IPv4 instead of the mac attribute`

src/components/networks/networkOverviewTab.jsx
```diff
@@ -149,3 +201,93 @@ export const NetworkOverviewTab = ({ network }) => {
...
+                    e.preventDefault();
+                }} isHorizontal>
+                    <FormGroup label={_("MAC address")} fieldId="add-new-static-entry-mac-address"
+                               helperTextInvalid={_("MAC address must not be empty")}
+                               validated={isSubmitted && !macAddress ? "error" : "default"}>
+                        <TextInput id="add-new-static-entry-mac-address"
+                                   validated={isSubmitted && !macAddress ? "error" : "default"}
+                                   value={macAddress}
+                                   onChange={setMacAddress} />
+                    </FormGroup>
```
skobyda, line 280: As mentioned in the comment below, There is no such thing as MAC address for IPv6. So when defining Static entries for IPv6, user should instead input "name", or alternatively "id".

skobyda (changes requested):

skobyda:
Another question, will there be separate PR also for removing such entries? Since just being able to add them but not able to edit nor remove the entries from UI seems quite incomplete.

src/components/networks/networkOverviewTab.jsx
skobyda, line 280: Also was there actually any specific reason why you decided to use MAC address even for IPv4? Since you can also identify the host by simple hostname (e.g. www.example.org), I think it's more convenient to let user configure entry identified by hostname instead of MAC (Or maybe have both MAC input and hostname input and let user choose how they want to identify the host).
But in short, if we want to have only one way of host identification, then I think hostname is more preferable to me than mac address.

KKoukiou: You can not have host elements without MAC for ipv4. Name attribute is for the DHCP to assign a specific hostname to the VM with the given mac address. So that it does not get the usual transient hostname.

test/check-machines-networks
KKoukiou: I think I will remove the ipv5 host entries adding from this PR.

src/libvirtApi/network.js
KKoukiou: Goood point :) did not see it!

KKoukiou:
> Another question, will there be separate PR also for removing such entries? Maybe we can introduce it also in this PR. Since just being able to add them but not able to edit nor remove the entries from UI seems quite incomplete.

I agree that's missing, but it's more complicated in matters of design so I decided to send a follow up for that.

src/components/networks/networkOverviewTab.jsx
KKoukiou: Removed feature for iPV6

src/components/networks/networkOverviewTab.jsx
KKoukiou: Fixed.

src/components/networks/networkOverviewTab.jsx
```diff
@@ -149,3 +192,93 @@ export const NetworkOverviewTab = ({ network }) => {
...
+            {_("add entry")}
+        </Button>
+    );
+    if (network.active) {
+        addEntry = (
+            <Tooltip content={_("Deactivate the network in order to edit it")}>
+                {addEntry}
+            </Tooltip>
+        );
+    }
```
skobyda, line 243: Why cannot you edit network while it's active? the Update API has flags for that: https://libvirt.org/html/libvirt-libvirt-network.html#virNetworkUpdateFlags
And I think it's a common use case to add host entries to an active network without the need to destroy it and activate it again?

src/components/networks/networkOverviewTab.jsx
```diff
@@ -21,6 +21,7 @@ import PropTypes from 'prop-types';
 import {
     DescriptionList, DescriptionListGroup, DescriptionListTerm, DescriptionListDescription,
     Flex, FlexItem,
+    List, ListItem,
```
skobyda, line 26: Maybe this commit, "Utilize PF4 List to display static dhcp mac -> ip mappings" fba4408bc1bd626719f30e9d04df3a1091d23dc1

Should be merged with the previous one? I don't see why it needs to be separate.

src/components/networks/networkOverviewTab.jsx
```diff
@@ -149,3 +191,82 @@ export const NetworkOverviewTab = ({ network }) => {
...
+
+    return (
+        <>
+            <Button variant="link"
+                    id={idPrefix + "-static-host-entries-add"}
+                    isInline
+                    onClick={() => setIsOpen(true)}>
+                {_("add entry")}
+            </Button>
+            <Modal position="top" variant="small" isOpen={isOpen} onClose={() => setIsOpen(false)}
```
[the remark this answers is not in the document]
KKoukiou: Good catch!

src/components/networks/networkOverviewTab.jsx
```diff
@@ -132,12 +157,31 @@ export const NetworkOverviewTab = ({ network }) => {
...
+                    <Text component={TextVariants.h5}>
+                        {_("DHCP Settings")}
+                    </Text>
+
+                    { ip[1].dhcp.range.start &&
+                    <DescriptionListGroup>
+                        <DescriptionListTerm> {_("Range")} </DescriptionListTerm>
+                        <DescriptionListDescription id={`${idPrefix}-ipv6-dhcp-range`}>
+                            {ip[1].dhcp.range.start + " - " + ip[1].dhcp.range.end}
+                        </DescriptionListDescription>
```
martinpitt, line 169: Duplicated code -- I think this is just a leftover from the IPv6 support, which you intended to remove? Or do you want to keep the *display* of existing IPv6 mappings, and just not edit them?

If you want to keep it, please move this into some `StaticDHCPSettings` component and parameterize it for IPv4/6.

src/components/networks/networkOverviewTab.jsx
martinpitt: Right, DHCP+DNS work exactly the other way around :grin: MAC → IP → NAME. Of course it would be much more convenient if the UI already knew the MAC address -- this configuration is specific to a particular (virtual) interface, is it not? So how much freedom does the user actually have here? Do they have to know the one-and-only correct value, or does this somehow simultaneously define the MAC and the IP for that interface?

skobyda (approved):
Thanks for all the work!

PR #400, opened 2021-09-27
Fix URL encoding and issue with URL after rename operation succeeded

KKoukiou:


test/check-machines-lifecycle
```diff
@@ -248,6 +248,13 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         b.click("#rename-dialog-confirm")
         b.wait_in_text(".pf-c-modal-box__footer .pf-c-alert.pf-m-danger", "Can't rename domain to itself")
 
+        self.goToVmPage("new")
+        self.performAction("new", "rename")
+
+        b.set_input_text("#rename-dialog-new-name", "test%")
+        b.click("#rename-dialog-confirm")
+        b.wait(lambda: "vm?name=test%25&connection=system" in b.eval_js("window.location.href"))
```
martinpitt, line 256: Ah, that's the test case for the first commit for correct URL encoding. I was about to complain there :grin: 

However, can you please ensure that the page actually works? I.e. check some expected element there. If the link would lead to the void or a page crash, that test case would still succeed, but we need to know that parsing and evaluating the URL actually works as intended.

martinpitt (changes requested):

src/components/vm/vmRenameDialog.jsx
```diff
@@ -40,7 +40,12 @@ export const RenameDialog = ({ vmName, vmId, connectionName, toggleModal }) => {
             return;
 
         return domainRename({ connectionName, id: vmId, newName: newName })
-                .then(toggleModal, exc => {
+                .then(() => {
+                    toggleModal();
+                    // If we are on the VMs details page change the URL to reflect the new name after the rename operation succeeded
+                    if (cockpit.location.path.length > 0)
+                        cockpit.location.go(["vm"], Object.assign(cockpit.location.options, { name: newName, connection: connectionName }));
```
martinpitt, line 47: This looks dangerous -- the [docs](https://cockpit-project.org/guide/latest/cockpit-location.html) say that cockpit.location can't be modified in-place, but that's what Object.assign does. I suggest to use the spread operator: `{ ...cockpit.location.options, name: newName, connection: connectionName }`, which will create a new object.

martinpitt (approved):
Danke!

PR #401, opened 2021-09-27
Call cockpit.spawn with superuser: try only for session connection guests

KKoukiou:
This fixes host device attach and detach operations for session qemu
VMs.

Previously the virt-xml operation would run against the VM with the
specified name, but always on the system connection.

This most likely will fail with `permision denied` anyway, but let's not
disable this from the UI, as with some more complex polikit permission
session VMs might also be able to use host devices.


martinpitt (approved):
Thanks! The commit makes sense, just the commit message is the wrong way around: it should say for *system* connection guests. This can be fixed on squash&merge, though.

martinpitt (commented):
On second thought, does that actually work? For PCI devices you need to be root to access them directly; for USB it might work for devices which have an udev ACL. That's probably what you meant with "will fail anyway". Could this get a test case? An USB device with a `chmod 666` ought to work, right? (That doesn't work with PCI)

KKoukiou:
> On second thought, does that actually work? For PCI devices you need to be root to access them directly; for USB it might work for devices which have an udev ACL. That's probably what you meant with "will fail anyway". Could this get a test case? An USB device with a `chmod 666` ought to work, right? (That doesn't work with PCI)

Let me add a test yes. Added with both device types, and without permission tweking this fails with permission denied. Will try to chmod.

test/check-machines-hostdevs
```diff
@@ -143,7 +153,7 @@ class TestMachinesHostDevs(VirtualMachinesCase):
                 b.wait_not_present("#vm-subVmTest1-hostdevs-dialog")
 
             def add(self):
-                m.execute("virsh dumpxml subVmTest1 > /tmp/vmxml1")
+                run_admin("virsh dumpxml subVmTest1 > /tmp/vmxml1")
```
martinpitt: That looks very clean now, nice! The function name is a bit debatable, but :shrug:

martinpitt (approved):
This looks good, assuming green tests. So actually testing USB passthrough (system+session) and PCI passthrough (system only) will be a follow-up, right? Cheers!

test/check-machines-hostdevs
KKoukiou: I named it just `run` but it's taken already. Any ideas welcome.

KKoukiou:
oh it fails now in debian based only with `bash: line 1: /tmp/vmxml1: Permission denied`

test/check-machines-hostdevs
```diff
@@ -146,7 +148,7 @@ class TestMachinesHostDevs(VirtualMachinesCase):
                 b.wait_not_present("#vm-subVmTest1-hostdevs-dialog")
 
             def add(self):
-                m.execute("virsh dumpxml subVmTest1 > /tmp/vmxml1")
+                self.run_admin("virsh dumpxml subVmTest1 > /tmp/vmxml1", connectionName)
```
martinpitt, line 151: The reason for the failure is likely that debian/ubuntu has /tmp protections enabled (i.e. for world-writable files), or possibly the tests just run in the reverse order. If /tmp/vmxml1 already exists and is owned by root, you can't rewrite it as user. You need to remove the file first (as root), or use a per-test random directory such as `self.vm_tmpdir`.

KKoukiou:
@martinpitt I decided to not run in this PR ubuntu-* for the problematic new test. I will debug seperately, not blocking the release.

martinpitt (dismissed):
Thanks! Let's land this once green, good enough. I added the remaining stuff to my todo.

martinpitt (approved):
Excellent, thanks! I'll clean up the things above next week, it put these on my todo.

PR #403, opened 2021-09-29
Add support for multiple pty consoles

KKoukiou:
### Machines: Support selecting between consoles of the same type

VM configurations can specify multiple pty consoles. For example: a VM can have an emulated serial console and a virtio serial console. The console menu now shows alias identifiers in the console menu, making it possible to choose between consoles of the same type.

140937551-4e7bb79d-54da-4bf7-9c67-b1ac85f1f074

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "eslint-config-standard-react": "^9.2.0",
     "eslint-plugin-flowtype": "^5.2.0",
     "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-jsx-a11y": "^6.3.1",
+    "eslint-plugin-jsx-a11y": "~6.4.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "^7.21.0",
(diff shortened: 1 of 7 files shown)


martinpitt:
Looks like the [consoles test](https://logs.cockpit-project.org/logs/pull-403-20211109-171556-fa8debfd-ubuntu-2004/log.html#71) needs an expected message, similar to the existing ones in -create and -lifecycle (due to closing the page while the console is still open).

martinpitt (changes requested):
Nice one! Looks good, but let's please fix the flake right away before it bites us more.

martinpitt:
This now fails to build in the same way that [killed the c-podman release](https://github.com/cockpit-project/cockpit-podman/actions/runs/1443181156):

    ERROR in Error while loading rule 'jsx-a11y/alt-text': rule.create is not a function

This must be a *very* recent eslint upgrade/regression on npmjs.org.

martinpitt:
I added the same hack as in https://github.com/cockpit-project/cockpit-podman/pull/813

martinpitt (approved):
LGTM now, pending green.

skobyda (approved):
Thanks,it looks good.

PR #405, opened 2021-09-30
packaging: Move Debian watch and packit to tar.xz

martinpitt:
These two places were forgotten in commit 544cb90b70.


marusak (approved):
Thanks!

PR #409, opened 2021-10-04
Fix persistent hotplug of vNIC to domain with MAC autogenerated

skobyda:
virt-xml contains a bug when a vNIC is gotplugged into a VM, with MAC
address unspecified, it generates different MAC addresses for online and
offline XML.
We can generate our own MAC address.

### DIFF
--- a/src/components/vm/nics/vmNicsCard.jsx
+++ b/src/components/vm/nics/vmNicsCard.jsx
@@ -68,7 +68,7 @@ export class VmNetworkActions extends React.Component {
     }
 
     render() {
-        const { vm, networks } = this.props;
+        const { vm, vms, networks } = this.props;
         const id = vmId(vm.name);
         const availableSources = {
             network: networks.map(network => network.name),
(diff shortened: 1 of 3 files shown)


KKoukiou (dismissed):
Ignore just saw the bug report in the tests.

src/components/vm/nics/vmNicsCard.jsx
```diff
@@ -359,7 +361,7 @@ export class VmNetworkTab extends React.Component {
                         objectType: _("network interface"),
                         objectName: network.mac,
                         onClose: () => this.setState({ deleteDialogProps: undefined }),
-                        deleteHandler: () => domainDetachIface({ connectionName: vm.connectionName, mac: network.mac, vmName: vm.name, live: vm.state === 'running', persistent: vm.persistent }),
+                        deleteHandler: () => domainDetachIface({ connectionName: vm.connectionName, mac: network.mac, vmName: vm.name, live: vm.state === 'running', persistent: nicPersistent }),
```
KKoukiou, line 364: I don't understand this change here, we used to check if the VM is not persistent, so that we do --no-define when running the virt-xml, That sounds correct, why you change this?

skobyda: That would not work in a situation when a VM is persistend, but the vNIC is not. When you have a running VM, and during attachment of vNIC you decide to not pick `Always attach` option, then vNIC is defined only in live XML. 
Therefore when removing such vNIC, not calling --no-define would result in an error.
Therefore for detaching vNICs, it's the persistence of vNIC we need to look at

PR #419, opened 2021-10-11
Some test cleanups

martinpitt:


martinpitt:
I am still trying to debug the ubuntu test skip in `TestMachinesHostDevs.testHostDevAddSessionConnection`. The root cause is that creating the VM silently seems to fail. The test does
```
+ su - admin -c 'export XDG_RUNTIME_DIR=/run/user/$(id -u admin); virt-install --connect qemu:///session --name subVmTest1 --os-variant cirros0.4.0 --boot hd,network --vcpus 1 --memory 128 --import --disk /home/admin/.local/share/libvirt/images/subVmTest1-2.img --graphics none --console file,target.type=serial,source.path=/var/log/libvirt/console-subVmTest1.log --print-step 1 > /tmp/xml-session && virsh -c qemu:///session define /tmp/xml-session'
Domain 'subVmTest1' defined from /tmp/xml-session
```

But when I log into the VM, and do `su - admin`, I don't really see it:

```
admin@ubuntu:~$ export XDG_RUNTIME_DIR=/run/user/1000
admin@ubuntu:~$ virsh list --all
 Id   Name   State
--------------------
```

I only see it with an explicit `-c`:

```
$ virsh -c qemu:///session list --all 
 Id   Name         State
-----------------------------
 -    subVmTest1   shut off

```

Likewise, cockpit shows subVmTest1 as **system** VM, so somehow being an user does not default to "session" as usual. 

And indeed:
```
admin@ubuntu$ virsh uri
qemu:///system
```

A-ha!

After `sudo deluser admin libvirt`, it fails more explicitly:
```
$ virsh uri
error: failed to connect to the hypervisor
error: Failed to connect socket to '/var/run/libvirt/libvirt-sock': Permission denied
```

This happens even if I `ssh admin@c`, and thus get a proper login session.

So something is really wrong -- this happens with a clean `./vm-run ubuntu-stable` from bots even, there is nothing in c-machines' test preparation.

martinpitt:
With adding `-c` explicitly, I now get a more direct error:
```
+ su - admin -c 'export XDG_RUNTIME_DIR=/run/user/$(id -u admin); virsh -c qemu:///session nodedev-list --cap 0'
error: Invalid capability type
```

With manual ssh:
```
$ virsh -c qemu:///session nodedev-list --cap 0
error: Invalid capability type

$ virsh -c qemu:///session nodedev-list 
block_vda_ROOT
computer
drm_card0
net_eth0_52_54_00_12_34_56
net_eth1_52_54_01_00_00_01
net_lo_00_00_00_00_00_00
pci_0000_00_00_0
pci_0000_00_01_0
pci_0000_00_01_1
pci_0000_00_01_2
pci_0000_00_01_3
pci_0000_00_02_0
pci_0000_00_03_0
pci_0000_00_04_0
pci_0000_00_05_0
pci_0000_00_06_0
pci_0000_00_0e_0
pci_0000_00_0f_0
scsi_0_0_1_0
scsi_generic_sg0
scsi_host0
scsi_host1
scsi_host2
scsi_target0_0_1
usb_1_0_1_0
usb_usb1
```

martinpitt:
The test [does not work on ubuntu-2004](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-419-20211011-064836-dbf623a4-ubuntu-2004/log.html#68), but does now on ubuntu stable. I'll take a quick look, and either skip or fix (most probably an unsupported feature on old libvirt)

KKoukiou (approved):
Thank you!!

PR #427, opened 2021-10-15
Arch testing support

jelly:


test/vm.install
```diff
@@ -43,8 +43,20 @@ if [ -d /var/tmp/debian ]; then
...
+elif [ -d /var/tmp/arch ]; then
+    # build source package
+    cd /var/tmp
+    TAR=$(ls cockpit-machines-*.tar.xz)
+    VERSION="${TAR#cockpit-machines-}"
+    VERSION="${VERSION%.tar.xz}"
 
+    cp arch/PKGBUILD .
+    sed -i "s/VERSION/$VERSION/" PKGBUILD
+    sed -i "s/SOURCE/$TAR/" PKGBUILD
```
martinpitt, line 55: Unimportant, but for the future: put both s// sed commands into one CLI call (semicolon separated).

test/check-machines-create
```diff
@@ -378,20 +378,21 @@ class TestMachinesCreate(VirtualMachinesCase):
...
 
         runner.createAndVerifyQemuConfParsedTest(
-                TestMachinesCreate.VmDialog(self, sourceType='file',
-                                            storage_pool="NoStorage",
-                                            location=config.NOVELL_MOCKUP_ISO_PATH),
-                vnc_listen, "127.0.0.1", vnc_passwd, None)
+            TestMachinesCreate.VmDialog(self, sourceType='file',
+                                        storage_pool="NoStorage",
+                                        location=config.NOVELL_MOCKUP_ISO_PATH),
+            vnc_listen, "127.0.0.1", vnc_passwd, None)
```
martinpitt, line 384: Unrelated whitespace change. Would be better as separate commit. (Same a bit below)

test/check-machines-create
```diff
@@ -1131,6 +1132,11 @@ class TestMachinesCreate(VirtualMachinesCase):
                     "cp {0}/cert.pem /usr/local/share/ca-certificates/cert.crt".format(self.test_obj.vm_tmpdir),
                     "update-ca-certificates"
                 ]
+            elif self.machine.image.startswith("arch"):
+                cmds += [
+                    "cp {0}/cert.pem /etc/ca-certificates/trust-source/anchors/cert.pem".format(self.test_obj.vm_tmpdir),
+                    "update-ca-trust"
+                ]
```
martinpitt, line 1139: This and the following arch specific code would fit better into the previous commit. This one is "test: skip failing tests due to missing dependencies", which really should just have the skipImage bits.

test/check-machines-consoles
```diff
@@ -99,6 +99,7 @@ class TestMachinesConsoles(VirtualMachinesCase):
     def testInlineConsoleWithUrlRoot(self, urlroot=""):
         self.testInlineConsole(urlroot="/webcon")
 
+    @skipImage('TODO: setlocale: No such file or directory', "arch")
```
martinpitt, line 102: How can we fix this without blocking on a new upstream glibc? Could we set `LANG=en_US.UTF-8` for arch?

FTR, I'm totally fine with this TODO skip, I just wouldn't want to keep it around for a year.

test/machineslib.py
```diff
@@ -277,6 +277,10 @@ def setUp(self):
         # Ignoring just to unbreak tests for now
         self.allow_journal_messages("Failed to get COMM: No such process")
 
+        # FIXME: Testing on Arch Linux fails randomly with networkmanager time outs while the test passes.
```
martinpitt, line 280: This commit message has a "networmanager" typo. Please fix it to unbreak commit log searches.

martinpitt (changes requested):
Great work! :partying_face: 

Just some minor commit shuffling to do. Can you please also rebase to latest main, to get packit results?

test/check-machines-consoles
jelly: Should be fixed in February, but yeah still need to figure this one out. Skipped it for now to not block the whole PR.

test/check-machines-create
jelly: Sorry about that, I made my editor lint/fix things which usually goes fine except when it doesn't :)

martinpitt (approved):
Rocking! branch was not rebased, so still no packit, but the changes here seem safe.

martinpitt:
@jelly : Can you please send a bots PR which enables arch for main?

PR #443, opened 2021-11-01
test: add a check about detaching NIC of a off VM

yunmingyang:


test/check-machines-nics
```diff
@@ -139,6 +140,20 @@ class TestMachinesNICs(VirtualMachinesCase):
         b.wait_not_present(".pf-c-modal-box")
         b.wait_not_present("#vm-subVmTest1-network-1-mac")
 
+        self.performAction("subVmTest1", "off")
```
KKoukiou, line 143: This is also tested at https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-nics#L266 

If you want to have it better organized we can move that deletion code into the testNICDelete test.

yunmingyang: Yeah, I think that is also better for align this verify case with polarion cases. So should I need to move the case in this PR? And should I need some other modifications?

KKoukiou: Moving the relevant code is enough.

yunmingyang: It seems that the new added check is the same with the original one, so I just remove the original check

KKoukiou: Actually the check is not the same. The previous code was covering  https://github.com/cockpit-project/cockpit/commit/7c041b8f8574cfadaf9ea4272d95da4f1f7cca80, so https://bugzilla.redhat.com/show_bug.cgi?id=1791543

There need to be two interfaces for this.

yunmingyang: Thanks. This issues is too old, so I provision a rhel 8.2 machine to reproduce this issue. And for reproducing the second "Actual results" of the issue, we need add another two network interfaces, and the type of the second one must be bridge. So, according above, I add a few lines. Please check and remind me if something is wrong, many thanks

test/check-machines-nics
```diff
@@ -139,6 +140,26 @@ class TestMachinesNICs(VirtualMachinesCase):
...
+        b.wait_in_text("body", "Virtual machines")
+        # Check NIC MAC addresses
+        b.wait_text("#vm-subVmTest1-network-1-mac", "52:54:00:4b:73:5f")
+        b.wait_text("#vm-subVmTest1-network-2-mac", "52:54:00:4b:73:6f")
+        b.wait_text("#vm-subVmTest1-network-3-mac", "52:54:00:4b:73:4f")
+        # Detach
+        self.deleteIface(2)
+        # Make sure popup dialog disappeared
+        b.wait_not_present(".pf-c-modal-box")
+        b.wait_not_present("#vm-subVmTest1-network-2-mac")
```
KKoukiou, line 161: I think this will fail, as the indexing is done according to the position in the list. So deleting the second element, will make the third sedond.

yunmingyang: OK, re-push. And I found a weird thing. Sometimes, after attaching three network interfaces, the VM disappeared and there is also a weird log in libvirtd(which is "internal error: End of file from qemu monitor"). I think that is the reason why the cases is pass during my last run in my laptop. Do you have any ideas about this libvirt behavior?

KKoukiou: Hey, sorry - just saw that. I have seen this error as well in the CI a few times, but not locally. If you can conistently reproduce it, please share the steps and we can debug it together.

KKoukiou (commented):
There are test failures https://logs.cockpit-project.org/logs/pull-443-20211109-104607-b20956bc-fedora-34/log.html#69 that are related.

test/check-machines-nics
yunmingyang: I’m trying to reproduce that, but it seems impossible reproduced manually, so I'm trying to reproduce it by a script, Once reproduced, I will leave a comment

yunmingyang: I am failed on reproducing that. Also it seems that it could not be reproduced every times, when running verify test( Probably reproduce 3 times in 5 times). Then I try to search suspicious points in the new case. I think the most suspicious point is "--current" since it will let the config take effect directly(get from help information of "virsh attach-interface").So I change "--current" to "--config". Then, I run the case five times, and they are all pass. So let's re-run  jobs for making sure the modification. Also, do you have another ideas?

test/check-machines-nics
```diff
@@ -139,6 +140,27 @@ class TestMachinesNICs(VirtualMachinesCase):
         b.wait_not_present(".pf-c-modal-box")
         b.wait_not_present("#vm-subVmTest1-network-1-mac")
 
+        # Detach NIC when the VM is shutoff, Also check this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1791543
+        self.performAction("subVmTest1", "off")
```
KKoukiou, line 144: Does this work? I think you need forceOff

test/check-machines-nics
```diff
@@ -139,6 +140,27 @@ class TestMachinesNICs(VirtualMachinesCase):
...
+        # Attach since no NIC now
+        m.execute("virsh attach-interface --domain subVmTest1 --type network --source default --model virtio --mac 52:54:00:4b:73:5f --config")
```
KKoukiou, line 146: --config and --current should be the same for offline Vms. So you can remove it without changing the functionality.

test/check-machines-nics
yunmingyang: It works, since there is a wait(lambda: "login as 'cirros' user." in self.machine.execute("cat {0}".format(args["logfile"])), delay=3) in previous lines

test/check-machines-nics
yunmingyang: OK, but in last test run, --current brought a behavior which is the VM disappear after refreshing. I could not reproduce it manually, and when automation, it seems not to be reproduced every time. So I try to use --config instead of --current, and the cases all passed in five times, so I guess maybe we meet a issue in libvirt, which will happen when attaching nic with --current multi-times, and not 100% reproducible. I am not sure, so I think we should run test to verify that. So what's your opinion?

KKoukiou (approved):
Super, that worked! Thanks!

PR #445, opened 2021-11-04
tests: rhel-9-0 now uses modular daemons similar to fedora-35

KKoukiou:


martinpitt (approved):
Thanks! I suppose this needs to land in lockstep with the image refresh.

PR #449, opened 2021-11-11
tests: invert the conditions for checking modular vs monolithic daemon distro

KKoukiou:
We should have a list that shrinks - not grows

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -57,7 +57,7 @@ systemctl start firewalld
 firewall-cmd --add-service=cockpit --permanent
 firewall-cmd --add-service=cockpit
 
-if grep -Eq 'PLATFORM_ID=.*(f35)' /etc/os-release; then
+if grep -Eq 'PLATFORM_ID=.*(f35|f36|el9)' /etc/os-release; then
     # HACK: new modular libvirt sockets are not running by default in f35
     # https://gitlab.com/libvirt/libvirt/-/issues/219
     systemctl start virtinterfaced.socket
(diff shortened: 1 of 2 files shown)


martinpitt:
Thanks!

> don't have m.image command

`m.image` should totally work, and this PR still relies on it. I suggest to just drop this sentence from the commit message, as it's confusing.

You need to force-push for the pycodestyle error anyway.

test/browser/browser.sh
```diff
@@ -57,7 +57,7 @@ systemctl start firewalld
 firewall-cmd --add-service=cockpit --permanent
 firewall-cmd --add-service=cockpit
 
-if grep -Eq 'PLATFORM_ID=.*(f35)' /etc/os-release; then
+if grep -Eq 'PLATFORM_ID=.*(f35|f36|el9)' /etc/os-release; then
```
martinpitt, line 60: I think we should invert that too, and *not* do it on f34 and el8. But that can also be a follow-up the next time we have to touch this file.

martinpitt (dismissed):
LGTM, bots have the last word.

martinpitt:
[This flake](https://logs.cockpit-project.org/logs/pull-449-20211111-081011-0b7b03c9-ubuntu-2004/log.html#4) is fixed by #448. Let's rebase this PR after that lands, to not totally ruin our test stats.

martinpitt (approved):

martinpitt:
Branch was out of date, so packit srpm build failure. Here it actually matters, so I pushed a rebase.

martinpitt (approved):
Thanks! The packit f35 failure is something entirely different (timeout? VM running out of memory?) but the tests until the hang look good.

KKoukiou:
tests succeeded also downstream.

PR #450, opened 2021-11-11
Cleanup css monthly routine :)

KKoukiou:


martinpitt (approved):
Niice! Some more roadblocks for removing PF3 removed. Any fa- or pficons-* left after this?

KKoukiou:
> Niice! Some more roadblocks for removing PF3 removed. Any fa- or pficons-* left after this?

No these were the last :)

PR #453, opened 2021-11-16
This reverts commit 6eb9f27a3d95bbfe9184c94ecf39741fb42cf1f7

KKoukiou:
When creating dummy ISO files don't use qemu-img, this does not
reflect the reality anyway.

### DIFF
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -182,7 +182,7 @@ class TestMachinesDisks(VirtualMachinesCase):
             cancel("sdb")
 
         # Virtio bus type should not be shown for CDROM devices
-        m.execute("qemu-img create -f raw /var/lib/libvirt/novell.iso 1M")
+        m.execute("touch /var/lib/libvirt/novell.iso")
         m.execute("virsh attach-disk --domain subVmTest1 --source /var/lib/libvirt/novell.iso --target hda --type cdrom --mode readonly --persistent")
         # virsh attach-disk want send an event for offline VM changes
         b.reload()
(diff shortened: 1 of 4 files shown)


test/check-machines-disks
```diff
@@ -800,8 +797,8 @@ class TestMachinesDisks(VirtualMachinesCase):
         self.createVm("subVmTest1")
 
         # Prepare file for Custom File disk type
-        m.execute("qemu-img create -f raw /var/lib/libvirt/novell.iso 1M")
-        m.execute("qemu-img create -f qcow2 /var/lib/libvirt/novell.qcow2 1M")
```
martinpitt, line 804: Was that by accident, or on purpose? I do see that using qemu-img for a fake iso file is wrong, but for a .qcow2 file it still seems right?

test/check-machines-settings
```diff
@@ -253,8 +253,7 @@ class TestMachinesSettings(VirtualMachinesCase):
         b = self.browser
 
         self.createVm("subVmTest1")
-        self.machine.execute("qemu-img create -f raw /var/lib/libvirt/images/phonycdrom 1M; "
-                             "virsh attach-disk subVmTest1 /var/lib/libvirt/images/phonycdrom sdb --type cdrom --config")
+        self.machine.execute("touch /var/lib/libvirt/images/phonycdrom; virsh attach-disk subVmTest1 /var/lib/libvirt/images/phonycdrom sdb --type cdrom --config")
```
martinpitt, line 256: The two-line version was easier to read/review.

martinpitt (commented):
Nice! One question/potential gotcha, otherwise, great debugging work!

test/check-machines-disks
KKoukiou: I just reverted the commit completely, with git revert. Yes. this should be fine.

test/check-machines-settings
KKoukiou: I just reverted the commit, but I can also partially revert it. and keep this and the comment from above.

martinpitt (approved):
I see, thanks!

PR #455, opened 2021-11-18
test: Fix race condition in TestMachinesFilesystems.testBasicSystemConnection

martinpitt:
The negative test for "Shared Directories card not present" is also able
to pass on OSes which do support it, but where detection takes longer
than reaching the check. This particularly affects centos-8-stream.

Wait until the page sufficiently initialized, i.e. when the "host
devices" is visible.

Update the test for current CentOS 8 stream. This fixes [this failure](http://artifacts.dev.testing-farm.io/4c9d5e09-adc0-4eed-bde0-875d8ba05d51/) in packit.

 - [ ] Requires https://github.com/cockpit-project/bots/issues/2639 so that our bots c8s image picks up the latest libvirt

### DIFF
--- a/test/check-machines-filesystems
+++ b/test/check-machines-filesystems
@@ -58,7 +58,7 @@ class TestMachinesFilesystems(VirtualMachinesCase):
         # wait until page initialized
         b.wait_visible("#vm-subVmTest1-hostdevs")
 
-        if m.image in ["rhel-8-5", "centos-8-stream", "ubuntu-2004", "fedora-33"] or connection == "session":
+        if m.image in ["rhel-8-5", "ubuntu-2004", "fedora-33"] or connection == "session":
             b.wait_not_present("#vm-subVmTest1-filesystems-add")
             return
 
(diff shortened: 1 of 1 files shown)


martinpitt:
The [centos-8-stream failure on our CI](https://logs.cockpit-project.org/logs/pull-455-20211118-152136-38426c83-centos-8-stream/log.html#71) here is expected, as the current image still has the old version where it *really* does not work. I also ran it against the updated image in https://github.com/cockpit-project/bots/pull/2639 [here](https://logs.cockpit-project.org/logs/pull-455-20211118-152810-38426c83-centos-8-stream-bots-2639/log.html), and that passed.

KKoukiou (approved):

KKoukiou:
centos-8-stream@bots#2639 already green!

martinpitt:
Tests are hard to get green here due to 3x affected. I think in the interests of making tests greener overall, we'll have to land this with some red :cry:

KKoukiou:
The original centos-8-stream is expected to fail, as it's running against the old image. Landing in lockstep with the bots PR.

martinpitt:
Nice, so the Fedora ones passed after all :+1:

PR #465, opened 2021-11-30
test: Drop obsolete vm.install hacks

martinpitt:
Our ubuntu-stable image is 21.04 now, and debian-stable is Debian 11,
so drop the now unused backports installation.

The virsh network works fine with the default configuration now, so drop
the firewalld config hack.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -6,18 +6,7 @@ set -eu
 # for Debian based images, build and install debs
 if [ -d /var/tmp/debian ]; then
     apt-get update
-    if grep -q 'VERSION_ID="21.04"' /etc/os-release; then
-        # HACK: nftables backend does not currently work with libvirt: https://launchpad.net/bugs/1799095
-        sed -i 's/FirewallBackend=nftables/FirewallBackend=iptables/' /etc/firewalld/firewalld.conf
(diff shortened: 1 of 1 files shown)


test/vm.install
```diff
@@ -6,18 +6,7 @@ set -eu
...
-    elif grep -q 'VERSION_ID="20.10"' /etc/os-release; then
-        # HACK: nftables backend does not currently work with libvirt: https://launchpad.net/bugs/1799095
-        sed -i 's/FirewallBackend=nftables/FirewallBackend=iptables/' /etc/firewalld/firewalld.conf
-
-        BACKPORTS="-t groovy-backports"
-    elif grep -q 'VERSION_ID="10"' /etc/os-release; then
-        BACKPORTS="-t buster-backports"
-    fi
-    eatmydata apt-get install ${BACKPORTS:-} ${APT_INSTALL_OPTIONS:-} -y cockpit-ws cockpit-system
+    eatmydata apt-get install ${APT_INSTALL_OPTIONS:-} -y cockpit-ws cockpit-system
```
martinpitt, line 9: FTR: I'm going to eliminate that next. We should really pre-install cockpit-ws/system on Debian/Ubuntu as well.

allisonkarlitskaya (approved):
groovy

PR #466, opened 2021-11-30
test: Enable NFS tests for Arch Linux

jelly:
The latest Arch image has nfs-utils which is required to set up an nfs
server.

### DIFF
--- a/test/check-machines-storage-pools
+++ b/test/check-machines-storage-pools
@@ -26,7 +26,7 @@ sys.path.append(os.path.join(TEST_DIR, "common"))
 sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 
 from machineslib import VirtualMachinesCase  # noqa
-from testlib import nondestructive, skipImage, test_main, wait  # noqa
+from testlib import nondestructive, test_main, wait  # noqa
 
 
 @nondestructive
@@ -248,7 +248,6 @@ class TestMachinesStoragePools(VirtualMachinesCase):
         # Check it's not present after deactivation
         self.waitPoolRow("myPoolThree", connectionName, False)
 
-    @skipImage("TODO: enable nfs support on the arch image", "arch")
     def testStoragePoolsCreate(self):
         b = self.browser
         m = self.machine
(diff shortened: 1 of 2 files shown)


test/check-machines-storage-pools
```diff
@@ -471,6 +470,9 @@ class TestMachinesStoragePools(VirtualMachinesCase):
         dev = self.add_ram_disk(2)
         m.execute(f"parted {dev} mklabel gpt")
 
+        if m.image == "arch":
+            m.execute("mkdir /media/ || true")
```
martinpitt, line 474: I don't like `|| true` in tests, unless it has a good justification (like opportunistic cleanup). /media is an FHS dir, so it ought to exist. If so, please drop the command; if not, please drop the `|| true`. If there really is a good reason that it *may* exist on our test images, use `mkdir -p`.

martinpitt (changes requested):
Nice! Let's see what the tests say, then just a little cleanup to do.

test/check-machines-storage-pools
jelly: Ok, seems the Arch `filesystem` package does not include it. As the tests are nondestructive I added the `|| true` so it doesn't fail on the second run.

KKoukiou: Adding the '-p' will make the command succeed even if the directory exists.

martinpitt: I understand, for nond tests. So I think `-p` is fine then.

martinpitt (dismissed):
Thanks!

martinpitt (approved):
Thanks! [f35 flaked](https://logs.cockpit-project.org/logs/pull-466-20211130-124502-fddca1dc-fedora-35/log.html#70-1), I retried this. packit can't work as the branch is out of date, but we can ignore that here.

jelly:
> Thanks! [f35 flaked](https://logs.cockpit-project.org/logs/pull-466-20211130-124502-fddca1dc-fedora-35/log.html#70-1), I retried this. packit can't work as the branch is out of date, but we can ignore that here.

Ok, centos-8-stream flaked too, so it should be good to go.

PR #467, opened 2021-11-30
test: Stop using the internet for Debian/Ubuntu tests

martinpitt:
Since https://github.com/cockpit-project/bots/commit/b06bd37b0ad our
Debian/Ubuntu test VMs have the basic cockpit packages pre-installed.
Drop the apt calls to install them, so that we stop getting surprise
breakages due to new cockpit versions. These will now properly appear on
image refreshes.

That was the last thing using the internet for these tests, so switch
them to run offline.


KKoukiou (approved):
Shall we do the same for c-podman then?

martinpitt:
@KKoukiou : Sure, I sent https://github.com/cockpit-project/cockpit-podman/pull/841 right after this PR (except there we can't do offline due to the `podman pull` calls)

martinpitt:
COPR is kaputt right now, hence no packit builds. But vm.install is only used in our own tests.

PR #469, opened 2021-12-02
test: Look past merge commit when downloading dist cache

martinpitt:
This fixes SRPM build in packit tests for non-current branches.

Adapted from https://github.com/cockpit-project/cockpit/commit/bbf218644c8c8a
Thanks to Allison Karlitskaya for figuring out the solution.

-----

This is intentionally *non-current* branch. This would have made packit tests fail in the "usual" way (failed to build SRPM).

### DIFF
--- a/test/download-dist
+++ b/test/download-dist
@@ -48,11 +48,17 @@ def download_dist(wait=False):
         message("download-dist: pre-built dist/ are for NODE_ENV=production")
         return False
 
+    # If the HEAD commit is a merge (when checking out a pull request /merge ref), use the merged
+    # commit ID to find the cached data, instead of the merge itself
     try:
-        sha = subprocess.check_output(["git", "rev-parse", "HEAD"], stderr=subprocess.DEVNULL).decode().strip()
+        sha = subprocess.check_output(["git", "rev-parse", "--verify", "HEAD^2"], stderr=subprocess.DEVNULL).decode().strip()
     except subprocess.CalledProcessError:
-        message("download-dist: not a git repository")
-        return False
+        # otherwise use the HEAD as-is
+        try:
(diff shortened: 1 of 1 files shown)


martinpitt:
So with the empty commit, [srpm build timed out](https://github.com/cockpit-project/cockpit-machines/pull/469/checks?check_run_id=4395463814) as expected, as it tried to wait for the wrong SHA.

I pushed the actual fix now. This is supposed to succeed in packit.

test/download-dist
```diff
@@ -48,11 +48,17 @@ def download_dist(wait=False):
         message("download-dist: pre-built dist/ are for NODE_ENV=production")
         return False
 
+    # If the HEAD commit is a merge (when checking out a pull request /merge ref), use the merged
+    # commit ID to find the cached data, instead of the merge itself
     try:
-        sha = subprocess.check_output(["git", "rev-parse", "HEAD"], stderr=subprocess.DEVNULL).decode().strip()
+        sha = subprocess.check_output(["git", "rev-parse", "--verify", "HEAD^2"], stderr=subprocess.DEVNULL).decode().strip()
```
allisonkarlitskaya, line 54: I'd be nervous about this, without some kind of assurance that the thing we download was built for (the equivalent of) the tree that we're actually inside of.  This specifically works in the cockpit case because we do a rebase to get our tree into the exact same shape as the one that built `dist/`.  Here I guess we don't do that.

allisonkarlitskaya: Specifically: the merge (which represents the state of the current tree) very possibly has a lot of commits in it that weren't present when `dist/` was built.

martinpitt:
Giving up on this. Last week we agreed to just drop the dist caching from c-machines.

PR #472, opened 2021-12-06
Fix node device attachment and detachment for usb devices without bus/device numbers

yunmingyang:
We also used to identify usb device by only 'vendor ID' and 'product ID'. But that will identify a wrong usb device when it has the same 'vendor ID' and 'product ID' with another usb device
And, in 'https://libvirt.org/formatnode.html', it seems that usb_device should always has bus and device numbere as there is no 'optional' label beside these two properties. So, it seems that we
won't miss anything

Relevant to https://bugzilla.redhat.com/show_bug.cgi?id=2018448

 - Land PR #601 to fix packit tests

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -393,8 +393,8 @@ export function timeoutedPromise(promise, delay, afterTimeoutHandler, afterTimeo
     return deferred.promise;
 }
 
-export function findHostNodeDevice(hostdev, nodeDevices) {
-    let nodeDev;
+export function findMatchingNodeDevices(hostdev, nodeDevices) {
+    let nodeDevs = [];
     switch (hostdev.type) {
     case "usb": {
         const vendorId = hostdev.source.vendor.id;
@@ -402,29 +402,23 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
         const device = hostdev.source.device;
         const bus = hostdev.source.bus;
 
-        nodeDev = nodeDevices.find(d => {
+        nodeDevs = nodeDevices.filter(d => {
             // vendor and product are properties used to identify correct device. But vendor and product
             // are not unique, and in some cases, multiple host devices with same vendor and product can exist.
-            // In such cases, optional properties (bus, device) are used in addition to product and vendor
-            // to identify correct device
-            if (bus &&
-                device &&
-                vendorId &&
-                productId &&
-                d.capability.vendor &&
-                d.capability.product &&
-                d.capability.bus == bus &&
-                d.capability.device == device &&
-                d.capability.vendor.id == vendorId &&
-                d.capability.product.id == productId)
-                return true;
-            else if (vendorId &&
+            // In such cases, optional properties (bus, device) are used in addition to product and vendor to identify correct device
+            // But there are cases when usb and device are not specified.
+            // If there are 2 usb devices without specified bus/device and same vendor/product,
+            // it's impossible to identify which one is the one referred in VM's XML, so we return an array of all matching
+            if (vendorId &&
                 productId &&
                 d.capability.vendor &&
                 d.capability.product &&
                 d.capability.vendor.id == vendorId &&
-                d.capability.product.id == productId)
-                return true;
+                d.capability.product.id == productId) {
+                if ((!bus && !device) ||
+                    (bus && device && d.capability.bus == bus && d.capability.device == device))
+                    return true;
+            }
         });
         break;
     }
@@ -435,7 +429,7 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
         const slot = parseInt(hostdev.source.address.slot, 16).toString();
         const func = parseInt(hostdev.source.address.func, 16).toString();
 
-        nodeDev = nodeDevices.find(d => {
+        nodeDevs = nodeDevices.filter(d => {
             // pci device is identified by bus, slot, domain, function
             if (bus &&
                 slot &&
@@ -454,7 +448,7 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
         const target = hostdev.source.address.target;
         const unit = hostdev.source.address.unit;
 
-        nodeDev = nodeDevices.find(d => {
+        nodeDevs = nodeDevices.filter(d => {
             if ((bus && target && unit) &&
                 d.capability.bus &&
                 d.capability.lun &&
@@ -473,7 +467,7 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
     case "mdev": {
         const uuid = hostdev.source.address.uuid;
 
-        nodeDev = nodeDevices.find(d => {
+        nodeDevs = nodeDevices.filter(d => {
             if ((uuid) &&
                 d.capability.uuid == uuid)
                 return true;
@@ -483,7 +477,7 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
     case "storage": {
         const block = hostdev.source.block;
 
-        nodeDev = nodeDevices.find(d => {
+        nodeDevs = nodeDevices.filter(d => {
             if ((block) &&
                 d.capability.block == block)
                 return true;
@@ -493,7 +487,7 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
     case "misc": {
         const ch = hostdev.source.char;
 
-        nodeDev = nodeDevices.find(d => {
+        nodeDevs = nodeDevices.filter(d => {
             if ((ch) &&
                 d.capability.char == ch)
                 return true;
@@ -503,7 +497,7 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
     case "net": {
         const iface = hostdev.source.interface;
 
-        nodeDev = nodeDevices.find(d => {
+        nodeDevs = nodeDevices.filter(d => {
             if ((iface) &&
                 d.capability.interface == iface)
                 return true;
@@ -511,7 +505,8 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
         break;
     }
     }
-    return nodeDev;
+
+    return nodeDevs;
 }
 
 /**
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -399,8 +399,8 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
     case "usb": {
         const vendorId = hostdev.source.vendor.id;
         const productId = hostdev.source.product.id;
-        const device = parseInt(hostdev.source.device, 16).toString();
-        const bus = parseInt(hostdev.source.bus, 16).toString();
+        const device = hostdev.source.device;
+        const bus = hostdev.source.bus;
 
         nodeDev = nodeDevices.find(d => {
             // vendor and product are properties used to identify correct device. But vendor and product
@@ -870,6 +870,32 @@ export function getNodeDevSource(dev) {
         const bus = dev.busnum;
 
         source = `${bus}.${device}`;
+    } else {
+        throw new Error(`getNodeDevSource: unsupport device type '${dev.type}'`);
+    }
+
+    return source;
+}
+
+export function getHostDevSourceObject(dev) {
+    let source;
+
+    if (dev.type === "pci") {
+        const domain = dev.source.address.domain;
+        const bus = dev.source.address.bus;
+        const slot = dev.source.address.slot;
+        const func = dev.source.address.func;
+
+        source = { domain, bus, slot, func };
+    } else if (dev.type === "usb") {
+        const device = dev.source.device;
+        const bus = dev.source.bus;
+        const vendor = dev.source.vendor.id;
+        const product = dev.source.product.id;
+
+        source = { vendor, product, bus, device };
+    } else {
+        throw new Error(`getHostDevSourceObject: unsupport device type '${dev.type}'`);
     }
 
     return source;
--- a/src/libvirt-xml-parse.js
+++ b/src/libvirt-xml-parse.js
@@ -30,6 +30,68 @@ export function getDiskElemByTarget(domxml, targetOriginal) {
     }
 }
 
+export function getHostDevElemBySource(domxml, source) {
+    const domainElem = getElem(domxml);
+
+    if (!domainElem) {
+        console.warn(`Can't parse dumpxml, input: "${domainElem}"`);
+        return;
+    }
+
+    const devicesElem = domainElem.getElementsByTagName('devices')[0];
+    if (!devicesElem) {
+        console.warn(`Can't parse dumpxml for host devices, devices element is not present`);
+        return;
+    }
+
+    const hostdevElems = devicesElem.getElementsByTagName('hostdev');
+
+    if (hostdevElems) {
+        for (let i = 0; i < hostdevElems.length; i++) {
+            const hostdevElem = hostdevElems[i];
+            const type = hostdevElem.getAttribute('type');
+            const sourceElem = hostdevElem.getElementsByTagName('source')[0];
+            if (!sourceElem)
+                continue;
+            const addressElem = sourceElem.getElementsByTagName('address')[0];
+
+            if (type === "usb" && "vendor" in source && "product" in source) {
+                const vendorElem = sourceElem.getElementsByTagName('vendor')[0];
+                const productElem = sourceElem.getElementsByTagName('product')[0];
+                if (!vendorElem || !productElem)
+                    continue;
+                const vendor = vendorElem.getAttribute('id');
+                const product = productElem.getAttribute('id');
+
+                if (vendor && product && vendor === source.vendor && product === source.product) {
+                    if (addressElem) {
+                        // If XML does contain bus/device numbers, we have to identify correct hostdev by them
+                        const bus = addressElem.getAttribute('bus');
+                        const device = addressElem.getAttribute('device');
+
+                        if (bus === source.bus && device === source.device)
+                            return new XMLSerializer().serializeToString(hostdevElem);
+                    } else {
+                        // If XML doesn't contain bus/device numbers, we can identify only by vendor/product ids
+                        return new XMLSerializer().serializeToString(hostdevElem);
+                    }
+                }
+            }
+
+            // PCI device
+            if (type === "pci" && "bus" in source && "domain" in source && "slot" in source && "func" in source) {
+                const bus = addressElem.getAttribute('bus');
+                const domain = addressElem.getAttribute('domain');
+                const slot = addressElem.getAttribute('slot');
+                const func = addressElem.getAttribute('function');
+
+                if (bus && domain && slot && func && domain === source.domain && slot === source.slot && bus === source.bus && func === source.func)
+                    return new XMLSerializer().serializeToString(hostdevElem);
+            }
+        }
+    }
+}
+
 export function getIfaceElemByMac(domxml, mac) {
     const domainElem = getElem(domxml);
 
--- a/src/libvirt-xml-parse.js
+++ b/src/libvirt-xml-parse.js
@@ -522,6 +522,7 @@ export function parseDumpxmlForHostDevices(devicesElem) {
             case "usb": {
                 const addressElem = getSingleOptionalElem(hostdevElem, 'address');
                 const sourceElem = getSingleOptionalElem(hostdevElem, 'source');
+                const sourceAddressElem = getSingleOptionalElem(sourceElem, 'address');
 
                 let vendorElem, productElem;
                 if (sourceElem) {
@@ -542,8 +543,8 @@ export function parseDumpxmlForHostDevices(devicesElem) {
                         product: {
                             id: productElem ? productElem.getAttribute('id') : undefined,
                         },
-                        device: addressElem.getAttribute('device'),
-                        bus: addressElem.getAttribute('bus'),
+                        device: sourceAddressElem ? sourceAddressElem.getAttribute('device') : undefined,
+                        bus: sourceAddressElem ? sourceAddressElem.getAttribute('bus') : undefined,
                     },
                 };
                 hostdevs.push(dev);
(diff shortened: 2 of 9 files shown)


KKoukiou:
This causes test failures. @yunmingyang, let's wait for @skobyda who wrote this code to explain maybe what's happening.

KKoukiou (commented):
'…Since virt-xml may miss bus number and device number when VM is shutoff' I did not check it in action, but is that a virt-xml bug? We generally want to keep virt-xml, for if there are bugs let's report them before making this change.

Also what you are trying to solve can be fixed probably in the virt-xml call directly, when providing the address there.

```
$ virt-xml --add-device --hostdev ?
--hostdev options:
  clearxml
  address.base
  address.bus
  address.controller
  address.cssid
  address.devno
  address.domain
  address.function
  address.iobase
  address.irq
  address.multifunction
  address.port
  address.reg
  address.slot
  address.ssid
  address.target
  address.type
  address.unit
  address.zpci.fid
  address.zpci.uid
  alias.name
  boot.order
  driver.name
  name
  rom.bar
  type
```

yunmingyang:
> …Since virt-xml may miss bus number and device number when VM is shutoff' I did not check it in action, but is that a virt-xml bug? We generally want to keep virt-xml, for if there are bugs let's report them before making this change.

Yes, it seems an issue. When the VM is shutoff, the XML generated by virt-xml only have vendorId and productId. And busNum and deviceNum appear after running the VM. But if creating with XML and set busNum and deviceNum in it, these number will be there whether the VM is running or not. So I try to attach with a XML. 

And in last test run, the failed cases is also because the shutoff VM did not contain busNum and deviceNum, then in findHostNodeDevice, host devices can not map any one in node devices. Then in machines page, "Model" and "Vendor" will be shown as there the id(like "0x1d6b") not the value(like "Integrated Rate Matching Hub"). Also in that condition, "Source" will be shown as empty.

There also another reason to use domainAttachDevice. It is that, when using virt-xml and selecting multi-devices to attach, we can only attach one of the devices, when the VM is shutoff. Also, when the VM is running, we can select multi-devices to attach, and this operation is successful, too. But after shutoff the VM, there is only one device left. Using domainAttachDevice seems could also resolve that problem. I suspect that problem is also an issue about virt-xml

> Also what you are trying to solve can be fixed probably in the virt-xml call directly, when providing the address there.

I think we use virt-xml like virt-xml ${vm_name} --add-device --host-dev 001.002(bus.device). I am not sure whether using "virt-xml test --add-device --hostdev address.bus=1,address.devno=2,type=usb" is OK

And for the new test run, I think the failed is about that virt-xml is not used when attaching so that no error raised

skobyda:
I tried this PR locally and it seems there are still some problems. I try it with my bluetooth usb device, and when i try to remove it fails, saying that a device with said device number/bus doesn't exist.
So I dig a bit deeper into it and I think that what basically happens is that when you plug a host device into a running VM (or plug into an offline VM and then start the VM), libvirt will detach the host device from the host. This is to prevent both VM and host from using the same device simultaneously. After it detaches it and reattaches, in some cases, it may end up with different device number (in my case this happens with a bluetooth usb device).

skobyda:
@yunmingyang I want to make sure that the bug I encounter with this PR is the same as described in https://bugzilla.redhat.com/show_bug.cgi?id=2018448 . So I asked in bugzilla for some additional logs.

skobyda:
@yunmingyang I believe that switching domainAttachHostDevice away from virt-xml was not necessarry, as that was not what was causing the bug. I believe the bug will fixed by changing domainDetachHostDevice to use libvirt's API.

> There also another reason to use domainAttachDevice. It is that, when using virt-xml and selecting multi-devices to attach, we can only attach one of the devices, when the VM is shutoff. Also, when the VM is running, we can select multi-devices to attach, and this operation is successful, too. But after shutoff the VM, there is only one device left. Using domainAttachDevice seems could also resolve that problem. I suspect that problem is also an issue about virt-xml

Also virt-xml can be used to attach multiple hostdevices at the same time, by just appending multiple "--add-device" option to virt-xml command. So I force-pushed that change.

yunmingyang:
> I believe that switching domainAttachHostDevice away from virt-xml was not necessarry, as that was not what was causing the bug. I believe the bug will fixed by changing domainDetachHostDevice to use libvirt's API. So I force pushed some changes about that.
> 
> > There also another reason to use domainAttachDevice. It is that, when using virt-xml and selecting multi-devices to attach, we can only attach one of the devices, when the VM is shutoff. Also, when the VM is running, we can select multi-devices to attach, and this operation is successful, too. But after shutoff the VM, there is only one device left. Using domainAttachDevice seems could also resolve that problem. I suspect that problem is also an issue about virt-xml
> 
> Also virt-xml can be used to attach multiple hostdevices at the same time, by just appending multiple "--add-device" option to virt-xml command. So I force-pushed that change.
> 
> @yunmingyang can you check if this PR now fixed the bug?

Sure, I have checked it simply, And I think the code fix the issue. But I think there should be still a thing which we need to make sure. It is that, when the VM is shutoff, some usb devices whose device number and bus number will be changed when the VM is running, will just show vendor ID and model ID, since virt-xml seems that just add vendor ID and model ID to a shutoff VM for this kind usb devices. But some usb devices(like 'Integrated Rate Matching Hub') could be shown correctly with their vendor and model name since virt-xml seems that add device number and bus number to the VM xml when the VM is shutoff . I think that may confuse some users. so do you think that make sense? @skobyda 

And another thing is it seems that getNodeDevSourceObject is not needed if the code is final version. so do you think we may need it in the future? If not, I think I will remove it and force push

skobyda:
> > I believe that switching domainAttachHostDevice away from virt-xml was not necessarry, as that was not what was causing the bug. I believe the bug will fixed by changing domainDetachHostDevice to use libvirt's API. So I force pushed some changes about that.
> > > There also another reason to use domainAttachDevice. It is that, when using virt-xml and selecting multi-devices to attach, we can only attach one of the devices, when the VM is shutoff. Also, when the VM is running, we can select multi-devices to attach, and this operation is successful, too. But after shutoff the VM, there is only one device left. Using domainAttachDevice seems could also resolve that problem. I suspect that problem is also an issue about virt-xml
> > 
> > 
> > Also virt-xml can be used to attach multiple hostdevices at the same time, by just appending multiple "--add-device" option to virt-xml command. So I force-pushed that change.
> > @yunmingyang can you check if this PR now fixed the bug?
> 
> Sure, I have checked it simply, And I think the code fix the issue. But I think there should be still a thing which we need to make sure. It is that, when the VM is shutoff, some usb devices whose device number and bus number will be changed when the VM is running, will just show vendor ID and model ID, since virt-xml seems that just add vendor ID and model ID to a shutoff VM for this kind usb devices. But some usb devices(like 'Integrated Rate Matching Hub') could be shown correctly with their vendor and model name since virt-xml seems that add device number and bus number to the VM xml when the VM is shutoff . I think that may confuse some users. so do you think that make sense? @skobyda

For IRC chat:
[16:42] <skobyda> It seems that's normal for usb device to change bus and device numbers
[16:43] <skobyda> so that's why virt-xml uses bus and device numbers in xml definition only when neccessarry, meaning that vendorId and productId are shared with other devices. When those are not shared with other devices, virt-xml doesn't put it into xml definition
[16:45] <skobyda> and I think that's correct. If usb number and device number are not gueranteed to be stable, then let's use it only when absolutely necessarry
[16:45] <skobyda> And when it's not neccessary, let's just have a host device attached to a VM without specifying bus and device number



> And another thing is it seems that getNodeDevSourceObject is not needed if the code is final version. so do you think we may need it in the future? If not, I think I will remove it and force push

src/helpers.js
```diff
@@ -424,13 +424,6 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
...
                 return true;
-            else if (vendorId &&
-                productId &&
-                d.capability.vendor &&
-                d.capability.product &&
-                d.capability.vendor.id == vendorId &&
-                d.capability.product.id == productId)
-                return true;
         });
         break;
```
skobyda, line 428: @yunmingyang  I'm going to drop this change and reintroduce it back.
According to libvirt xml schema, source bus and device numbers are optional, therefore we should be able to look up host device even without them present: https://gitlab.com/libvirt/libvirt/-/blob/master/docs/schemas/domaincommon.rng#L3292

skobyda: This was the change which was causing the c-machines not being able to look up a correct node device, therefore it could not properly show vendor name, product name and bus and device number. (The problem we talked about on IRC that user could consider a bus and device number missing a bug)

src/libvirtApi/domain.js
```diff
@@ -99,15 +101,13 @@ function buildConsoleVVFile(consoleDetail) {
...
-        flags |= Enum.VIR_DOMAIN_AFFECT_CONFIG;
+function domainAttachDevice({ connectionName, vmId, xmlDesc, live }) {
+    let attachflag = Enum.VIR_DOMAIN_AFFECT_CONFIG;
+    if (live) {
+        attachflag |= Enum.VIR_DOMAIN_AFFECT_LIVE;
+    }
 
-    // Error handling is done from the calling side
-    return call(connectionName, vmId, 'org.libvirt.Domain', 'AttachDevice', [xmlDesc, flags], { timeout, type: 'su' });
+    return call(connectionName, vmId, 'org.libvirt.Domain', 'AttachDevice', [xmlDesc, attachflag], { timeout, type: 'su' });
```
skobyda, line 110: A lot of test failures https://logs.cockpit-project.org/logs/pull-472-20220114-124851-4e91fab0-debian-testing/log.html. My gut tells me it's probably caused by this change, but will have to investigate futher

src/helpers.js
yunmingyang: But, that will still cause an issue that c-machines filter a wrong device from nodeDevices, and that will also cause deletion error

src/libvirtApi/domain.js
yunmingyang: That is because I modified the this function at the first and missed parameter "permanent", as we don't use this method when attaching device, I revert this method

src/helpers.js
skobyda: I slightly modified the function, so now it works for both cases: When bus and device number is present, and when it is not present (and there is a risk of 2 node devices sharing same vendor and product ids)

skobyda:
So after many discussions with @yunmingyang , I think we found a state where both of us are happy with current PR, as it fixes some bugs but still offers full functionality
However, since now like half of the code changes in this PR are from me, I think I'm not unbiased reviewer anymore. @martinpitt mind giving it a code-review?

martinpitt:
I didn't review yet, but let's give this a round of testing. This needs a rebase for packit, and a tests-trigger allow for our CI. I pushed a clean rebase.

martinpitt:
[testCPUModel](https://logs.cockpit-project.org/logs/pull-472-20220121-063942-b0c0cb82-fedora-34/log.html#63-2) failure, retrying for comparison

src/libvirt-xml-parse.js
```diff
@@ -30,6 +30,60 @@ export function getDiskElemByTarget(domxml, targetOriginal) {
...
+                    }
+                }
+            }
+
+            // PCI device
+            if (type === "pci" && "bus" in source && "domain" in source && "slot" in source && "func" in source) {
+                const sourceElem = hostdevElem.getElementsByTagName('source')[0];
+                const addressElem = sourceElem.getElementsByTagName('address')[0];
+
+                const bus = addressElem.getAttribute('bus');
```
martinpitt, line 75: Again, check addressElem for being valid before using it.

src/libvirtApi/domain.js
```diff
@@ -452,18 +454,34 @@ export function domainDetachDisk({
             });
 }
 
-export function domainDetachHostDevice({ connectionName, vmName, live, dev }) {
-    const options = { err: "message" };
-    const source = getNodeDevSource(dev);
-    const args = ["virt-xml", "-c", `qemu:///${connectionName}`, vmName, "--remove-device", "--hostdev", source];
+export function domainDetachHostDevice({ connectionName, vmName, vmId, live, dev }) {
```
martinpitt, line 457: It's redundant to pass `vmName` as the function already requires `vmId` anyway.  It also does not use `vmName` anywhere, so please drop it.

martinpitt (changes requested):
Thanks! This is really complicated code, so it desperately needs tests to make sure the new logic works. Both for the second commit (let the test create/share two devices with same vendor/product, but different bus address), and the third commit (calculating the wrong address).

src/helpers.js
```diff
@@ -399,8 +399,8 @@ export function findHostNodeDevice(hostdev, nodeDevices) {
     case "usb": {
         const vendorId = hostdev.source.vendor.id;
         const productId = hostdev.source.product.id;
-        const device = parseInt(hostdev.source.device, 16).toString();
-        const bus = parseInt(hostdev.source.bus, 16).toString();
+        const device = hostdev.source.device;
+        const bus = hostdev.source.bus;
```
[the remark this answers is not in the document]
skobyda: These variables are not values which are shown in the UI (therefore not related to the bug). They merely are there so we can compare output of virsh dumpxml and virsh nodedev-dumpxml (i.e. match hostdev in VM's XML to a nodedev provided by libvirt)
Example of output of virsh nodedev-dumpxml:
```
<device>
  <name>usb_1_10</name>
  <path>/sys/devices/pci0000:00/0000:00:14.0/usb1/1-10</path>
  <devnode type='dev'>/dev/bus/usb/001/003</devnode>
  <parent>usb_usb1</parent>
  <driver>
    <name>usb</name>
  </driver>
  <capability type='usb_device'>
    <bus>1</bus>
    <device>3</device>
    <product id='0x0026'>AX201 Bluetooth</product>
    <vendor id='0x8087'>Intel Corp.</vendor>
  </capability>
</device>
```

Example of hostdev from virsh dumpxml:
```
    <hostdev mode='subsystem' type='usb' managed='yes'>
      <source>
        <vendor id='0x8087'/>
        <product id='0x0026'/>
        <address bus='1' device='3'/>
      </source>
      <address type='usb' bus='0' port='4'/>
    </hostdev>
```

Since both bus and device values are decimal numbers, we can just take them as strings and compare them.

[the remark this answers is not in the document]
martinpitt: I see, thanks; resolving

src/libvirt-xml-parse.js
```diff
@@ -30,6 +30,60 @@ export function getDiskElemByTarget(domxml, targetOriginal) {
...
+export function getHostDevElemBySource(domxml, source) {
+    const domainElem = getElem(domxml);
+
+    if (!domainElem) {
+        console.warn(`Can't parse dumpxml, input: "${domainElem}"`);
+        return;
+    }
+
+    const devicesElem = domainElem.getElementsByTagName('devices')[0];
+    const hostdevElems = devicesElem.getElementsByTagName('hostdev');
```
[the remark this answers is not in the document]
skobyda: This schema https://gitlab.com/libvirt/libvirt/-/blob/master/docs/schemas/domaincommon.rng#L6036 which VM'x XML is validated against defines that devices is not optional, and will be always present.
But better be safe than sorry.
Yes, it's a step backwards, I'm going to report bugzilla now

src/libvirt-xml-parse.js
```diff
@@ -30,6 +30,60 @@ export function getDiskElemByTarget(domxml, targetOriginal) {
...
+    if (hostdevElems) {
+        for (let i = 0; i < hostdevElems.length; i++) {
+            const hostdevElem = hostdevElems[i];
+            const type = hostdevElem.getAttribute('type');
+
+            if (type === "usb" && "vendor" in source && "product" in source) {
+                const sourceElem = hostdevElem.getElementsByTagName('source')[0];
+                const addressElem = sourceElem.getElementsByTagName('address')[0];
+                const vendor = sourceElem.getElementsByTagName('vendor')[0].getAttribute('id');
+                const product = sourceElem.getElementsByTagName('product')[0].getAttribute('id');
```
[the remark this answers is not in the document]
skobyda: Similar as above, libvirt xml schema specify that source is always present.

src/libvirt-xml-parse.js
```diff
@@ -30,6 +30,60 @@ export function getDiskElemByTarget(domxml, targetOriginal) {
...
+        console.warn(`Can't parse dumpxml, input: "${domainElem}"`);
+        return;
+    }
+
+    const devicesElem = domainElem.getElementsByTagName('devices')[0];
+    const hostdevElems = devicesElem.getElementsByTagName('hostdev');
+
+    if (hostdevElems) {
+        for (let i = 0; i < hostdevElems.length; i++) {
+            const hostdevElem = hostdevElems[i];
```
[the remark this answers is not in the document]
skobyda: That owuld break the code below, since when we find correct hostdevElem, we return it `return new XMLSerializer().serializeToString(hostdevElem);`
So I would have to create a tmp variable, put it outside of `forEach(() => {...});, `and return it at the end, at which point I think it's simpler to have the for... cycle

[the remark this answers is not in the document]
martinpitt: Right, I was missing that, thanks.

skobyda:
Hey @yunmingyang , I think we just had here force-push collision, which removed some changes which solved @martinpitt review comments. 
With this push I guess you wanted to add a commit which throws an assertion in case of unsupported host dev type? If that is the only change, I can combine it with my changes. (Also I think that should not be a separate commit, but rather part of f29a9e40285b880362a5f98c5b76293de632bf61)

yunmingyang:
> Hey @yunmingyang , I think we just had here force-push collision, which removed some changes which solved @martinpitt review comments. With this push I guess you wanted to add a commit which throws an assertion in case of unsupported host dev type? If that is the only change, I can combine it with my changes. (Also I think that should not be a separate commit, but rather part of [f29a9e4](https://github.com/cockpit-project/cockpit-machines/commit/f29a9e40285b880362a5f98c5b76293de632bf61))

Sure, many thanks

src/helpers.js
```diff
@@ -870,6 +870,32 @@ export function getNodeDevSource(dev) {
...
+        source = { domain, bus, slot, func };
+    } else if (dev.type === "usb") {
+        const device = dev.source.device;
+        const bus = dev.source.bus;
+        const vendor = dev.source.vendor.id;
+        const product = dev.source.product.id;
+
+        source = { vendor, product, bus, device };
+    } else {
+        throw new Error(`getHostDevSourceObject: unsupport device type '${dev.type}'`);
```
skobyda, line 893: @yunmingyang I squashed your commit about throwing this error message to already existing commit. Also changed the error message to also show which type is not supported.

yunmingyang: OK, thanks

martinpitt:
Some [relevant failures](http://artifacts.dev.testing-farm.io/a43ecc7a-865f-42ce-8033-188323257a59/) in `testHostDevAddMultipleDevices` and `testHostDevicesList` on c8s?

martinpitt (changes requested):
Thanks! A few more nitpicks, and the test failures need to be investigated.

martinpitt:
@skobyda : Still pixel test issues here -- apparently new tests without reference images?

martinpitt:
And the new test also doesn't [work on the testing farm](http://artifacts.dev.testing-farm.io/d7d089be-2329-46b2-a2e4-ac97fbd028cc/)

martinpitt:
Updated pixel refs for the renamed test, and rebased to main to run packit.

martinpitt:
Force-pushed again to get a clean re-run of packit. My previous wrong push and revert thoroughly confused it.

martinpitt:
The [c8s packit run](https://artifacts.dev.testing-farm.io/637d46c8-af9b-4543-a5a8-ffbbf2e915c4/) failed `testHostDevAddMultipleDevices` and `testHostDevicesList` (in the part which the test changes). Our test/browser/run-test.sh excludes `testHostDevAdd` which got renamed to `testHostDevAddSingleDevice`, and curiously that one did succeed -- but I wouldn't want to bet on this being robust enough.

`testHostDevicesList` expects this:

    b.wait_in_text("#vm-subVmTest1-hostdev-1-source #1-device", "Unspecified")

The [screenshot](https://artifacts.dev.testing-farm.io/637d46c8-af9b-4543-a5a8-ffbbf2e915c4/work-all5zFnvm/plans/all/execute/data/test/browser/data/TestMachinesHostDevs-testHostDevicesList-centos-8-stream-localhost-22-FAIL.png) has that part cut off, but the [HTML dump](https://artifacts.dev.testing-farm.io/637d46c8-af9b-4543-a5a8-ffbbf2e915c4/work-all5zFnvm/plans/all/execute/data/test/browser/data/TestMachinesHostDevs-testHostDevicesList-centos-8-stream-localhost-22-FAIL.html) shows that the source does not have such a curiously named DOM sub-node at all:

```html
<td data-label="Source" class=""><div id="vm-subVmTest1-hostdev-1-source"><dl class="pf-c-description-list pf-m-horizontal"></dl></div></td>
```

However, I just dealt with a similar problem in PR #601, and I believe this is simply testing against the wrong cockpit-machines RPM. Let's land #601 first, then rebase and run again. We still need to adjust the `$EXCLUDES`, but it can't hurt to run the three tests on the TF to see how far they get.

The [c9s run](https://artifacts.dev.testing-farm.io/a880a31c-cf3b-40b3-b97c-460972cf6178/) is an unmitigated disaster, something corrupted the VM. Let's re-run and see if it clears up.

martinpitt:
Rebased on top of #601, to get test results now instead of having to wait. I also adjusted `$EXCLUDES`.

martinpitt:
#601 landed, rebased.

martinpitt:
This looks good enough for me now, thank you! I rebased again to get green on packit (as the Rawhide and C9s fixes landed)

martinpitt (approved):

PR #473, opened 2021-12-06
Fix tests, and more of it

KKoukiou:
…ud-init options

This will also fix the failing, centos-8-stream packit test, which
already has the virt-install new version.

### DIFF
--- a/src/libvirt-xml-create.js
+++ b/src/libvirt-xml-create.js
@@ -230,33 +230,3 @@ export function getSnapshotXML(name, description) {
 
     return new XMLSerializer().serializeToString(doc.documentElement);
 }
-
-export function getFilesystemXML(source, target, xattr) {
-    const doc = document.implementation.createDocument('', '', null);
-
-    const filesystemElem = doc.createElement('filesystem');
-
-    const driverElem = doc.createElement('driver');
-    driverElem.setAttribute('type', 'virtiofs');
-    filesystemElem.appendChild(driverElem);
-
-    if (xattr) {
-        const binaryElem = doc.createElement('binary');
-        binaryElem.setAttribute('xattr', 'on');
-        filesystemElem.appendChild(binaryElem);
-    }
-
-    const sourceElem = doc.createElement('source');
-    sourceElem.appendChild(doc.createTextNode(name));
-    sourceElem.setAttribute('dir', source);
-    filesystemElem.appendChild(sourceElem);
-
-    const targetElem = doc.createElement('target');
-    targetElem.appendChild(doc.createTextNode(name));
-    targetElem.setAttribute('dir', target);
-    filesystemElem.appendChild(targetElem);
-
-    doc.appendChild(filesystemElem);
-
-    return new XMLSerializer().serializeToString(doc.documentElement);
-}
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -40,7 +40,6 @@ import {
 } from '../actions/store-actions.js';
 import {
     getDiskXML,
-    getFilesystemXML,
 } from '../libvirt-xml-create.js';
 import {
     finishVmCreateInProgress,
@@ -59,7 +58,6 @@ import {
 import {
     getDiskElemByTarget,
     getDoc,
-    getElem,
     getSingleOptionalElem,
     parseDomainDumpxml,
 } from '../libvirt-xml-parse.js';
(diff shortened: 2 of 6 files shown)


martinpitt (approved):
Thanks! Adding blocked as this presumably breaks on the current centos-8-stream image.

KKoukiou:
> Thanks! Adding blocked as this presumably breaks on the current centos-8-stream image.

Correct.

martinpitt:
[rhel-8-6@bots#2681](https://logs.cockpit-project.org/logs/pull-473-20211208-182533-8c14c3d0-rhel-8-6-bots-2681/log.html#16-2) failed two tests, [centos-8-stream@bots#2690](https://logs.cockpit-project.org/logs/pull-473-20211208-182512-8c14c3d0-centos-8-stream-bots-2690/log.html) passed. I'm going to push a rebase to see how it fares on testing farm.

mvollmer:
@KKoukiou, I am trying to get rhel-8-6 to work at the same time as centos-8-stream, and I found this:

```
class TestMachinesFilesystems(VirtualMachinesCase):
    def testBasicSystemConnection(self):
        m = self.machine
        if "rhel-8-6" in m.image:
            # This should warn us when virtmanager package is updated in rhel-8-6. When that happens, we can update domainCreateFilesystem() API
            # Attaching filesystem to a VM was fully implemented to virt-xml in virtmanager 3.0.
            # RHEL 8.6 still uses virtmanager with version lesser than 3.0, therefore we use fallback for the old API
            output = m.execute("virt-xml --add-device --filesystem=?")
            unsupported_parameters = ['driver.type', 'source.dir', 'target.dir', 'binary.xattr']
            for param in unsupported_parameters:
                self.assertFalse(param in output)
        self._testBasic()
```

This fails now: https://logs.cockpit-project.org/logs/pull-473-20211209-053847-b106dcc2-rhel-8-6-bots-2681/log.html#16-2

So is there some extra work to be done in cockpit-machines?

mvollmer:
> So is there some extra work to be done in cockpit-machines?

Found it, working on it... :-)

martinpitt:
@mvollmer : Note that we also still test c-machines on rhel-8-5, which likely still has an older virt-manager? However, this code cleanup is nice, and the probability that we'll do an 8.5 update is very low. So from my side I'd actually be fine with dropping rhel-8-5 from our testmap, and clean up the old special cases.

KKoukiou (commented):
The changes from @mvollmer commit are correct. Thanks for cleaning this up!

mvollmer:
This looks good now.  Let's have a full test run.

mvollmer:
The TestMachinesFilesystems.testBasicSystemConnection failure on debian-stable is hartnäckig. I'll have a look locally.

mvollmer:
> The TestMachinesFilesystems.testBasicSystemConnection failure on debian-stable is hartnäckig. I'll have a look locally.

Got it.  The test would not wait for the dialog to close before using it again. It would only wait for the created thing to appear in the page, but the dialog was still open and busy at that time. The action would fully finish later, in the middle of filling out the dialog for the next test step, and close the dialog only then.

martinpitt:
*All* packit tests were red, I retried them this morning as some were "just" timeouts. I sent PR #478 with a README update to get a comparison with current main. This PR is supposed to at least fix some failures in centos-stream-8.

The centos-8-stream and rhel-8-6 failures from our CI are expected -- the corresponding runs against the image refreshes are green. (That's the "blocked" label).  rhel-8-5 ought to get green, but it's still running.

There is a [storage pool pixel test failure](https://logs.cockpit-project.org/logs/pull-473-20211210-054858-8aa7d15b-fedora-34-firefox/pixeldiff.html#TestMachinesStoragePools-testStoragePools-storage-pools-page) due to the button being disabled vs. enabled -- that just smells like a missing wait condition. I retry that one to compare, and will investigate this locally.

martinpitt:
The pixel test failed again, and it also failed in #478. I pushed the fix here. I have a different chromium, so it'll likely fail again, and then I'll grab the reference image from  CI.

martinpitt:
There is [another broken pixel test](https://logs.cockpit-project.org/logs/pull-478-20211210-061307-95a56707-fedora-34-mobile/pixeldiff.html#TestMachinesNetworks-testNetworkState-networks-page-mobile), and even the current reference images are both bogus.  I fixed this as well.

martinpitt:
I was fighting with [this pixel test failure](https://logs.cockpit-project.org/logs/pull-473-20211210-072200-edd845cb-fedora-34-firefox/pixeldiff.html#TestMachinesStoragePools-testStoragePools-storage-pools-page) locally as well. There is some infuriating subtle noise in the font rendering. My hunch is that this might be related to animations again? I'll make some more experiments, this might be another case of a sleep() :cry: until we have a "wait for animations" API.

martinpitt:
I pacified (or should I say "gutted") the storage pool pixel tests enough that it locally succeeds several times now, with chromium, firefox, and mobile. 23rd time is the charm! :crossed_fingers:

martinpitt (approved):
I think we'll have to land this with all packit tests being red, otherwise it's just too hard to climb out of that hole. I compared the logs with #478, and at least the tests are strictly better, so it's an improvement already, and no regression.

@mvollmer, want to cross-check?

KKoukiou (commented):
Heh, I really thought that network didn't contain all the configuration options we show thus the expanded part of the row in the screenshot was empty. How naive!

PR #481, opened 2021-12-14
test: Update test/common, mostly for animation waiting

mvollmer:
Changes:

- Now assert_pixels waits for animations to be over before taking the
  screenshot.

- assert_pixels also removes focus before screenshotting, so we need
  to adjust some references.

- We can decide locally which image does the pixel tests and changes
  to TEST_OS_DEFAULT wont affect that.


martinpitt:
The changes look good, especially the testSerialConsole reference image update -- all of a sudden this makes sense!

Needs ref image updating.

mvollmer:
We can stop ignoring the triangles now, I think.

mvollmer:
> The changes look good, especially the testSerialConsole reference image update -- all of a sudden this makes sense!

Hmm, is the text in the console stable enough?  Let's see!

martinpitt:
> is the text in the console stable enough?

I am convinced it is not. At least the generated host key will differ between each run, The IP addresses also sound brittle. We may need to ignore the console contents.

test/check-machines-create
```diff
@@ -1060,8 +1060,6 @@ class TestMachinesCreate(VirtualMachinesCase):
             return self
 
         def assert_pixels(self, subtag=None):
-            # focus away from any input field, to avoid the text cursor in screenshots
-            self.browser.focus("#create-vm-dialog .btn-cancel")
```
martinpitt, line 1064: That is correct, but will again change the pixel tests. The update commit was earlier than this change, but let's see if you included it there already.

mvollmer: No changes were necessary. :-)

mvollmer:
> Yup, both IP addresses and uptime change, see [failed pixel test](https://logs.cockpit-project.org/logs/pull-481-20211214-131722-3c08113d-fedora-35-firefox/pixeldiff.html#TestMachinesConsoles-testSerialConsole-vm-details-console-serial)

Ignored.  

I was surprised that the serialConsole test acutally shows the VNC console. Is that expected?

mvollmer:
> I was surprised that the serialConsole test acutally shows the VNC console. Is that expected?

Clicking the "Expand" button doesn't keep the selection from the dropdown, but seems to go back to some default, maybe.

martinpitt:
> I was surprised that the serialConsole test acutally shows the VNC console. Is that expected?

Good point -- no, that's surely a bug.

martinpitt (approved):
Thank you! packit tests are doomed until PR #484, but they don't do pixel tests. So I think it's safe to land this.

PR #482, opened 2021-12-14
Fix typos and spelling errors found by codespell

garrett:
Most of these changes are in comments. I'll note where it affects the code inline.

### DIFF
--- a/test/check-machines-migrate
+++ b/test/check-machines-migrate
@@ -90,47 +90,47 @@ class TestMachinesMigration(VirtualMachinesCase):
     def testSharedStorageMigration(self):
         self._testMigrationGeneric(False, False, None)
 
-    @skipImage('RHEL doesnt provide support for copy-storage migration', "rhel-8-4", "rhel-8-5", "rhel-8-6", "rhel-9-0", "centos-9-stream")
+    @skipImage('RHEL does not provide support for copy-storage migration', "rhel-8-4", "rhel-8-5", "rhel-8-6", "rhel-9-0", "centos-9-stream")
     def testCopyStorageMigration(self):
         self._testMigrationGeneric(True, False, None)
 
     def testMoveTemporarilyMigration(self):
         self._testMigrationGeneric(False, True, None)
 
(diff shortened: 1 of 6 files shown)


test/check-machines-migrate
```diff
@@ -90,47 +90,47 @@ class TestMachinesMigration(VirtualMachinesCase):
     def testSharedStorageMigration(self):
         self._testMigrationGeneric(False, False, None)
 
-    @skipImage('RHEL doesnt provide support for copy-storage migration', "rhel-8-4", "rhel-8-5", "rhel-8-6", "rhel-9-0", "centos-9-stream")
+    @skipImage('RHEL does not provide support for copy-storage migration', "rhel-8-4", "rhel-8-5", "rhel-8-6", "rhel-9-0", "centos-9-stream")
```
garrett, line 93: Either `doesn't` needs to be used (and the single tick quotes need to be double tick quotes) or we expand the word to `does not`, which is what I did in this file. I think it's just text to explain what's happening in the test, so it's fine to change, correct? There are several other occurrences in this file.

test/check-machines-migrate
```diff
@@ -158,7 +158,7 @@ class TestMachinesMigration(VirtualMachinesCase):
 
         if fail == "uri":
             # test failure for syntactically incorrect destination uri
-            b.set_input_text("#dest-uri-input", "qemu+ssh://root@machine2/sistem")
+            b.set_input_text("#dest-uri-input", "qemu+ssh://root@machine2/system")
```
KKoukiou, line 161: This is used as expected failure.

KKoukiou (changes requested):

test/check-machines-migrate
KKoukiou: This is fine to change indeed.

test/check-machines-migrate
garrett, line 161: Oh, is this intentionally a spelling error to fail? I guess I should back out this change.

KKoukiou (approved):

PR #490, opened 2021-12-17
Update serial console when its alias changes

skobyda:
In order to create a serial console, we call 'virsh console [domain]
--devname [alias]' command. Sometimes however, alias might be undefined.
In that case, don't pass any devname parameter at all, instead of
passing empty string as parameter.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2032099

### DIFF
--- a/src/components/vm/consoles/consoles.jsx
+++ b/src/components/vm/consoles/consoles.jsx
@@ -48,10 +48,21 @@ class Consoles extends React.Component {
 
         this.state = {
             consoleDetail: undefined,
+            serial: props.vm.displays && props.vm.displays.filter(display => display.type == 'pty'),
         };
 
         this.getDefaultConsole = this.getDefaultConsole.bind(this);
         this.onDesktopConsoleDownload = this.onDesktopConsoleDownload.bind(this);
+    }
+
+    static getDerivedStateFromProps(nextProps, prevState) {
+        const oldSerial = prevState.serial;
+        const newSerial = nextProps.vm.displays && nextProps.vm.displays.filter(display => display.type == 'pty');
+
+        if (newSerial.length !== oldSerial.length || oldSerial.some((pty, index) => pty.alias !== newSerial[index].alias))
+            return { serial: newSerial };
+
+        return null;
     }
 
     getDefaultConsole () {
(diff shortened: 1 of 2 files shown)


martinpitt (commented):
The change itself makes total sense -- an extra empty positional argument is bogus. However, I still don't understand at all how this actually works:

  - `virsh console --help` has no concept of an "alias". Supposedly this is `--devname <string>`, to select the particular console device that you want to attach to? But c-machines does not contain "devname" anywhere, how is that CLI even built? Is that relying on some obsolete/undocumented positional parameter for the devname instead of using the `--devname` option? If so, let's fix this.

 - How is that timing related? The bz report says that it works fine after a few seconds, but if you switch to the console immediately it fails. Are the device names something that get initialized  during VM bootup only? It sounds like it should be a static property of the VM definition.
 
Thanks!

KKoukiou:
> The change itself makes total sense -- an extra empty positional argument is bogus. However, I still don't understand at all how this actually works:
> 
>     * `virsh console --help` has no concept of an "alias". Supposedly this is `--devname <string>`, to select the particular console device that you want to attach to? But c-machines does not contain "devname" anywhere, how is that CLI even built? Is that relying on some obsolete/undocumented positional parameter for the devname instead of using the `--devname` option? If so, let's fix this.
> 
>     * How is that timing related? The bz report says that it works fine after a few seconds, but if you switch to the console immediately it fails. Are the device names something that get initialized  during VM bootup only? It sounds like it should be a static property of the VM definition.
> 
> 
> Thanks!



> The change itself makes total sense -- an extra empty positional argument is bogus. However, I still don't understand at all how this actually works:
> 
>     * `virsh console --help` has no concept of an "alias". Supposedly this is `--devname <string>`, to select the particular console device that you want to attach to? But c-machines does not contain "devname" anywhere, how is that CLI even built? Is that relying on some obsolete/undocumented positional parameter for the devname instead of using the `--devname` option? If so, let's fix this.

This is from the man virsh page.

```
   console
       Syntax:

          console domain [devname] [--safe] [--force]

       Connect the virtual serial console for the guest. The optional devname parameter refers to the device alias of an alternate console, serial or parallel device configured for the guest.  If omitted, the primary console will be
       opened.
```

So it's positional parameter, documented on the man page but not on the command help message.

> 
>     * How is that timing related? The bz report says that it works fine after a few seconds, but if you switch to the console immediately it fails. Are the device names something that get initialized  during VM bootup only? It sounds like it should be a static property of the VM definition.
> 
> 
> Thanks!

skobyda:
A test failure which at first seems unrelated, retrying https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-490-20220114-124627-3b00d5ca-arch/log.html#65-2

martinpitt (approved):
Thank you!

PR #491, opened 2021-12-19
Enable dynamic logging

martinpitt:
Drop the static `isDev` configuration flag and move to Cockpit's
standard way [1] of enabling debugging at runtime through setting
`window.debugging`.

Also drop the unused `logError()` helper, our code uses
`console.error()` directly and there is little reason to do otherwise.

[1] https://github.com/cockpit-project/cockpit/blob/main/HACKING.md#debug-logging-in-javascript-console

----

This would have helped me a lot to debug issue #488 on the downstream RHEL infra, where it's difficult to rebuild the webpack. Thus marking release-blocker.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -36,8 +36,6 @@ const VMS_CONFIG = {
             }
         }
     },
-    // TODO: make it configurable via config file
-    isDev: false, // Never commit with 'true'
 };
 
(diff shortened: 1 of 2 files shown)


marusak (approved):
Thanks!

PR #494, opened 2021-12-22
Do not change "startVm" value when selecting Unattended installation

skobyda:
This was causing several in several bugs in Create VM dialog:
1. uncheck "Start VM" checkbox, select OS with aunattended installation,
   click "X" button, "Start VM" button is checked.
2. uncheck "Start VM" checkbox, check "Unattended installation", uncheck
   "Unattended installation", "Start VM" checkbox is checked.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2033603

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -888,13 +888,15 @@ class CreateVmModal extends React.Component {
                 const converted = convertToUnit(stateDelta.minimumStorage, units.B, bestUnit);
                 this.setState({ storageSizeUnit: bestUnit.name }, () => this.onValueChanged("storageSize", converted));
             }
+
             if (!value || !value.unattendedInstallable)
                 this.onValueChanged('unattendedInstallation', false);
+
             this.setState(stateDelta);
             break;
         }
         case 'unattendedInstallation':
-            this.setState({ unattendedInstallation: value, startVm: true });
+            this.setState({ unattendedInstallation: value });
             break;
         default:
             this.setState({ [key]: value });
@@ -925,7 +927,7 @@ class CreateVmModal extends React.Component {
                 storageSize: convertToUnit(this.state.storageSize, this.state.storageSizeUnit, units.GiB),
                 storagePool: this.state.storagePool,
                 storageVolume: this.state.storageVolume,
-                startVm: this.state.startVm,
+                startVm: this.state.unattendedInstallation || this.state.startVm,
                 unattended: this.state.unattendedInstallation,
                 userPassword: this.state.userPassword,
                 rootPassword: this.state.rootPassword,
@@ -960,7 +962,7 @@ class CreateVmModal extends React.Component {
         let startVmCheckbox = (
             <FormGroup fieldId="start-vm" label={_("Immediately start VM")} hasNoPaddingTop>
                 <Checkbox id="start-vm"
-                    isChecked={this.state.startVm}
+                    isChecked={this.state.unattendedInstallation || this.state.startVm}
                     isDisabled={this.state.unattendedInstallation}
                     onChange={checked => this.onValueChanged('startVm', checked)} />
             </FormGroup>
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -696,6 +696,7 @@ class TestMachinesCreate(VirtualMachinesCase):
             self.browser = test_obj.browser
             self.machine = test_obj.machine
             self.assertTrue = test_obj.assertTrue
+            self.assertFalse = test_obj.assertFalse
             self.assertIn = test_obj.assertIn
             self.assertEqual = test_obj.assertEqual
             self.goToVmPage = test_obj.goToVmPage
@@ -798,9 +799,12 @@ class TestMachinesCreate(VirtualMachinesCase):
             b.wait_not_present("#navbar-oops")
 
             # click the 'X' button to clear the OS input and check there is no Ooops
+            # Also check start VM checkbox was not changed (https://bugzilla.redhat.com/show_bug.cgi?id=2033603)
+            b.set_checked("#start-vm", False)
             b.click("#os-select-group button[aria-label=\"Clear all\"]")
             b.wait_attr("#os-select-group input", "value", "")
             b.wait_not_present("#navbar-oops")
+            self.assertFalse(b.get_checked("#start-vm"))
 
             return self
 


jelly (approved):

martinpitt:
test/check-machines-create:800:85: W291 trailing whitespace

src/components/vm/hostdevs/hostDevAdd.jsx
```diff
@@ -88,10 +88,10 @@ const DevRow = ({ idPrefix, type, selectableDevices, setSelectableDevices }) =>
...
             cells.push(getOptionalValue(bus, `${id}-bus`, _("Bus")));
         } else if (nodeDev.capability.type === "pci") {
-            let domain = nodeDev.capability.domain;
-            let bus = nodeDev.capability.bus;
-            let slot = nodeDev.capability.slot;
-            let func = nodeDev.capability.function;
+            let domain = Number(nodeDev.capability.domain);
+            let bus = Number(nodeDev.capability.bus);
+            let slot = Number(nodeDev.capability.slot);
+            let func = Number(nodeDev.capability.function);
```
martinpitt, line 94: This is an unrelated change -- can you please move it to a separate commit?

test/check-machines-create
```diff
@@ -798,9 +798,12 @@ class TestMachinesCreate(VirtualMachinesCase):
             b.wait_not_present("#navbar-oops")
 
             # click the 'X' button to clear the OS input and check there is no Ooops
+            # Also check start VM checkbox was not changed (https://bugzilla.redhat.com/show_bug.cgi?id=2033603)
+            b.set_checked("#start-vm", False)
             b.click("#os-select-group button[aria-label=\"Clear all\"]")
             b.wait_attr("#os-select-group input", "value", "")
             b.wait_not_present("#navbar-oops")
+            self.assertTrue(not b.get_checked("#start-vm"))
```
martinpitt, line 806: assertFalse() and drop the `not`

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -925,7 +927,7 @@ class CreateVmModal extends React.Component {
                 storageSize: convertToUnit(this.state.storageSize, this.state.storageSizeUnit, units.GiB),
                 storagePool: this.state.storagePool,
                 storageVolume: this.state.storageVolume,
-                startVm: this.state.startVm,
+                startVm: this.state.unattendedInstallation || this.state.startVm,
```
martinpitt, line 930: This is weird UI-wise -- it means you can uncheck the "immediately start" checkbox, but it will still auto-start the VM anyway, because it's unattended? If unattended install really doesn't work with a delayed start, then please add the reason as a comment here. But then the previous behaviour was actually right, and the checkbox *should* be auto-checked when you select unattended. But then it should also be disabled in that case.

martinpitt (changes requested):
Thanks! Please rebase this, so that packit tests can run.

src/components/vm/hostdevs/hostDevAdd.jsx
skobyda: Ahhh, how this got here, this should not even be in his PR

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: > This is weird UI-wise -- it means you can uncheck the "immediately start" checkbox, but it will still auto-start the VM anyway, because it's unattended? If unattended install really doesn't work with a delayed start, then please add the reason as a comment here. But then the previous behaviour was actually right, and the checkbox _should_ be auto-checked when you select unattended. But then it should also be disabled in that case.

No, that is exactly current behavior which this PR doesn't ammend, it was like that before this PR and will still be even with this PR. When unattended installation checkbox is checked, then the startVm checkbox is disabled and automatically shown as checked.

What I'm trying to change in this PR is that startVm value doesn't get overridden in two bugged cases:
1. uncheck "Start VM" checkbox, select OS with unattended installation, check "Unattended installation", uncheck "Unattended installation", "Start VM" checkbox is checked.
2. uncheck "Start VM" checkbox, select OS with unattended installation,  click "X" button, "Start VM" button is checked.
So to fix both cases, I changed
```
-  this.setState({ unattendedInstallation: value, startVm: true });
+ this.setState({ unattendedInstallation: value });
```
Because when you click on "X" button, it calls onValueChanged("os", null), which in turns call onValueChanged("unattendedInstallation", false), which calls the above mentioned code, which I think wasn't logical because what it basically did was that whenever state of "unattendedInstallation" changed, it just set startVm to true which was cases both above mentioned bugs.

src/components/vm/hostdevs/hostDevAdd.jsx
skobyda: fixed

test/check-machines-create
skobyda: fixed

martinpitt:
[This](https://logs.cockpit-project.org/logs/pull-494-20220114-141226-ec0e0c74-fedora-35-firefox/log.html#29-2) is an interesting flake, given the [screenshot](https://logs.cockpit-project.org/logs/pull-494-20220114-141226-ec0e0c74-fedora-35-firefox/TestMachinesMigration-testCopyStorageMigration-fedora-35-127.0.0.2-2501-FAIL.png) -- VM migration fails due to an SSH key prompt and then an "input/output error". Either the former always happens and something went wrong in the networking (although that's all local), or the SSH prompt is what actually breaks the copy, as nothing answers it.

However, it is unrelated. I'll push a clean rebase, to also run the packit tests.

src/components/create-vm-dialog/createVmDialog.jsx
martinpitt: Makes sense, thanks! I tested this now, and it works correctly.

martinpitt (approved):
Cheers!

martinpitt:
Hmm, the delete race is [really](https://logs.cockpit-project.org/logs/pull-494-20220118-063123-a344679c-arch/log.html#65-1) [bad](https://logs.cockpit-project.org/logs/pull-494-20220118-063125-a344679c-rhel-8-5/log.html#67) today -- I already retried once, trying again.

martinpitt:
No really, [it's *hard*](https://logs.cockpit-project.org/logs/pull-494-20220118-092456-a344679c-rhel-8-5/log.html)!

PR #502, opened 2022-01-04
Delete disks only after VM is successfully undefined and destroyed

skobyda:
A relevant bug was observed on s390x arch.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2031678

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -324,13 +324,8 @@ export function domainDelete({
     options,
     storagePools
 }) {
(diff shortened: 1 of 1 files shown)


martinpitt:
Breaks all tests. Please rebase to current main when you push next, so that packit tests can run. Thanks!

skobyda:
This passes checks now.

martinpitt (approved):
Thanks! This makes sense, destroying storage with a running VM sounds brittle.

PR #503, opened 2022-01-04
Show install_machine.py invocation in debug logs

martinpitt:
That makes it possible to reproduce and investigate creation failures on
the command line, by calling install_machine.py with the logged JSON
string.

---

This helped me with investigating https://bugzilla.redhat.com/show_bug.cgi?id=2032462 . Originally I thought this was our bug, but it's not, so only that little change remained..


skobyda (approved):
LGTM

PR #508, opened 2022-01-06
Support the webpack rsync plugin

jelly:
These changes allow build artifacts to be synced to a test virtual machine using `RSYNC=c npm run watch`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,7 +200,7 @@ test/reference: test/common
 # checkout Cockpit's PF/React/build library; again this has no API stability guarantee, so check out a stable tag
 $(LIB_TEST):
 	flock Makefile sh -ec '\
-	    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 257; \
+	    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 260; \
 	    git checkout --force FETCH_HEAD -- pkg/lib; \
 	    git reset -- pkg/lib'
 	mv pkg/lib src/ && rmdir -p pkg
--- a/src/components/storagePools/storagePoolOverviewTab.jsx
+++ b/src/components/storagePools/storagePoolOverviewTab.jsx
@@ -23,8 +23,6 @@ import { DescriptionList, DescriptionListGroup, DescriptionListTerm, Description
 import { storagePoolId } from '../../helpers.js';
 import cockpit from 'cockpit';
 
-import 'form-layout.scss';
-
 const _ = cockpit.gettext;
 
 export const StoragePoolOverviewTab = ({ storagePool }) => {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,3 +1,4 @@
+const fs = require("fs");
 const path = require("path");
 
 const copy = require("copy-webpack-plugin");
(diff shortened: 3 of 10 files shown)


src/components/storagePools/storagePoolOverviewTab.jsx
```diff
@@ -23,8 +23,6 @@ import { DescriptionList, DescriptionListGroup, DescriptionListTerm, Description
 import { storagePoolId } from '../../helpers.js';
 import cockpit from 'cockpit';
 
-import 'form-layout.scss';
```
martinpitt, line 26: This is a bit unclear and buggy in the current code -- a lot of files (such as this one) include the CSS, but the only place where it is actually being used is src/components/vm/overview/bootOrder.jsx . That one would need to be ported first.

Makefile
```diff
@@ -200,7 +200,7 @@ test/reference: test/common
 # checkout Cockpit's PF/React/build library; again this has no API stability guarantee, so check out a stable tag
```
martinpitt, line 200: The "Update cockpit's lib to 260" commit mentions "podman"; perhaps say "our CSS"?

webpack.config.js
```diff
@@ -7,12 +8,16 @@ const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const CompressionPlugin = require("compression-webpack-plugin");
 const ESLintPlugin = require('eslint-webpack-plugin');
 const CockpitPoPlugin = require("./src/lib/cockpit-po-plugin");
+const CockpitRsyncPlugin = require("./src/lib/cockpit-rsync-plugin");
```
martinpitt, line 11: This should get the corresponding README changes.

martinpitt (changes requested):
Thanks! Sad that this is blocked on form-layout, though :cry:

martinpitt:
The e2e failures about `npm install` are about the internal npm mirror going down; investigating. If there is no quick fix, I'll just drop it.

jelly:
> Thanks! Sad that this is blocked on form-layout, though cry

I'll take a look at porting, I dropped the include and visually it looked ok to me so I wasn't 100% sure if the include was required.

martinpitt:
> I dropped the include and visually it looked ok to me

If there's no visual difference, then perhaps it actually is obsolete; then I suppose we should just rename `ct-form` class to something else, along with the rule in src/components/vm/overview/bootOrder.css ?

jelly:
> > I dropped the include and visually it looked ok to me
> 
> If there's no visual difference, then perhaps it actually is obsolete; then I suppose we should just rename `ct-form` class to something else, along with the rule in src/components/vm/overview/bootOrder.css ?

Ok, this is now blocked on https://github.com/cockpit-project/cockpit-machines/pull/191/files or at least porting it to use a DataList and DescriptionList. Leaving the draggable for later.

src/components/storagePools/storagePoolOverviewTab.jsx
jelly: bootOrder.jsx is now ported in 1d680c5833cbd7bcf57d26cd67bff3cc3a976a26

Makefile
jelly: Copy pasta-error, adjusted.

webpack.config.js
jelly: Done.

jelly:
Everything should be ok now, let's see if the pixel tests agree.

martinpitt (approved):
!! Thanks! :100:

PR #510, opened 2022-01-07
test: Robustify TestMachinesCreate.testCreateCloudBaseImage

martinpitt:
After destroying the VM, wait until the UI caught up before clicking
"Delete".

----

This fixes [this failure](https://logs.cockpit-project.org/logs/pull-2729-20211216-133824-fd3dfae8-rhel-8-6-cockpit-project-cockpit-machines/log.html#9), which also happens quite often in downstream gating.

See  #488 for debugging details.

### DIFF
--- a/src/libvirtApi/common.js
+++ b/src/libvirtApi/common.js
@@ -147,7 +147,7 @@ function delayPolling(action, timeout) {
     }
 }
 
-// Undefined the VM from Redux store only if it"s not transient
+// Undefined the VM from Redux store only if it's not transient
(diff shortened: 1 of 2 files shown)


martinpitt:
This is [not the full story yet](https://logs.cockpit-project.org/logs/pull-510-20220107-084813-1965acfb-rhel-9-0/log.html#65-2), but I know that -- debugging still going on in #488. But this at least helps the test work in 10/10 cases for me locally (before, it usually failed after the second to third run)

marusak (approved):
Danke!

PR #511, opened 2022-01-07
Use Patternfly's react-icons for up/down icons

jelly:
The icons are currently not shown most likely due to an patternfly
update, as react-icons are to be used these days.


martinpitt (approved):
Thanks!

PR #512, opened 2022-01-07
Port bootorder dialog

jelly:


KKoukiou (dismissed):
Very cool thanks! 

I find the up/down buttons slightly large, so I would rather use the isSmall variant, but fine for going in as is.

jelly:
> Very cool thanks!
> 
> I find the up/down buttons slightly large, so I would rather use the isSmall variant, but fine for going in as is.

I've looked at adding `isSmall` but it does not visually change the button neither does `<AngleUpIcon size="sm"/>` make it smaller. I'd however like to merge it to unblock the other open PR.

martinpitt (approved):

PR #515, opened 2022-01-12
Drop patternfly-react

jelly:
After the bootorder dialog was ported to use PF4 we can now finally drop
it. Add deep-equal as a new dependency previously provided by
patternfly-react and required for src/lib/{hooks.js,notifications.js}.


martinpitt:
Nice! However, you can clean up more -- change src/index.js to import patternfly-4-cockpit.scss only; patternfly-cockpit.scss is both PF 3 and 4. And then you can kill all the hacks in webpack.config.js for `/patternfly-cockpit.scss`, and drop the "FIXME: Ignore warnings from PF3 table" bit. Probably best to diff it against starter-kit's webpack.config.js and see what is still necessary.

webpack.config.js
```diff
@@ -102,71 +97,14 @@ module.exports = {
...
-            },
             {
                 test: /patternfly-4-cockpit.scss$/,
                 use: [
                     extract.loader,
                     {
                         loader: 'css-loader',
                         options: {
-                            sourceMap: !production,
+                            sourceMap: true,
```
martinpitt, line 98: This looks odd, but starter-kit and machines do it that way as well. I just wonder why? But that should be in another commit, it's not related to dropping PF3.

webpack.config.js
```diff
@@ -190,7 +128,6 @@ module.exports = {
                         options: {
                             sourceMap: !production,
                             sassOptions: {
-                                quietDeps: true,
```
martinpitt, line 193: Was this a workaround for PF3? We have quietDeps in cockpit and cockpit-composer as well. Mind if you back that out and investigate separately?

webpack.config.js
```diff
@@ -25,10 +23,7 @@ const copy_files = [
 const plugins = [
     new copy({ patterns: copy_files }),
     new extract({filename: "[name].css"}),
-    new ESLintPlugin({
-        extensions: ["js", "jsx"],
-        failOnWarning: true,
```
martinpitt, line 30: That seems useful though -- perhaps we should rather enable that in starter-kit as well?

martinpitt (changes requested):
Thanks! Some minor clean-ups, but let's wait for a round of test results.

webpack.config.js
jelly: I'm not sure, everything compiled fine and looked fine so hmmm :eyes:

webpack.config.js
jelly: Sure, I'll leave it out for now. The whole point of the exercise was to get the WebpackRsyncPlugin in :+1:

martinpitt:
Nice! Pretty much all the pixel tests fail now; tests are still running, hopefully just some minor noise. We had the same when dropping PF3 from cockpit as well -- i.e. unless there is a dramatic change, we can just update to the new ref images

jelly:
> Nice! Pretty much all the pixel tests fail now; tests are still running, hopefully just some minor noise. We had the same when dropping PF3 from cockpit as well -- i.e. unless there is a dramatic change, we can just update to the new ref images

Not entirely, this introduces a regression https://logs.cockpit-project.org/logs/pull-515-20220112-154443-c80c6cfa-fedora-35-firefox/pixeldiff.html#TestMachinesCreate-testCreateBasicValidation-auto

jelly:
> > Nice! Pretty much all the pixel tests fail now; tests are still running, hopefully just some minor noise. We had the same when dropping PF3 from cockpit as well -- i.e. unless there is a dramatic change, we can just update to the new ref images
> 
> Not entirely, this introduces a regression https://logs.cockpit-project.org/logs/pull-515-20220112-154443-c80c6cfa-fedora-35-firefox/pixeldiff.html#TestMachinesCreate-testCreateBasicValidation-auto

Fixed, by re-adding the webpack workaround, I added the patternfly issue with details so we at least know why it's there.

https://logs.cockpit-project.org/logs/pull-515-20220112-164341-da799715-fedora-35-firefox/pixeldiff.html#TestMachinesDisks-testAddDiskNFS-vm-add-disk-modal-nfs

This seems to be a valid change, all our other dialogs don't have a black rounded corner around the cancel button.

https://logs.cockpit-project.org/logs/pull-515-20220112-164341-da799715-fedora-35-firefox/pixeldiff.html#TestMachinesCreate-testCreateBasicValidation-iso

Shows a nice issue in the test where it might have flaked? 10.4 versus 10.5 Gb? Can't say with certainty which one to pick.

martinpitt:
Nice! I spot-checked the other pixel failures, and they seem like okay-ish noise to me.

martinpitt:
> 10.4 versus 10.5 Gb? 

The delta is green, which means it's part of the ignored selectors. You can ignore that one, it just failed on the slightly different label positions (the red delta)

webpack.config.js
```diff
@@ -199,13 +136,13 @@ module.exports = {
...
                 test: /\.s?css$/,
-                exclude: /patternfly-(4-)?cockpit.scss/,
+                exclude: /patternfly-4-cockpit.scss/,
                 use: [
                     extract.loader,
                     {
                         loader: 'css-loader',
                         options: {
-                            sourceMap: !production,
+                            sourceMap: true,
```
KKoukiou, line 145: This change is not correct.

KKoukiou (changes requested):

package.json
```diff
@@ -59,8 +59,8 @@
     "@patternfly/react-styles": "4.25.4",
     "@patternfly/react-table": "4.44.4",
     "date-fns": "2.28.0",
+    "deep-equal": "^2.0.5",
     "patternfly": "3.59.5",
```
KKoukiou: You can also drop patternfly

webpack.config.js
jelly: Thanks, fixed!

martinpitt:
Remaining 4 pixel diffs look reasonable and should be updated. Plus, this needs a rebase to fix a conflict. Thanks!

martinpitt:
The [f34](http://artifacts.dev.testing-farm.io/cd07031e-4b8d-408c-bae0-23c61f78f971/) and [rawhide](http://artifacts.dev.testing-farm.io/780aa96e-c47a-4570-8fa5-3830e018756d/) packit tests were completely broken, retrying.

martinpitt:
TF results are a disaster right now, but I don't think it's related to this PR -- I see it in PR #518 as well. Something changed between yesterday (#516) and today which makes it go bonkers.

martinpitt (approved):
Thanks a lot for the cleanups and bearing all that noise! :hugs:

martinpitt:
I sent PR #520 to ensure that this happens on current main as well.

martinpitt:
#520 has the same problem, investigating it there.

PR #489, opened 2021-12-17
Debug delete race in issue #488

martinpitt:
Debug PR for issue #488

PR #516, opened 2022-01-12
Fix race condition on deleting a VM (alternative to #518)

### LINKED ISSUE #488: TestMachinesCreate.testCreateUrlSource race condition on deleting VM
Various creation tests are very unstable particular in downstream RHEL gating. We got a [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2033367) about that, it also happens [in packit](http://artifacts.dev.testing-farm.io/54c6774e-4557-48b4-8cef-fc1fb5db7ce3/) and in [our own CI](https://logs.cockpit-project.org/logs/pull-2729-20211216-133824-fd3dfae8-rhel-8-6-cockpit-project-cockpit-machines/log.html#9) ([another example](https://logs.cockpit-project.org/logs/pull-2729-20211216-122348-fd3dfae8-rhel-8-6-cockpit-project-cockpit-machines/log.html#9), [yet another one](https://logs.cockpit-project.org/logs/pull-504-20220107-065530-b22e4b0a-rhel-8-6/log.html#9)). The pattern is this:

```
  File "/var/ARTIFACTS/work-allR0438v/plans/all/discover/default/tests/test/check-machines-create", line 173, in testCreateCloudBaseImage
    runner.createCloudBaseImageTest(TestMachinesCreate.VmDialog(self, sourceType='cloud',
  File "/var/ARTIFACTS/work-allR0438v/plans/all/discover/default/tests/test/check-machines-create", line 1421, in createCloudBaseImageTest
    self._deleteVm(dialog) \
  File "/var/ARTIFACTS/work-allR0438v/plans/all/discover/default/tests/test/check-machines-create", line 1269, in _deleteVm
    self.test_obj.waitVmRow(dialog.name, "system", False)
[...]
wait_js_cond(!ph_is_present("tbody tr[data-row-id=vm-subVmTestCreate2-system]")): timeout
```

What happens is:

 * VM gets created, `virt-install` is started and keeps running.
 * VM moves to running state. The test clicks "delete".
 * This aborts virt-install, so the UI shows a notification "Installation failed", and the console shows a warning

    > warning: ERROR    Domain not found: no domain with matching uuid '1eb4dcc8-fb5d-4def-ac8c-af2fd01d1c90' (subVmTestCreate7)

    This part *always* happens. It's unclean, but not what breaks the test.

 * *sometimes* going back to the overview is doing a `domainGet()` call which fails:

    > warning: GET_VM action failed for path /org/libvirt/QEMU/domain/_010f40e4_e9bd_479f_bf7c_19fa037f525f Domain not found: no domain with matching uuid '010f40e4-e9bd-479f-bf7c-19fa037f525f'

   In that case, the "ghost" VM stays around, and the test times out waiting for the VM row to go away.

I went through all usages of domainGet() and domainGetAll() and none seem to apply to deleting a VM. src/components/vm/deleteDialog.jsx's delete() just calls src/libvirtApi/domain.js's domainDelete() and then just cockpit.location.go(["vms"]), i.e. it relies on the list page to update the VM state. This now requires drilling through the redux stuff, which I just find annoyingly unintuitive/indirect/hard to understand.

martinpitt:
When deleting a VM in the UI, it gets both destroyed and undefined at
the same time. The async handling in domainEventUndefined() and
domainEventStopped() can run in dozens of combinations, but there is one
common case which fails:

 - domainEvent.Undefined signal gets received first, and the handler
   starts
 - domainEvent.Undefined finishes its ListDomains call in the meantime,
   and as the domain is not destroyed yet, it is still in the result; so
   the handler issues a domainGet() to update properties
 - the domain gets destroyed now, so the domainGet() fails and logs the
   `GET_VM action failed` warning; as a result, properties in redux do
   not get updated -- in particular, `persistent` remains `true` (it
   would have been updated to `false` if the domainGet() call succeeded)
 - the corresponding domainEvent.Stopped signal arrives and
   domainEventStopped() gets called; as the domain does not exist any
   more, it is not in the ListDomains() result, and thus it calls
   undefineVm() with `transientOnly: true`. However, as the redux model
   still thinks the domain is persistent, this is a no-op.

In *general*, the logic in domainEvent{Stopped,Undefine}() is correct
for the cases where the handlers don't run in parallel. The bug is that
it entirely ignores the `GET_VM` action -- if that fails, the domain is
gone, and we should always update our redux model accordingly.

To fix that, dispatch an undefineVm() in domainGet() if it fails --
ignoring bugs in libvirt, this could only have happened as a race
condition for trying to read a domain while it is being destroyed.

Fixes #488
https://bugzilla.redhat.com/show_bug.cgi?id=2033367

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -35,6 +35,7 @@ import {
 } from '../libvirtUtils.js';
 import {
     deleteUnlistedVMs,
+    undefineVm,
     updateOrAddVm,
     updateVm,
 } from '../actions/store-actions.js';
@@ -502,7 +503,14 @@ export function domainGet({
     let props = {};
     let domainXML;
 
-    return call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_SECURE], { timeout, type: 'u' })
+    const promiseGetXML = call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_SECURE], { timeout, type: 'u' });
+
+    promiseGetXML.catch(ex => {
+        logDebug("GET_VM action GetXMLDesc failed for path", objPath, ex.toString(), "undefining VM");
+        store.dispatch(undefineVm({ connectionName, id: objPath }));
+    });
+
+    return promiseGetXML
             .then(domXml => {
                 domainXML = domXml[0];
                 return call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_SECURE | Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' });
--- a/src/libvirtApi/common.js
+++ b/src/libvirtApi/common.js
@@ -154,7 +154,7 @@ function domainEventUndefined(connectionName, domPath) {
                 if (!objPaths[0].includes(domPath))
                     store.dispatch(undefineVm({ connectionName, id: domPath }));
                 else
-                    domainGet({ connectionName, id:domPath, updateOnly: true });
+                    domainGet({ connectionName, id: domPath, updateOnly: true });
             })
             .catch(ex => console.warn("ListDomains action failed:", ex.toString()));
 }
@@ -164,7 +164,7 @@ function domainEventStopped(connectionName, domPath) {
     call(connectionName, "/org/libvirt/QEMU", "org.libvirt.Connect", "ListDomains", [0], { timeout, type: "u" })
             .then(objPaths => {
                 if (objPaths[0].includes(domPath))
-                    domainGet({ connectionName, id:domPath, updateOnly: true });
+                    domainGet({ connectionName, id: domPath, updateOnly: true });
                 else // Transient vm will get undefined when stopped
                     store.dispatch(undefineVm({ connectionName, id:domPath, transientOnly: true }));
             })


src/libvirtApi/domain.js
```diff
@@ -547,7 +548,10 @@ export function domainGet({
 
                 snapshotGetAll({ connectionName, domainPath: objPath });
             })
-            .catch(ex => console.warn("GET_VM action failed for path", objPath, ex.toString()));
+            .catch(ex => {
+                console.warn("GET_VM action failed for path", objPath, ex.toString());
+                store.dispatch(undefineVm({ connectionName, id: objPath }));
```
martinpitt, line 553: If this is too much/wide, we could alternatively let domainGet() return the promise, and leave the error handling to all callers. @skobyda, @KKoukiou , can you think of a situation where this function would fail in a condition other than the VM actually going away?

martinpitt:
It seems testCreateUrlSource does not like much being [run three times in a row](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-516-20220112-130519-8d39014f-debian-stable/log.html#71-1), it seems to corrupt the VM? This failure is already on creation, not deletion.

But also, I want to test an alternate bug fix, see #489

src/libvirtApi/domain.js
martinpitt: I filed #518 with that alternative implementation. I can honestly not judge which one would be conceptually better, as I am not sure if a GET_VM is expected under normal circumstances other than during domain deletion.

skobyda: I looked at https://gitlab.com/libvirt/libvirt/-/blob/master/src/conf/domain_conf.c#L28124 which is basically where the magic behind `GetXMLDesc` API happens, and although there are many cases defined when error status is returned, deep down most of them are just result of errors which may happen during string manipulation, when XML string is formatted.
And no, I did not seen a situation when domainGet failed apart from just being non-existent (like in the bug which this fix).

> as I am not sure if a GET_VM is expected under normal circumstances other than during domain deletion. 

What do you mean? If GET_VM function is called in under places apart from deletion, or if just error message is expected to happen in other situations apart from deletion?

skobyda: What gets me a bit more worried is a case when `snapshotGetAll` fails, and as a result of this failure, the VM will be undefined from redux store.

But since the GET_VM in our bug is coming only `GetXMLDesc`, maybe we can limit `store.dispatch(undefineVm({ connectionName, id: objPath }));` only to failures caused by `GetXMLDesc`? And I think it would be safe to just undefine VM from store if `GetXMLDesc` fails. Since if it fails, that means we don't really have any (or updated) data about the VM, and there is nothing to show about the VM in the UI, so practically there is no reason to have it in our redux store.

martinpitt: > What do you mean? If GET_VM function is called in under places apart from deletion, or if just error message is expected to happen in other situations apart from deletion?

I mean the latter -- if we e.g. expect GET_VM/GetXMLDesc to fail on creation, in particular in conjuction with the placeholder domains that we are (or were?) creating in some cases. At least the tests don't run into such a situation.

> What gets me a bit more worried is a case when snapshotGetAll fails, and as a result of this failure, the VM will be undefined from redux store.

That's a good point.

> maybe we can limit `store.dispatch(undefineVm({ connectionName, id: objPath }));` only to failures caused by GetXMLDesc?

That's a good idea! I'll change the PR accordingly.

martinpitt: I checked the code, and snapshotGetAll() can't really fail. I get this *all the time* on RHEL 8.6, as it's unsupported:

> LIST_DOMAIN_SNAPSHOTS action failed for domain /org/libvirt/QEMU/domain/_9728f5f7_fe69_412a_8c99_51fdc8bb34a8: {"problem":null,"name":"org.freedesktop.DBus.Error.UnknownMethod","message":"No such method 'ListDomainSnapshots'"} index.js:2:1422511

and snapshotGetAll() also just has a catch-all and ignores all errors. But that is bad design -- it should check the error and *only* ignore `UnknownMethod`, not all the others.

But it's still conceptually better to catch GetXML failures only, of course.

martinpitt: Updated to your suggestion.

martinpitt:
Updated to @skobyda 's suggestion from above. This works pretty well, the current run in the "torture PR" #488 are all successful, and the [debug log](https://logs.cockpit-project.org/logs/pull-489-20220114-062508-02d63f79-rhel-8-6/log.html) shows a few cases of this race condition happening and now getting handled:

```
> debug: signal on /org/libvirt/QEMU: org.libvirt.Connect.DomainEvent(["/org/libvirt/QEMU/domain/_8cae25b3_78ae_4413_86a6_84da7c906968",5,1])
> debug: domainEventStopped system /org/libvirt/QEMU/domain/_8cae25b3_78ae_4413_86a6_84da7c906968
> debug: signal on /org/libvirt/QEMU: org.libvirt.Connect.DomainEvent(["/org/libvirt/QEMU/domain/_8cae25b3_78ae_4413_86a6_84da7c906968",1,0])
> debug: domainEventUndefined system /org/libvirt/QEMU/domain/_8cae25b3_78ae_4413_86a6_84da7c906968
> debug: GET_VM action GetXMLDesc failed for path /org/libvirt/QEMU/domain/_8cae25b3_78ae_4413_86a6_84da7c906968 Domain not found: no domain with matching uuid '8cae25b3-78ae-4413-86a6-84da7c906968' undefining VM
> debug: reducer vms: action={"type":"UNDEFINE_VM","id":"/org/libvirt/QEMU/domain/_8cae25b3_78ae_4413_86a6_84da7c906968","connectionName":"system"}
```

Tests are green here as well. @skobyda , PTAL?

src/libvirtApi/domain.js
```diff
@@ -502,7 +503,14 @@ export function domainGet({
     let props = {};
     let domainXML;
 
-    return call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_SECURE], { timeout, type: 'u' })
+    const p_getxml = call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_SECURE], { timeout, type: 'u' });
```
skobyda, line 506: Just a little nitpick, we don't use this_naming_practive almost anywhere in JS code for variables and functions, only for enums and python tests. Instead thisNamingPractice could fit better

skobyda (dismissed):
This looks good. and I prefer it more than #518, since as I described in comment above, if GetXMLDesc fails, the we would not really have any or have non-updated data about VM, so I think it's fine to undefine it from store for any domainGet function calls, not just few specific ones.

src/libvirtApi/domain.js
martinpitt: Sure! Renamed to `promiseGetXML`.

skobyda:
There is a test failure: https://logs.cockpit-project.org/logs/pull-516-20220114-130323-a8487b3c-rhel-8-6/log.html#9-2
It seems that after VM creation, the test tries to cleanup and delete the VM, but the VM is still present in the UI.

martinpitt:
@skobyda: Yes, that's next on my list -- see https://github.com/cockpit-project/cockpit-machines/issues/488#issuecomment-1010924881 . That's yet another race condition which I didn't investigate closely yet, but I see that all the time.

src/libvirtApi/domain.js
martinpitt: This was wrong after all -- it only closed the race condition for the first GetXML call, but the VM could just as well go away on the second one or the GetState(). I'll send another PR.

PR #518, opened 2022-01-13
Fix race condition on deleting a domain (alternative to #516)

### LINKED ISSUE #488: TestMachinesCreate.testCreateUrlSource race condition on deleting VM

martinpitt:
When deleting a domain in the UI, it gets both destroyed and undefined at
the same time. The async handling in domainEventUndefined() and
domainEventStopped() can run in dozens of combinations, but there is one
common case which fails:

 - domainEvent.Undefined signal gets received first, and the handler
   starts
 - domainEvent.Undefined finishes its ListDomains call in the meantime,
   and as the domain is not destroyed yet, it is still in the result; so
   the handler issues a domainGet() to update properties
 - the domain gets destroyed now, so the domainGet() fails and logs the
   `GET_VM action failed` warning; as a result, properties in redux do
   not get updated -- in particular, `persistent` remains `true` (it
   would have been updated to `false` if the domainGet() call succeeded)
 - the corresponding domainEvent.Stopped signal arrives and
   domainEventStopped() gets called; as the domain does not exist any
   more, it is not in the ListDomains() result, and thus it calls
   undefineVm() with `transientOnly: true`. However, as the redux model
   still thinks the domain is persistent, this is a no-op.

In *general*, the logic in domainEvent{Stopped,Undefine}() is correct
for the cases where the handlers don't run in parallel. The bug is that
it entirely ignores the `GET_VM` action -- if that fails, the domain is
gone, and we should always update our redux model accordingly.

To fix that, add a flag to domainGet() to actually return promise
rejections instead of ignoring them, and dispatch an undefineVm() in
the Undefined/Stopped handlers() if it fails -- then the domain went
away while the handler was running, and we should forget about it.

Fixes #488
https://bugzilla.redhat.com/show_bug.cgi?id=2033367

----

This is an alternative implementation to #516 which is more targeted, but also more complicated.

### DIFF
--- a/src/libvirtApi/common.js
+++ b/src/libvirtApi/common.js
@@ -154,7 +154,7 @@ function domainEventUndefined(connectionName, domPath) {
                 if (!objPaths[0].includes(domPath))
                     store.dispatch(undefineVm({ connectionName, id: domPath }));
                 else
-                    domainGet({ connectionName, id:domPath, updateOnly: true });
+                    domainGet({ connectionName, id: domPath, updateOnly: true });
             })
             .catch(ex => console.warn("ListDomains action failed:", ex.toString()));
 }
@@ -164,7 +164,7 @@ function domainEventStopped(connectionName, domPath) {
     call(connectionName, "/org/libvirt/QEMU", "org.libvirt.Connect", "ListDomains", [0], { timeout, type: "u" })
             .then(objPaths => {
                 if (objPaths[0].includes(domPath))
-                    domainGet({ connectionName, id:domPath, updateOnly: true });
+                    domainGet({ connectionName, id: domPath, updateOnly: true });
                 else // Transient vm will get undefined when stopped
                     store.dispatch(undefineVm({ connectionName, id:domPath, transientOnly: true }));
             })
--- a/src/libvirtApi/common.js
+++ b/src/libvirtApi/common.js
@@ -151,10 +151,15 @@ function delayPolling(action, timeout) {
 function domainEventUndefined(connectionName, domPath) {
     call(connectionName, "/org/libvirt/QEMU", "org.libvirt.Connect", "ListDomains", [Enum.VIR_CONNECT_LIST_DOMAINS_TRANSIENT], { timeout, type: "u" })
             .then(objPaths => {
-                if (!objPaths[0].includes(domPath))
+                if (!objPaths[0].includes(domPath)) {
                     store.dispatch(undefineVm({ connectionName, id: domPath }));
-                else
-                    domainGet({ connectionName, id: domPath, updateOnly: true });
+                } else {
+                    domainGet({ connectionName, id: domPath, updateOnly: true, reportErrors: true })
+                            .catch(e => {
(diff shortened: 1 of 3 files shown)


martinpitt:
So this is obviously unhappy. But before I go ahead and debug this, I'm interested in @skobyda and @KKoukiou 's opinion about which approach you like more? This one or #516?

martinpitt:
We'll land #516.

PR #520, opened 2022-01-13
Speed up build workflow, README update

martinpitt:


martinpitt:
OK, TF has the [same problem](http://artifacts.dev.testing-farm.io/7f256f0c-3e6e-47d1-bb63-1db6619347ee/). I don't get the issue with running with  `TEST_BROWSER=firefox` locally, even though my "real" firefox was also affected by today's [http3 FUBAR](https://bugzilla.mozilla.org/show_bug.cgi?id=1749908). But let's try to disable http3 and see if that makes a difference. I pushed a quick hack which tests that theory.

martinpitt:
srpm build got stuck, so re-pushed. I used the opportunity to try and [speed up build-dist](https://github.com/cockpit-project/cockpit-machines/runs/4801460317?check_suite_focus=true). Setting up the dependencies took 1:11 minutes, while the actual build takes 1:54 minutes. Let's make the former faster.

martinpitt:
The hack worked, so I sent it to https://github.com/cockpit-project/cockpit/pull/16828 and will replace the seddery here with a test/common bump once that lands.

martinpitt:
The firefox networking issue got fixed upstream, so presumably we don't really need the hack any more. Doing another run without.

jelly (approved):

PR #523, opened 2022-01-18
Fix VM deletion race condition with domainGet() harder

martinpitt:
The fix in commit a724e2e29c933 was incomplete -- the VM deletion can
just as well happen after the first, but before the second GetXMLDesc()
call in domainGet(), or the GetState() and property get -- all of these
mean that the VM is really gone.

The only thing we need to be concerned about is the snapshotGetAll()
call, but that does its own error handling (and ignores them).
So go back to undefining the VM whenever anything in domainGet() fails,
as in earlier versions of PR #516.

-----

I also added some small cleanups.

I am still debugging this in #489, with "loudness 11" debug level.

### DIFF
--- a/src/libvirtApi/common.js
+++ b/src/libvirtApi/common.js
@@ -239,7 +239,7 @@ function getNodeMaxMemory({ connectionName }) {
     return call(connectionName, "/org/libvirt/QEMU", "org.libvirt.Connect", "NodeGetMemoryStats", [-1, 0], { timeout, type: "iu" })
             .then(stats => store.dispatch(setNodeMaxMemory({ memory: stats[0].total })))
             .catch(ex => {
-                console.warn("NodeGetMemoryStats failed: %s", ex);
+                console.warn("NodeGetMemoryStats failed:", ex.toString());
                 return Promise.reject(ex);
             });
 }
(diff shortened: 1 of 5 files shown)


martinpitt:
Better, but still not good. The `testCreateCloudBaseImage` race has existed for a long time, and I created issue #524 about it. But I have to debug this in #489, the failures [here](https://logs.cockpit-project.org/logs/pull-523-20220118-134737-7860eda0-rhel-8-6/log.html) are not nearly detailed enough.

martinpitt:
packit tests are unrelated, see PR #527

martinpitt:
I added an attempted fix for `testCreateCloudBaseImage()` (see #526) and rebased to pick up the packit fix PR #527

martinpitt:
OK, progress -- there are still a bunch of #488 - type failures [here](https://logs.cockpit-project.org/logs/pull-523-20220119-125645-d2c01e40-rhel-8-5/log.html#71) and [here](https://logs.cockpit-project.org/logs/pull-523-20220119-124105-d2c01e40-ubuntu-2004/log.html#67), due to the "3x affected" behaviour.

But not a *single* `testCreateCloudBase` failure. Hence I'm eager to land this PR and the accompanying fixes, and continue to debug the other race in #488 and #489.

@skobyda, can you please review? I'll retry the tests so that we don't have to land red.

skobyda (approved):
LGTM

martinpitt:
One [testCreateUrlSource](https://logs.cockpit-project.org/logs/pull-523-20220119-164344-d2c01e40-rhel-8-5/log.html#71) flake left, everything else passed. One last retry before I merge, to make sure I didn't break that.

Issue #526, opened 2022-01-19
TestMachinesCreate.testCreateThenInstall often fails on Testing Farm


martinpitt:
On CentOS-9-Stream, we are getting an awful lot of failures:
```
Traceback (most recent call last):
  File "/usr/lib64/python3.9/unittest/case.py", line 550, in _callTestMethod
    method()
  File "/var/ARTIFACTS/work-allfLW1WD/plans/all/discover/default/tests/test/check-machines-create", line 416, in testCreateThenInstall
    runner.createThenInstallTest(TestMachinesCreate.VmDialog(self, sourceType='file',
  File "/var/ARTIFACTS/work-allfLW1WD/plans/all/discover/default/tests/test/check-machines-create", line 1453, in createThenInstallTest
    self._tryCreateThenInstall(dialog, installFromVmDetails, tryWithFailInstall) \
  File "/var/ARTIFACTS/work-allfLW1WD/plans/all/discover/default/tests/test/check-machines-create", line 1347, in _assertDomainDefined
    wait(lambda: "7" in self.run_admin(dumpxmlCmd + " | grep cockpit_machines | wc -l", connection))
  File "/var/ARTIFACTS/work-allfLW1WD/plans/all/discover/default/tests/test/common/testlib.py", line 1993, in wait
    raise Error(msg or "Condition did not become true.")
testlib.Error: Condition did not become true.
```

[example 1](http://artifacts.dev.testing-farm.io/540595d1-8378-486d-a0da-cd64570db6ba/), [example 2](http://artifacts.dev.testing-farm.io/26e777f2-ea3c-472a-b839-8c0f362be187/), [example 3](http://artifacts.dev.testing-farm.io/a7618146-1917-492b-99af-024e7db4297c/). There are still no artifacts, so hard to tell what's wrong -- but this needs debugging. Most likely we'll need to disable that test from fmf for the time being.

martinpitt:
Let's discuss in #581, this has more information.

PR #527, opened 2022-01-19
fmf: Disable TestMachinesCreate.testCreateThenInstall

martinpitt:
It fails way too often (three times in a row), and is too hard to debug
without artifacts.

See issue #526


martinpitt:
[more](http://artifacts.dev.testing-farm.io/dfb874e0-83b0-4bd6-8589-6c31a4bebfc0/) [failures](http://artifacts.dev.testing-farm.io/fba15c2e-d383-49f5-8c22-ff1ac4750913/) and affects rawhide as well. I'll adjust this, and need to rebase anyway.

marusak (approved):

PR #529, opened 2022-01-20
Fix two more VM delete races in tests on rhel-8-5 and ubuntu-2004

martinpitt:
See https://github.com/cockpit-project/cockpit-machines/issues/488#issuecomment-1017168058 for the ongoing saga. There are *still* more problems, but piece by piece :grin:


skobyda (approved):
Thanks. A fix for this should be present in virt-install since 3.2.0, which newer versions of ubuntu already have, and hopefully also RHEL anytime soon, so this should be a very temporary workaround.

martinpitt:
Landing with the two `testCreateCloudBaseImage` failures on affected/repeated tests. That one is the last on my list to investigate in #488, and this PR does not actually affect that test.

PR #531, opened 2022-01-20
test: Skip ISO URL test on RHEL/CentOS 9

martinpitt:
This is not a supported feature, not a bug, and thus not quite adequate
to track as a naughty. That naughty is very brittle, on top of that.

This will replace https://github.com/cockpit-project/bots/issues/2538

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -604,11 +604,13 @@ class TestMachinesCreate(VirtualMachinesCase):
                                                       start_vm=False))
 
         # Test detection of ISO file in URL
-        runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='url',
-                                                      location=config.ISO_URL,
-                                                      memory_size=128, memory_size_unit='MiB',
-                                                      storage_pool="NoStorage",
-                                                      start_vm=True))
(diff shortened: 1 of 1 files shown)


martinpitt:
This will fix the unsightly permanent skips in [rhel-9-0](https://logs.cockpit-project.org/logs/pull-530-20220120-134056-1d0078f5-rhel-9-0/log.html) and artifacts in [c9s packit](http://artifacts.dev.testing-farm.io/99a5f7db-f066-4e95-8089-23bd1fbcc56e/)

marusak (changes requested):
it is a bug. They will fix it but it seems only in 9.1 and in the BZ they now discuss what should be done for 9.0. Also it is very prominent in our UI as it is one of the main ways how one would get a new VM. So I would not just sweep it under the rug as "not supported".
cockpit-machines might temporarily depend on the curl plugin for example.

I agree that we should not skip half of the test just because, but we should not settle with "not supported".

martinpitt:
> it is a bug. They will fix it 

Oh, ok -- the "WONTFIX" sounded otherwise. But sure!

martinpitt:
> I agree that we should not skip half of the test

Not that bad, really. The one thing after it simulates that situation of missing curl block driver for Debian; that doesn't run on RHEL anyway.

PR #533, opened 2022-01-21
Simplify handling of Undefined/Stopped events

martinpitt:
Since commit a724e2e29c9 and 514e62a707d, domainGet() is able to handle
removal events. With that there is no point in doing a "look before you
leap" approach for handling these events: Drop the ListDomains() call
and complicated prediction, and just always call domainGet() on all
Started/Stopped/Defined/Undefined lifecycle events, and let that sort it
out.

With that, we also don't need domainGet()'s `updateOnly` mode any more. If we
get e.g. a "TrayChange" event on a previously unknown domain, that obviously
exists -- so let's detect and add it.

This makes the whole lifecycle event handling more robust, more
efficient, and easier to understand.


martinpitt:
I want to clean up the commits a bit and have some further small cleanups. But let's wait for a round of tests first.

src/libvirtApi/common.js
```diff
@@ -315,14 +290,9 @@ function startEventMonitorDomains(connectionName) {
...
             case domainEvent.Defined:
-                domainGet({ connectionName, id:objPath });
-                break;
-
             case domainEvent.Undefined:
-                domainEventUndefined(connectionName, objPath);
-                break;
-
             case domainEvent.Started:
+            case domainEvent.Stopped:
```
skobyda, line 295: Because of this change, Stopped event now doesn't call `getVm` with `updateOnly = true`, which results in `updateOrAddVm` being called instead of `updateVm`. 
The difference this makes is that, as seen [here](https://github.com/cockpit-project/cockpit-machines/blob/main/src/reducers.js#L231), `updateVm` calls `clearUsageData`, which `updateOrAddVm` doesn't call. That leaves us with a bug that when VM is stopped (e.g. by clicking on "Force shutoff", it leaves metrics in "Usage" stuck:

Screenshot from 2022-01-21 12-17-30

If we want to bring this change, we should also call `clearUsageData` for `updateOrAddVm` too.

skobyda (changes requested):

src/libvirtApi/common.js
skobyda: PS: It seems usage metrics clear after few moments anyway, but that is probably caused by different getVm refresh, so it's not really a big bug ;)

martinpitt: Thanks for spotting this! I can reproduce. So this would already happen in main in some cases, when the recently fixed race condition would hit which got fixed in commit a724e2e29c9. So indeed it seems to make sense to do that in UPDATE_OR_ADD as well.

Fixed. But please double-check my logic, I'm not that well versed with the Redux stuff yet.

src/libvirtApi/common.js
```diff
@@ -291,22 +291,23 @@ function startEventMonitor({ connectionName }) {
...
+    Undefined: 1,
+    Started: 2,
+    Suspended: 3,
+    Resumed: 4,
+    Stopped: 5,
+    Shutdown: 6,
+    PMsuspended: 7,
+    Crashed: 8
+};
+
```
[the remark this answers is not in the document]
martinpitt: Done, thanks!

src/libvirtApi/helpers.js
```diff
@@ -89,6 +89,18 @@ export const Enum = {
...
+    Undefined: 1,
+    Started: 2,
+    Suspended: 3,
+    Resumed: 4,
+    Stopped: 5,
+    Shutdown: 6,
+    PMsuspended: 7,
+    Crashed: 8
+};
+
```
skobyda, line 103: I was more thinking of incorporating it into the `Enum` variable above as following:
```
VIR_DOMAIN_EVENT_DEFINED: 0,
VIR_DOMAIN_EVENT_UNDEFINED: 1,
VIR_DOMAIN_EVENT_STARTED: 2,
VIR_DOMAIN_EVENT_SUSPENDED: 3,
VIR_DOMAIN_EVENT_RESUMED: 4,
VIR_DOMAIN_EVENT_STOPPED: 5,
VIR_DOMAIN_EVENT_SHUTDOWN: 6,
VIR_DOMAIN_EVENT_PMSUSPENDED: 7,
VIR_DOMAIN_EVENT_CRASHED: 8,
VIR_DOMAIN_EVENT_LAST: 9,
```
source: https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainEventType

skobyda (changes requested):

src/libvirtApi/helpers.js
martinpitt: Ah, I see. will fix. Seemed a little weird to mash everything into the same type... the whole point of enums is to be *different* types?

skobyda (approved):
LGTM

PR #535, opened 2022-01-21
Fix incorrect warning about volume being in use during VM creation

skobyda:
This fixes a race condition which was visible only for few moments
during VM creation.
How to reproduce:
1. Have some available storage volume
2. In VM creation dialog, choose that available volume as storage
3. Clicking button "Create"

There will be few moments when VM creation dialog is still open,
but the volume is already in use (by the VM which is being created)
and dialog shows "This volume is already used by another VM."

This is confusing and can lead user to think that something went
wrong during VM creation. Therefore ideally do not show this warning
during VM creation.

Screenshot from 2022-01-21 14-57-07

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -635,7 +635,7 @@ const MemoryRow = ({ memorySize, memorySizeUnit, nodeMaxMemory, minimumMemory, o
     );
 };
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks!

skobyda:
What the heck happend with tests?

tests "Aborted without status": https://logs.cockpit-project.org/logs/pull-535-20220121-143009-70575b63-debian-stable/log

PR #538, opened 2022-01-25
Don't use MAC address when detaching network interface

skobyda:
Libvirt allows to have multiple network interfaces attached to the same VM to have the same MAC address, therefore that's not a good parameter to uniquely identify an attached vNIC. Ideally we should identify a vNIC to detach by a number of slot, bus, function and domain. Such detachment is however broken, so instead let's detach it by the index of <interface> in array of VM's XML <devices>
This serves as a workaround for https://github.com/virt-manager/virt-manager/issues/356
It also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1867478

### DIFF
--- a/src/components/vm/nics/vmNicsCard.jsx
+++ b/src/components/vm/nics/vmNicsCard.jsx
@@ -376,7 +376,7 @@ export class VmNetworkTab extends React.Component {
                         objectType: _("network interface"),
                         objectName: network.mac,
                         onClose: () => this.setState({ deleteDialogProps: undefined }),
-                        deleteHandler: () => domainDetachIface({ connectionName: vm.connectionName, mac: network.mac, vmName: vm.name, live: vm.state === 'running', persistent: nicPersistent }),
+                        deleteHandler: () => domainDetachIface({ connectionName: vm.connectionName, index: network.index, vmName: vm.name, live: vm.state === 'running', persistent: nicPersistent }),
                     };
                     const deleteNICAction = (
                         <DeleteResourceButton objectId={`${id}-iface-${networkId}`}
@@ -413,7 +413,11 @@ export class VmNetworkTab extends React.Component {
             else
                 return 0;
         };
-        const rows = vm.interfaces.sort(sortIfaces).map(target => {
+        // Normally we should identify a vNIC to detach by a number of slot, bus, function and domain.
+        // Such detachment is however broken in virt-xml, so instead let's detach it by the index of <interface> in array of VM's XML <devices>
+        // This serves as workaround for https://github.com/virt-manager/virt-manager/issues/356
+        const ifaces = vm.interfaces.map((iface, index) => ({ ...iface, index }));
+        const rows = ifaces.sort(sortIfaces).map(target => {
             const columns = detailMap.map(d => {
                 let column = null;
                 if (typeof d.value === 'string') {
--- a/src/components/vm/nics/vmNicsCard.jsx
+++ b/src/components/vm/nics/vmNicsCard.jsx
@@ -405,7 +405,15 @@ export class VmNetworkTab extends React.Component {
         detailMap = detailMap.filter(d => !d.hidden);
 
         const columnTitles = detailMap.map(target => target.name);
-        const rows = vm.interfaces.sort().map(target => {
+        const sortIfaces = (a, b) => {
+            if (a.type !== b.type)
+                return a.type > b.type ? 1 : -1;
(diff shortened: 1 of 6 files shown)


martinpitt:
Ugh, [what happened here](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-538-20220125-014225-f227477f-debian-stable/log.html), retrying.

src/components/vm/nics/vmNicsCard.jsx
```diff
@@ -389,7 +389,11 @@ export class VmNetworkTab extends React.Component {
         detailMap = detailMap.filter(d => !d.hidden);
 
         const columnTitles = detailMap.map(target => target.name);
-        const rows = vm.interfaces.sort().map(target => {
+        // Normally we should identify a vNIC to detach by a number of slot, bus, function and domain.
+        // Such detachment is however broken in virt-xml, so instead let's detach it by the index of <interface> in array of VM's XML <devices>
+        // This serves as workaround for https://github.com/virt-manager/virt-manager/issues/356
+        const ifaces = vm.interfaces.map((iface, index) => ({ ...iface, index }));
+        const rows = ifaces.sort().map(target => {
```
martinpitt, line 396: `ifaces` is a list of objects with no obvious search key, so it will be sorted by some arbitrary object hash or pointer or so. Shouldn't you give a comparison function to pick out the interface name, or sort by connectionName, then by VM name (to group related interfaces together), then by name or so?

This isn't new in this PR, but let's fix this.

martinpitt (changes requested):
Thanks! The fix looks ok, but please add a test for this, i.e. two interfaces with the same MAC, and make sure it removes the right one. E.g. the `index + 1` is not at all obvious, and such things tend to change over time.

src/components/vm/nics/vmNicsCard.jsx
skobyda: Yeah this sort() there doesn't make sense. I added a custom sort function which first tries to sort interfaces by type and then by MAC.

test/check-machines-nics
```diff
@@ -146,6 +146,7 @@ class TestMachinesNICs(VirtualMachinesCase):
         m.execute("virsh attach-interface --domain subVmTest1 --type network --source default --model virtio --mac 52:54:00:4b:73:5f --config")
         m.execute("virsh attach-interface --domain subVmTest1 --type bridge --source virbr0 --model virtio --mac 52:54:00:4b:73:6f --config")
         m.execute("virsh attach-interface --domain subVmTest1 --type network --source default --model virtio --mac 52:54:00:4b:73:4f --config")
+        m.execute("virsh attach-interface --domain subVmTest1 --type network --source default --model e1000e --mac 52:54:00:4b:73:4f --config")
```
martinpitt, line 149: Please add a comment here that this is intentionally the same MAC address. Just for our future selves :grin:

test/check-machines-nics
```diff
@@ -154,12 +155,20 @@ class TestMachinesNICs(VirtualMachinesCase):
...
         self.deleteIface(2)
         # Make sure popup dialog disappeared
         b.wait_not_present(".pf-c-modal-box")
         b.wait_text_not("#vm-subVmTest1-network-2-mac", "52:54:00:4b:73:6f")
         b.wait_text("#vm-subVmTest1-network-2-mac", "52:54:00:4b:73:4f")
+        b.wait_text("#vm-subVmTest1-network-2-model", "virtio")
+        # Test deleting the interface with same MAC address as the other vNIC will delete the correct one
+        self.deleteIface(2)
+        b.wait_text("#vm-subVmTest1-network-2-mac", "52:54:00:4b:73:4f")
+        b.wait_text("#vm-subVmTest1-network-2-model", "e1000e")
```
martinpitt, line 171: Ah, so the (previous) third interface gets shifted up to be the second.  Can you please add a `wait_not_present()` assertion that there is no third interface any more?

martinpitt (changes requested):
Thanks! This nicely resolves the sort() bit, resolving that. Tests are still failing.

src/components/vm/nics/vmNicsCard.jsx
martinpitt: Nice, thanks!

martinpitt (approved):
Thanks! The virt-xml bug got fixed, but supposedly it'll still take some time to get into all the distros. In the meantime, detaching by current index is a bit racy, but IMHO acceptable for this use case.

PR #539, opened 2022-01-25
Workaround tmt regression with artifacts syncing

thrix:
Unfortunately tmt broke this feature if one of the releases,
with https://github.com/psss/tmt/pull/1004 this is being fixed
and should bring back the old behaviour, until use the only
path that is now pulled back from the guest.

Signed-off-by: Miroslav Vadkerti <mvadkert@redhat.com>

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -8,7 +8,9 @@ if [ -d source ]; then
 else
     SOURCE="$(realpath $TESTS/../..)"
 fi
-LOGS="$(pwd)/logs"
+# workaround until tmt-1.10 is out with https://github.com/psss/tmt/pull/1004
+# LOGS="$(pwd)/logs"
+LOGS="$TMT_TREE/../execute/data/test/browser/logs"
 mkdir -p "$LOGS"
 chmod a+w "$LOGS"
 


martinpitt:
Thanks @thrix ! however, all tests fail with `./browser.sh: line 13: TMT_TREE: unbound variable`. So maybe the tmt version in TF doesn't support that variable yet, or so?

thrix:
@martinpitt oh my, lost track of the release, it is `1.10`, hmm, will try to backport ...

https://github.com/psss/tmt/pull/984/files

martinpitt:
@thrix : No worries, it's not *that* urgent. We can just wait for the official fix. Many thanks for figuring this out! :heart:

thrix:
We now have 1.10 tmt preview on testing farm, it is fairly stable, so let's try again

martinpitt:
The [c9s run](http://artifacts.dev.testing-farm.io/7a127cb7-3559-4231-b4fa-d94a9bd5b44d/) had a flake and thus an artifact. They now get exported [here](http://artifacts.dev.testing-farm.io/7a127cb7-3559-4231-b4fa-d94a9bd5b44d/work-all0ft6Bv/plans/all/execute/data/test/browser/logs/), to the advertised path. Yay! 

Thank you so much @thrix! So we could keep that workaround for a bit; not sure if it would actively break once the tmt fix is out, but it looks like it could work.

thrix:
@martinpitt cool so it works \o/

> Thank you so much @thrix! So we could keep that workaround for a bit; not sure if it would actively break once the tmt fix is out, but it looks like it could work.

No, it should not break, just also the old way should later work ...

thrix:
let me check on that F35 ...

martinpitt:
@thrix: [F35 failure](http://artifacts.dev.testing-farm.io/0c995763-e94c-47b1-ae68-cccac9a9502b/) is because of the bad selinux-policy-35.10-1.fc35.noarch. 35.11 has been out since last week which fixes that, but for some reason it crept back into the TF image?

martinpitt (approved):
Let's land this, in case the actual rollout takes longer. It's useful for our upstream PRs in the meantime, harmless, and sufficiently documented. Thanks again!

PR #542, opened 2022-01-28
test: Update to reworked pixel test machinery

mvollmer:
TEST_MOBILE is not a thing anymore, but assert_pixels switches to the
"mobile" layout temporarily.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -931,6 +931,8 @@ class TestMachinesCreate(VirtualMachinesCase):
             if self.sourceTypeSecondChoice:
                 b.select_from_dropdown("#source-type", self.sourceTypeSecondChoice)
 
+            # Data loading doesn't start immediately
+            time.sleep(1)
             b.wait_attr("#os-select-group", "data-loading", "false")
             # For the mocked fedora URL installation wait to see that the OS section is automatically filled
             if self.sourceType == "url" and self.location == TestMachinesCreate.TestCreateConfig.TREE_URL:
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1106,6 +1106,9 @@ class TestMachinesCreate(VirtualMachinesCase):
             self._fakeOsDBInfo()
             self._fakeFedoraTree()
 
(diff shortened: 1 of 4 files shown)


mvollmer:
Draft because we might catch up with more features in one go, like the "medium" layout.

mvollmer:
> Draft because we might catch up with more features in one go, like the "medium" layout.

Let's not, that will take a bit longer I guess.

test/check-machines-create
```diff
@@ -931,6 +931,8 @@ class TestMachinesCreate(VirtualMachinesCase):
             if self.sourceTypeSecondChoice:
                 b.select_from_dropdown("#source-type", self.sourceTypeSecondChoice)
 
+            # Data loading doesn't start immediately
+            time.sleep(1)
             b.wait_attr("#os-select-group", "data-loading", "false")
             # For the mocked fedora URL installation wait to see that the OS section is automatically filled
```
skobyda, line 937: Just a question, but with this sleep, are we waiting for OS data to be loaded? I thought that's what the following              `b.wait_attr("#os-select-group", "data-loading", "false")` does? Or maybe it's broken?

mvollmer: This sleep waits for the loading to _start_.  The following `wait_attr` then waits for the loading to be finished.

If you look at the current pixel test reference in main, you can see that the test catches the dialog with the OS field disabled, which means that the data is currently being loaded, no?

image

skobyda: Ah right I see. It seems it's a result of how `onOsAutodetect` sets `autodetectOSInProgress` value. Perhaps that could be somehow rewritten so there is no sleep, but that's outside of this PR.

skobyda (approved):

test/check-machines-create
skobyda: Ah right I see. It seems it's a result of how onOsAutodetect sets autodetectOSInProgress value. Perhaps that could be somehow rewritten so there is no sleep, but that's outside of this PR.

martinpitt (approved):
Many thanks! packit is busted right now, landing.

PR #544, opened 2022-02-01
Move spec file into packaging/

martinpitt:
So that it lives together with the Debian and Arch packaging.

Factorize the spec name/path in the Makefile.
    
Adopted from https://github.com/cockpit-project/cockpit-podman/commit/dbde3813c27c


marusak (approved):
One commit message went away from you `Rrop the "installed separately" comment about teh application RPM`. It is fine to land as is or even feel free to squash commits on landing and just mention both changes in one commit message.

martinpitt:
@marusak : Eww yes, let's fix that. I pushed, no code changes. Can you please re-approve?

marusak (approved):
Thanks!

PR #545, opened 2022-02-01
Use image-customize --build to build/install distribution packages

martinpitt:
This centralizes/factorizes the rpm/deb/arch package builds, and now also
builds RPM packages in the VM instead of on the host, which is cleaner.

---

Dependencies:
 - PR #544 
 - https://github.com/cockpit-project/bots/pull/2875


martinpitt:
The bots PR landed, and I rebased to fix the conflict. @marusak , @KKoukiou , @skobyda : I was quite bold here with removing stuff. However, we don't *have* to remove the `make srpm` and `make rpm` targets -- if they are useful to you for some cases, I'm happy to put them back. But they wouldn't run in CI any more, so wouldn't be tested -- we'd need to do that separately.

Please let me know what you think!

FTR, https://github.com/cockpit-project/cockpit-podman/pull/901 is the same thing for podman, but let's get a round of feedback here first, before I do the same changes too many times (starter-kit, sub-man, etc.)

marusak:
How would I then prepare rpm? I never prepared one for machines, but I did for podman and for cockpit. Can we document then how would one go about creating and rpm?

martinpitt:
@marusak: Ah, so you *do* use `make rpm`? I never do, that's why I asked. But that's enough information for me; do you also care about srpm, or just rpm?

I'll then put back the rules, simplify them, and add a little smoke test that building the RPM works.

marusak:
>  Ah, so you do use make rpm? I never do, that's why I asked. But that's enough information for me; do you also care about srpm, or just rpm?

I podman I used it once or twice. And in Cockpit I used `./tools/make-rpms` bunch of times. I had no need for srpm yet.

martinpitt:
Ack, I put back `make rpm`.

marusak (approved):
Thanks!

PR #550, opened 2022-02-02
Documentation updates

martinpitt:
This will help external contributors to find their way around more easily.


jelly (dismissed):

HACKING.md
```diff
@@ -46,3 +46,24 @@ Violations of some rules can be fixed automatically by:
     $ npm run eslint:fix
 
 Rules configuration can be found in the `.eslintrc.json` file.
+
+# Running tests locally
+
+Run `make check` to build an RPM, install it into a standard Cockpit test VM
```
marusak, line 52: Yeah, just yesterday Tomas was looking into [cockpit hacking](https://github.com/cockpit-project/cockpit/tree/main/test) and there is this section:

>To run the integration tests run the following (do NOT run the integration tests as root):
>
> `$ ./test/common/run-tests`

and he copied it and ran it. It feel very misleading that the second command on the page is the command that would run everything and you would eventually wait for hours for it to finish, even though you most likely never would do that on your machine. Same case here, we promote `make check` as the first thing, but you don't really want to do that. You want to do `make vm && make test/common` and then run one specific test.

P.S. that command in cockpit fails now as it need `--test-dir`

HACKING.md
```diff
@@ -46,3 +46,24 @@ Violations of some rules can be fixed automatically by:
...
+(fedora-35 by default), and run the `test/check-*` integration tests on
+it. This uses Cockpit's Chrome DevTools Protocol based browser tests, through a
+Python API abstraction.
+
+After the test VM is prepared, you can manually run a single test without
```
marusak, line 57: How do I `prepare test VM`?

HACKING.md
martinpitt: Good point. I changed the instructions to focus on a single test, and only mention the full `make check` with a caveat.

I'll fix cockpit's HACKING in a cockpit PR.

HACKING.md
martinpitt: This should be more clear now.

marusak (approved):
Yes, thank you, now it is much more new-user friendly :)

PR #556, opened 2022-02-04
Disk serial number

skobyda:
Normally QEMU just makes up an arbitrary serial number on the fly. However,
there are many reasons why users might want to specify a stable serial
number:
- User might have licensed functionality inside the VM tied to a specific
disk serial number (most common with Windows guests)
- Setting up multipath storage (https://sharkcz.livejournal.com/12846.html)
- This disk will be accessible from inside the guest by /dev/disk/by-id.
... and probably more

There are certain limitations about what the serial number must look like.
1. Libvirt specifies that it must match regex test: [A-Za-z0-9_\.\+\- ]+
2. Various disk types and hypervisors have limitation on length:
"IDE/SATA devices are commonly limited to 20 characters. SCSI devices
depending on hypervisor version are limited to 20, 36 or 247 characters."

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2036330

Screenshot from 2022-02-04 14-24-07
Screenshot from 2022-02-04 14-09-27
Screenshot from 2022-02-04 14-09-11
Screenshot from 2022-02-04 14-07-31


garrett (changes requested):
Neat! This is a great feature and I like where this is going.

However, we should state what's expected similar to [a password pattern](https://www.patternfly.org/v4/demos/password-strength/html-demos), using [PatternFly's helper-text](https://www.patternfly.org/v4/components/helper-text#dynamic-list).

It looks like this:

image


>    Libvirt specifies that it must match regex test: [A-Za-z0-9_.+- ]+

Say something like this:

- Letters, numbers, and limited punctuation (`-`, `_`, `+`, `.`) are all allowed

We can either filter out any characters or turn this info into an error if something wrong is typed. (I think filtering it out makes more sense.)

>    Various disk types and hypervisors have limitation on length:
>    "IDE/SATA devices are commonly limited to 20 characters. SCSI devices
>    depending on hypervisor version are limited to 20, 36 or 247 characters."

Detect the type of disk and show only that type's limitation:

- This device's serial number is limited to 20 characters
- This device's serial number is limited to 247 characters

If we allow them to type more, then it should be an error. If we limit the field to the # of characters, then show an information thing when hitting the # of characters. I think we should probably just limit the # of text typed, but can be persuaded either way.

If disks cannot be detected, then we'd want something like this showing up at the minimum maximum of characters (in this case, >= 20 characters):

⚠ Some disks are limited by length, such as 20, 36, or 247 characters

And, if course, if there's an error detected, we should prevent them from submitting the entry. (This is why I like the idea of just filtering and removing the incorrect characters, so it'll be successful. Regardless of the way we approach it, we need to communicate the limitations as stated above.)

skobyda:
I like the approach about filtering out incorrect characters and truncating overly long one while user is typing (at least that's what I understood)

> If disks cannot be detected, then we'd want something like this showing up at the minimum maximum of characters (in this case, >= 20 characters):

So I looked a bit at QEMU code (it seems they don't have the serial number limits mentioned in their docs) and QEMU does is that they truncate the serial of SCSI disks to be 36 characters long and it seems that for the vast majority of other types of disks the limit is 20.

I think if we just show this warning

> warning Some disks are limited by length, such as 20, 36, or 247 characters

then we could make the user confused since in most cases they would not know which one is correct (unless they want to look at QEMU code) and would just fall to the safer, 20 characters maximum option anyway.

So I think we should be on the safe side, be strict and enforce a limit of 20 characters for all the types of disks (with an exception for SCSI disks where the limit is 36) and not allow users to write any longer serial number. 
Also, libvirt docs recommend something similar: "Hypervisors may also start rejecting overly long serials instead of truncating them in the future so it's advised to avoid the implicit truncation by testing the desired serial length range with the desired device and hypervisor combination."

So what about:

- If the user inputs an incorrect character, such character is immediately filtered out of the text input field, and error is shown:

Screenshot from 2022-02-08 14-33-42

- If the user inputs over the limit serial, it's truncated and error is shown:
Screenshot from 2022-02-08 14-33-27
Screenshot from 2022-02-08 14-33-08

garrett:
> then we could make the user confused since in most cases they would not know which one to use (unless they want to look at QEMU code) and would just fall to the safer, 20 characters maximum option anyway.
>
> So I think we should be on the safe side, be strict and enforce a limit of 20 characters for all the types of disks (with an exception for SCSI disks where the limit is 36) and not allow users to write any longer serial number.

I agree that it's best if we don't show the confusing messages... but I was saying that can't we somehow detect if longer labels are supported?

If we cannot, then that's sad. We could revise the _warning_ to show up when more than 20 characters are used (and omit the other characters), so they know that there might be an error if they submit it.

skobyda:
> > then we could make the user confused since in most cases they would not know which one to use (unless they want to look at QEMU code) and would just fall to the safer, 20 characters maximum option anyway.
> > So I think we should be on the safe side, be strict and enforce a limit of 20 characters for all the types of disks (with an exception for SCSI disks where the limit is 36) and not allow users to write any longer serial number.
> 
> I agree that it's best if we don't show the confusing messages... but I was saying that can't we somehow detect if longer labels are supported?

No we cannot detect it automatically. It's not even specified in qemu docs (only libvirt mentions that varios hypervisors may have varios lengths for varios disk types). The best I could do is to create a mapping in our code of allowed lengths for each disk type, but since we cannot guarantee that QEMU won't change the length for any disk type in the future, I don't think that's good idea

> If we cannot, then that's sad. We could revise the _warning_ to show up when more than 20 characters are used (and omit the other characters), so they know that there might be an error if they submit it.

Yeah I think just showing up a warning everytime when user tries to input more than 20 characters is good and keeps us on the safe side in case QEMU changes allowed limit for some disk type.

garrett:
Are you sure it's qemu related and it's 20 characters?

https://wiki.archlinux.org/title/persistent_block_device_naming#by-label

In the notes subsection of the "by-label" section:

> Labels can be up to 16 characters long.

I think this is oldschool MBR related. (I remember the 16 characters being a thing from DOS days, which is where MBR came from. It's 1980s tech.)

---

https://wiki.archlinux.org/title/persistent_block_device_naming#by-partlabel

Later in the page, there's the "by-partlabel" section. It's specifically for the new-style GPT disk partitioning style, where "new" is relative and means "from the 1990s". It was included in the EFI spec.

The notes mention:

> According to the specification, GPT partition labels can be up to 72 characters long.

---

And then there's Advanced Format (AF) from 2010: 

- https://wiki.archlinux.org/title/Advanced_Format 
- https://en.wikipedia.org/wiki/Advanced_Format

… but most of the time, these are used in GPT emulation mode, so it's really tech under the partition format. Even when it's handled natively, the difference is still handled transparently. (In other words, it's basically compatible and doesn't do anything different enough for us to care in the UI, and probably doesn't do anything different at the emulation layer of Qemu either.)

---

I _think_ we really only need to be concerned with MBR (16 characters) and GPT (72 characters), and that QEMU falls out from this.

---

So, really, what I'm asking is: We can tell the difference between MBR and GPT and show the different length warnings based on that, right?

---

However, I'm testing this out locally on a USB disk.

Here's a USB stick formatted with MBR. When I add a new partition in GNOME Disks, it limits the volume name to 11 characters when it's FAT-formatted:

Screenshot from 2022-02-09 10-06-18

But if I select NTFS, it lets me use many more characters, even though this is on an MBR partition scheme:

Screenshot from 2022-02-09 10-07-33

But it might not have liked that. :grin: 

Screenshot from 2022-02-09 10-08-18

(Oh, I think this is my broken USB stick. Still, what GNOME Disks thinks is correct is interesting. There are probably reasons.)

---

Basically, I'm saying that there are certain limits and we should be able to figure out why. It's probably tied to the partitioning scheme (MBR versus GPT) and quite possibly the filesystem too.

garrett:
I wiped out the same USB stick and created a FAT, NTFS, and EXT4 partition using MBR using GNOME Disks:

Screenshot from 2022-02-09 10-11-47

Screenshot from 2022-02-09 10-12-54

NTFS lets me use longer labels, so I guess there's some metadata in the partition that overcomes the MBR label length.

Oddly, when I try to create a long label with ext4 on MBR, GNOME Disks lets me, but returns this error:

Screenshot from 2022-02-09 10-17-31

That's 16 characters.

So my guess is that NTFS does some fancy thing similar in idea to LFN ([Long File Name](https://en.wikipedia.org/wiki/Long_filename), a hack used by Windows 95 and later to map long filenames back to 8.3 length filenames on DOS+Windows using VFAT, so it's still FAT16 compatible, but with extended attributes) like mapping for the labels when on MBR and Linux supports this. Yet ext4 (and probably other Linux filesystems) do not support this.

(In other words, my "educated guess" is that NTFS has a normal MBR label for compatibility and a label in the filesystem itself, whereas at least ext4 seems to only use the system label.)

garrett:
According to ntfslabel https://linux.die.net/man/8/ntfslabel, it supports at most 128 characters.

And e2label https://linux.die.net/man/8/e2label says the length is 16 characters.

Here's the result of trying to format an ext4 partition on GPT:

Screenshot from 2022-02-09 10-30-20

Again: 16 characters for ext4, even on GPT.

I don't think the label used by Qemu is any different from the partition labeling scheme on normal disks. In fact, it looks like the partition type limits the label, not the partitioning scheme (which does also limit it in some certain circumstances, but filesystems themselves seem to either overcome the MBR limit or not use the potential of GPT, depending on the filesystem).

garrett:
Oh, this is for disk label, not partition label? Whoops! :facepalm:

skobyda:
Updating the PR based on our IRC dicussion
(context for future readers: In the end, it's QEMU that throws above-mentioned limitations out of the window and just chooses arbitrary lengths for various disk data types, e.g. scsi: https://github.com/qemu/qemu/commit/48b6206305b8d56524ac2ee347b68e6e0a528559 . So after some discussion on IRC, decision wa made to show warning in case user types more than 36 characters for scsi disks, and more then 20 characters for all the other disks)
Screenshot from 2022-02-09 12-47-36
Screenshot from 2022-02-09 12-47-27

garrett (changes requested):
Looks good functionality wise, however there is not adequate spacing between the help text and the text entry and it looks too low contrast (but perhaps it's OK if that's the default PF chose?)

It'd be awesome if we could somehow show which letters are within the limit and which are not. I'm not entirely sure how that could work though. (I have a few ideas, like a hacky pseudo element that's a no-click gradient overlay that goes # of chars, like 20, as transparent and then the rest as a screen... but this is a bad hack and we should forget it. :wink)

**Summary**: Let's fix the spacing!

garrett:
https://www.patternfly.org/v4/components/helper-text/html-demos has appropriate spacing:

image

garrett:
The implementation here is 4 pixels too high. It should be 4 pixels lower. Just a guess: This is probably due to a differing implementation from what is in the demo?

Unfortunately, the demo is just in HTML:

```html
    <div class="pf-c-form__group-control">
      <input
        class="pf-c-form-control pf-m-success"
        value="This is a valid comment"
        type="text"
        id="helper-text-form-comment"
        name="helper-text-form-comment"
        aria-describedby="helper-text-form-comment-helper"
      />
      <div
        class="pf-c-form__helper-text"
        id="helper-text-form-comment-helper"
        aria-live="polite"
      >
        <div class="pf-c-helper-text">
          <div class="pf-c-helper-text__item pf-m-dynamic pf-m-success">
            <span class="pf-c-helper-text__item-icon">
              <i class="fas fa-fw fa-check-circle" aria-hidden="true"></i>
            </span>
            <span
              class="pf-c-helper-text__item-text"
            >This is dynamic helper text with an icon showing success.</span>
          </div>
        </div>
      </div>
    </div>
```

But I looked at the layout for `pf-c-form__helper-text`:

Screenshot from 2022-02-10 17-37-04

image

The top margin is set to `--pf-c-form__helper-text--MarginTop`, which eventually resolves to `--pf-global--spacer--xs`, which happens to be 4px.

So our version here in the PR is lacking the top margin of an xs spacer for whatever odd reason.

(BTW: The text color is consistent with upstream PF.)

skobyda:
> It'd be awesome if we could somehow show which letters are within the limit and which are not. I'm not entirely sure how that could work though. (I have a few ideas, like a hacky pseudo element that's a no-click gradient overlay that goes # of chars, like 20, as transparent and then the rest as a screen... but this is a bad hack and we should forget it. :wink)

That would also not work properly if user inputs letters into middle/beginning of the input field? Maybe we could mark the incorrect letter in input field in red? But that would also require some hacky tweaking.

garrett:
> That would also not work properly if user inputs letters into middle/beginning of the input field? Maybe we could mark the incorrect letter in input field in red? But that would also require some hacky tweaking.

Yeah, there's not really a good, straightforward solution for showing which characters are problematic. I have some ideas, and it looks like you do too... but there's nothing really great.

We could show what the label would be truncated to in the text below, I guess, if we want to communicate that. It's not ideal, but it probably would work?

skobyda:
Updated the spacing with https://github.com/cockpit-project/cockpit-machines/pull/556/commits/c482af18cf2c30496d6b00ae2a9024cd0bc01fd8
Screenshot from 2022-02-16 12-00-27

skobyda:
> We could show what the label would be truncated to in the text below, I guess, if we want to communicate that. It's not ideal, but it probably would work?

Yeah, that can be done.
Perhaps something like this?
Screenshot from 2022-02-16 12-19-24

garrett:
> "Truncated" Is not the correct word maybe, since we are removing characters also in the middle of the word. Perhaps "Cleaned" is better word?

Cleaned isn't right. It isn't dirty.

A better word in this case would be "shortened". But I thought we would forbid adding those characters in the first place?

The Machines page does this already, by converting spaces to underscores in the create VM dialog. (However, underscores break when doing an automated install. They should be dashes. But that's a bug and not exactly what I'm pointing at here.)

The place where this happens is https://github.com/cockpit-project/cockpit-machines/blob/a9cd38bfae3122ae6cfc75b4afbe30c809dd41b0/src/components/create-vm-dialog/createVmDialog.jsx#L802-L806 ...as the `onValueChanged` function is passed into the onChange attribute for the machine's name https://github.com/cockpit-project/cockpit-machines/blob/a9cd38bfae3122ae6cfc75b4afbe30c809dd41b0/src/components/create-vm-dialog/createVmDialog.jsx#L206-L211 to filter the input when typed. In this case, it does a split+join to change spaces to underscores, but to filter out, one could construct a regular expression for either passing through just specific characters (which is dangerous for language support, unless generic regular expressions for words are used, like `\w` instead of something too restrictive like `[A-Za-z]`).

So you might want something like

```js
.replace(/ /, '_').replace(/[^\w-_,+.]/,'')
```

This should swap out spaces to underscores and drop all characters not related to words, numbers, or the limited set of special characters.

> Screenshot from 2022-02-16 12-19-24

What I was suggesting was actually supposed to be different lines, like [how I suggested it above](https://github.com/cockpit-project/cockpit-machines/pull/556#pullrequestreview-874654786) and how PatternFly shows it in their demos (also linked above in the same comment).

Plus neither of these are really warnings. At best, they're information blurbs instead.

The > 20 (etc.) is a warning, but these aren't.

Also, labeling as "Serial" is a bit odd. We'd need "Serial number" (but this is more than a number) or something like "Unique identifier".

skobyda:
@garrett 

- So reverting to filtering out unfit characters (I think we can safely use libvirt provided regex to see which serial passes and which not.
- Message saying what characters are allowed is present in "indeterminate" variant, unless user inputs a character which doesn't pass the regex
- Message about possible silent truncation is warning
- Renaming serial to "Disk identifier" as that's also common and I guess more precise name for it

Screenshot from 2022-03-01 02-28-58

garrett (changes requested):
This is looking great in your explanation and screenshot. I'll build here to check it out myself too.

As this is a space limited field and used in technical places, I wonder if we should make it use a monospace font? (This is optional. I'm just "thinking out loud" here. Perhaps we shouldn't.)

After digging through the code, we _know_ that it will be truncated to 20 (except for SCSI); saying "may" is misleading, right? We might want to enforce the limit with an actual truncation in the UI, remove the "may be" part, and show the warning when it's hit.

WDYT?

skobyda:
> This is looking great in your explanation and screenshot. I'll build here to check it out myself too.
> 
> As this is a space limited field and used in technical places, I wonder if we should make it use a monospace font? (This is optional. I'm just "thinking out loud" here. Perhaps we shouldn't.)

Sounds good. Monospace font makes easier to read each letter instead of a whole word, which I think is helpful for disk identifiers where a typo in 1 letter might be problematic.

> After digging through the code, we _know_ that it will be truncated to 20 (except for SCSI); saying "may" is misleading, right? We might want to enforce the limit with an actual truncation in the UI, remove the "may be" part, and show the warning when it's hit.

I don't think we can know it for sure. I just tried playing around, attaching various disks to the VM, and it acts weird.
As expected, virtio and sata disks serials are truncated to 20 characters.

However, no usb disk serial was truncated to any number of characters. (However as libvirt mentions, hypervisors such as qemu might introduce it in the future and it's safe to not rely on it)

And only some scsi disk serials were truncated to 36 characters. I'm looking at the XML definition of the 2 scsi disks (one's serial was truncated, other's not) and both disks are basically equal, they just have different target identifier and different serial.

I'll contact libvirt/qemu people about this behavior, and if there will be a serial length limit for USB disks. But I think trying to have some futher more complicated code in cockpit which predicts which scsi/usb disks will even be truncated, and if so, to what length, is just a doubling of code from qemu and may be a suspect to changes in few years anyway.

So in summary, I think it's better to give a warning, but don't enforce truncation.

garrett:
Machines - add disk (3)


- Automatically strip letters that aren't allowed.
- Automatically convert spaces to underscores.
- Show the allowed characters when it applies and keep it around after as a reminder. (Don't show by default unless it's triggered.)
- Only show the truncation message when over the limit. Have it disappear when it's equal to or lower than the limit.
- Show the truncated form at the end of the message, so people have a sense of what it might be and what to shorten it to.

garrett:
Truncation length should be 20 except where we know it's not... so: 36 for SCSI, 20 otherwise.

skobyda:
I think I updated everything to be up with design review @garrett . Also fixed the tests locally, so @martinpitt I think it's time for some code review if you have time

Screenshot from 2022-03-02 14-23-18
Screenshot from 2022-03-02 14-23-04
Screenshot from 2022-03-02 14-22-33
Screenshot from 2022-03-02 14-22-19

skobyda:
Few pixel tests broken, updated

skobyda:
Ups forgot to add monospace for serial. Here is update:
Screenshot from 2022-03-02 14-58-55

garrett:
I've noticed that sometimes when I type in the entry, it seems to randomly selects all the text and replaces it with the rest of what I'm typing.

garrett (changes requested):
Overall, it looks great!

However: I've noticed that sometimes when I type in the entry, it seems to randomly selects all the text and replaces it with the rest of what I'm typing.

Can you fix this please?

garrett:
I looked at the code and it seems consistent with the rest of Machines. If I understand it correctly, it's changing the value every time the value is changed? Shouldn't it only rewrite the value if the value doesn't match with the filter? And only when the user changes it (so it doesn't try to change twice)?

Or is this something that React is supposed to handle and my fast typing is hitting a corner case?

test/check-machines-disks
```diff
@@ -416,6 +422,23 @@ class TestMachinesDisks(VirtualMachinesCase):
                 b.wait_visible("div.pf-c-modal-box button[aria-expanded=true]:contains(Hide additional options)")
                 b.select_from_dropdown(f"div.pf-c-modal-box #vm-{self.vm_name}-disks-adddisk-bus-type", self.bus_type)
 
+            # Configure serial number
+            if self.serial and b.is_visible("div.pf-c-modal-box button:contains(Show additional options)"):
```
martinpitt, line 426: I don't like the `b.is_visible` here -- it should not be a surprise whether or not this is visible, but predictable. Ideally we could just drop this, what happens if you do?

test/check-machines-disks
```diff
@@ -416,6 +422,23 @@ class TestMachinesDisks(VirtualMachinesCase):
...
+                b.click("div.pf-c-modal-box button:contains(Show additional options)")
+                b.wait_visible("div.pf-c-modal-box button[aria-expanded=true]:contains(Hide additional options)")
+                print(self.xwarning_object)
```
martinpitt, line 429: Debugging leftover

test/check-machines-disks
```diff
@@ -954,6 +980,42 @@ class TestMachinesDisks(VirtualMachinesCase):
...
+        self.VMAddDiskDialog(
+            self,
+            pool_name='myPoolOne',
+            volume_name='disk_with_serial_warning1',
+            mode='create-new',
+            serial='íňválíď;šéříáľ123',
+            expected_serial='vl123',
+            expected_target='vdd',
+            xwarning_object="serial-characters",
+            xwarning_message="Allowed characters: basic Latin alphabet, numbers, and limited punctuation (-, _, +, .)"
```
martinpitt, line 1003: Repeating the whole message here is a bit weird; as it uses substring matching anyway, "Allowed characters" seems enough here.

martinpitt (changes requested):
Thanks! A few nitpicks, but by and large this looks fine.

I also reviewed the pixel diffs, and functionality-wise these make sense. However, there is now [less space before the "Show additional options](https://github.com/cockpit-project/pixel-test-reference/compare/cb69f0ff3781ed9a3c0b9a7b63f95f31ea55b299..b332230e9bc3f7ff21a86f76cc04bbe9842b8a8f#diff-e3045165df48adf224702aa4eb7c77224d5ed1ce6e25f2f80f812666a2f5dc80), and IMHO this looks a bit crammed now. I guess it's related to splitting this into two forms. Was that intended? @garrett

martinpitt:
@skobyda Please rebase on your next push, so that the packit tests can run. I also fixed c9s today (rawhide still todo)

garrett (changes requested):
Oh, that expander thingy wasn't intended. There should be more space there, like before. Please fix.

When I did the mockup, I based it on the screenshot and assumed that it's an existing layout bug and wanted to call it out for a fixup. I didn't realize this PR actually causes the issue.

Thanks for spotting! Yay for having reference screenshots in the test pipeline!

test/check-machines-disks
skobyda: Right, looking at the code, it should be always present

skobyda:
I fixed @martinpitt comments
I also fixed the space between expander and the rest of the form. I used PF's `padding-top: var(--pf-global--gutter--md);` variable, which is normally used for spacing between form items.
TestMachinesDisks-testAddDiskAdditionalOptions-vm-add-disk-modal-additional-options-pixels

garrett:
>  I also fixed the space between expander and the rest of the form. I used PF's --pf-global--gutter--md, which is normally used for spacing between form items to add top padding to expander padding-top: var(--pf-global--gutter--md);

Why not use utilities?

https://www.patternfly.org/v4/utilities/spacing

skobyda:
> Why not use utilities?
> 
> https://www.patternfly.org/v4/utilities/spacing

Good Idea, didn't think of that. Updated it to use utilities
Screenshot from 2022-03-09 01-54-55

src/components/vm/disks/diskAdd.jsx
```diff
@@ -149,37 +151,39 @@ class AdditionalOptions extends React.Component {
...
+                    <Grid hasGutter md={6}>
+                        <FormGroup fieldId='cache-mode' label={_("Cache")}>
+                            <FormSelect id='cache-mode'
+                                onChange={value => this.props.onValueChanged('cacheMode', value)}
+                                value={this.props.cacheMode}>
+                                {diskCacheModes.map(cacheMode => {
+                                    return (
+                                        <FormSelectOption value={cacheMode} key={cacheMode}
+                                                          label={cacheMode} />
+                                    );
```
martinpitt: Please avoid introducing imperative code blocks when you just return a value. A clean `k => v` mapping is functional and nicer than `k => { return v }`. I know the code was just moved, but let's clean it up while we are touching it.

src/components/vm/disks/diskAdd.scss
```diff
@@ -0,0 +1 @@
+@use "../../../../node_modules/@patternfly/patternfly/utilities/Spacing/spacing.css";
```
martinpitt, line 1: Where is that used? Is that the `pf-u-pt-md` class?

martinpitt (changes requested):
Thanks! Some very minor nitpicks, but this by and large needs pixel refs updated (some dialogs changed size, presumably due to the form rearrangements?) and there is a `TypeError: Cannot read properties of undefined (reading 'volumes')` crash [here](https://logs.cockpit-project.org/logs/pull-556-20220309-005709-ea4d051c-debian-stable/log.html#68).

src/components/vm/disks/diskAdd.scss
martinpitt: No need to, I just wanted to confirm that *this* commit actually needs it. Thanks!

src/components/vm/disks/diskAdd.jsx
martinpitt: If not, you can also do `=> ( ...` and then enclose a multi-line value. However, for a single JSX element you can just do `k => <Element ...>` and add multiple properties on the next line (with the usual alignment).

martinpitt: Sure, that's fine.

skobyda:
> in the entry, it seems to randomly selects all the text and replaces it with the rest of what I'm typing.

I think I missed this comment. Can you explain in more detail, I don't understand what do you mean.
When I try to type fast, the text is inputted into the field correctly, and all unfit characters are not present.
There is also a [test](https://github.com/cockpit-project/cockpit-machines/pull/556/files#diff-be5d654a51be086b508c4708e9bdbbec977952d654d7ab1f29fc35535d3bfdf8R1002) where unfit characters are spread all over the string, so if it would lead to mixing up the string, it would catch it.

skobyda:
Only one failure for fedora-rawhide packit test. The test failures [here](https://artifacts.dev.testing-farm.io/8b17a9ec-e66c-4c88-8979-9d9360df9c9b/) are exactly [same](https://artifacts.dev.testing-farm.io/ed061a0e-9a4b-41ce-959d-fde850d75313/) as in other open [PR](https://github.com/cockpit-project/cockpit-machines/pull/615) (testAddDiskPool, testCreateUrlSource), so they are unrelated to this PR.
So tests are relatively green, I fixed all the comments, so I think this is ready for another round of review.

martinpitt (dismissed):
LGTM, thanks! But waiting for @garrett 's review as he seemed to have some trouble with "randomly selects all the text". Not sure if that's specific to this PR or some generic PF issue.

garrett:
>> in the entry, it seems to randomly selects all the text and replaces it with the rest of what I'm typing.
>
> I think I missed this comment. Can you explain in more detail, I don't understand what do you mean.

As I said, I type, it selects the text while I'm typing sometimes, and then it replaces the text with the rest of the text I am typing. That's it. That's the issue.

It's super frustrating to type something and have half of it disappear when typing.

> But waiting for @garrett 's review as he seemed to have some trouble with "randomly selects all the text". Not sure if that's specific to this PR or some generic PF issue.

Yes, I'm not sure either... but I haven't seen the behavior anywhere else in Cockpit in several years. Either it's a new issue or it's specific to this PR. Or, perhaps, I haven't encountered it elsewhere in the Machines page and it's specific to the way Machines rewrites the input.

Is it changing the input at every keystroke or only when things differ after the filter is applied?

garrett (dismissed):
I can't reproduce the issue even on my own computer. Perhaps it was under heavy load at the time and had issues on the page once in a while? I don't know... I can type as fast as I want (with or without spaces, even with mashing tons of keys as fast as possible) and cannot trigger it today.

If it's a corner case, it's an extreme corner case apparently.

skobyda:
@garrett did it look like this? I achieved this either by pressing "Ctrl+A" or clicking on "select all". Maybe somehow, in your environment was triggering the same action. 
ezgif-2-f0549407b6

martinpitt:
@skobyda : Sorry, this is conflict-y now, due to the changed pixel tests. As this has two approvals, can you rebase and we land this quickly?

skobyda:
Test finished. As expecteed, fedora-rawh=hide is only one failing, I think all of failures there are known fails: https://artifacts.dev.testing-farm.io/689a84e2-fee7-4768-84e4-83ddc80b3546/

I updated only pixel tests screenshots, the code is the same as during the last review.

@martinpitt  mind give another re-review?

martinpitt (approved):
Dakujem!

Issue #431, opened 2021-10-18
Allow adding authorized SSH keys when using cloud base images


currybullen:
Would it be possible to add the option to include one or more authorized SSH keys for the non-root user when creating a VM using the "Cloud base image" method? This would enable the user to SSH to the VM immediately after creation using pubkey authentication.

This could be as simple as adding another input field below the "User password" field: "Authorized SSH keys" where zero or more public keys can be supplied.

![image](https://user-images.githubusercontent.com/8695615/137801117-8aa6f897-facc-40db-89f9-e365ac4a9c57.png)

The authorized key(s) could then be passed along with the rest of the user data to `virt-install` using the `ssh_authorized_keys` field: https://cloudinit.readthedocs.io/en/latest/topics/examples.html#configure-instances-ssh-keys.

Another option I've been thinking about is if it would be possible to simply allow the user to supply their own cloud-init user data file, as it would provide the user a high degree of customization at their fingertips. The scope of such a solution would of course however be a bit larger.

chrisweeksnz:
Adding the ability for users to supply their own cloud-init user data would also solve the problem of Fedora CoreOS not getting configured using a "Cloud based image'. Currently the VM gets deployed correctly, but as Fedora CoreOS requires an ignition file to be provided during cloud-init, the deployed host is unusable when ticking "Set cloud init parameters" and filling in details.

garrett:
(@skobyda pinged me about this in the morning, so I'm replying here with my thoughts on it for now.)

I think it's a great idea to add this. However, I also think the create VM dialog is quite large already so we can't just stuff these features into our current design. But we could readjust our dialog to make it more similar to [Cockpit-Podman's new container dialog](https://cockpit-project.org/blog/cockpit-257.html#podman-new-create-container-workflow) so we could have optional tabs for things like this.

The interface for Podman looks like these mockups:

![image](https://user-images.githubusercontent.com/10246/142425066-f6b9e47a-e871-442b-9fde-3d8d85b73183.png)

![image](https://user-images.githubusercontent.com/10246/142425133-bdab0f71-395c-448e-b77c-e87550c4f4e5.png)

And actual screenshots (from the blog post above):

![image](https://user-images.githubusercontent.com/10246/142425199-30f4c0b9-cac5-42c9-955c-bf1e6f6eb007.png)

![image](https://user-images.githubusercontent.com/10246/142425219-5dbf9ed7-3111-4623-a7f9-2ca2abd4374f.png)

[...]

PR #560, opened 2022-02-07
Redesign VM Creation dialog

skobyda:
The VM creation dialog has started to become overfilled. At this point, it probably even blocks adding new functionalities to the dialog, e.g. Downloading RHEL images or https://github.com/cockpit-project/cockpit-machines/issues/431 .
As a solution, let's divide the content of dialog into tabs, as hinted in https://github.com/cockpit-project/cockpit-machines/issues/431#issuecomment-972890349 .

## Redesign virtual machine creation

The content of the Virtual machine creation dialog is split between a mandatory Details configuration and an optional Automation configuration.
Clicking on "Create and run" will create, immediately start, and install the VM. Clicking on "Create and edit" will create the VM in a shut-off state and redirect the user to the VM's page where the further configuration of the VM may be done before the installation is started.

details_tab
automation_tab

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -229,6 +229,17 @@ class AppActive extends React.Component {
                                      }
                                      icon={ExclamationCircleIcon} />
                 );
+            } else if (vm.createInProgress) {
+                return (
+                    <EmptyStatePanel title={ cockpit.format(_("Creating VM $0"), cockpit.location.options.name) }
+                                     action={
+                                         <Button variant="link"
+                                                 onClick={() => cockpit.location.go(["vms"])}>
+                                             {_("Go to VMs list")}
+                                         </Button>
+                                     }
+                                     loading />
+                );
             }
 
             const connectionName = vm.connectionName;
--- a/src/components/common/machinesConnectionSelector.jsx
+++ b/src/components/common/machinesConnectionSelector.jsx
@@ -40,7 +40,7 @@ export const MachinesConnectionSelector = ({ onValueChanged, loggedUser, connect
                    onChange={() => onValueChanged('connectionName', LIBVIRT_SESSION_CONNECTION)}
                    name="connectionName"
                    id="connectionName-session"
-                   label={_("Session")} />
+                   label={_("User session")} />
         </FormGroup>
     );
 };
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -183,7 +183,7 @@ function validateParams(vmParams) {
 
     if (vmParams.nodeMaxMemory && vmParams.memorySize > convertToUnit(vmParams.nodeMaxMemory, units.KiB, vmParams.memorySizeUnit)) {
         validationFailed.memory = cockpit.format(
-            _("Up to $0 $1 available on the host"),
+            _("$0 $1 available on host"),
             toReadableNumber(convertToUnit(vmParams.nodeMaxMemory, units.KiB, vmParams.memorySizeUnit)),
             vmParams.memorySizeUnit
         );
@@ -596,7 +596,7 @@ const MemoryRow = ({ memorySize, memorySizeUnit, nodeMaxMemory, minimumMemory, o
     let helperText = (
         nodeMaxMemory
             ? cockpit.format(
-                _("Up to $0 $1 available on the host"),
+                _("$0 $1 available on host"),
                 toReadableNumber(convertToUnit(nodeMaxMemory, units.KiB, memorySizeUnit)),
                 memorySizeUnit
             ) : ""
@@ -644,7 +644,7 @@ const StorageRow = ({ connectionName, allowNoDisk, storageSize, storageSizeUnit,
     let helperTextNewVolume = (
         poolSpaceAvailable
             ? cockpit.format(
-                _("Up to $0 $1 available on the default location"),
+                _("$0 $1 available at default location"),
                 toReadableNumber(convertToUnit(poolSpaceAvailable, units.B, storageSizeUnit)),
                 storageSizeUnit
             )
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -21,14 +21,14 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { debounce } from 'throttle-debounce';
 import {
-    Checkbox,
     Form, FormGroup,
     FormSelect, FormSelectOption, FormSelectOptionGroup,
     InputGroup,
     Modal,
     Select as PFSelect, SelectOption, SelectVariant,
+    Tabs, Tab, TabTitleText,
     TextInput,
-    Button, Tooltip, TooltipPosition
+    Button, Tooltip
 } from '@patternfly/react-core';
 
 import cockpit from 'cockpit';
@@ -188,10 +188,11 @@ function validateParams(vmParams) {
             vmParams.memorySizeUnit
         );
     }
-    if (vmParams.unattendedInstallation && !vmParams.userPassword && vmParams.userLogin) {
+
+    if (!vmParams.userPassword && vmParams.userLogin) {
         validationFailed.userPassword = _("User password must not be empty when user login is set");
     }
-    if (vmParams.unattendedInstallation && vmParams.userPassword && !vmParams.userLogin) {
+    if (vmParams.userPassword && !vmParams.userLogin) {
         validationFailed.userLogin = _("User login must not be empty when user password is set");
     }
 
@@ -412,63 +413,42 @@ const UnattendedRow = ({
     onValueChanged,
     os, profile,
     rootPassword,
-    unattendedDisabled,
-    unattendedInstallation,
     unattendedUserLogin,
     userLogin, userPassword,
     validationFailed,
 }) => {
-    let unattendedInstallationCheckbox = (
-        <Checkbox id="unattended-installation"
-                  isChecked={unattendedInstallation}
-                  isDisabled={unattendedDisabled}
-                  onChange={checked => onValueChanged('unattendedInstallation', checked)} />
-    );
-    if (unattendedDisabled) {
-        unattendedInstallationCheckbox = (
-            <Tooltip id='os-unattended-installation-tooltip' content={_("The selected operating system does not support unattended installation")} position={TooltipPosition.left}>
-                {unattendedInstallationCheckbox}
-            </Tooltip>
-        );
-    }
-
     return (
-        <FormGroup fieldId="unattended-installation"
-                   label={_("Run unattended installation")}
-                   hasNoPaddingTop>
-            {unattendedInstallationCheckbox}
-            {!unattendedDisabled && unattendedInstallation && <>
-                {os.profiles.length > 0 &&
-                <FormGroup fieldId="profile-select"
-                           label={_("Profile")}>
-                    <FormSelect id="profile-select"
-                                value={profile || (os.profiles && os.profiles[0])}
-                                onChange={e => onValueChanged('profile', e)}>
-                        { (os.profiles || []).sort()
-                                .reverse() // Let jeos (Server) appear always first on the list since in osinfo-db it's not consistent
-                                .map(profile => {
-                                    let profileName;
-                                    if (profile == 'jeos')
-                                        profileName = 'Server';
-                                    else if (profile == 'desktop')
-                                        profileName = 'Workstation';
-                                    else
-                                        profileName = profile;
-                                    return <FormSelectOption value={profile}
-                                                             key={profile}
-                                                             label={profileName} />;
-                                }) }
-                    </FormSelect>
-                </FormGroup>}
-                <UsersConfigurationRow rootPassword={rootPassword}
-                                       rootPasswordLabelInfo={_("Leave the password blank if you do not wish to have a root account created")}
-                                       showUserFields={unattendedUserLogin}
-                                       userLogin={userLogin}
-                                       userPassword={userPassword}
-                                       validationFailed={validationFailed}
-                                       onValueChanged={onValueChanged} />
-            </>}
-        </FormGroup>
+        <>
+            {os.profiles.length > 0 &&
+            <FormGroup fieldId="profile-select"
+                       label={_("Profile")}>
+                <FormSelect id="profile-select"
+                            value={profile || (os.profiles && os.profiles[0])}
+                            onChange={e => onValueChanged('profile', e)}>
+                    { (os.profiles || []).sort()
+                            .reverse() // Let jeos (Server) appear always first on the list since in osinfo-db it's not consistent
+                            .map(profile => {
+                                let profileName;
+                                if (profile == 'jeos')
+                                    profileName = 'Server';
+                                else if (profile == 'desktop')
+                                    profileName = 'Workstation';
+                                else
+                                    profileName = profile;
+                                return <FormSelectOption value={profile}
+                                                         key={profile}
+                                                         label={profileName} />;
+                            }) }
+                </FormSelect>
+            </FormGroup>}
+            <UsersConfigurationRow rootPassword={rootPassword}
+                                   rootPasswordLabelInfo={_("Leave the password blank if you do not wish to have a root account created")}
+                                   showUserFields={unattendedUserLogin}
+                                   userLogin={userLogin}
+                                   userPassword={userPassword}
+                                   validationFailed={validationFailed}
+                                   onValueChanged={onValueChanged} />
+        </>
     );
 };
 
@@ -572,22 +552,14 @@ const CloudInitOptionsRow = ({
     userLogin, userPassword,
     validationFailed,
 }) => {
-    const [showOptions, setShowOptions] = useState(false);
-
     return (
-        <FormGroup fieldId="cloud-init-checkbox">
-            <Checkbox id="cloud-init-checkbox"
-                      isChecked={showOptions}
-                      onChange={setShowOptions}
-                      label={_("Set cloud init parameters")} />
-            {showOptions && <UsersConfigurationRow rootPassword={rootPassword}
-                                                    rootPasswordLabelInfo={_("Leave the password blank if you do not wish to set a root password")}
-                                                    showUserFields
-                                                    userLogin={userLogin}
-                                                    userPassword={userPassword}
-                                                    validationFailed={validationFailed}
-                                                    onValueChanged={onValueChanged} />}
-        </FormGroup>
+        <UsersConfigurationRow rootPassword={rootPassword}
+                               rootPasswordLabelInfo={_("Leave the password blank if you do not wish to set a root password")}
+                               showUserFields
+                               userLogin={userLogin}
+                               userPassword={userPassword}
+                               validationFailed={validationFailed}
+                               onValueChanged={onValueChanged} />
     );
 };
 
@@ -749,11 +721,11 @@ class CreateVmModal extends React.Component {
         super(props);
         this.state = {
             createMode: NONE,
+            activeTabKey: 0,
             validate: false,
             vmName: '',
             connectionName: LIBVIRT_SYSTEM_CONNECTION,
             sourceType: defaultSourceType,
-            unattendedInstallation: false,
             source: '',
             os: undefined,
             memorySize: props.nodeMaxMemory ? Math.min(1, convertToUnit(props.nodeMaxMemory, units.KiB, units.GiB)) : 1,
@@ -801,6 +773,14 @@ class CreateVmModal extends React.Component {
             });
         });
     }
+
+    handleTabClick = (event, tabIndex) => {
+        // Prevent the form from being submitted.
+        event.preventDefault();
+        this.setState({
+            activeTabKey: tabIndex,
+        });
+    };
 
     onValueChanged(key, value) {
         switch (key) {
@@ -894,9 +874,6 @@ class CreateVmModal extends React.Component {
             this.setState(stateDelta);
             break;
         }
-        case 'unattendedInstallation':
-            this.setState({ unattendedInstallation: value });
-            break;
         default:
             this.setState({ [key]: value });
             break;
@@ -915,6 +892,7 @@ class CreateVmModal extends React.Component {
             // then show errors in the notification area
             this.setState({ createMode: startVm ? RUN : EDIT, validate: false });
 
+            const unattendedInstallation = this.state.rootPassword || this.state.userLogin || this.state.userPassword;
             const vmParams = {
                 connectionName: this.state.connectionName,
                 vmName: this.state.vmName,
@@ -926,7 +904,7 @@ class CreateVmModal extends React.Component {
                 storageSize: convertToUnit(this.state.storageSize, this.state.storageSizeUnit, units.GiB),
                 storagePool: this.state.storagePool,
                 storageVolume: this.state.storageVolume,
-                unattended: this.state.unattendedInstallation,
+                unattended: unattendedInstallation,
                 userPassword: this.state.userPassword,
                 rootPassword: this.state.rootPassword,
                 userLogin: this.state.userLogin,
@@ -970,6 +948,10 @@ class CreateVmModal extends React.Component {
         const { nodeMaxMemory, nodeDevices, networks, osInfoList, loggedUser, storagePools, vms } = this.props;
         const validationFailed = this.state.validate && validateParams({ ...this.state, osInfoList, nodeMaxMemory, vms: vms.filter(vm => vm.connectionName == this.state.connectionName) });
 
+        const unattendedInstructionsMessage = _("Enter root and/or user information to enable unattended installation.");
+        const unattendedUnavailableMessage = _("Automated installs are only available when downloading an image or using cloud-init.");
+        const unattendedOsUnsupportedMessageFormat = os => cockpit.format(_("$0 does not support unattended installation."), os);
+
         let unattendedDisabled = true;
         if ((this.state.sourceType == URL_SOURCE || this.state.sourceType == LOCAL_INSTALL_MEDIA_SOURCE) && this.state.os) {
             if (this.state.os.medias && this.state.sourceMediaID in this.state.os.medias)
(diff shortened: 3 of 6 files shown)


garrett (changes requested):
I don't approve.

When redesigning the create container dialog for podman, the point was to have as many default options as possible up-front and put the secondary optional settings in a separate tab. People can set up most common containers without having to switch to the integration tab.

This PR needlessly splits up options into sepratate tabs so people do actually have to jump into additional tabs.

This PR can provide the base for splitting up features, for sure, but not like this. We'll have to look at the features we currently provide and the ones we want to provide and try to move everything to the frontmost tab. Then, we should only have one (or possibly two) additional tab(s) with more optional settings.

And we still shouldn't put every setting into the create dialog.

Additionally, we might want to consider changing run setting into two actions, so the create dialog could be something like:

[Create and run] [Create only] Cancel

Create and run would be primary, create only would be secondary, and cancel should be a cancel link-styled-button. Splitting it like this would probably be more logical and would free up a little bit of space inside the dialog.

skobyda:
> This PR needlessly splits up options into sepratate tabs so people do actually have to jump into additional tabs.

Ah I see now.

> This PR can provide the base for splitting up features, for sure, but not like this. 

Definitely, that was my idea with this Draft, to get the creative discussion started here. 

> We'll have to look at the features we currently provide and the ones we want to provide and try to move everything to the frontmost tab.

I see, so ideally we would have Source, Storage, and Memory in 1 tab.
Actually, do we even want to have a memory configuration in the creation dialog? Similar to vCPUs, it's an option that can be changed easily after VM is created. And it was added to the VM creation dialog long before we had the ability to predict recommended or minimum memory based on Operating System.

> And we still shouldn't put every setting into the create dialog.

Of course. Downloading RHEL images is requested feature, so I'm afraid that one will have to be integrated into the creation dialog.

> Additionally, we might want to consider changing run setting into two actions, so the create dialog could be something like:
> 
> [Create and run] [Create only] Cancel
> 
> Create and run would be primary, create only would be secondary, and cancel should be a cancel link-styled-button. Splitting it like this would probably be more logical and would free up a little bit of space inside the dialog.

Nice Idea. I was unsure about "Immediatelly start VM" anyway.

skobyda:
@garrett What about something in a sense of having "Details" and "Optional tabs":
### With memory
Screenshot from 2022-02-07 14-39-25

### Without memory
Screenshot from 2022-02-07 14-40-07

skobyda:
Additionally, we could also move connection row into "Additional/Optional" tab. A vast majority of users use system connection (There was time when session connection VMs had significant bugs in the UI but we got no bugzilla or github issue report, which leaves me to believe that only a small fraction of users use session connection).
Therefore we could maybe move Connection Selector do "Additional" tab, away from the the up-front default view:

Screenshot from 2022-02-07 14-45-15
Screenshot from 2022-02-07 14-45-19

garrett:
Thanks for starting this work. We've talked about it happening for a few years in the team at one point or another... but haven't moved toward it until your work to do so just now, so even your initial proposal was really useful, even if we aren't merging it in exactly as it was. :wink: 

That's getting better! One thing I did in the podman create dialog was to say what they are instead of "additional". And I think "details" is sort of strange for the main configuration items.

But I would like to try to get my GitHub notification queue under control before more properly working on a design... I don't want to get too behind on that. Afterward, I'm happy to work with you further on the design of this dialog. Meanwhile, if you want to experiment further or suggest additional things, that would be great too.

I like the direction it's headed, and this is definitely something we have been needing to do for quite a while!

garrett:
Notes:

- System / session isn't really optional.
- And having a tab with one option, unattended, doesn't make much sense.
- I like the RAM and disk size in the main tab, as people will need them.

Let's consider what else we need and should have in the VM and try to think about how it all could/should fit together.

skobyda:
> I think "details" is sort of strange for the main configuration items.

maybe "General" / "General configuration"?

skobyda:
>     * System / session isn't really optional.

Yeah, but I would say that the configuration option is rarely used, as only a small fraction of users use session connection. So perhaps we could somehow take the average's user attention away from that configuration option?

>     * I like the RAM and disk size in the main tab, as people will need them.
 

Disk size is definitely needed, and you cannot change disk size later thru cockpit UI. 
But for RAM size, you can change that one after creation. 
Do we know people often change RAM configuration? I would say that RAM configuration is much more close to vCPUs configuration, which is also not in the creation dialog and there are no big requests for that.
Also since RAM can be autodetected form the OS, maybe it could pop-up only if auto-detection fails?

 * And having a tab with one option, unattended, doesn't make much sense.

Agreed

> Let's consider what else we need and should have in the VM and try to think about how it all could/should fit together.

To clear up my mind

| Configuration option | Can be amended later from c-machines? |  Is optional? |  Can be auto-detected? |
| ----------- | ----------- | ----------- |
| Name      | No       |  No   | No (only randomly autogenerated?) |
| Connection      | No       |  No   | Maybe (We can check if the user has administrative access or not) |
| Installation type (source)      | No       |   No  | No  |
| Operating system   | No        |   No  | Yes |
| Destination (storage)      | No       |  No   | No |
| Disk size   | No        |  No   | Yes    |
| Memory      | Yes       |  No   | Yes   |
| Unattended installation      | No       |  Yes   | No       |
| Start VM after creation  | No        |   Yes  | No       |

martinpitt:
> Yeah, but I would say that the configuration option is rarely used

Ugh, is it? On a server one surely mostly uses the system connection. But then again, nobody should deploy a VM using Cockpit on a production server. This should be for interactive experiments.

On a desktop/private machine, I actually think the session connection makes more sense?

garrett:
Let's take a step back: What are we trying to accomplish here? We want to add space for new features, yes? But where are the new features?

As far as I see, the only things that make sense right now are:

- Express install ("Unattended installation" for Download an OS, "Cloud init" for when a Cloud base image is selected)

...and that's it?

The system / user session toggle isn't optional. It wouldn't be shown unless you have admin access, but then when you do, it matters a lot. It influences where the VM will be stored, how it will connect to the network, what rights it has, etc.

We're talking about making create and run the default action and have a secondary action to only create a VM (without running). So that option goes away.

What's left? Just express install, which can change on context on what's selected. (It's basically always user and password regardless of tech implementing it. Download also has a profile setting, for server or workstation.... but that's the only difference.)

So what other options could we have there? I seem to remember things like vCPU, hardware redirection (like dedicating USB/PCI), Anything else? Is it fine for these to be set on the page before first run? How many settings do we want to include in the create step? (Please give examples.)

Here's a first draft:

create-vm
[🠟 create-vm.excalidraw.zip](https://github.com/cockpit-project/cockpit-machines/files/8072429/create-vm.excalidraw.zip)
https://excalidraw.com/#json=YyuFn-8uiLugi-f7Qkion,gv__cvt4RPDHSXeLWJfoYw

If we did have a second category, then we could make the "Express install" (which is the generic terminology that GNOME Boxes uses, which would cover both "Unattended installation" and "Cloud init") only show up at the end when relevant. Otherwise, it's probably better to have a tab strip with a disabled tab (with a hover tooltip like, "Express install is only available when downloading an operating system or using a cloud base image.") versus having just 1 tab. (Yes, PF does support disabled tabs, even with tooltips... https://www.patternfly.org/v4/components/tabs#default)

garrett:
Note: @Jelle and I have talked about moving the system / user session location down to the same area for podman too. This isn't being inconsistent; it's based on a design iteration of what we talked about. :wink:

garrett:
We could make the actions "Create and run" and "Create and edit", which would create the VM and immediately take you to the VM's details page for adjusting before running. That's probably even better.

I don't think there's really a use case to create but don't run and don't edit. It's create and either run or edit.

skobyda:
With the previous IRC discussion and this mockup, I think it's getting much better. Thanks for your work @garrett 
Updated according to mockup and a previous IRC discussion:

- What does it mean that a VM name will be suggested? Does it mean it will be autogenerated? In that case, I put together a function which will autogenerate VM name in format of `[Location]-[Operating_system]-[Datetime]`
- In order to be consistent, I think we should rename "Connection" to "Location" everywhere in c-machines. So I've done that.
- So far the second tab is called "Express Install", but in IRC we had a bit of a discussion about that, since in some cases, like could base image, technically there is no installation. But so is not during importing of an image, neither during network boot PXE. Other altenatives metioned were "Second stage set-up", "Automation"...
- Update the buttons to "Create and run" and "Create and edit" 

Screenshot from 2022-03-01 14-44-16

garrett:
Firstly: 

- The tabs probably shouldn't be stretched across. We might get more tabs, eventually.

> What does it mean that a VM name will be suggested? Does it mean it will be autogenerated? In that case, I put together an implementation that will autogenerate VM name in format of [Location]-[Operating_system]-[Datetime] when user clicks "Create ..." button and immediately use that autogenerated name to create a.

Yeah, I think that's about right. 

The downside is that people really should name their VMs based on app purpose and/or app names most of the time (apache, sql, mariadb) and we can't really guess the purpose of the VM.

It might make sense to change the placeholder to the suggested name once we have enough information to generate the name.

> In order to be consistent, I think we should rename "Connection" to "Location" everywhere in c-machines. So I've done that.

Yeah, we should be consistent. However, I seem to remember something about @KKoukiou possibly having reasons on why it should be called connection instead of location? (It's been a while. I might be misremembering too.)

Regardless, it should be consistent, no matter name we use.

Also, system and session look very similar at a glance; I've suggested "user session" before, which explains what "session" is.

> So far the second tab is called "Express Install", but in IRC we had a bit of a discussion about that, since in some cases, like could base image, technically there is no installation. But so is not during importing of an image, neither during network boot PXE. Other ideas about a tab name were "Second stage set-up", "Automation"...

I had a suggestion that people seemed fine with, but I don't remember what it was. But it was to that effect. Perhaps it was "Automation"? I don't really remember the exact word at the moment.

> Updated the buttons to "Create and run" and "Create and edit"

Cool! We'll want to immediately visit the VM page when that's clicked. Hopefully it'd be created ASAP so we could switch to it.

skobyda:
So I hope I managed to attend all the comments:

- Renamed it to Location. The only problem I could see with it is that user could confuse it with "storage". Although technically choosing system/user session also affects where the VM will be stored (where its XML configuration and disk will be stored). And previous "connection" was maybe even more ambigious.
- Renamed session to 'user session'. I'm fine with this change and I agree that this is more precise
- Renamed the tab to Automation. Don't really have a strong opinion how this tab should be called, automation sounds right and described the functionalities there
- and lastly, I changed the auto-generation of VM name

**** This is the helper text if user doesn't prefill any VM name
Screenshot from 2022-03-08 14-58-41

**** Once they select Operating system, we have enough information to generate the VM name, so we will inform what the auto-generated name will look like:
Screenshot from 2022-03-08 14-58-58

WDYT @garrett

garrett (changes requested):
It looks great! :+1: 

Although the tabs shouldn't expand like that. Please fix the tab stretching.

I don't have time this evening to build it and check locally, but I do like what I see and I'm looking forward to trying it out tomorrow! Hopefully we can get this merged in soon (and backport some of these adjustments to the UI to Cockpit-Podman too).

(I did have the toggle as a PF Toggle Group (https://www.patternfly.org/v4/components/toggle-group) in the mockup — but how it is here, a radio button group (as you currently have it) might be clearer. We'll want to change that for Podman as well. The two should be consistent.)

skobyda:
Ahh, I knew I missed something. Updated it it doesn't use stretched tabs anymore.

And about the PF toggle group, I didn't add it because I didn't notice it in mockup. I'm not really against it, but to be honest I don't really see what improvement such change would bring compared to the radio buttons, and PF guidelines mentioned it should be used more to change context or views. But again, I'm not really against it if it could bring some improvement.

Screenshot from 2022-03-09 01-59-01

garrett:
> And about the PF toggle group, I didn't add it because I didn't notice it in mockup. I'm not really against it, but to be honest I don't really see what improvement such change would bring compared to the radio buttons, and PF guidelines mentioned it should be used more to change context or views. But again, I'm not really against it if it could bring some improvement.

It's changing context of where it goes, but yeah; I think the radios are clearer... especially in the form area.

I've been talking with @jelly about it too; I think we'll just go with radio buttons in both Machines and Podman UIs. He's working on updating the "Create container" modal dialog with changes that we've been working on here for "Create VM", so hopefully we'll have both updated and consistent in the same release. :+1:

garrett:
Ah, automation is problematic in this PR right now.

Screenshot from 2022-03-09 11-51-38

1. Ideally, switching wouldn't affect the modal height.
    - But if that's not an easy thing, then whatever. 
    - _This is totally optional_... 
    - and probably not even necessary if we show something else here and/or set a min height (which might not be the same height even, but would be less short than what it currently is).
2. The Automation tab should probably either be disabled, hidden, or show a message in the tab itself about not being supported.
   -  However, if it's hidden, then showing just 1 tab is weird...
   -  and growing a tab would cause the tab bar to be displayed, which would shift everything down, which is also odd...
   -  and being disabled is visually jarring...
   - So I guess we should have the tab exist but display an unsupported message (similar to an empty state pattern)
3. The message should say what _does_ support automation.
   - (There are specific types of installation methods that do.)
4. This is using an improper label / check combo.
   - There shouldn't be a label and a check; it should be a check and a label.
5. Should we not have an opt-in for automation, but not automate things when it's not filled out and do when it is?
   - That is, treat the inputs as optional fields.
   - And probably have a message at the top about this?
   - And it'd get rid of the need for a check (the above major point, # 4)

I'll make a mockup. But it's lunchtime, so I'll work on this after lunch.

garrett:
image

> If left blank, a name will be set to 'system-almalinux8-2022/3/9'

Can the system name have slashes?

The text should probably be shortened to:

> If blank, name will be 'system-almalinux8-2022/3/9'

KKoukiou:
> Also, why does it say "Up to... on the..."? Why not just "31 GB available"? (And it should be GB, not GiB, right?)
> 
> Or at least, it it needs to say more, it could be "312 GB available at the default location" and "32 GB available on the host". And then what is the default location?

... available on the host.. refers to the available memory not storage. When we show the hint for the storage, we mention the 'available in the default location'.

> 
> Additionally, the available space message goes away when user session is created. But this is probably outside of the scope of this PR, unless this PR happens to change that. (Which I'm guessing it doesn't.)
>

garrett:
Right, but "up to" is implied and unnecessary. I was talking about both the storage and memory. They're both needlessly verbose.

garrett:
New mockups.

## General

create-vm-modal


## Unattended

create-vm-modal-unattended


## Cloud-init

create-vm-modal-cloud_init


## Unavailable


create-vm-modal-unavailable

 
## Source

https://excalidraw.com/#json=1m7ZoAKQusFjItmY-YXUR,9nX6PKnvnhSsJ2Qkn8jTRg

**[🠟 create-vm-modal.excalidraw.zip](https://github.com/cockpit-project/cockpit-machines/files/8215062/create-vm-modal.excalidraw.zip)**

skobyda:
### When user doesn't select an installation type not supporting automation:

Screenshot from 2022-03-29 13-38-30

Screenshot from 2022-03-29 13-38-38

### When user selects an OS not supporting automation or doesn't select any OS:

Screenshot from 2022-03-29 13-38-00
Screenshot from 2022-03-29 13-38-14

### When selects "Download OS" with an OS which supports Automation:
Screenshot from 2022-03-29 13-38-53
Screenshot from 2022-03-29 13-39-01

### When selects "Cloud base image"
Screenshot from 2022-03-29 13-39-10
Screenshot from 2022-03-29 13-39-16


### After user clicks on "Create and edit", they are redirected to a page of the VM. The page shows empty state panel until the VM is created:
Screenshot from 2022-03-29 13-39-36

skobyda:
I'm mostly unsure about the case when user selects an OS which doesn't support Automation or doesn't select any OS, I just show message `The selected operating system does not support unattended installation.`. In the old design, this message was present for a disabled checkbox as a Tooltip. But maybe there should be a different message "You need to select operating system which supports unattended installation" in case user doesn't select any OS?

garrett:
> But maybe there should be a different message "You need to select operating system which supports unattended installation" in case user doesn't select any OS?

Yeah, I had a mockup for that above:

the mockup

I suppose we could expand that message:

- "Automated installs are only available for certain operating systems when downloading an image or using cloud-init."

But this doesn't handle the case where general type supports automation, but the OS does not.

I guess in this case, we need to say something like:

"ABC does not support XYZ."

Such as:

- "CentOS 7 does not support cloud-init."
- "Windows does not support unattended installation."

It feels like we should list what's supported there and possibly inidicate whether it's supported or not when someone's on the other tab. We don't want to have 1 tab and the disabled state for the second tab looks pretty heavyweight. I'm not sure of a good solution at the moment. If we had more than 2 tabs, then it might be OK to make the automation one disappear when it's not supported and appear when it is. But even still, it's unknown what could cause it to show up, so that leads me back to us showing it somehow.

(Then, of course, we need to determine how much of that needs to be in this PR versus a follow-up.)

garrett:
What would happen if we set an isDisabled flag on the automation tab when there's nothing there? And then we could have a tooltip (with the same text) when someone hovers over it, I guess?

image

garrett:
I've checked out this PR in its current state and noticed that when you create a VM, the spinner shows up on both the "Create and run" and the "Create and edit" buttons. It should only show up on the button you clicked.

@jelly fixed a recent bug like that too on the crypto-policies. We can probably use the same technique here.

garrett:
@skobyda: BTW, I think the disabled tab approach is probably better and more discoverable (as you can see availability when you adjust things). However, the look of it isn't fantastic, but it is what upstream PF uses, so I guess that's OK.

jelly:
> I've checked out this PR in its current state and noticed that when you create a VM, the spinner shows up on both the "Create and run" and the "Create and edit" buttons. It should only show up on the button you clicked.
> 
> @jelly fixed a recent bug like that too on the crypto-policies. We can probably use the same technique here.

All what is required is to make the button's inProgress take into account another boolean `createclicked` which  is set when  `this.onCreateClicked` is called using setState.

skobyda:
### After clicking "Create and run"
Screenshot from 2022-04-04 11-55-58

### After clicking "Create and edit"
Screenshot from 2022-04-04 11-55-37

skobyda:
@garrett Fixed the issue with Create buttons being incorrectly disabled. Mind giving it another look?

skobyda:
Ups, didn't notice that Tabs were present in Import VM Dialog. Removed them from there.

### How Import VM dialog looked before this PR:
Screenshot from 2022-04-04 12-23-58


### How Import VM dialog will look after this PR:
Screenshot from 2022-04-04 12-23-43

skobyda:
Apart from pixel tests failure, tests so far look green. @martinpitt wanna give it a code review?

garrett (dismissed):
Looks great! Awesome work! :+1: 

Now it just needs code review and an updated description and screenshots at the top. :wink:

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -130,12 +130,33 @@ function getSpaceAvailable(storagePools, connectionName) {
     return space;
 }
 
+function getVmName(vmName, connectionName, vms, os) {
+    if (!isEmpty(vmName.trim()))
+        return vmName;
+
+    let retName = connectionName;
```
martinpitt, line 139: It's redundant to include the connection name into the domain name. These are already namespaced, and it will make them even longer.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -130,12 +130,33 @@ function getSpaceAvailable(storagePools, connectionName) {
...
+    }
+
+    return tmpRetName;
+}
+
 function validateParams(vmParams) {
     const validationFailed = {};
 
-    if (isEmpty(vmParams.vmName.trim()))
-        validationFailed.vmName = _("Name must not be empty");
```
martinpitt, line 137: getVmName() can fail and return an empty string, so it seems we still actually need this?

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -195,13 +216,19 @@ function validateParams(vmParams) {
...
 }
 
-const NameRow = ({ vmName, onValueChanged, validationFailed }) => {
+const NameRow = ({ vmName, suggestedVmName, os, onValueChanged, validationFailed }) => {
     const validationStateName = validationFailed.vmName ? 'error' : 'default';
 
+    let helperText;
+    // Don't show when user inputs any vm name or when we are not able to generate any suggestedVmName from operating system
+    if (isEmpty(vmName.trim()) && (!os || (os && !isEmpty(suggestedVmName.trim()))))
+        helperText = isEmpty(suggestedVmName.trim()) ? _("If blank, a name will be suggested") : _(`If blank, a name will be set to '${suggestedVmName}'`);
```
martinpitt, line 228: "a name will be suggested" is too surprising. This would only appear until the user selects an OS, right? and that is mandatory? We need to avoid a situation where you click "create" with "a name will be suggested".

Still, I feel this is too magic/jumpy, as a helper text. Easier/less surprising behaviour proposal: set the name input to an autogenerated name once the user chooses an OS and the name is empty still, and then just keep the current "is empty" validation.

Also, grammar: If you keep this sentence, it's "If blank, name will be set.." (drop the "a"). But still, just do it, and avoid the indirection.

As this whole auto-naming business is quite hairy, I also recommend splitting this out into its own PR. I feel it's getting unnecessarily difficult to handle all these independent changes in a single PR.

test/check-machines-create
```diff
@@ -96,10 +96,6 @@ class TestMachinesCreate(VirtualMachinesCase):
         # Memory unit conversion
         runner.unitConversionTest(TestMachinesCreate.VmDialog(self, sourceType='disk-image'))
 
-        # try to CREATE WITH DIALOG ERROR
-        # name
-        runner.checkDialogFormValidationTest(TestMachinesCreate.VmDialog(self, "", storage_size=1), {"vm-name": "Name must not be empty"})
```
martinpitt, line 101: This should be replaced with the new expected behaviour of auto-creating a name and ensuring that VM creation actually works and has a name of the expected form.

src/app.jsx
```diff
@@ -231,7 +231,7 @@ class AppActive extends React.Component {
                 );
             } else if (vm.createInProgress) {
                 return (
-                    <EmptyStatePanel title={ cockpit.format(_("VM $0 is being created"), cockpit.location.options.name, cockpit.location.options.connection) }
+                    <EmptyStatePanel title={ cockpit.format(_("Creating VM $0"), cockpit.location.options.name, cockpit.location.options.connection) }
```
martinpitt, line 234: This belongs into the previous commit, please rebase.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -772,7 +774,7 @@ class CreateVmModal extends React.Component {
         }
         super(props);
         this.state = {
-            inProgress: false,
+            inProgress: 0,
```
martinpitt, line 750: Please define and use a named constant for "0" as well; like "INVALID" or "DEFINE"; and `inProgress` still sounds boolean-y, maybe rename to "stage" or "step"?

test/check-machines-create
```diff
@@ -820,7 +817,10 @@ vnc_password= "{vnc_passwd}"
             return self
 
         def createAndVerifyVirtInstallArgsCloudInit(self):
-            self.browser.click(".pf-c-modal-box__footer button:contains(Create)")
+            if (self.create_and_run):
```
martinpitt, line 820: Python style: no () around if condition.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -203,7 +203,7 @@ function validateParams(vmParams) {
 
     if (vmParams.nodeMaxMemory && vmParams.memorySize > convertToUnit(vmParams.nodeMaxMemory, units.KiB, vmParams.memorySizeUnit)) {
         validationFailed.memory = cockpit.format(
-            _("Up to $0 $1 available on the host"),
+            _("$0 $1 available on host"),
```
martinpitt, line 206: What's the rationale for this? This breaks 2x38 translations.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -957,7 +932,7 @@ class CreateVmModal extends React.Component {
                 storageSize: convertToUnit(this.state.storageSize, this.state.storageSizeUnit, units.GiB),
                 storagePool: this.state.storagePool,
                 storageVolume: this.state.storageVolume,
-                unattended: this.state.unattendedInstallation,
+                unattended: unattendedInstallation, // TODO
```
martinpitt, line 935: This still needs to be resolved?

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1001,6 +976,10 @@ class CreateVmModal extends React.Component {
         const { nodeMaxMemory, nodeDevices, networks, osInfoList, loggedUser, storagePools, vms } = this.props;
         const validationFailed = this.state.validate && validateParams({ ...this.state, osInfoList, nodeMaxMemory, vms: vms.filter(vm => vm.connectionName == this.state.connectionName) });
 
+        const unattendedInstructionsMessage = _("Enter root and/or user information to enable unattended installation.");
+        const unattendedUnavailableMessage = _("Automated installs are only available when downloading an image or using cloud-init.");
+        const unattendedOsUnsupportedMessageFormat = os => _(`${os} does not support unattended installation.`);
```
martinpitt, line 981: This i18n pattern does not work. *Never* use _() with backtick-formatting, as it's not translatable. You *first* need to translate the template and *then* fill it in with values. Please use cockpit.format() (of course not here in the constant, but where you use it)

martinpitt (changes requested):
Thanks! First round of review. I suggest to split this up into several smaller PRs.

src/components/create-vm-dialog/createVmDialog.jsx
martinpitt: I filed PR #655 to fix and spot that.

src/components/common/machinesConnectionSelector.jsx
```diff
@@ -30,7 +30,7 @@ export const MachinesConnectionSelector = ({ onValueChanged, loggedUser, connect
         return null;
 
     return (
-        <FormGroup label={_("Connection")} isInline hasNoPaddingTop id={id} className="machines-connection-selector">
+        <FormGroup label={_("Location")} isInline hasNoPaddingTop id={id} className="machines-connection-selector">
```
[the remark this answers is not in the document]
skobyda: I supported the change because I can see that "connection" can be just as confusing. It's true that connection is the term used by libvirt, but that's it to be honest. It's not really used elsewhere, like in qemu, kvm, or other virtualization technologies.

For a user who didn't really use libvirt before, the term "connection" might be even more confusing. Since if I just google "VM connection" or "Virtualization connection", all the articles are about either connecting to the VM (e.g. ssh, VNC and its alternatives) or about VM's connectivity network. So for a user who didn't use libvirt before, maybe they would confuse "connection" with connectivity to the VM or network connection.

Also what we are referring in the UI as "connection" is not really the whole thing. Libvirt refers to the connection as [driver][+optionalTransportMode]://[pathToLocation]/[system/user], so the examples of "connection" according to libvirt are:
```
qemu:///system
xen:///session
test+tcp://localhost:5000/default
qemu+ssh://root@hail.cloud.example.com/system
```

In our UI, we are using the term "connection" for only 1 part of that, i.e. system/session, and libvirt actually refers to that as "daemon mode", as it specifies under which "mode" (user and its privileges) it will connect to the libvirtd daemon.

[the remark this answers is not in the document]
skobyda: In short, my point is that I'm open to dropping the term "connection", but also open to trying to figure out alternative term.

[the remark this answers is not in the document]
skobyda: @garrett this Martin's comment may interest you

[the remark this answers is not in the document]
skobyda: Also let's probably move this into a separate PR so it doesn't block this, what do you all say?

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: Let's move this whole functionality about generating VM names to a separate PR.

src/components/common/machinesConnectionSelector.jsx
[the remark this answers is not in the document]
martinpitt: Yes, fully agreed -- that's what I meant by splitting the PRs; this label rename, the auto-naming, and the tabification are all independent from each other. And at least the auto-naming is rather complex.

Maybe "connection" is not ideal, but replacing it with an even less specific name (and breaking translations alongside) isn't that useful either IMHO.

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: Ah no, that was from ancient times when I started this PR. Resolved already

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: It was discussed in [this comment](https://github.com/cockpit-project/cockpit-machines/pull/560#issuecomment-1062915042) and 2 comments above.

src/app.jsx
```diff
@@ -229,6 +229,17 @@ class AppActive extends React.Component {
                                      }
                                      icon={ExclamationCircleIcon} />
                 );
+            } else if (vm.createInProgress) {
+                return (
+                    <EmptyStatePanel title={ cockpit.format(_("Creating VM $0"), cockpit.location.options.name, cockpit.location.options.connection) }
```
martinpitt, line 234: Arguments don't match up -- template just has `$0` and no `$1`, but you supply two values.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -745,7 +748,7 @@ class CreateVmModal extends React.Component {
         }
         super(props);
         this.state = {
-            inProgress: false,
+            inProgress: INVALID,
```
martinpitt, line 723: As the other comment was resolved: `inProgress` sounds boolean-y. Please rename to e.g. `createMode`, or similar.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -970,6 +948,10 @@ class CreateVmModal extends React.Component {
         const { nodeMaxMemory, nodeDevices, networks, osInfoList, loggedUser, storagePools, vms } = this.props;
         const validationFailed = this.state.validate && validateParams({ ...this.state, osInfoList, nodeMaxMemory, vms: vms.filter(vm => vm.connectionName == this.state.connectionName) });
 
+        const unattendedInstructionsMessage = _("Enter root and/or user information to enable unattended installation.");
+        const unattendedUnavailableMessage = _("Automated installs are only available when downloading an image or using cloud-init.");
+        const unattendedOsUnsupportedMessageFormat = os => cockpit.format("$0 does not support unattended installation.", os);
```
martinpitt, line 953: Still needs gettext call around the template string.

martinpitt (changes requested):
Thanks! Some little details, and the tests broke - looks like there's a remaining `startVm` where which didn't get renamed.

skobyda:
testCreateAndVerifyQemuConf fails at [fedora-rawhide](https://artifacts.dev.testing-farm.io/2acbdb0c-d73d-4af6-a3bc-48e4f71df7e2/). But at first glance that seems unrelated with the redesign. @martinpitt Have you seen the failure before?

martinpitt:
@skobyda : Yes, I've seen it quite a lot of times. It seems to be some kind of race condition. #660 should help, although only "by accident" -- I wanted to run the tests without qxl, just not because of this flake :-)

martinpitt:
@skobyda Dang, there's a conflict, can you please rebase?

I otherwise re-reviewed the code and tested it interactively, and :+1: So let's land this quickly after rebase and green. Thanks!

martinpitt (approved):
Cheers! Tests still running (please wait for green), but otherwise LGTM!

PR #562, opened 2022-02-08
Use image-customize's new --fresh function

martinpitt:
See https://github.com/cockpit-project/bots/commit/468ee1b39e2

This avoids hardcoding the test image overlay paths.


skobyda (approved):

martinpitt:
[unrelated pixel test failure](https://logs.cockpit-project.org/logs/pull-562-20220208-101932-83cd3d41-fedora-35-firefox/log.html) -- this is the second case of "slightly jumping fonts" that I saw. We need to apply some harder measure for these :cry:

PR #568, opened 2022-02-14
Fix broken virt-install XML without "Immediately start VM"

martinpitt:
Whenever there is an install phase (e.g. booting install media from an
iso or network and using that to install an OS into the guest),
virt-install will generate two slightly different guest
definitions: one for the very first boot (where the OS is installed to
the guest) and one for subsequent boots (after the OS is installed on
the guest). By default, --print-xml will print both the install phase
XML and the final XML, which resulted in a failure due to

    xml.etree.ElementTree.ParseError: junk after document element: line 61, column 0

Avoid that by only printing the XML for the first phase.

https://bugzilla.redhat.com/show_bug.cgi?id=2032462

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -287,14 +287,14 @@ class TestMachinesCreate(VirtualMachinesCase):
         b.enter_page('/machines')
         b.wait_in_text("body", "Virtual machines")
 
-        # First create the PXE VM but do not start it. We 'll need to tweak a bit the XML
-        # to have serial console at bios and also redirect serial console to a file
+        # Create with immediate starting
         runner.createTest(TestMachinesCreate.VmDialog(self, name='pxe-guest', sourceType='pxe',
                                                       location="network=pxe-nat",
                                                       memory_size=128, memory_size_unit='MiB',
                                                       storage_pool="NoStorage",
                                                       start_vm=True, delete=False))
 
+        # Stop it, tweak the XML to have serial console at bios and also redirect serial console to a file
         # We don't want to use start_vm == False because if we get a separate install phase
         # virt-install will overwrite our changes.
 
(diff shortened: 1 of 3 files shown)


test/check-machines-create
```diff
@@ -287,14 +287,14 @@ class TestMachinesCreate(VirtualMachinesCase):
         b.enter_page('/machines')
         b.wait_in_text("body", "Virtual machines")
 
-        # First create the PXE VM but do not start it. We 'll need to tweak a bit the XML
```
KKoukiou, line 290: I see that this comment is after a few years of writing detaching from it's original code, but it's still referring to the XML editing done a few lines bellow. Line 317 and after.

I don't encourage keeping it, just clarifying that's actually still relevant.

test/check-machines-create
```diff
@@ -375,6 +375,15 @@ class TestMachinesCreate(VirtualMachinesCase):
         self.machine.execute("virsh undefine pxe-guest")
         runner.checkEnvIsEmpty()
 
+        # Check postponed starting
```
KKoukiou, line 378: There is a whole test for this already, testCreateThenInstall.

Maybe we want to just extend that one for the state check?

test/check-machines-create
martinpitt: Right, but the "do not start it" is outright wrong. That's why I moved the comment about the XML fudging a little further down.

test/check-machines-create
martinpitt: I'm happy to move it there, sure! I'll still let the tests finish here, to check if the changed option breaks anything else.

martinpitt: Done. I validated that this still reproduces the original bug.

martinpitt:
Whoops, [this run](https://logs.cockpit-project.org/logs/pull-568-20220214-141443-5fb00703-fedora-35-firefox/log.html#9) already caught the updated container. See https://github.com/cockpit-project/cockpituous/pull/455 , I backed this out, we need to re-run this.

martinpitt:
Ugh, with [that commit](https://github.com/cockpit-project/cockpit-machines/commit/9af9a5e78844f308b6ebb9f984398cfea5301f1c) we got tons of flakes. Let's try if that's really only to blame on the "3x affected" (seems like it) or the increased parallelization (if not, let's get hard data that it isn't).

martinpitt:
This looks much better now. I am investigating the infra problems, but until that gets solved, let's revert the increased parallelism.

KKoukiou (approved):

PR #574, opened 2022-02-16
fmf: Fix modular libvirt socket startup on Rawhide

martinpitt:
Rawhide is Fedora 37 now.

Invert the condition to only exclude the old platforms which are known
not to have modular services.

---

See [current failure](http://artifacts.dev.testing-farm.io/5e3c5c7e-757b-4bf1-9ec3-89952739e74a/)

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -56,7 +56,9 @@ systemctl start firewalld
 firewall-cmd --add-service=cockpit --permanent
 firewall-cmd --add-service=cockpit
 
-if grep -Eq 'PLATFORM_ID=.*(f35|f36|el9)' /etc/os-release; then
+. /usr/lib/os-release
+
+if [ "${PLATFORM_ID:-}" != "platform:f34" ] && [ "${PLATFORM_ID:-}" != "platform:el8" ]; then
     # HACK: new modular libvirt sockets are not running by default in f35
     # https://gitlab.com/libvirt/libvirt/-/issues/219
     systemctl start virtinterfaced.socket


martinpitt:
FTR, [c8s also fails](http://artifacts.dev.testing-farm.io/eb551efa-f42f-46b7-bccc-468b18ed709f/). This is directly related to #568, it almost looks like the installed package does not have that fix yet? But it's a completely unrelated problem, I'll look at that separately.

martinpitt:
[still two failures](http://artifacts.dev.testing-farm.io/c75137d3-33a6-4e8a-b8e3-526b58d37fa7/), but at least that's progress -- right now, everything fails.

skobyda (commented):
Code changes looks good, but the test failures should be looked at (I have not seen some of them yet, so maybe the one on fedora-rawhide?)

skobyda (approved):

martinpitt:
Looking into the c8s failure. Keeping notes here. Reproducer:
```
tmt run --until prepare provision --how virtual --image CentOS-Stream-8
sshe -p 10022 -i /var/tmp/tmt/run-002/plans/all/provision/default/id_ecdsa root@127.0.0.1 dnf install https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-machines-574/centos-stream-8-x86_64/03511079-cockpit-machines/cockpit-machines-260-1.20220216071018304082.pr574.31.g7494ecd.el8.noarch.rpm
tmt run --until report -l
```

This succeeds, hmm.. So once again some weird difference between the local VMs and the Testing Farm images.

PR #576, opened 2022-02-17
fmf: Use official TMT test data directory for artifacts

martinpitt:
See https://tmt.readthedocs.io/en/stable/overview.html#variables

---

Exact same fix as in https://github.com/cockpit-project/cockpit-podman/pull/922 , where I tested it with a broken commit. As rawhide currently fails, we should get some artifacts on that.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -8,9 +8,9 @@ if [ -d source ]; then
 else
     SOURCE="$(realpath $TESTS/../..)"
 fi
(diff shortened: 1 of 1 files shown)


jelly (approved):

martinpitt:
The (known) failing [rawhide run](http://artifacts.dev.testing-farm.io/719af354-6bef-41ae-ae91-1586f775846f/) has [artifacts](http://artifacts.dev.testing-farm.io/719af354-6bef-41ae-ae91-1586f775846f/work-allGpUZUi/plans/all/execute/data/test/browser/data/), yay!

PR #580, opened 2022-02-21
packit: Disable Fedora 34, enable Fedora 36, fix test regressions

martinpitt:
Corresponding to cockpituous changes in commit 72318f86194d1.

Investigating test regressions, also on Rawhide:

 - [x] `TestMachinesDisks.testAddDiskPool`: Genuinely broken, see https://github.com/cockpit-project/bots/pull/2977 for adding naughty
 - [x] `TestMachinesFilesystems:testBasicSystemConnection`: Fixed in separate commit
 - `TestMachinesCreate.testCreateThenInstall` (rawhide only): Moved to issue #581, this is more complex

These all reproduce in local tmt VM.

[f36 failure log](http://artifacts.dev.testing-farm.io/b402bb5c-7754-4dab-8743-2de515356db1/)
[rawhide failure log](http://artifacts.dev.testing-farm.io/438058c3-e338-4fd9-b055-293f6ac96adc/)

### DIFF
--- a/packaging/cockpit-machines.spec.in
+++ b/packaging/cockpit-machines.spec.in
@@ -39,6 +39,7 @@ Requires: libvirt-dbus >= 1.2.0
 Recommends: virt-install
 Recommends: libosinfo
 Recommends: python3-gobject-base
+Suggests: qemu-virtiofsd
 
 %description
 Cockpit component for managing virtual machines.
--- a/packit.yaml
+++ b/packit.yaml
@@ -13,8 +13,8 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      - fedora-34
       - fedora-35
+      - fedora-36
       - fedora-rawhide
       - centos-stream-8
       - centos-stream-9
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -70,6 +70,11 @@ if [ "${PLATFORM_ID:-}" != "platform:f34" ] && [ "${PLATFORM_ID:-}" != "platform
     systemctl start virtstoraged.socket
 fi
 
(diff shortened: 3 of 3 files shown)


martinpitt:
`TestMachinesFilesystems:testBasicSystemConnection`: The domain fails to start:
```
virtqemud[22408]: End of file while reading data: Input/output error
virtqemud[22408]: operation failed: Unable to find a satisfying virtiofsd
virtqemud[22408]: End of file while reading data: Input/output error
```

At first this smells like https://bugzilla.redhat.com/show_bug.cgi?id=1966580 , but now /usr/libexec/virtiofsd does not exist at all any more. It is now a separate package `qemu-virtiofsd`, when I install it, it works.

martinpitt:
`TestMachinesCreate.testCreateThenInstall`: (I suppose) that grep is meant to confirm the presence of the `METADATA` added by src/scripts/install_machine.py, and indeed it is completely absent.

I added a `sit()` in `_tryCreateThenInstall()` right after

    wait(lambda: dialog.name in self.machine.execute("virsh list --persistent"), delay=3)

and another one after the forceOff.

The `virt-install` command never finishes by itself, and the metadata never appears in dumpXml. Only after forceOff'ing the VM these two happen. So this is *not* a race condition (I can sit for quite long on either point) or a behaviour change in virt-install.

This still seems to be the case for F37, i.e. no change. However, it seems `inject_metadata()` from src/scripts/install_machine.py now does not get called any more.

This seems to affect rawhide only, I now saw at least two F36 logs where this test worked. So let's land this one first, to cover F36.

jelly (approved):

martinpitt:
Wohoo, all green, how come? #581 is very persistent for me.. Anyway,  not complaining :grin:

PR #583, opened 2022-02-21
Use `virt-install --reinstall` for implementing the 'Install' button functionality

KKoukiou:
- [x] https://github.com/cockpit-project/bots/pull/2992
 - [x] https://github.com/cockpit-project/bots/pull/3033
 - [x] Land and rebase after PR #627
 - [x] Fix in [Ubuntu jammy](https://launchpad.net/ubuntu/+source/libvirt/8.0.0-1ubuntu5) (next LTS)
 - [x] ~Wait for 8.1.0-1 to go into [Debian testing](https://tracker.debian.org/pkg/libvirt), that version contains the AppArmor fix~ -- obsolete, current libvirt already fails with current c-machines, see https://github.com/cockpit-project/bots/pull/3108
 - [x] Fix tests enough so that we can have confidence in them again: #630, #631, #633, #651
 - [x] Fix [testCreateUrlSource](https://logs.cockpit-project.org/logs/pull-583-20220404-065517-8611b971-debian-testing/log.html#72) race condition with unpredictable Alert
 - Fix [testCreateUrlSource on arch](https://logs.cockpit-project.org/logs/pull-583-20220404-103722-1f6bcec2-arch/log.html#72): https://github.com/cockpit-project/bots/pull/3187
 - goes on top of PR #656
 - goes on top of PR #787


test/check-machines-create
```diff
@@ -1478,8 +1478,11 @@ class TestMachinesCreate(VirtualMachinesCase):
 
         def createThenInstallTest(self, dialog, installFromVmDetails=False, tryWithFailInstall=False):
             self._tryCreateThenInstall(dialog, installFromVmDetails, tryWithFailInstall) \
-                ._assertScriptFinished() \
-                ._assertDomainDefined(dialog.name, dialog.connection) \
+                ._assertScriptFinished()
+
+            wait(lambda: "shut off" if tryWithFailInstall else "running" in self.machine.execute(f"virsh dominfo --domain {dialog.name} | grep State"), delay=3)
```
martinpitt, line 1483: Note, in my local test runs it was always shut off here, as `_tryCreateThenInstall()` runs `forceOff`. (Before that, it's running)

KKoukiou:
Debians and ubuntus now fail on two tests because of Apparmor. <- Add naugthy

KKoukiou:
rawhide and f36 on the CI, fail in the testCreateThenInstall, testConfigureBeforeInstall with:

ERROR    \n--os-variant/--osinfo OS name is required, but no value was\nset or detected.\n\nThis is now a fatal error.

I am debugging this currently.

martinpitt:
[ubuntu-2004 is a sea of red](https://logs.cockpit-project.org/logs/pull-583-20220225-114640-8bf947a6-ubuntu-2004/log.html) as `systemctl start virtqemud.service` fails -- that version does not have split daemons yet. However, I thought we agreed on bumping the `virtinst` dependency (rpm and deb) to ensure that the new `--reinstall` option is available, and dropping ubuntu-2004 from testmap.

The Debian and ubuntu-stable failures are due to the apparmor issue, handled by https://github.com/cockpit-project/bots/pull/2992

src/scripts/install_machine.py
```diff
@@ -181,12 +180,18 @@ def prepare_virt_install_params(args):
...
             "--connect", f"qemu:///{args['connectionName']}",
             "--quiet",
-            "--name", args['vmName'],
-            "--os-variant", args['os'],
-            "--memory", str(args['memorySize']),
         ]
 
-        if args['type'] == 'install' or ('storagePool' in args and args['storagePool'] != 'NewVolume'):
+        if args['type'] == 'install':
+            params += ['--reinstall', args['vmName']]
```
martinpitt, line 187: Repeating it here to have it as part of the review: Please bump the virt-install dependency to `>= 3.0.0` in packaging/debian/control and the spec.

src/libvirtApi/domain.js
```diff
@@ -588,6 +588,7 @@ export function domainInstall({ onAddErrorNotification, vm }) {
     return cockpit
             .spawn([pythonPath, "--", "-", args], opts)
             .input(installVmScript)
+            .fail(ex => console.error(JSON.stringify(ex)))
```
martinpitt, line 591: .catch() please, let's avoid introducing jQuery promise API.

test/machineslib.py
```diff
@@ -290,3 +290,9 @@ def setUp(self):
 
         # HACK: older c-ws versions always log an assertion, fixed in PR cockpit#16765
         self.allow_journal_messages("json_object_get_string_member: assertion 'node != NULL' failed")
+
+    def tearDown(self):
+        b = self.browser
+        if (b.is_present("#button.alert-link.more-button")):
+            b.click("button.alert-link.more-button")
```
martinpitt, line 293: Thanks for this! The messages should also appear in the test log, but this is nice.

martinpitt (changes requested):
Amazing cleanup, epharisto! Let's get this over the finish line soon, as it's quite intrusive.

martinpitt:
@KKoukiou : FTR: if you are busy with anaconda, I'm happy to help out here (bots testmap and the few cleanups above)

src/libvirtApi/domain.js
KKoukiou: catch and fail are slightly different, in that catch will return a new (resolved) promise, whereas fail will return the original promise

In this case I want to propagate the original failure, for the UI error handling to show the error as well.

KKoukiou: I think I can use then instead.

martinpitt: Interesting, I was not aware of that. Thanks!

KKoukiou:
Hmm, we have some failures on centos-9 and rawhide packit:

```
> info: spawn 'vm creation' returned error: "{"problem":null,"exit_status":1,"exit_signal":null,"message":"error: failed to get emulator capabilities\nerror: internal error: missing 'nvram-template' in '/usr/share/qemu/firmware/50-edk2-ovmf-amdsev.json'\nTraceback (most recent call last):\n  File \"<stdin>\", line 341, in <module>\n  File \"<stdin>\", line 249, in create_vm\n  File \"/usr/lib64/python3.9/contextlib.py\", line 119, in __enter__\n    return next(self.gen)\n  File \"<stdin>\", line 222, in prepare_virt_install_params\n  File \"<stdin>\", line 74, in prepare_graphics_params\n  File \"<stdin>\", line 31, in get_graphics_capabilies\n  File \"<stdin>\", line 18, in virsh\n  File \"/usr/lib64/python3.9/subprocess.py\", line 424, in check_output\n    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,\n  File \"/usr/lib64/python3.9/subprocess.py\", line 528, in run\n    raise CalledProcessError(retcode, process.args,\nsubprocess.CalledProcessError: Command '('virsh', '-c', 'qemu:///system', 'domcapabilities')' returned non-zero exit status 1."}"
```

martinpitt:
@KKoukiou : c9s is already addressed, see PR #602 (please rebase). rawhide is really broken, I'm investigating in #604, but as rawhide is broken in multiple ways, that's hard to investigate :cry:

KKoukiou:
The failure in rawhide for testCreateThenInstall is because of a segfault:

https://artifacts.dev.testing-farm.io/1a5b2bc9-e1dc-4360-bc6c-082bccbfe0b1/work-allDzMryG/plans/all/execute/data/test/browser/data/TestMachinesCreate-testCreateThenInstall-fedora-37-localhost-22-FAIL.png

And in the journal you can see:
89944 Mar 03 12:16:34 ip-172-31-16-196.us-east-2.compute.internal kernel: rpc-virtqemud[280144]: segfault at 18 ip 00007f3e4c565621 sp 00007f3e4ce8a7b0 error 4 in libvirt_driver_qemu.so[7f3e4c4d4000+130000]

We need to add the debuginfo packages and create a bug report.

martinpitt:
@KKoukiou : I already investigated that yesterday in #604. I filed https://bugzilla.redhat.com/show_bug.cgi?id=2060412 and it just got fixed today, so hopefully rawhide should be ok again from tomorrow on.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/3033 to drop ubuntu-2004 from testmap. The [new naughty](https://github.com/cockpit-project/bots/pull/2992) will take care of debian and ubuntu-stable, and the (independent) rawhide failure got fixed last week.

So what remains here is the [testCreateUrlSource failure on debian-stable](https://logs.cockpit-project.org/logs/pull-583-20220303-104035-07639b87-debian-stable/log.html#44-2). I'll push a rebase to get a current result.

martinpitt:
I also squashed the commits a little bit. If this goes green, and after https://github.com/cockpit-project/bots/pull/3033 gets approved and landed, I'll land https://github.com/cockpit-project/bots/pull/2992 and re-run the tests here.

martinpitt:
Oh, the [naughty pattern does not match](https://logs.cockpit-project.org/logs/pull-583-20220307-060237-dcf0b7c8-debian-stable-bots-2992/log.html#35-2). It seems to expect logging an error message to the console, but we only see the apparmor error. I updated the naughty pattern accordingly.

martinpitt:
There is [testCreateUrlSource failure](https://logs.cockpit-project.org/logs/pull-583-20220307-060025-dcf0b7c8-debian-testing/log.html#44-2) again, this time on debian-testing instead of debian-stable. This also [failed testCreateFileSource](https://logs.cockpit-project.org/logs/pull-583-20220307-060025-dcf0b7c8-debian-testing/log.html#44-2) (but a retry worked)

KKoukiou:
> There is [testCreateUrlSource failure](https://logs.cockpit-project.org/logs/pull-583-20220307-060025-dcf0b7c8-debian-testing/log.html#44-2) again, this time on debian-testing instead of debian-stable. This also [failed testCreateFileSource](https://logs.cockpit-project.org/logs/pull-583-20220307-060025-dcf0b7c8-debian-testing/log.html#44-2) (but a retry worked)

Seems like the `testCreateURLSource` fails on an ambiguous selector:
wait_js_cond(ph_is_visible(".pf-c-alert-group li .pf-c-alert")): .pf-c-alert-group li .pf-c-alert is ambiguous

test/browser/run-test.sh
```diff
@@ -48,8 +48,8 @@ if [ "$TEST_OS" = "centos-8-stream" ]; then
     EXCLUDES="$EXCLUDES TestMachinesConsoles.testExternalConsole"
 fi
 
-# FIXME: https://github.com/cockpit-project/cockpit-machines/issues/581
-EXCLUDES="$EXCLUDES TestMachinesCreate.testCreateThenInstall"
+# depends on available host devices, TF machines not predictable enough
+EXCLUDES="$EXCLUDES TestMachinesHostDevs.testHostDevAdd"
```
martinpitt, line 52: That may be necessary, but it should then go into its own commit (or possibly PR) with separate rationale. Did you see this fail recently?

test/machineslib.py
```diff
@@ -290,3 +290,9 @@ def setUp(self):
 
         # HACK: older c-ws versions always log an assertion, fixed in PR cockpit#16765
         self.allow_journal_messages("json_object_get_string_member: assertion 'node != NULL' failed")
+
+    def tearDown(self):
+        b = self.browser
+        if (b.is_present("#button.alert-link.more-button")):
```
martinpitt, line 293: This also needs to check `b.cdp.valid`, so that it doesn't try to do that if a test fails in the setup phase, before starting the browser and logging into cockpit.

martinpitt (changes requested):
This looks very nice now, aside from two minor issues. The tests should be green once we land the pending naughty bots PR, and rawhide got fixed yesterday.

@KKoukiou , is there anything missing from your POV? If not, I'm happy to take over and drive this to completion.

KKoukiou:
> This looks very nice now, aside from two minor issues. The tests should be green once we land the pending naughty bots PR, and rawhide got fixed yesterday.
> 
> @KKoukiou , is there anything missing from your POV? If not, I'm happy to take over and drive this to completion.

There is nothing else missing from my POV. Sure, if it's fine for you :)

martinpitt:
I pulled the test fixes/cleanups into PR #627, as these are really desirable and are ready to land. I fixed the two issues above in that PR.

test/browser/run-test.sh
martinpitt: Fixed in #627.

test/machineslib.py
martinpitt: Fixed in #627

martinpitt:
Rebased.

martinpitt:
Debian-testing is now broken even with the current code, see https://github.com/cockpit-project/bots/pull/3108 . So this is not blocked any more, we can fix/land it, and stop backporting to Debian stable and Ubuntu impish.

martinpitt:
I retried this too many times already. I think there's some new flakes here, e.g. [this one](https://logs.cockpit-project.org/logs/pull-583-20220322-095355-edd6aec0-centos-8-stream/log.html#64) looks unfamiliar. However, before we go ahead with anything new, we really really need to fix the dramatically bad situation of main first. I work on that in PR #630, blocked until then.

martinpitt:
There are now some real regressions here. The `CURL: Error opening file` could be https://github.com/cockpit-project/bots/issues/3066 . But it also happens on [debian-stable](https://logs.cockpit-project.org/logs/pull-583-20220325-052802-85e79c7d-debian-stable/log.html#72):

> info: spawn 'vm creation' returned error: "{"problem":null,"exit_status":1,"exit_signal":null,"message":"setlocale: No such file or directory\nERROR    internal error: process exited while connecting to monitor: 2022-03-25T05:33:29.137161Z qemu-system-x86_64: -blockdev {\"driver\":\"https\",\"url\":\"https://archive.fedoraproject.org:443/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso\",\"node-name\":\"libvirt-1-storage\",\"auto-read-only\":true,\"discard\":\"unmap\"}: CURL: Error opening file: \nDomain installation does not appear to have been successful.\nIf it was, you can restart your domain by running:\n  virsh --connect qemu:///system start subVmTestCreate7\notherwise, please restart your installation.\nTraceback (most recent call last):\n  File \"<stdin>\", line 341, in <module>\n  File \"<stdin>\", line 250, in create_vm\n  File \"/usr/lib/python3.10/subprocess.py\", line 420, in check_output\n    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,\n  File \"/usr/lib/python3.10/subprocess.py\", line 524, in run\n    raise CalledProcessError(retcode, process.args,\nsubprocess.CalledProcessError: Command '['virt-install', '--connect', 'qemu:///system', '--quiet', '--os-variant', 'fedora28', '--memory', '128', '--name', 'subVmTestCreate7', '--check', 'path_in_use=off', '--wait', '-1', '--noautoconsole', '--disk', 'none', '--graphics', 'vnc,listen=127.0.0.1', '--graphics', 'spice,listen=127.0.0.1', '--cdrom', 'https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso']' returned non-zero exit status 1."}"

This points to some incomplete setup of the mock tree.

martinpitt:
Some [arch flakes](https://logs.cockpit-project.org/logs/pull-583-20220404-090019-1f6bcec2-arch/log.html#68-1) due to 3x affected retry, but I think they are unrelated.

test/check-machines-create
```diff
@@ -593,23 +593,28 @@ vnc_password= "{vnc_passwd}"
                                                       storage_size=128, storage_size_unit='MiB',
                                                       start_vm=False))
 
-        # Test detection of ISO file in URL
-        runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='url',
-                                                      location=config.ISO_URL,
-                                                      memory_size=128, memory_size_unit='MiB',
-                                                      storage_pool="NoStorage",
-                                                      start_vm=True))
+        self.browser.eval_js('window.debugging = "machines"')
```
martinpitt, line 596: Whoops, debugging leftover. Although it's quite nice to investigate the arch failure :grin:

martinpitt:
Oh, got it -- [this arch failure](https://logs.cockpit-project.org/logs/pull-583-20220404-103722-1f6bcec2-arch/log.html#72) is nothing more than our [recent curl breakage](https://github.com/cockpit-project/bots/issues/3066), so just needs an updated naughty.

martinpitt:
@skobyda : This is one depending bots PR away from going green. It looks fine to me, and all the heavy lifting has been done (see checklist in the description). Do you want to review this?

martinpitt:
There is [another](https://logs.cockpit-project.org/logs/pull-583-20220404-162508-839156e7-arch/log.html#68) [race](https://logs.cockpit-project.org/logs/pull-583-20220404-162508-839156e7-fedora-35-firefox/log.html#67-2) with `checkEnvIsEmpty()` -- the Alert happens after that function clicks away all alerts. This is a similar root cause as for the testCreateUrlSource above -- we really need to figure out when to expect a failure and when not, and wait for it. The current stuff is too probabilistic. Sorry @KKoukiou , this PR is a gift that keeps on giving :-/

martinpitt (dismissed):

KKoukiou (commented):
Can't approve but changes look great! Pleased to see this finally getting in!

src/libvirtApi/domain.js
```diff
@@ -613,17 +604,11 @@ export function domainInstall({ onAddErrorNotification, vm }) {
     const args = JSON.stringify({
         autostart: vm.autostart,
```
skobyda, line 605: Is this option necessarry? I believed this also configured by virt-install --reinstall. I just tried deleting it from the python script, then created VM, changed status of autostart, installed the VM and the correct status of autostart persisted.

src/scripts/install_machine.py
```diff
@@ -322,10 +305,15 @@ def install_vm(args):
...
+    ET.register_namespace("cockpit_machines", ns["cockpit_machines"])
+    ET.register_namespace("libosinfo", "http://libosinfo.org/xmlns/libvirt/domain/1.0")
+
     # ET.fromstring() already wants UTF-8 encoded bytes
     root = ET.fromstring(xml)
     metadata = root.find('metadata')
-    cockpit_machines_metadata = metadata.find('cockpit_machines')
+    cockpit_machines_metadata = metadata.find('cockpit_machines:data', ns)
     if cockpit_machines_metadata:
         metadata.remove(cockpit_machines_metadata)
```
skobyda, line 318: I think this part is broken. Following the creation of VM, I run `virsh dumpxml vmName` both before and after I clicked on "Install", but in both cases the metadata seems untouched:
```
  <metadata xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0" xmlns:cockpit_machines="https://github.com/cockpit-project/cockpit-machines">
    <libosinfo:libosinfo>
      <libosinfo:os id="http://centos.org/centos-stream/9"/>
    </libosinfo:libosinfo>
    <cockpit_machines:data>
      <cockpit_machines:has_install_phase>true</cockpit_machines:has_install_phase>
      <cockpit_machines:install_source_type>os</cockpit_machines:install_source_type>
      <cockpit_machines:install_source/>
      <cockpit_machines:os_variant>centos-stream9</cockpit_machines:os_variant>
    </cockpit_machines:data>
  </metadata>
```

skobyda (changes requested):

test/check-machines-create
```diff
@@ -1217,10 +1217,11 @@ vnc_password= "{vnc_passwd}"
...
-            second_state = "Running" if dialog.start_vm else "Shut off"
 
-            self._assertVmStates(dialog.name, init_state, second_state)
+            if dialog.start_vm:
+                b.wait_text(f"#vm-{dialog.name}-state", "Running")
+            else:
+                self._assertVmStates(dialog.name, "Creating VM", "Shut off")
             b.wait_not_present("#create-vm-dialog")
 
         def _tryCreate(self, dialog):
```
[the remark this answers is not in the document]
martinpitt: The implementation may have moved, but IMHO we should still check that this use case (changing settings before starting the VM) actually works. "We trust that this works" has bitten us way too often :grin:

src/scripts/install_machine.py
KKoukiou: I am not able to reproduce this. Before the "install" the metadata are expected to be there. After clicking install, the metadata are removed.

src/libvirtApi/domain.js
KKoukiou: I indeed verified manually as well that is option is also propagated.

KKoukiou: Honestly however, to me it seems like a side effect. As it's not part of the VM xml definition.

skobyda (changes requested):

src/libvirtApi/domain.js
skobyda: It's true that VM's autostart configuration, is part not part of XML, but it's part of some flags/boolean values which libvirt keeps track for existing VMs.

Before this PR, we lost the autostart configuration, because for install process we undefined and then defined again the VM, so wherever libvirt kept the state of autostart, it was removed with VM undefining.
```
event 'lifecycle' for domain 'centos': Defined Added
event 'lifecycle' for domain 'centos': Undefined Removed
event 'agent-lifecycle' for domain 'centos': state: 'disconnected' reason: 'domain started'
event 'lifecycle' for domain 'centos': Resumed Unpaused
event 'lifecycle' for domain 'centos': Started Booted
event 'lifecycle' for domain 'centos': Defined Added
event 'graphics' for domain 'centos': connect local[IPv4 127.0.0.1 5900] remote[IPv4 127.0.0.1 40200] none
event 'graphics' for domain 'centos': initialize local[IPv4 127.0.0.1 5900] remote[IPv4 127.0.0.1 40200] none
```

With this PR, it seems that virt-install is only pausing the VM for installation. That means that VM with all its non-XML configuration seems to persist:
```
event 'lifecycle' for domain 'centos': Defined Added
event 'agent-lifecycle' for domain 'centos': state: 'disconnected' reason: 'domain started'
event 'lifecycle' for domain 'centos': Resumed Unpaused
event 'lifecycle' for domain 'centos': Started Booted
event 'lifecycle' for domain 'centos': Defined Updated
event 'graphics' for domain 'centos': connect local[IPv4 127.0.0.1 5900] remote[IPv4 127.0.0.1 40198] none
event 'graphics' for domain 'centos': initialize local[IPv4 127.0.0.1 5900] remote[IPv4 127.0.0.1 40198] none
```

So my point is, If virt-xml --reinstall guarantees that VM will not get undefined/redefined in the process, then I think it's not a side effect but expected behavior.

src/scripts/install_machine.py
skobyda: Hmm that's strange, I checked again and for me it persists there. Does your metadata look the same before installation begins?

test/check-machines-create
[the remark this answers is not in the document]
skobyda: Alright, then at least I would change how we check that installation has started. [Here](https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-create#L1580) we check for that by looking if VM is persistent. Which was fine because before this PR, we undefined and then defined the VM (signal `Defined Added`), so for a short period the VM was not persistent.
But as I described in [this comment](https://github.com/cockpit-project/cockpit-machines/pull/583#discussion_r843704457), `virt-install --reinstall` doesn't undefine the VM, merely just updates it (signal `Defined Updated`), so the VM if persistent throughout the whole installation process, so the above-mentioned check will always be true.
Maybe we could remove that check, or instead listen for the signal `Defined Updated`.

src/libvirtApi/domain.js
```diff
@@ -613,17 +604,11 @@ export function domainInstall({ onAddErrorNotification, vm }) {
...
         connectionName: vm.connectionName,
-        disks: prepareDisksParam(vm.disks),
-        firmware: vm.firmware == "efi" ? 'uefi' : '',
```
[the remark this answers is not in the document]
martinpitt: So we should first introduce a test that switches back from uefi to bios, make sure that works on main, and that should then fail here, right?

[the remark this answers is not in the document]
skobyda: yeah, in our test we only tests we currently only switch from BIOS -> UEFI. But This bug is mostly about switching from BIOS -> BIOS, and I think that way maybe incorrect value gets written into XML for BIOS (didn't investigate further, not sure by that claim).
I'm not sure if a test for BIOS -> UEFI -> BIOS would bring anything meaningful, but it would not be hard to code.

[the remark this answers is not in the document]
martinpitt: I sent PR #656 for this -- that detects this regression.

test/check-machines-create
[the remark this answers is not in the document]
martinpitt: I sit()ed right before the `b.click("#vm-VmNotInstalled-install")`, and `virsh list --persistent` does not show anything. It only appears in `virsh list --all`, as it is not running (and `--persistent` only shows running VMs). I think the point of that check was to wait for the VM to start/run? So I'm not sure what you mean. Do you have time to rework the test in this PR to what you have in mind?

src/libvirtApi/domain.js
martinpitt: This is convincing. I dropped it from the install_machine.py script and domain.js. `testConfigureBeforeInstall` covers this and is still happy.

src/libvirtApi/domain.js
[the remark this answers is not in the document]
martinpitt: Merely adding back the property setting here does not fix this. So this needs some more work. I included the test case here, so that we have a reminder.

src/libvirtApi/domain.js
skobyda: Thanks. Since we still keep the test, if in the future it will start failing, at least we will know that something broke in virt-install and not with us.

martinpitt:
Rebased to fix conflicts.

src/libvirtApi/domain.js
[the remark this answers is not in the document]
martinpitt: I fixed the state vs. xml values more properly now. Works fine locally/interactively, I'll let the tests confirm.

martinpitt:
[some](https://logs.cockpit-project.org/logs/pull-583-20220411-071426-99af1ed5-debian-testing/log.html) [flakes](https://logs.cockpit-project.org/logs/pull-583-20220411-071426-99af1ed5-fedora-35/log.html) -- apparently the new behavior aggravates the unpredictable "failed to install" alerts. This bug/tech debt has been looming over us for a while now, seems this PR is resistant enough that we finally need to dig  into this :cry:

martinpitt:
Rebased to fix conflict, and cleaned up the VM state check a bit.

skobyda:
Force pushing again to see if the flake is really gone.

src/scripts/install_machine.py
skobyda: Nope, I was wrong, this is actually a huge blocker, since if this metadata is left there, then after VM is installed, and you turn the VM off, UI will offer the option to install the VM again.

skobyda: Nope, I was wrong, this is actually a huge blocker, since if this metadata is left there, then after VM is installed, and you turn the VM off, UI will offer the option to install the VM again.

martinpitt:
Rebased to current main, this was conflicty.

src/scripts/install_machine.py
martinpitt: @skobyda : I tested this: I created a VM as above, then clicked "Install". I tries to PXE boot (fails and keeps trying, of course), I do "force shut down". Then the state button changes to "Run", not "Install". This is even true after logging out and back in. Turns out I didn't look hard enough back then.. Right after "create and edit", the metadata has

      <cockpit_machines:has_install_phase>true</cockpit_machines:has_install_phase>

This is true even after logging out and in, or reboot. After "Run", it is still true. But after shutting down, it changes to

      <cockpit_machines:has_install_phase>false</cockpit_machines:has_install_phase>

which explains why we don't get the Install button again. I'll write an integration test for this to make sure it keeps working.

Do you mean you got the "Install" button again, or that it tries to do something different in the background?

martinpitt:
Ugh, [what on earth is this mess](https://artifacts.dev.testing-farm.io/3281b9bf-0e1a-4728-a7af-883e3f669def/) on the TF? All OSes fail all tests with `Error: Cannot find module 'chrome-remote-interface'`. This still worked two days ago. PR #787 is also running, I'll see if it happens there also. In the meantime I rebase on top of #787 to get the additional install phase checks.

skobyda:
Finally I got back to this, sorry for keeping it open for so long.
So the flake is a race condition. We create VM using `virt-install --wait -1`, which will keep virt-install running in order to run some cleanup operations after installation is finished.
However, if you Force off the VM, and then imediatelly undefine it (which is what we do during VM deletion), the spawned virt-install process will exit with an exception because of a race condition.
I sent an issue to virt-install: https://github.com/virt-manager/virt-manager/issues/455

Now, we can work around this by either adding some timeout into our VM deletion flow, between VM force off and VM undefine,
Or we can ignore virt-install exiting with this particular exception.

test/check-machines-create
```diff
@@ -1851,14 +1851,16 @@ vnc_password= "{vnc_passwd}"
...
-        # Change the os boot firmware configuration
-        b.wait_in_text("#vm-VmNotInstalled-firmware", "BIOS")
-        b.click("#vm-VmNotInstalled-firmware")
-        b.wait_visible(".pf-c-modal-box__body")
-        b.select_from_dropdown(".pf-c-modal-box__body select", "efi")
-        b.click("#firmware-dialog-apply")
-        b.wait_not_present(".pf-c-modal-box__body")
-        b.wait_in_text("#vm-VmNotInstalled-firmware", "UEFI")
+        # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006324
+        if "debian" in m.image and m.execute("virsh --version") >= "8.1.0":
```
martinpitt, line 1856: This would skip the test on Fedora, RHEL etc. I don't think that's intended?

skobyda (dismissed):

test/check-machines-create
skobyda: Fixed

martinpitt (dismissed):
Thank you for tracking down the virt-install race! If you need to push again for some reason, please clean it up a bit further, but good enough.

test/check-machines-create
```diff
@@ -1851,14 +1851,16 @@ vnc_password= "{vnc_passwd}"
...
-        # Change the os boot firmware configuration
-        b.wait_in_text("#vm-VmNotInstalled-firmware", "BIOS")
-        b.click("#vm-VmNotInstalled-firmware")
-        b.wait_visible(".pf-c-modal-box__body")
-        b.select_from_dropdown(".pf-c-modal-box__body select", "efi")
-        b.click("#firmware-dialog-apply")
-        b.wait_not_present(".pf-c-modal-box__body")
-        b.wait_in_text("#vm-VmNotInstalled-firmware", "UEFI")
+        # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006324
+        if "debian-stable" != m.image or m.execute("virsh --version") >= "8.1.0":
```
martinpitt, line 1856: I meant if you check for precise image names, you should drop the virsh --version comparison.

martinpitt:
Hmm, `testConfigureBeforeInstall` now fails on several OSes, some checks need to be adjusted? It's plausible that these OSes also have an earlier libvirt version.

skobyda:
> Hmm, `testConfigureBeforeInstall` now fails on several OSes, some checks need to be adjusted? It's plausible that these OSes also have an earlier libvirt version.

Can't reproduce locally, and I believe they are unrelated flakes, because I remember I saw them before. I'll retrigger and see if it persists

martinpitt (approved):
I see, due to the 3x affected retry. OK!

PR #585, opened 2022-02-22
fmf: Update test exclusions

martinpitt:
- `testCreateThenInstall` and `testHostDevAdd` still don't work in
   Fedora 37, so drop the OS restriction. Update the issue link for the
   former.

- Drop exclusions of `testNICDelete` and `testSerialConsole`, these work
  now.

- Read /usr/lib/os-release instead of the /etc/os-release symlink, as
  the former has been the canonical location for a while.


marusak (approved):
Thanks!

PR #595, opened 2022-02-24
tmt: Put back absolute log path for older tmt versions

martinpitt:
Commit a1fceb256e20 accidentally dropped setting `$LOGS` to an absolute
path if `$TMT_TEST_DATA` is not set. This causes the test to fail with

    test/browser/run-test.sh: line 75: logs/exitcode: No such file or directory

when running against older `tmt` versions. This is the case on the
internal RHEL testing farm, but not on the public Fedora/CentOS one
(which is why we didn't see this in packit).

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -10,7 +10,7 @@ else
 fi
 
 # https://tmt.readthedocs.io/en/stable/overview.html#variables
-LOGS="${TMT_TEST_DATA:-logs}"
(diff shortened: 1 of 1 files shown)


martinpitt:
See internal https://src.osci.redhat.com/rpms/cockpit-machines/pull-request/19 for details, I'm testing the fix there.

KKoukiou (approved):

martinpitt:
Downstream test run passed \o/

PR #607, opened 2022-03-04
Use correct source for iface of type direct for InactiveWarning tooltip

### LINKED ISSUE #606: There should no icon beside the source of network interface after changing the type of network interface
Version:
libvirt-dbus-1.4.0-4.fc35.x86_64
cockpit-264-1.fc35.x86_64
cockpit-machines-260-1.fc35.noarch

Description:
If only changing the type of network interface,  there will be also an icon beside the source of network interface.

Steps:
1 Create a linux network bridge, like bridge0
2 Create a shutoff VM, then enter the VM detail page
3 Change the type and source of network interface to "Bridge to LAN" and "bridge0"
4 Run the VM, then change the type of network interface to "Direct attachment", keep the source of network interface as "bridge0"

Actual results:
1 After step 4, there will be also an icon beside the source of the network interface

Expected results:
1 After step 4, there will be no icon beside the source of the network interface since it is not changed

skobyda:
When editing a source of iface of a running VM, a tooltip is shown to
    inform the user that changes will take effect after reboot.
    However, for a iface of type direct, we source 'source.direct' instead
    of 'source.dev', which resulted in a tooltip being shown even when not
    necessary.
    
Fixes #606

### DIFF
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -411,6 +411,28 @@ class TestMachinesNICs(VirtualMachinesCase):
                 f"#vm-subVmTest1-network-{self.nic_num}-mac",
                 self.mac if self.mac and self.vm_state == "shut off" else self.mac_current
             )
+            if self.vm_state == "running":
+                if self.source_type is not None and self.source_type != self.source_type_current:
+                    b.wait_visible(f"#vm-subVmTest1-network-{self.nic_num}-type-tooltip")
+                else:
+                    b.wait_not_present(f"#vm-subVmTest1-network-{self.nic_num}-type-tooltip")
+
+                if self.source is not None and self.source != self.source_current:
+                    b.wait_visible(f"#vm-subVmTest1-network-{self.nic_num}-source-tooltip")
+                # Changing source_type may in effect also change source, so tooltip being present is correct
+                # Only changing source type beteen bridge and direct should not affect source
+                elif self.source_type in ["direct", "bridge"] and self.source_type_current in ["direct", "bridge"]:
+                    b.wait_not_present(f"#vm-subVmTest1-network-{self.nic_num}-source-tooltip")
+
+                if self.mac is not None and self.mac != self.mac_current:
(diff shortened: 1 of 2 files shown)


test/check-machines-nics
```diff
@@ -411,6 +411,28 @@ class TestMachinesNICs(VirtualMachinesCase):
...
+            if self.vm_state == "running":
+                if self.source_type is not None and self.source_type != self.source_type_current:
+                    b.wait_visible(f"#vm-subVmTest1-network-{self.nic_num}-type-tooltip")
+                else:
+                    b.wait_not_present(f"#vm-subVmTest1-network-{self.nic_num}-type-tooltip")
+
+                if self.source is not None and self.source != self.source_current:
+                    b.wait_visible(f"#vm-subVmTest1-network-{self.nic_num}-source-tooltip")
+                # Changing source_type may in effect also change source, so tooltip being present is correct
+                # Only changing source type beteen bridge and direct should not affect source
```
martinpitt, line 423: What makes this direct/bridge type transition special? Could there be other combinations where this is the case?

martinpitt (commented):
Thanks!

test/check-machines-nics
skobyda: On system connection where this test runs, currently you can only choose 3 types of virtual network interface: `Bridge to LAN, Direct attachement and Virtual Network.`
`Bridge to LAN` and `Direct attachemen`t share the same source set, therefore it's possible to change interface type between the two without source changing:
Screenshot from 2022-03-07 11-02-32
Screenshot from 2022-03-07 10-57-56

However virtual network has a difference set of sources, therefore when you switch to or from `virtual network`, source will reset and will always be different than in `bridge` or `direct attachement`
Screenshot from 2022-03-07 10-58-48

martinpitt: Thanks for the detailed explanation!

martinpitt (approved):
packit failed as this branch is not rebased, but it was green this morning.

PR #608, opened 2022-03-07
test: a fix in testRename

yunmingyang:
Should verify there is no VM whose name is "old", not "old-name"


martinpitt (approved):
Well spotted, thank you!

PR #611, opened 2022-03-09
packit: build SRPM in Copr

TomasTomecek:
and be able to specify the precise list of deps needed to create the
SRPM

Details: https://packit.dev/posts/copr-srpms/

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -3,6 +3,9 @@ synced_files:
   - cockpit-machines.spec
 upstream_package_name: cockpit-machines
 downstream_package_name: cockpit-machines
+srpm_build_deps:
+- npm
+- make
 actions:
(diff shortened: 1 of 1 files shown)


martinpitt:
@TomasTomecek : Sorry, that was the "approve workflow run for first-time contributors" issue. We don't need that in c-podman as sandcastle was able to run webpack, but it was too small for c-machine's webpack. We now actually *could* run webpack as copr has more power, but IMHO it's still more efficient to just build it once, and re-use it from CI.

martinpitt:
[rawhide failure](https://artifacts.dev.testing-farm.io/44b260b7-1e95-4f13-bc44-7607731684b7/), but I think that's not your fault.

martinpitt (approved):
Dekuji!

PR #621, opened 2022-03-14
Bump testlib to 264+d1a410cf9

martinpitt:
This will include https://github.com/cockpit-project/cockpit/commit/f7038fbae483f
to fix the annoying crashes with `--trace`.

Fixes #620

PR #624, opened 2022-03-15
IGNORE: debug virt-install --cdrom bug https://bugzilla.redhat.com/show_bug.cgi?id=2063853

martinpitt:
Yesterday I [could still reproduce this](https://github.com/cockpit-project/cockpit-machines/pull/621#issuecomment-1066732646) on an upgraded F36 → 37 VM, but not today any more. It still seems to happen on the TF though, let's do a few debug runs. I still can't run F36/F37 in tmt locally.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -16,8 +16,5 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      - fedora-35
       - fedora-36
       - fedora-rawhide
-      - centos-stream-8
-      - centos-stream-9
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -14,10 +14,6 @@ LOGS="${TMT_TEST_DATA:-$(pwd)/logs}"
 mkdir -p "$LOGS"
 chmod a+w "$LOGS"
 
-# install firefox (available everywhere in Fedora and RHEL)
-# we don't need the H.264 codec, and it is sometimes not available (rhbz#2005760)
-dnf install --disablerepo=fedora-cisco-openh264 -y firefox
-
 #HACK: unbreak rhel-9-0's default choice of 999999999 rounds, see https://bugzilla.redhat.com/show_bug.cgi?id=1993919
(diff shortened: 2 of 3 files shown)


martinpitt:
First result: [qemu-img info works on F36](https://artifacts.dev.testing-farm.io/ab85d18f-0f81-4147-b473-db7ef5adb819/). virt-install fails, but it [does the same thing on F35](https://artifacts.dev.testing-farm.io/26c982c8-6feb-4402-9e26-cb346f9d3320/). I added `--connect qemu:///system`, let's see.

martinpitt:
:thinking:  [this works](https://artifacts.dev.testing-farm.io/43e20dc6-a896-427d-b3af-ca2ade3adf17/) -- WTH?

martinpitt:
In https://github.com/cockpit-project/bots/pull/3060 this started to happen on our fedora-36 image as well, so it's easier to investigate now.

martinpitt:
I'm investigating this on our latest fedora-36 image now, as that now inherited the bug. Just like in packit, directly running this after a fresh boot works fine:
```
virt-install --name test1 --os-variant fedora28 --memory 128 --check path_in_use=off --wait -1 --noautoconsole --disk none --cdrom https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso
```
With `vm-run --no-network fedora-36` and adding `127.0.0.1 archive.fedoraproject.org` to /etc/hosts (what the tests do) I get that error:
```
ERROR    internal error: process exited while connecting to monitor: 2022-03-16T09:50:12.997018Z qemu-system-x86_64: -blockdev {"driver":"https","url":"https://archive.fedoraproject.org:443/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso","node-name":"libvirt-1-storage","auto-read-only":true,"discard":"unmap"}: CURL: Error opening file: 
```

which now also explains why this fails on the TF machines -- it will not actually try to talk to the network any more.

So I think we'll need to up our game with mocking the boot.iso file, and put it on an actual web server.

martinpitt:
curl got fixed in https://bodhi.fedoraproject.org/updates/FEDORA-2022-5037545546 and our tests are happy with that.

PR #625, opened 2022-03-17
if volume.format is iso change device to cdrom and bus to sata.

gaionim:
Changes seems simple and not harmful to me.
When I put some iso in my pool i like use them like cdrom, not disks ;-)

NB
I'm using old version 239 in debian 11 and changes seems good enogh to my purposes.


gaionim:
My wrong. Changed branch, but not rebuild, so still testing my old code.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -441,8 +441,15 @@ export class AddDiskModalBody extends React.Component {
         stateDelta.format = getDefaultVolumeFormat(pool);
         if (['dir', 'fs', 'netfs', 'gluster', 'vstorage'].indexOf(pool.type) > -1) {
             const volume = pool.volumes.find(vol => vol.name === value);
-            if (volume && volume.format)
+            if (volume && volume.format) {
                 stateDelta.format = volume.format;
+                if (volume.format === "iso") {
```
KKoukiou, line 446: I agree with the change, I just the a similar check already present. https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/disks/diskAdd.jsx#L506

The difference is that currently we check th filename suffix to be 'iso', but you check the volume.format attribute from the xml.

KKoukiou:
> With new version from bullseye-backports upstream is already working fine, so this pull is useless.

oh I guess then the fix I pointed above was include with the update! Thanks for taking a look how to contribute to, heavily appreciated. 
In this case feel free to close the pr.

src/components/vm/disks/diskAdd.jsx
gaionim: the other difference is that I change disk type too, not only bus type.

KKoukiou: Since your check is actually more robust, maybe it worths removing the other condition I pointed you to and just keepiing only your special case handling.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -441,8 +441,15 @@ export class AddDiskModalBody extends React.Component {
...
+                    this.onValueChanged("busType", "sata");
+                    this.onValueChanged("device", "cdrom");
+                } else {
+                    this.onValueChanged("device", "disk");
+                }
```
KKoukiou, line 451: Calling the onValueChanged inside this function, is suboptimal, as this function is just supposed to return a stateDelta as the name implies. So I actually will approve this, and hopefully we can get into cleaning up this file in the future, but rewriting this to a functional component, and use 'useEffect' hook to handle dependent state updates.

@skobyda are you fine with that?

gaionim: I agree, but I need more time to get into the "big picture" of this nice project.

KKoukiou: thanks for taking the time to contribute! Will leave it up to @skobyda to run some manual testing on this.

skobyda: I think this.onValueChanged needs to stay here, because onValueChanged has some additional logic when changing busType and device (e.g. for changing busType, it will regenerate target, or when changing device, it will also regenerate busType).

src/components/vm/disks/diskAdd.jsx
```diff
@@ -441,8 +441,15 @@ export class AddDiskModalBody extends React.Component {
         stateDelta.format = getDefaultVolumeFormat(pool);
         if (['dir', 'fs', 'netfs', 'gluster', 'vstorage'].indexOf(pool.type) > -1) {
             const volume = pool.volumes.find(vol => vol.name === value);
-            if (volume && volume.format)
+            if (volume && volume.format) {
                 stateDelta.format = volume.format;
+                if (volume.format === "iso") {
+                    this.onValueChanged("busType", "sata");
```
skobyda, line 447: I personally think that "scsi" should be default. Of course, both sata and scsi are valid bus types.
We already have existing code which sets "scsi" for iso files supplied thru "Custom path" option, and I think the same logic should be used for ISO storage volums.
An just reading Red Hat and IBM articles, it seems to me that scsi is more common configuration for cdrom virtual disks.

skobyda: So calling simple `this.onValueChanged("device", "cdrom");` will do the job and set default bus type to scsi. You can still of course change in the UI.

src/components/vm/disks/diskAdd.jsx
skobyda: I think simple this.onValueChanged("device", "cdrom"); is enough. As I described in the [comment](https://github.com/cockpit-project/cockpit-machines/pull/625/commits/1c0ba9f601acca50f7e34ea35b86732e1e525465#r845996000), we should use scsi busType instead. Which is defaultly set [here](https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/disks/diskAdd.jsx#L514) behind it.

src/components/vm/disks/diskAdd.jsx
gaionim: agree

skobyda:
@gaionim I'm right now writing tests fo this PR, and I'm actually having a bit trouble creating ISO volume in our tests. Can you tell me how is your ISO volume configured? (e.g. how do you attach it to storage volume or what is the XML if this storage volume if you run `$virsh vol-dumpxml volName --pool poolName `

gaionim:
Hi,
I simply has a pool  (directory named utility ) and copyed some iso, as [virtio-win.iso](https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/stable-virtio/virtio-win.iso)  
than from cockpit
 - add disk
 - source -> use existing 
 - pool -> select pool utility
 - volume -> select my iso  
 - ADD
 
Check new disk is cdrom

gaionim:
Not sure it is useful ... but here is the dump
```
virsh vol-dumpxml virtio-win.iso --pool Utility
<volume type='file'>
  <name>virtio-win.iso</name>
  <key>/mnt/vm/disks/utility/virtio-win.iso</key>
  <capacity unit='bytes'>541001728</capacity>
  <allocation unit='bytes'>541536256</allocation>
  <physical unit='bytes'>541001728</physical>
  <target>
    <path>/mnt/vm/disks/utility/virtio-win.iso</path>
    <format type='iso'/>
    <permissions>
      <mode>0644</mode>
      <owner>64055</owner>
      <group>64055</group>
    </permissions>
    <timestamps>
      <atime>1649347132.203857953</atime>
      <mtime>1642052381</mtime>
      <ctime>1647341128.992404632</ctime>
      <btime>0</btime>
    </timestamps>
  </target>
</volume>
```

test/check-machines-disks
```diff
@@ -694,6 +698,7 @@ class TestMachinesDisks(VirtualMachinesCase):
         m.execute(f"virsh pool-define-as myPoolOne --type dir --target {v1} && virsh pool-start myPoolOne")
         m.execute(f"virsh pool-define-as myPoolTwo --type dir --target {v2} && virsh pool-start myPoolTwo")
 
+        m.upload([os.path.join(BOTS_DIR, "machine/cloud-init.iso")], os.path.join(default_tmp, "defaultVol.iso"))
```
martinpitt, line 701: :+1:

martinpitt (approved):
Cheers!

PR #627, opened 2022-03-18
Various test fixes, drop special cases for ubuntu-2004, rhel-8-{4,5}

martinpitt:
Split out from #583, as that one is still blocked a bit.


martinpitt:
I'm :+1 with this, but I can't formally approve as I created the PR (even though the commits are from @KKoukiou )

marusak (approved):

PR #628, opened 2022-03-18
Debian-testing now uses cpu host-passthrough by default on qemu x86

skobyda:
Should fix 2 failures at debian-testing image-refresh: https://github.com/cockpit-project/bots/pull/3108

### DIFF
--- a/test/check-machines-settings
+++ b/test/check-machines-settings
@@ -35,7 +35,7 @@ from testlib import nondestructive, wait, test_main  # noqa
 # virt-install changed the default to "host-passthrough"
 # https://github.com/virt-manager/virt-manager/commit/2c477f330244e04614e174f50fbf37260c535705
 distrosWithDefaultHostModel = ["fedora-34", "fedora-35", "rhel-8-4", "rhel-8-5", "rhel-8-6", "rhel-9-0",
-                               "ubuntu-stable", "ubuntu-2004", "debian-testing", "debian-stable", "centos-8-stream", "arch"]
+                               "ubuntu-stable", "ubuntu-2004", "debian-stable", "centos-8-stream", "arch"]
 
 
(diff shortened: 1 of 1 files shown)


test/check-machines-settings
```diff
@@ -406,6 +406,7 @@ class TestMachinesSettings(VirtualMachinesCase):
 
         # Change vCPUs setting
         b.click("#vm-subVmTest1-vcpus-count button")  # Open dialog
+        b.wait_visible("#machines-vcpu-modal-dialog")
```
martinpitt, line 409: How does that fix a race? It's a  good check on its own (we do expect the dialog), but the following set_input_text() will wait for the vcpu-max-field to appear anyway, so it shouldn't make much difference. Also, the test in this PR fail on the tooltip

martinpitt:
The "3x affected" retries really kill the show, as it's rather impossible to get these two tests to green that often. That said, this PR's changes *do* look good, and will unblock https://github.com/cockpit-project/bots/pull/3108 . I'll cycle this for a bit.

martinpitt:
[this run against the new image](https://logs.cockpit-project.org/logs/pull-628-20220319-061134-5b0dfc7a-debian-testing-bots-3108/log.html#70) fixes testMultipleSettings, but testCPUModel still fails. It tries to operate a dialog which doesn't exist, that sounds like an old problem. Retrying to compare, as it *seems* to be unrelated.

martinpitt:
Nice, [debian-testing@bots#3108](https://logs.cockpit-project.org/logs/pull-628-20220319-064730-5b0dfc7a-debian-testing-bots-3108/log.html) went green! So let's land these two in lockstep, and look at the flakes later.

martinpitt (approved):

PR #629, opened 2022-03-21
test: add a check which is deleting a VM and keeping its volume

yunmingyang:


test/check-machines-lifecycle
```diff
@@ -420,6 +420,16 @@ class TestMachinesLifecycle(VirtualMachinesCase):
...
 
+        # Try to delete a VM and keep its volume
+        name = "vm-keep-vol"
+        args = self.createVm(name)
+        self.waitVmRow(name)
+        self.performAction(name, "delete")
+        b.set_checked("#vm-{}-delete-modal-dialog input[type=checkbox]".format(name), False)
+        b.click("#vm-{}-delete-modal-dialog button:contains(Delete)".format(name))
+        self.waitVmRow(name, present=False)
+        wait(lambda: m.execute("test -f {} && echo $?".format(args["image"])))
```
skobyda, line 431: Thanks, this looks good. One comment is that we are trying to phase out "{}".format(var) for a f"{var}", which is not used almost everywhere in this file. 
```suggestion
        # Try to delete a VM and keep its volume
        name = "vm-keep-vol"
        args = self.createVm(name)
        self.waitVmRow(name)
        self.performAction(name, "delete")
        b.set_checked(f"#vm-{name}-delete-modal-dialog input[type=checkbox]", False)
        b.click("#vm-{name}-delete-modal-dialog button:contains(Delete)")
        self.waitVmRow(name, present=False)
        wait(lambda: m.execute(f"test -f {args["image"]} && echo $?"))
```

test/check-machines-lifecycle
```diff
@@ -420,6 +420,16 @@ class TestMachinesLifecycle(VirtualMachinesCase):
...
 
+        # Try to delete a VM and keep its volume
+        name = "vm-keep-vol"
+        args = self.createVm(name)
+        self.waitVmRow(name)
+        self.performAction(name, "delete")
+        b.set_checked(f"#vm-{name}-delete-modal-dialog input[type=checkbox]", False)
+        b.click(f"#vm-{name}-delete-modal-dialog button:contains(Delete)")
+        self.waitVmRow(name, present=False)
+        wait(lambda: m.execute(f"test -f {args['image']} && echo $?"))
```
martinpitt, line 431: I don't understand this condition. We want to assert that the file is still present even after the domain got removed and undefined, right? *Waiting* for the file to exist is then the wrong direction -- it will already exist when creating the VM, so the worst thing that can happen is that it will be removed. To be extra sure about that, I'd rather recommend to use `checkEnvIsEmpty()`, or waiting until `virsh list --all` is empty.

Also, `test && echo $?` does not achieve anything, it will *always* succeed. If the file exists, it will succeed and print "0", and if the file does not exist, the command will succeed (`false && true` is true in POSIX shell!) and not print anything except perhaps a line break. It may *accidentally* work if the m.execute() really returns "" as opposed to `\n`, as the wait() considers the empty string as false-y.

But pelase fix this to

    m.execute(f"test -f {args['image']}")

which is much cleaner and avoids making funny assumptions about terminating line breaks.

martinpitt (changes requested):
Thank you!

test/check-machines-lifecycle
yunmingyang: Oh, sure, that's a miss during checking, sorry.

martinpitt:
Thank you! This looks good now. I force-pushed a rebase to your branch, to fix packit tests. This should also take care of the flakes which I fixed in the last two days.

martinpitt (approved):

PR #633, opened 2022-03-23
test: Fix TestMachinesCreate.testCreate{PXE,AndVerifyQemuConf}

martinpitt:
Destroying pxe-guest does not immediatel stop virt-install. Instead,
this interferes with our redefinition and restarting of the domain, so
that sometimes either failed. This resulted in the domain sometimes not
actually having the /tmp/serial.txt console, or already running when
the test tried to start it.

----

This fails [very](https://logs.cockpit-project.org/logs/pull-631-20220322-175241-f9703f15-arch/log.html#65-1) [often](https://logs.cockpit-project.org/logs/pull-630-20220322-131300-6b7ecc3e-fedora-36/log.html#13) like this:
```
sed: can't read /tmp/serial.txt: No such file or directory
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/work/bots/make-checkout-workdir/test/check-machines-create", line 321, in testCreatePXE
    wait(lambda: self.machine.execute(r"sed 's,\x1B\[[0-9;]*[a-zA-Z],,g' /tmp/serial.txt | grep 'Rebooting in 60'"), delay=3)
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 2131, in wait
    val = func()
  File "/work/bots/make-checkout-workdir/test/check-machines-create", line 321, in <lambda>
    wait(lambda: self.machine.execute(r"sed 's,\x1B\[[0-9;]*[a-zA-Z],,g' /tmp/serial.txt | grep 'Rebooting in 60'"), delay=3)
  File "/work/bots/make-checkout-workdir/bots/machine/machine_core/ssh_connection.py", line 368, in execute
    raise subprocess.CalledProcessError(proc.returncode, command, output=output)
subprocess.CalledProcessError: Command 'sed 's,\x1B\[[0-9;]*[a-zA-Z],,g' /tmp/serial.txt | grep 'Rebooting in 60'' returned non-zero exit status 1.
```

i.e. /tmp/serial.txt never gets created. There seems to be a race condition when defining the VM, as the VNC console in the [screenshot](https://logs.cockpit-project.org/logs/pull-631-20220322-175241-f9703f15-arch/TestMachinesCreate-testCreatePXE-arch-127.0.0.2-2401-FAIL.png)  shows the successful PXE boot just fine.

I can't reproduce this locally, so debugging on the infra.

 - Builds on top of PR #631

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1352,14 +1352,11 @@ class TestMachinesCreate(VirtualMachinesCase):
             self.machine.execute("until test -z $(ls /home/admin/.local/share/libvirt/images/ 2>/dev/null); do sleep 1; done")
 
             # When we delete a VM while virt-install is running there will be an error
-            if dialog and dialog.start_vm and dialog.sourceType in ["url", "os"]:
-                b.wait_visible(".pf-c-alert-group")
-
             alerts = self.browser.call_js_func("ph_count", ".pf-c-alert-group li")
             for i in range(alerts):
                 b.click(".pf-c-alert-group li:first-of-type .pf-c-alert button.pf-m-plain")
 
-            b.wait_not_present(".pf-c-alert-group li .pf-c-alert")
+            b.wait_not_present(".pf-c-alert-group")
 
             return self
 
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -349,13 +349,17 @@ class TestMachinesCreate(VirtualMachinesCase):
         config = TestMachinesCreate.TestCreateConfig
         m = self.machine
 
+        self.restore_file("/etc/libvirt/qemu.conf")
+
         self.login_and_go("/machines")
         self.browser.wait_in_text("body", "Virtual machines")
 
         vnc_listen = "192.168.2.6"
-        m.execute(f"echo 'vnc_listen = \"{vnc_listen}\"' >> /etc/libvirt/qemu.conf")
         vnc_passwd = "mypass"
-        m.execute(f"echo 'vnc_password= \"{vnc_passwd}\"' >> /etc/libvirt/qemu.conf")
+
+        m.write("/etc/libvirt/qemu.conf", f'''vnc_listen = "{vnc_listen}"
+vnc_password= "{vnc_passwd}"
+''', append=True)
 
         runner.createAndVerifyQemuConfParsedTest(
             TestMachinesCreate.VmDialog(self, sourceType='file',
@@ -364,7 +368,6 @@ class TestMachinesCreate(VirtualMachinesCase):
             vnc_listen, "127.0.0.1", vnc_passwd, None)
 
         # Ensure that missing qemu.conf would not crash the script but just pick the 127.0.0.1 default value
-        self.restore_file("/etc/libvirt/qemu.conf")
         m.execute("rm /etc/libvirt/qemu.conf")
         runner.createAndVerifyQemuConfParsedTest(
             TestMachinesCreate.VmDialog(self, sourceType='file',
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -857,6 +857,7 @@ class TestMachinesCreate(VirtualMachinesCase):
         def createRespectQemuConfConsoleConfig(self, vnc_listen, spice_listen, vnc_passwd, spice_passwd):
             self.browser.click(".pf-c-modal-box__footer button:contains(Create)")
             self.browser.wait_not_present("#create-vm-dialog")
+            self.browser.wait_text(f"#vm-{self.name}-state", "Shut off")
 
             domainXML = self.machine.execute(f"virsh dumpxml --security-info {self.name}")
             root = ET.fromstring(domainXML)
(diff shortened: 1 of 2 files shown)


martinpitt:
[one hit](https://logs.cockpit-project.org/logs/pull-633-20220323-060023-53751841-fedora-35-firefox/log.html#12), with the [relevant log](https://github.com/cockpit-project/cockpit-machines/files/8330257/log.txt). This confirms that pxe-guest does *not* have the serial console defined. It also still has all the default `<console>` objects, so the XML manipulation did not work, or the re-definition of the domain silently failed. I'll add some more debugging to show the generated /tmp/domain.xml, to see which one it is.

martinpitt:
Interesting -- now there are several failures [like this](https://logs.cockpit-project.org/logs/pull-633-20220323-064845-409ea6b4-rhel-9-0/log.html#17), where the dumpxml from the regenerated domain is correct, but trying to start it fails:
```
+ virsh start pxe-guest
error: Domain is already active

  File "/work/bots/make-checkout-workdir/test/check-machines-create", line 329, in testCreatePXE
    self.machine.execute("virsh start pxe-guest")
```
I haven't seen this before, so I find it relatively unlikely that it happened before (as the original failure happens later on). So somehow undefining the domain triggered a (re)start of the domain. I [added a commit](https://github.com/cockpit-project/cockpit-machines/pull/633/commits/1d2ffc6201634f605de232ff61680bba88a0db70) which waits for the `destroy` to actually happen -- it might be async?

martinpitt:
Bazinga! [all passed](https://github.com/cockpit-project/cockpit-machines/commit/9205bb52db3bb3318f2be13ca07d74a788137887) on that commit. I'll now push a cleaned-up version of this.

martinpitt:
All good in the previous run, except two instances of [testCreateAndVerifyQemuConf](https://logs.cockpit-project.org/logs/pull-633-20220323-095551-43cedf87-centos-8-stream/log.html#64) [failure](https://logs.cockpit-project.org/logs/pull-633-20220323-095551-43cedf87-debian-testing/log.html#64-1). That's unrelated -- I'll look at this, but not as part of this PR.

martinpitt:
That [testCreateAndVerifyQemuConf](https://logs.cockpit-project.org/logs/pull-633-20220323-103314-d765fc14-centos-8-stream/log.html#59) test keeps failing, and the reason is so obvious that I'll just fix it here right away after all.

test/check-machines-create
```diff
@@ -313,7 +313,8 @@ class TestMachinesCreate(VirtualMachinesCase):
         # Redefine the domain with the new XML
         xmlstr = ET.tostring(root, encoding='unicode', method='xml')
 
-        self.machine.execute(f"echo '{xmlstr}' > /tmp/domain.xml && virsh define --file /tmp/domain.xml")
+        self.write_file("/tmp/domain.xml", xmlstr)
+        self.machine.execute("virsh define --file /tmp/domain.xml")
```
skobyda, line 320: So the problem here, if I understand it correctly was that virsh define was reading the old file domain.xml?

test/check-machines-create
```diff
@@ -275,9 +275,12 @@ class TestMachinesCreate(VirtualMachinesCase):
         # We don't want to use start_vm == False because if we get a separate install phase
         # virt-install will overwrite our changes.
 
-        # Wait for virt-install to define the VM and then stop it
-        wait(lambda: "pxe-guest" in self.machine.execute("virsh list --persistent"), delay=3)
+        # Wait for virt-install to define the VM, stop it, and wait for virt-install to be done
+        wait(lambda: "pxe-guest" in self.machine.execute("virsh list --name --persistent"), delay=3)
+        wait(lambda: "pxe-guest" in self.machine.execute("virsh list --name"), delay=3)
         self.machine.execute("virsh destroy pxe-guest")
+        self.machine.execute("while pgrep virt-install; do sleep 1; done")
```
skobyda, line 282: Is this correct. Here we are waiting until virt-install to finish and process to cease existing, right?
But according to manpages, pgrep return 0 when process exists, and 1 when it doesn't exist, so the while condition should be negated, no?
```
EXIT STATUS
       0      One or more processes matched the criteria. For pkill and pidwait, one or more processes must also have been successfully signalled or waited for.
       1      No processes matched or none of them could be signalled.
```

test/check-machines-create
martinpitt: No, /tmp/domain.xml has always been okay. But the running virt-install in the background and this `virsh define` collided, and sometimes virt-install won -- hence this `virsh define` was a no-op.

test/check-machines-create
martinpitt: Right, the point of that is to wait while virt-install is running -- i.e. wait until it is *not* running any more. So this works as intended.

martinpitt: No, it's already running and it works until it stops running. Precisely, while it is running (pgrep finds a process) it sleeps.

PR #635, opened 2022-03-23
fmf: Trim tests to only exercise external APIs and be faster

martinpitt:
packit tests currently take ~ 80 minutes (SRPM+RPM builds, plus the
actual tests over an hour), which is too long for our "result after max.
1 hour" goal. Many of them are also not that useful in downstream
gating, as they only exercise internal functionality -- it's enough to
do that in upstream CI where we have a lot more performance (due to
parallelization).

Exclude some tests which don't exercise external APIs, and are thus not
much useful for reverse dependency gating.

Lower the maximum test duration to 50 minutes, so that we will notice
when tests are piling up too much again. That leaves ten minutes for the
webpack+srpm+rpm builds.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -44,9 +44,42 @@ if [ "$ID" = "rhel" ]; then
     "
 fi
 
-if [ "$TEST_OS" = "centos-8-stream" ]; then
-    EXCLUDES="$EXCLUDES TestMachinesConsoles.testExternalConsole"
-fi
+# We only have one VM and tests should take at most one hour. So run those tests which exercise external API
+# (and thus are useful for reverse dependency testing and gating), and exclude those which test internal
+# functionality -- upstream CI covers that.
+EXCLUDES="$EXCLUDES
+          TestMachinesCreate.testConfigureBeforeInstall
+          TestMachinesCreate.testCreateBasicValidation
+          TestMachinesCreate.testDisabledCreate
+
+          TestMachinesConsoles.testExternalConsole
+          TestMachinesConsoles.testInlineConsole
+          TestMachinesConsoles.testInlineConsoleWithUrlRoot
+          TestMachinesConsoles.testSerialConsole
+          TestMachinesConsoles.testSwitchConsoleFromSerialToGraphical
+
+          TestMachinesDisks.testAddDiskAdditionalOptions
+          TestMachinesDisks.testAddDiskCustomPath
+          TestMachinesDisks.testDetachDisk
+          TestMachinesDisks.testDiskEdit
+
+          TestMachinesLifecycle.testBasicLibvirtUserUnprivileged
+          TestMachinesLifecycle.testBasicWheelUserUnprivileged
+          TestMachinesLifecycle.testDelete
+          TestMachinesLifecycle.testLibvirt
+
+          TestMachinesHostDevs.testHostDevAddMultipleDevices
+
+          TestMachinesMigration.testFailMigrationUriIncorrect
+          TestMachinesMigration.testFailMigrationDomainUnknown
+
+          TestMachinesNetworks.testNetworkSettings
+          TestMachinesNetworks.testNICPlugingAndUnpluging
+
+          TestMachinesNICs.NICAddDialog
+
+          TestMachinesSettings.testMultipleSettings
+          "
 
 if [ "$ID" = "fedora" ]; then
     # Testing Farm machines are really slow in European evenings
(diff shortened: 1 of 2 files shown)


test/browser/run-test.sh
```diff
@@ -44,9 +44,31 @@ if [ "$ID" = "rhel" ]; then
...
+          TestMachinesDisks.testDiskEdit
+
+          TestMachinesLifecycle.testBasicLibvirtUserUnprivileged
+          TestMachinesLifecycle.testBasicWheelUserUnprivileged
+          TestMachinesLifecycle.testDelete
+          TestMachinesLifecycle.testLibvirt
+
+          TestMachinesNetworks.testNetworkSettings
+          TestMachinesNetworks.testNetworkState
+          "
```
skobyda, line 72: I would recommend adding also
`TestMachinesHostDevs.testHostDevAddMultipleDevices` as it is using the same API as `TestMachinesHostDevs.testHostDevAddSingleDevice`. `testHostDevAddSessionConnection` again using the same API, only with different connection parameter: system/session

test/browser/run-test.sh
```diff
@@ -44,9 +44,31 @@ if [ "$ID" = "rhel" ]; then
...
+# We only have one VM and tests should take at most one hour. So run those tests which exercise external API
+# (and thus are useful for reverse dependency testing and gating), and exclude those which test internal
+# functionality -- upstream CI covers that.
+EXCLUDES="$EXCLUDES
+          TestMachinesCreate.testDisabledCreate
+
+          TestMachinesConsoles.testExternalConsole
+          TestMachinesConsoles.testInlineConsole
+          TestMachinesConsoles.testInlineConsoleWithUrlRoot
+          TestMachinesConsoles.testSerialConsole
```
skobyda, line 56: While for other types of consoles, we are using a patternfly components, serial console is from cockpit's own `cockpit-components-terminal.jsx`, so technically it's a bit different. So maybe that one should be tested? Or is it enough that  `cockpit-components-terminal.jsx`is tested with packit tests in cockpit repo?

test/browser/run-test.sh
martinpitt: Thanks, I'll exclude `testHostDevAddMultipleDevices`. The session one feels more prone to regressions, as there's a lot of device/service permissions in between. We can cut that once we get over the 50 minute line, but until then I'd rather leave it running.

test/browser/run-test.sh
martinpitt: It seems enough to run that upstream. JSX components don't change with e.g. a newer kernel or libvirt version in Fedora/RHEL, this is all tightly under our upstream control. So this seems safe to skip.

test/browser/run-test.sh
```diff
@@ -44,9 +44,31 @@ if [ "$ID" = "rhel" ]; then
...
+          TestMachinesDisks.testDetachDisk
+          TestMachinesDisks.testDiskEdit
+
+          TestMachinesLifecycle.testBasicLibvirtUserUnprivileged
+          TestMachinesLifecycle.testBasicWheelUserUnprivileged
+          TestMachinesLifecycle.testDelete
+          TestMachinesLifecycle.testLibvirt
+
+          TestMachinesNetworks.testNetworkSettings
+          TestMachinesNetworks.testNetworkState
```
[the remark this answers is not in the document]
martinpitt: Thanks! Added back. I'm grateful that you reviewed this so carefully, as I don't have a deep knowledge about our test coverage yet.

skobyda:
If it would be still way too long, there is a lot of subtests. For example `testNetworksCreate` contains a dozen `NetworkCreateDialog` subtests, but they are using the same API, only testing various combinations of network configuration, so we could skip all subtests but one. The same can be applied to tests like testVmCreate, storagePoolCreate, storageVolumeCreate... But I think that would require some amendments in the code of tests itself, so let's see if that is necessary.

martinpitt:
@skobyda , no, it's not. The previous round was already 38 minutes only, this run should be even shorter. That's good enough for now.

martinpitt:
TestMachinesCreate.testCreateThenInstall [failed in f36/packit](https://artifacts.dev.testing-farm.io/6fdfeac4-79e0-4122-a142-893d6c5a2dee/). Re-running to check if it's a new regression or a flake. [f35](https://artifacts.dev.testing-farm.io/45cc04bd-b669-41e7-ae81-05a624e182e5/work-allsHH0Jb/log.txt) ran in 41 minutes - that's more than I expected, but still under the limit.

martinpitt:
Now `TestMachinesCreate.testCreateThenInstall` failed in [c9s](https://artifacts.dev.testing-farm.io/237d681f-3da2-4c04-85d6-d95f597efa93/). That's apparently the next flake we need to look at. Still, I think this is good enough to land.

skobyda (approved):

PR #636, opened 2022-03-24
Reword shared directory xattr checkbox

### LINKED ISSUE #347: machines: There should be no "enable/disable" in xattr text
Version: 
cockpit-250-1.fc34.x86_64

Steps:
1. Prepare a shutoff VM
2. Enter the VM detail page
3. Click "Add shared directory", then click "Extend additional options"
4. Check the text of "Extended attributes", if checking the checkbox of "Extended attributes", I think it means enable "extended attributes (xattr) on files and directories". So it is a little duplicate to add "enable/disable" in "Enable/disable extended attributes (xattr) on files and directories"

jelly:
Remove the Enable/disable text from the shared directory dialog and
clear up the duplicate "Extended Attributes" in the checkbox label and
from label.

Fixes #347

### DIFF
--- a/src/components/vm/filesystems/vmFilesystemsCard.jsx
+++ b/src/components/vm/filesystems/vmFilesystemsCard.jsx
@@ -238,10 +238,10 @@ const VmFilesystemAddModal = ({ connectionName, setIsOpen, objPath, vmName, vmSt
                                    isExpanded={additionalOptionsExpanded}>
                     <FormGroup hasNoPaddingTop
                                fieldId={`${idPrefix}-modal-xattr`}
-                               label="Extended attributes">
+                               label={_("Extended attributes")}>
                         <Checkbox id={`${idPrefix}-modal-xattr`}
                                   isChecked={xattr}
-                                  label={_("Enable/disable extended attributes (xattr) on files and directories")}
+                                  label={_("Use extended attributes on files and directories")}
                                   onChange={xattr => setXattr(xattr)} />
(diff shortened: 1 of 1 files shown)


src/components/vm/filesystems/vmFilesystemsCard.jsx
```diff
@@ -238,10 +238,10 @@ const VmFilesystemAddModal = ({ connectionName, setIsOpen, objPath, vmName, vmSt
                                    isExpanded={additionalOptionsExpanded}>
                     <FormGroup hasNoPaddingTop
                                fieldId={`${idPrefix}-modal-xattr`}
-                               label="Extended attributes">
+                               label="Options">
```
skobyda, line 241: I think original `Extended attributes` is just fine. There are more options which may be added in the future for shared directories such as `cache mode`, `lock`, so lavelling only Extended attributes as "options" can be confusing.

skobyda (dismissed):

src/components/vm/filesystems/vmFilesystemsCard.jsx
jelly: Ok, I also just noticed the label was not translated so fixed that as well.

KKoukiou (approved):

martinpitt (approved):
@KKoukiou already approved previously, @skobyda 's feedback got addressed, and Jelle only rebased to fix packit.

PR #638, opened 2022-03-25
test: Drop obsolete special cases, enable Ubuntu 22.04

martinpitt:
We don't release to Fedora 34, Ubuntu 20.04, RHEL 8.4/8.5 any more.

Commit 0e5507720e already cleaned up a lot of these, but was missing
some spots.

----

This is an initial PR to get a place to run and fix ubuntu-2204.

### DIFF
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -19,7 +19,7 @@ from testlib import MachineCase
 from netlib import NetworkHelpers
 from storagelib import StorageHelpers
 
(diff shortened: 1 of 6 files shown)


test/machineslib.py
```diff
@@ -1,4 +1,4 @@
-# This file is part of Cockpit.
+# This fila is part of Cockpit.
```
martinpitt, line 1: Err, what? I quickly need to get a cat, to claim that it walked over the keyboard..

marusak (approved):
Thanks! Assuming green tests of course

PR #640, opened 2022-03-28
Removing virtual network's DHCP static host entries

### LINKED ISSUE #617: Can't delete dhcp static host entries
I have an "open" network, and I have static host entries configured for all the machines on the network, but I cant seem to delete them, only add them.

skobyda:
Closes #617  

Screenshot from 2022-03-28 13-28-32

Screenshot from 2022-03-28 13-24-41


garrett (changes requested):
Cool!

1. However, what's "this"? Confirming something unknown isn't great. Show what will be removed.
2. "Confirm this action" is redundant with everything else in the modal.
3. This is a remove action, not delete. Data isn't being erased; it's just removing the definition for the entry. 
   - https://www.patternfly.org/v4/ux-writing/terminology/
   - **Delete**: (v.) Use to describe completely removing an object. Delete is a destructive action.
   - **Remove**: (v.) Use to describe removing an item from a list, group, view, or other container element without completely deleting it.
      - (We're not deleting the machine.)
 
PF guidelines:
https://www.patternfly.org/v4/components/modal/design-guidelines/#confirmation-dialogs

Perhaps something more like this:

vm-dhcp-remove

skobyda:
Updated. I used a `DescriptionList` to present name-value pairs of IP and MAC in the dialog:

Screenshot from 2022-04-04 13-54-41

skobyda:
>     1. However, what's "this"? Confirming something unknown isn't great. Show what will be removed.
> 
>     2. "Confirm this action" is redundant with everything else in the modal.

We break these suggestions in basically most of "Remove {object}" dialogs. I think all of those dialogs need a redesign. I'm currently working on a separate PR fir a general overhaul of all "Remove {object}" dialogs. I will use some commits from this PR for that but I don't think it should block this PR. Whatever PR gets merge first will bring those commit to the `main`

Fex examples of dialogs which are also incorrect:
Screenshot from 2022-04-04 13-59-41

garrett:
> Updated. I used a DescriptionList to present name-value pairs of IP and MAC in the dialog:

Much better! Thanks!

> We break these suggestions in basically most of "Remove {object}" dialogs.

OK, we definitely don't do that in Cockpit in any part I've seen. These dialogs must've "flown under the radar" and have been merged in without design approval.

> I think all of those dialogs need a redesign. I'm currently working on a separate PR for a general overhaul of all "Remove {object}" dialogs.

Yes! Definitely. Thanks for working on that!

skobyda:
Hey @martinpitt @jelly Mind giving it a code review?

src/machines.scss
```diff
@@ -77,3 +80,7 @@
 .ct-external-docs-link.pf-c-button.pf-m-link.pf-m-inline {
   padding-left: var(--pf-global--spacer--sm);
 }
+
+.ct-monospace {
+    font-family: var(--pf-global--FontFamily--redhat--monospace);
+}
```
garrett, line 86: With a `.ct-`* level, we should probably have this elsewhere in Cockpit, probably under `lib/`* which gets re-imported to Machines.

And, then, eventually, when PatternFly has their own `.pf-` to do this, we switch to that instead.

But, for this PR, I think this is OK and that can be handled in a follow-up.

What do others here think?

garrett (approved):
:+1: We might want to move `ct-monospace` elsewhere, but it might be OK for this PR.

(I think we're currently setting monospace manually in most places and should probably change those too, where it's just a one-off like this, instead of being part of the overall CSS for a widget.)

src/components/common/deleteResource.jsx
```diff
@@ -70,7 +75,15 @@ export class DeleteResourceModal extends React.Component {
                            </Button>
                        </>
                    }>
-                { actionDescription || cockpit.format(_("Confirm this action")) }
+                <DescriptionList isHorizontal>
+                    {actionDescription || cockpit.format(_("Confirm this action"))}
```
martinpitt, line 79: The designs looked like the "Confirm this action" string should disappear, and be replaced with something like "These ... will be removed:"?

test/check-machines-networks
```diff
@@ -795,6 +795,72 @@ class TestMachinesNetworks(VirtualMachinesCase):
...
+                else:
+                    b.wait_not_present("#add-new-static-entry")
+
+            def verify_dialog(self):
+                # Check that the defined network is now visible
+                b.wait_in_text("body", "Networks")
+
+                # Verify libvirt XML
+                net_xml = f"virsh -c qemu:///system net-dumpxml {self.network_name}"
+                xmllint_element = f"{net_xml} | xmllint --xpath 'string(//network/ip/dhcp/host/{{prop}})' - 2>&1 || true"
```
martinpitt, line 851: Why would this fail legitimately? I really don't like alternative truths in tests, they should be predictable. Also, it's usually better to keep stderr as-is, so that errors appear in the log instead of being swallowed into the string comparison. Can you please drop this? If there is a case where the XML does not have a dhcp/host/prop path (the plausible way to fail this), it's fine if the test fails here already?

src/components/common/deleteResource.jsx
```diff
@@ -114,7 +116,9 @@ export const DeleteResourceButton = ({ objectId, disabled, overlayText, actionNa
     } else {
         return (
             <Button id={`delete-${objectId}`}
-                variant='danger'
+                variant={variant || 'danger'}
+                isDanger={!!variant}
```
martinpitt, line 122: Shouldn't that be `{variant === 'danger'}`? (Also, why is that not the default in PF? Specifying both `variant="danger"` and `isDanger` seems redundant)

src/components/common/deleteResource.jsx
```diff
@@ -97,14 +97,16 @@ DeleteResourceModal.propTypes = {
...
     if (disabled) {
         return (
             <Tooltip id={`delete-${objectId}-tooltip`}
                      content={overlayText}>
                 <span>
                     <Button id={`delete-${objectId}`}
-                        variant='danger'
+                        variant={variant || 'danger'}
+                        isDanger={!!variant}
+                        isInline={isInline}
```
martinpitt, line 110: DRY. Please create the `<Button>` once with `isDisabled={disabled}`, and wrap it into the `<Tooltip>` conditionally.

src/machines.scss
martinpitt: Agreed -- this should also become a cockpit PR, and then at some point c-m can bump the pkg/lib version and drop its own copy. But let's not block on that indeed.

src/components/networks/networkOverviewTab.jsx
```diff
@@ -53,7 +56,47 @@ const DHCPHost = (host, index, family, idPrefix) => {
...
+                objectName: "",
+                actionDescription: _(`The static host entry for ${host.ip} will be removed:`),
+                objectDescription: [
+                    { name: _("IP"), value: <span className="ct-monospace">{host.ip}</span> },
+                    { name: _("MAC"), value: <span className="ct-monospace">{host.mac}</span> }
+                ],
+                actionName: _("Remove"),
+                onClose: () => setDeleteDialogProps(undefined),
+                deleteHandler: () => {
+                    return networkRemoveStaticHostEntries({
```
martinpitt, line 75: Please drop the `{ return ... }` wrapper, and keep this functional (`() => networkRemoveStaticHostEntries({...`)

src/components/networks/networkOverviewTab.jsx
```diff
@@ -53,7 +56,47 @@ const DHCPHost = (host, index, family, idPrefix) => {
...
-    return <ListItem key={index} id={id}> {hostInfo} </ListItem>;
+    const removeDHCPHostButton = (
+        <DeleteResourceButton objectId={`${id}-button`}
+            actionName={_("remove")}
+            variant="link"
+            isInline
+            showDialog={() => setDeleteDialogProps({
+                objectType: _("static host from DHCP"),
+                objectName: "",
+                actionDescription: _(`The static host entry for ${host.ip} will be removed:`),
```
martinpitt, line 67: This does not work for i18n, as the gettext call happens after the string mutation. You need to use cockpit.format() for this.

    git grep '_(`'

This ^ shows two other instances of this, can you please fix them in a separate commit or PR?

martinpitt (changes requested):
Thanks! Some small issues which should be quick to fix.

src/components/common/deleteResource.jsx
martinpitt: Or, I suppose that's just a fallback, and this will be cleaned up in #652?

src/components/networks/networkOverviewTab.jsx
martinpitt: I filed PR #655 to fix and spot that.

test/check-machines-networks
skobyda: Yes, it's totally normal that virtual network doesn't have whole <dhcp> elemenent subtree, and even if it has, it could also not have <dhcp> subtree. 
If in the future, ours (or libvirt's) code for adding DHCP hosts regresses in a way that dhcp hosts it would not be present in the XML, then with ` - 2>&1 || true` the error message would look like this:
```
+ virsh -c qemu:///system net-dumpxml default | xmllint --xpath 'string(//network/ip/dhcp/host/@mac)' - 2>&1 || true

  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 887, in testNetworkAddStaticDCHPHosts
    ).execute()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 817, in execute
    self.verify_dialog()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 853, in verify_dialog
    self.test_obj.assertEqual(self.mac, m.execute(xmllint_element.format(prop='@mac')).strip())
  File "/usr/lib64/python3.10/unittest/case.py", line 845, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib64/python3.10/unittest/case.py", line 1226, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/usr/lib64/python3.10/unittest/case.py", line 675, in fail
    raise self.failureException(msg)
AssertionError: '52:54:00:6b:70:af' != ''
- 52:54:00:6b:70:af
+ 

```

Without `- 2>&1 || true` the error message would look like this:

```
+ virsh -c qemu:///system net-dumpxml default | xmllint --xpath 'string(//network/ip/dhcp/host/@mac)'
Usage : xmllint [options] XMLfiles ...
	Parse the XML files and output the result of the parsing
	--version : display the version of the XML library used
	--debug : dump a debug tree of the in-memory document
	--shell : run a navigating shell
	--debugent : debug the entities defined in the document
	--copy : used to test the internal copy implementation
	--recover : output what was parsable on broken XML documents
	--huge : remove any internal arbitrary parser limits
	--noent : substitute entity references by their value
	--noenc : ignore any encoding specified inside the document
	--noout : don't output the result tree
	--path 'paths': provide a set of paths for resources
	--load-trace : print trace of all external entities loaded
	--nonet : refuse to fetch DTDs or entities over network
	--nocompact : do not generate compact text nodes
	--htmlout : output results as HTML
	--nowrap : do not put HTML doc wrapper
	--valid : validate the document in addition to std well-formed check
	--postvalid : do a posteriori validation, i.e after parsing
	--dtdvalid URL : do a posteriori validation against a given DTD
	--dtdvalidfpi FPI : same but name the DTD with a Public Identifier
	--quiet : be quiet when succeeded
	--timing : print some timings
	--output file or -o file: save to a given file
	--repeat : repeat 100 times, for timing or profiling
	--insert : ad-hoc test for valid insertions
	--compress : turn on gzip compression of output
	--html : use the HTML parser
	--xmlout : force to use the XML serializer when using --html
	--nodefdtd : do not default HTML doctype
	--push : use the push mode of the parser
	--pushsmall : use the push mode of the parser using tiny increments
	--memory : parse from memory
	--maxmem nbbytes : limits memory allocation to nbbytes bytes
	--nowarning : do not emit warnings from parser/validator
	--noblanks : drop (ignorable?) blanks spaces
	--nocdata : replace cdata section with text nodes
	--format : reformat/reindent the output
	--encode encoding : output in the given encoding
	--dropdtd : remove the DOCTYPE of the input docs
	--pretty STYLE : pretty-print in a particular style
	                 0 Do not pretty print
	                 1 Format the XML content, as --format
	                 2 Add whitespace inside tags, preserving content
	--c14n : save in W3C canonical format v1.0 (with comments)
	--c14n11 : save in W3C canonical format v1.1 (with comments)
	--exc-c14n : save in W3C exclusive canonical format (with comments)
	--nsclean : remove redundant namespace declarations
	--testIO : test user I/O support
	--catalogs : use SGML catalogs from $SGML_CATALOG_FILES
	             otherwise XML Catalogs starting from 
	         file:///etc/xml/catalog are activated by default
	--nocatalogs: deactivate all catalogs
	--auto : generate a small doc on the fly
	--xinclude : do XInclude processing
	--noxincludenode : same but do not generate XInclude nodes
	--nofixup-base-uris : do not fixup xml:base uris
	--loaddtd : fetch external DTD
	--dtdattr : loaddtd + populate the tree with inherited attributes 
	--stream : use the streaming interface to process very large files
	--walker : create a reader and walk though the resulting doc
	--pattern pattern_value : test the pattern support
	--chkregister : verify the node registration code
	--relaxng schema : do RelaxNG validation against the schema
	--schema schema : do validation against the WXS schema
	--schematron schema : do validation against a schematron
	--sax1: use the old SAX1 interfaces for processing
	--sax: do not build a tree but work just at the SAX level
	--oldxml10: use XML-1.0 parsing rules before the 5th edition
	--xpath expr: evaluate the XPath expression, imply --noout

Libxml project home page: https://gitlab.gnome.org/GNOME/libxml2
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 887, in testNetworkAddStaticDCHPHosts
    ).execute()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 817, in execute
    self.verify_dialog()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 853, in verify_dialog
    self.test_obj.assertEqual(self.mac, m.execute(xmllint_element.format(prop='@mac')).strip())
  File "/home/skobyda/Work/machines/bots/machine/machine_core/ssh_connection.py", line 368, in execute
    raise subprocess.CalledProcessError(proc.returncode, command, output=output)
subprocess.CalledProcessError: Command 'virsh -c qemu:///system net-dumpxml default | xmllint --xpath 'string(//network/ip/dhcp/host/@mac)'' returned non-zero exit status 1.
```

skobyda: The `- 2>&1 || true` is used all over our place, and I think it was probably used to get clearer error message from assertions instead of just Command returned non-zero exit. Both error message are don't tell you where is the problem though, so I personally don't mind removing it

martinpitt: Urgh, that's the worst conceivable output from xmllint for a "pattern not found" situation.. Thanks, nevermind then.

src/components/common/deleteResource.jsx
skobyda: I'm trying to work here basically with 2 types of button. Danger button, and Danger link button.

1. `<Button variant="link" isDanger>`:
Screenshot from 2022-04-07 23-56-31

2. `<Button variant="danger">`: 
Screenshot from 2022-04-07 23-56-57

skobyda: But no I see that I programmed it a bit confusingly. Maybe instead of `variant` parameter being passed to `DeleteResourceButton`, I can have parameter `isLink`, and then this button could be:
`<Button variant={isLink ? 'link' : 'danger' isDanger={isLink} />`
That would probably be clearer.

test/check-machines-networks
skobyda: In the end, I didn't really like error messages the both approaches were generating, so I adjusted the code a bit to create this output:
```
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 892, in testNetworkAddStaticDCHPHosts
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 818, in execute
    self.verify_dialog()
  File "/home/skobyda/Work/machines/./test/check-machines-networks", line 859, in verify_dialog
Exception: No DHCP host with IP address 128.132.123.1 found in network XML:
<network>
  <name>default</name>
  <uuid>4c3eb707-770b-4d8e-b000-fabec9cff59e</uuid>
  <forward mode='nat'>
    <nat>
      <port start='1024' end='65535'/>
    </nat>
  </forward>
  <bridge name='virbr0' stp='on' delay='0'/>
  <mac address='52:54:00:0e:e7:7b'/>
  <ip address='192.168.122.1' netmask='255.255.255.0'>
    <dhcp>
      <range start='192.168.122.2' end='192.168.122.254'/>
      <host mac='52:54:00:e6:6b:9b' ip='192.168.122.222'/>
    </dhcp>
  </ip>
</network>
```

martinpitt: Thanks! Indeed that's more explicit.

test/check-machines-networks
```diff
@@ -795,6 +796,75 @@ class TestMachinesNetworks(VirtualMachinesCase):
...
+            def verify_dialog(self):
+                # Check that the defined network is now visible
+                b.wait_in_text("body", "Networks")
+
+                # Verify libvirt XML
+                net_xml = m.execute(f"virsh -c qemu:///system net-dumpxml {self.network_name}")
+                network_node = ET.fromstring(net_xml)
+                host_by_mac = network_node.find(f'./ip/dhcp/host[@mac="{self.mac}"]')
+                if host_by_mac is None:
+                    raise Exception(f'No DHCP host with MAC address {self.mac} found in network XML:\n{net_xml}')
```
martinpitt, line 855: A cleaner way to write this would be `self.assertTrue(host_by_mac, f'...error')`. This would also avoid raising the overly generic Exception base class.

src/components/common/deleteResource.jsx
```diff
@@ -97,14 +97,16 @@ DeleteResourceModal.propTypes = {
...
+export const DeleteResourceButton = ({ objectId, disabled, overlayText, actionName, showDialog, isLink, isInline }) => {
     if (disabled) {
         return (
             <Tooltip id={`delete-${objectId}-tooltip`}
                      content={overlayText}>
                 <span>
                     <Button id={`delete-${objectId}`}
-                        variant='danger'
+                        variant={isLink ? 'link' : 'danger'}
+                        isDanger={isLink}
```
martinpitt: This still looks rather confusing .. but indeed that's just the [API](https://www.patternfly.org/v4/components/button/).

martinpitt (approved):
Thanks! The assertion thing isn't important enough to delay this for another round. Nice work!

martinpitt:
The [rawhide test](https://artifacts.dev.testing-farm.io/ff9cbb16-f8d1-4cf6-8d5b-0a741038ed19/) [failed](https://artifacts.dev.testing-farm.io/ff9cbb16-f8d1-4cf6-8d5b-0a741038ed19/work-allnG5Ias/log.txt):

```
23:17:12                 out:   File "/var/ARTIFACTS/work-allnG5Ias/plans/all/discover/default/tests/test/check-machines-create", line 362, in testCreateAndVerifyQemuConf
23:17:12                 out:     runner.createAndVerifyQemuConfParsedTest(
23:17:12                 out:   File "/var/ARTIFACTS/work-allnG5Ias/plans/all/discover/default/tests/test/check-machines-create", line 1451, in createAndVerifyQemuConfParsedTest
23:17:12                 out:     .createRespectQemuConfConsoleConfig(vnc_listen, spice_listen, vnc_passwd, spice_passwd)
23:17:12                 out:   File "/var/ARTIFACTS/work-allnG5Ias/plans/all/discover/default/tests/test/check-machines-create", line 873, in createRespectQemuConfConsoleConfig
23:17:12                 out:     spice = root.findall(".//graphics[@type='spice']")[0]
23:17:12                 out: IndexError: list index out of range
23:17:12                 out: 
23:17:12                 out: # Result testCreateAndVerifyQemuConf (__main__.TestMachinesCreate) failed
```
This is unrelated, possibly they finally removed spice from libvirt (as it is on its way out). But I'm looking at this anyway, see #622.

PR #641, opened 2022-03-29
Test improvements, fix testCreateThenInstall flake

martinpitt:
I'm trying to debug https://artifacts.dev.testing-farm.io/19531be9-b707-4fec-8b9b-b4795df9fe0b/ , but the test fails in a mostly useless way. Supposedly the domain XML does not have any metadata, but it could also have a different count. Let's make this more useful in the future.

This also fixes this different [testCreateThenInstall failure](https://artifacts.dev.testing-farm.io/237d681f-3da2-4c04-85d6-d95f597efa93/).

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1335,12 +1335,21 @@ vnc_password= "{vnc_passwd}"
             return self
 
         def _assertDomainDefined(self, name, connection):
-            dumpxmlCmd = f"virsh -c qemu:///system dumpxml --inactive {name}"
-
-            wait(lambda: "7" in self.run_admin(dumpxmlCmd + " | grep cockpit_machines | wc -l", connection))
-            # The running XML has always substituted port
-            # This checks that the defined domain is using the proper inactive XML
-            wait(lambda: self.run_admin(dumpxmlCmd + " | grep \"type='vnc' port='-1'\"", connection))
+            for retry in range(5):
+                dumpxml = self.run_admin(f"virsh -c qemu:///system dumpxml --inactive {name}", connection)
+                try:
+                    # has cockpit-machines specific metadata
+                    self.test_obj.assertIn("<cockpit_machines:has_install_phase>", dumpxml)
+                    self.test_obj.assertIn("<cockpit_machines:install_source>", dumpxml)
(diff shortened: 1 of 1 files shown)


martinpitt:
I landed the [targetcli naughty](https://github.com/cockpit-project/bots/pull/3164), and c8s should be okay again, so let's re-run packit.

martinpitt:
Oh right, needs rebase.

martinpitt:
Something [went wery wrong](https://artifacts.dev.testing-farm.io/547c0f10-57fc-4d08-a43f-66370c598acf/) with the packit c9s test. No artiffacts. The others were green (so at least rawhide works again :tada: ). I'll wait for @skobyda 's review, if he has change suggestions, they'll re-run anyway.

test/check-machines-create
```diff
@@ -1335,12 +1335,21 @@ vnc_password= "{vnc_passwd}"
...
+                    # has cockpit-machines specific metadata
+                    self.test_obj.assertIn("<cockpit_machines:has_install_phase>", dumpxml)
+                    self.test_obj.assertIn("<cockpit_machines:install_source>", dumpxml)
+                    # The running XML has always substituted port
+                    # This checks that the defined domain is using the proper inactive XML
+                    self.test_obj.assertIn("type='vnc' port='-1'", dumpxml)
+                    break
+                except AssertionError as e:
+                    fail = e
+                    time.sleep(2)
```
skobyda, line 1349: What is the reason for this sleep?

martinpitt: So that it doesn't retry the virsh dumpxml and assertions 5 times in a row without delay -- the purpose is to give it 10 seconds (5 retries times 2s sleep) for the metadata to appear. Unfortunately this is asynchronous -- I first tried to immediately assert that the XML is as expected, but that failed. So the original wrapping into `wait()` was justified.

skobyda (approved):
Thanks, let's give packit tests one more try?

martinpitt:
@skobyda : c9s failed the same way in #642, so I'm afraid this broke independently. I'm happy to investigate after this and #642 land.

PR #643, opened 2022-03-29
Work around virt-install mis-detecting SPICE graphics support

martinpitt:
Currently, spice graphics support requires QXL graphics, mostly
accidentally [1]. When uninstalling the qemu-device-display-qxl package,
`virsh domcapabilities` still includes "spice", but creating the domain
fails with:

    unsupported configuration: domain configuration does not support video model 'qxl'

Work around that by explicitly asking QEMU whether it supports QXL, and
disable spice if not. Only check qemu-system-x86_84, and ignore other
architectures for the time being. Spice is being deprecated anyway.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2064594

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -545,8 +545,14 @@ vnc_password= "{vnc_passwd}"
                                                       start_vm=True))
 
     def testCreateUrlSource(self):
+        m = self.machine
         runner = TestMachinesCreate.CreateVmRunner(self)
         config = TestMachinesCreate.TestCreateConfig
+
+        # disable QXL driver for Fedora, which builds this as a module -- should work and not add a spice console then
+        if 'fedora' in m.image:
+            m.execute("mount -o bind /dev/null /usr/lib64/qemu/hw-display-qxl.so")
+            self.addCleanup(m.execute, "umount /usr/lib64/qemu/hw-display-qxl.so")
 
         self.login_and_go("/machines")
         self.browser.wait_in_text("body", "Virtual machines")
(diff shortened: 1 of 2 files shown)


martinpitt:
Caveat: rhel-8-6 has /usr/libexec/qemu-kvm only, and QXL builtin. Let's dial this back to only disable spice if the command exists.

martinpitt:
[rhel-8-6](https://logs.cockpit-project.org/logs/pull-643-20220329-142308-c252d281-rhel-8-6/log.html#2) is rather unhappy. I can't reproduce this locally, retrying to get a log to compare, especially for how to shuffle the tests across the runners.

martinpitt:
Looks like [rawhide has broken iSCSI](https://artifacts.dev.testing-farm.io/284d3439-a3a4-4bf6-a387-1b82574ba6ef/)? Unrelated, queueing up for tomorrow for investigation.

test/check-machines-create
```diff
@@ -547,9 +547,15 @@ vnc_password= "{vnc_passwd}"
                                                       start_vm=True))
 
     def testCreateUrlSource(self):
+        m = self.machine
```
skobyda, line 550: Since you added this variable here, I think it would be nice for it to be also used in rest of the function (rest of the function calls `self.machine.execute`), or not declare it at all and use self.machine.execute for workaround. Either way let's be consistent

skobyda (changes requested):

test/check-machines-create
martinpitt: You are right, thanks! Fixed

skobyda (approved):

martinpitt:
[rawhide failed](https://artifacts.dev.testing-farm.io/fcf79cbc-2d00-46d3-a403-6e4842b43eb3/work-allulPAr6/log.txt) on the issue fixed in #648, so that's fine.

PR #646, opened 2022-03-30
Autodetect image type when adding a disk via a custom path

jelly:


test/check-machines-disks
```diff
@@ -321,7 +321,7 @@ class TestMachinesDisks(VirtualMachinesCase):
             self.xwarning_object = xwarning_object
             self.xwarning_message = xwarning_message
 
-        def getExpectedFormat(selfi, pool_type):
+        def getExpectedFormat(self, pool_type):
```
martinpitt, line 324: Heh, this should even be a `@staticmethod` then.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -543,9 +543,10 @@ export class AddDiskModalBody extends React.Component {
         this.setState({ dialogError: text, dialogErrorDetail: detail });
     }
 
-    onAddClicked() {
+    async onAddClicked() {
```
martinpitt, line 546: It's the first time we use `async` in c-machines. I don't have anything against it, I just wonder what it gives you here? What's the difference to just returning the spawn promise?

src/components/vm/disks/diskAdd.jsx
```diff
@@ -583,6 +584,23 @@ export class AddDiskModalBody extends React.Component {
             storagePool = storagePools.find(pool => pool.name === this.state.storagePoolName);
             volume = storagePool.volumes.find(vol => vol.name === this.state.existingVolumeName);
             isVolumeUsed = getStorageVolumesUsage(vms, storagePool);
+        } else if (this.state.mode === CUSTOM_PATH) {
+            // fallback to raw.
+            format = "raw";
```
martinpitt, line 589: Let's move this to the `.catch()`, to avoid race conditions and make the code a bit cleaner.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -583,6 +584,23 @@ export class AddDiskModalBody extends React.Component {
...
+            await cockpit.spawn(["qemu-img", "info", "--output", "json", this.state.file], { err: "message" })
+                    .then(output => {
+                        try {
+                            const obj = JSON.parse(output);
+                            if (obj.format) {
+                                format = obj.format;
+                            }
+                        } catch (e) {
+                            console.error("failed to parse qemu-img json output", e);
```
martinpitt, line 598: Let's also print the output here for debugging.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -583,6 +584,23 @@ export class AddDiskModalBody extends React.Component {
...
+        } else if (this.state.mode === CUSTOM_PATH) {
+            // fallback to raw.
+            format = "raw";
+            await cockpit.spawn(["qemu-img", "info", "--output", "json", this.state.file], { err: "message" })
+                    .then(output => {
+                        try {
+                            const obj = JSON.parse(output);
+                            if (obj.format) {
+                                format = obj.format;
+                            }
```
martinpitt, line 596: There should also be an else here with a `console.error` -- we really don't expect this to not return a format, do we? Or can this happen with some unknown file types? (doesn't even an entirely blank file default to "raw"?)

test/check-machines-disks
```diff
@@ -835,6 +835,16 @@ class TestMachinesDisks(VirtualMachinesCase):
             xfail=True, xfail_error_title="Disk failed to be attached",
         ).execute()
 
+        # Using directory path for disk should detect disk type
+        m.execute("qemu-img create -f qcow2 /tmp/foobar.qcow2 1M")
+        self.VMAddDiskDialog(
+            self,
+            device='disk',
+            expected_target='vdb',
```
martinpitt, line 843: Two blocks below there is a test case which adds a .qcow2 file already. (1) How does this work right now? and (2) how is qcow vs. raw related to the "vdb" vs. "sdb" target? bus type and image format should be completely independent from each other?

Can we also have a test case for raw format?

martinpitt (changes requested):
Thanks for fixing this!

src/components/vm/disks/diskAdd.jsx
martinpitt: Ack, so let's log an error then, and default to `raw` ourselves in that case.

src/components/vm/disks/diskAdd.jsx
martinpitt: As I said, I'm totally fine with using `async`/`await` -- I just don't understand what it does here. *If* you use these, then you shouldn't use `.then()`/`.catch()` on the cockpit.spawn() promise, but just await it and write the whole block as "synchronous" code. *That* would make it look nicer. But merely making the event handler async as seen from the outside seems strange -- does anything in React care?

martinpitt:
@jelly : Please rebase on your next push, to pick up the fix for c9s (landed yesterday). Thanks!

src/components/vm/disks/diskAdd.jsx
```diff
@@ -583,16 +585,35 @@ export class AddDiskModalBody extends React.Component {
             storagePool = storagePools.find(pool => pool.name === this.state.storagePoolName);
             volume = storagePool.volumes.find(vol => vol.name === this.state.existingVolumeName);
             isVolumeUsed = getStorageVolumesUsage(vms, storagePool);
+        } else if (this.state.mode === CUSTOM_PATH) {
+            optionallyResolveFormat = cockpit.spawn(["head", "--bytes", "3", this.state.file], { binary: true, err: "message" })
+                    .then(output => {
+                        let magic = "";
+                        output.map(m => { magic += String.fromCharCode(m) });
+                        switch (magic) {
+                        case "QFI":
```
skobyda, line 594: According to QCOW specification [docs](https://forensicswiki.xyz/wiki/index.php?title=QCOW_Image_Format), the magic is first 4 bytes: `QFI\xfb` (or `QFI.` in ASCII)

src/components/vm/disks/diskAdd.jsx
```diff
@@ -583,16 +585,35 @@ export class AddDiskModalBody extends React.Component {
...
+                        default:
+                            format = "raw";
+                            break;
+                        }
+                    })
+                    .catch(e => {
+                        console.warn("could not execute qemu-img", e.toString());
+                        // fallback to raw.
+                        format = "raw";
+                    });
```
skobyda, line 606: From the conversation we had at #virt IRC, I believed it was recommended that we should in the end always allow for users to specify a format, and use this autodetection. So in practice, this autodetection could be use to change the default value of "Format" in the UI, but in the end allow the user to change it, no? As there are other security issues (e.g. you can put qcow2 header over raw wile aswell)
Screenshot from 2022-03-31 18-02-26

skobyda (changes requested):

src/components/vm/disks/diskAdd.jsx
martinpitt: I'd like to circle back on this. How exactly is a manual format specification better? How would the user decide what format it is? They can decide either by file extension (which is not reliable for malicious images either), or by running `qemu-img info` (which we got told we shouldn't do). So looking at the magic seems like the safest way. We can certainly *show* the format somewhere, so that at least there's some raised eyebrow if a user specifies `somedisk.raw` and the format is actually "qcow" or "vmdk".

The potential security issue is not about the format autodetection, but about a malicious qcow image with an arbitrary backing image. That's what we should detect, but that requires `qemu-img info`. And frankly, if you can run arbitrary code on the host with that, that's a bug in qemu-img. That should just dissect the format, not actually run anything.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -583,16 +585,35 @@ export class AddDiskModalBody extends React.Component {
...
+                        let magic = "";
+                        output.map(m => { magic += String.fromCharCode(m) });
+                        switch (magic) {
+                        case "QFI":
+                            format = "qcow2";
+                            break;
+                        default:
+                            format = "raw";
+                            break;
+                        }
```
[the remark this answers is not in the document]
jelly: Ah, I thought about adding vmdk in the future

src/components/vm/disks/diskAdd.jsx
jelly: Hmm that's not what I see:

```
[jelle@t14s][~/projects/cockpit-machines]%head --bytes=4 ~/Downloads/disk.qcow2
QFI%
```

```
magic
'QFIû'
output
Uint8Array(4) [81, 70, 73, 251, buffer: ArrayBuffer(4), byteLength: 4, byteOffset: 0, length: 4, Symbol(Symbol.toStringTag): 'Uint8Array']
```

So that's strange? I've grabbed the QFI from https://github.com/cockpit-project/bots/blob/main/machine/machine_core/machine_virtual.py#L321 hmm

martinpitt: It's not `.` in ASCII. It's really `\xfb`, so you need to compare it as binary (that's why I said the channel must be binary)

martinpitt:
Moving from https://bugzilla.redhat.com/show_bug.cgi?id=2070560:

Simon and I discussed that on IRC. Let's do it like this:

- Detect top-level file format automatically (raw/qcow/vmdk). We should use `qemu-img info` for that. Reading and parsing the [qcow format](https://github.com/qemu/qemu/blob/master/docs/interop/qcow2.txt) ourselves is easy for the "is this qcow?" and "does it have a backing file?" questions, but not for "what is the backing file name/URL".
- If top-level file is not qcow: done, all good
- If top-level file is qcow: Check if it has a backing file
  * If no: done, all good 
  *  If yes: Show the backing file name in the dialog, and require the user to specify the backing format in a dropdown. The default value is invalid ("Unspecified" or so), and while it is not specified, the "Import" action button of the dialog remains disabled.

With that we both stop surprises ("I didn't expect this to have a backing file, cancel") as well as the detection problem (we can't detect format of non-file backing images).

skobyda:
> Moving from https://bugzilla.redhat.com/show_bug.cgi?id=2070560:
> 
> Simon and I discussed that on IRC. Let's do it like this:
> 
>     * Detect top-level file format automatically (raw/qcow/vmdk). We should use `qemu-img info` for that. Reading and parsing the [qcow format](https://github.com/qemu/qemu/blob/master/docs/interop/qcow2.txt) ourselves is easy for the "is this qcow?" and "does it have a backing file?" questions, but not for "what is the backing file name/URL".

I think we are now diving deeper into a bigger issue which can be left for other PR.
Let's maybe step back a little: What are the causes and problems?

###  Causes
- User is oblivious and downloaded disk from an untrusworthy source. **We can only assume user was oblivious and doesn't suspect anything.**

**What can we do about it**
We cannot possibly identify the cause, as we cannot know where the user got the disk from. So nothing really we can do there.

### Possible Fallouts
(source: #virt IRC and https://libvirt.org/kbase/secureusage.html_
1. The malicious disk can during detection of format with `qemu-disk info` over-consume CPU or fill the RAM.
**Possible prevention**: We can either autodetect data ourself, or if using qemu-disk for autodetection, use ulimit to limit CPU and RAM. 
2. Disk can point to thru a backing chain to a network, and thus compromise sensitive data. This however also requires us to specify backing chain format, otherwise libvirt will refuse the disk.
**Possible prevention**: We can detect if there is a backing file set, show where the backing file points, and prompt the user to input a format of the file backing.
3. Disk can have logical size of several terabytes and thus can baloon-up and take all the space.
**Possible prevention**: We can also probe the logical size ourself from disk metadata, or use **qemu-disk**, and reject the disk if it's bigger than available space on host (or at least warn user).

I really think that in this PR, we are concerned only about 1st fallout. The other two should be solved separately.

skobyda:
> Right now, none of these possible falllouts are possible, as c-m only supports raw disks. 

Oh wait, I didn't realize all the other options for adding non-raw disk are adding disks are thru libvirt's storage pool, where they check for such security holes. 
So yeah, since this is the PR where this security hole made available, then I agree we need to check for those fallouts here in this PR.

jelly:
> I think we need to address them all. Right now, none of these possible falllouts are possible, as c-m only supports raw disks. So if we ignore 2, then we risk unintended data exfiltration and similar. Also, adding that bit of the UI is not a lot of work, so my vote is for handling at least 1 and 2 as a blocker for this PR. 3 is more obscure indeed, and can easily be mitigated by stopping and deleting the VM. This could in principle also happen to legit VMs.

So this circles then back down on how do we detect the image type, and for qcow2 specifically the image type.

If we can't use `qemu-img` I believe we either have to parse use magic to detect the file type. We concluded that parsing the whole header is required for figuring out the backing file and I feel that we shouldn't write this ourself. 

An alternative could be using libguestsfs utils? From what I've heard it spawns a VM and mounts the the raw device in a VM for inspecting. But this adds another huge dependency to c-machines.

If limiting using `ulimit` is good enough to for our use case with qemu-img then I'm fine to apply that in this PR and show a backing file in the dialog.

martinpitt:
I quickly played around with ulimit, and it seems to work quite nicely:
```
❱❱❱ time bash -ec 'ulimit -t 2; dd if=/dev/urandom of=/dev/null'
Killed

real	0m2.006s

❱❱ ❱ bash -ec 'ulimit -v 100; /bin/echo hello'
Segmentation fault

❱❱❱ bash -ec 'ulimit -v 10000; /bin/echo hello'
hello
```

jelly:
When there is a backing file, show the that importing a backing file is now not supported. In the future we should support importing backing files and selecting the image type of the backing file. We DBus api for adding disks

jelly:
Seems I introduced a race condition which I didn't have locally https://logs.cockpit-project.org/logs/pull-646-20220411-100616-4d38e4f4-fedora-36/log.html

src/components/vm/disks/diskAdd.jsx
```diff
@@ -504,11 +505,38 @@ export class AddDiskModalBody extends React.Component {
...
+                cockpit.spawn(["/bin/bash", "-c", "ulimit -t 2 -v 100000; qemu-img info --output json " + value], { err: "message" })
+                        .then(output => {
+                            try {
+                                const data = JSON.parse(output);
+                                if (data['backing-filename']) {
+                                    this.dialogErrorSet(_("Importing an image with a backing file is unsupported"),
+                                                        cockpit.format(_("Backing file: $0"), data['backing-filename']));
+                                    this.setState({ customDiskVerified: false });
+                                } else {
+                                    this.setState({ customDiskVerified: true, format: data.format, dialogError: "" });
```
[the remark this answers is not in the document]
skobyda: And then add a separate case for "format" which handles iso files:
```
          case 'format': {
              if (value === "iso")
                  // use onValueChange instead of setState in order to perform subsequent state change logic
                  this.onValueChanged("device", "cdrom");
              this.setState({ format: value });
              break;
          }
```

src/components/vm/disks/diskAdd.jsx
```diff
@@ -504,11 +505,38 @@ export class AddDiskModalBody extends React.Component {
             break;
         }
         case 'file': {
+            this.setState({ file: value });
+
             if (value.endsWith(".iso")) {
                 // use onValueChange instead of setState in order to perform subsequent state change logic
                 this.onValueChanged("device", "cdrom");
             }
```
skobyda, line 513: If approach in my other comment gets implemented, this can be removed

test/check-machines-disks
```diff
@@ -898,6 +937,7 @@ class TestMachinesDisks(VirtualMachinesCase):
         m.execute("for pool in $(virsh pool-list --all --name); do virsh pool-destroy $pool || true; virsh pool-undefine $pool; done")
         b.wait_visible("#vm-details[data-pools-count=0]")
         b.click(prefix)
+        b.set_file_autocomplete_val("#vm-subVmTest1-disks-adddisk-file", "/var/lib/libvirt/novell.iso")
```
skobyda, line 941: What is the aim of this?

skobyda (changes requested):

src/components/vm/disks/diskAdd.jsx
```diff
@@ -504,11 +505,38 @@ export class AddDiskModalBody extends React.Component {
...
-            this.setState({ file: value });
+
+            if (value) {
+                // Limit process time to 2 seconds and ram to 100 MB
+                cockpit.spawn(["/bin/bash", "-c", "ulimit -t 2 -v 100000; qemu-img info --output json " + value], { err: "message" })
```
martinpitt, line 517: Please add `-e` bash option, to ensure that the ulimit call succeeds.

But moreover, `value` must be shell-escaped here, to work with quotes and spaces. (Please add a test for a file name containing these). I'm not sure ATM how to do that in a sensible manner, it's rather awkward to have to go through a shell string. Perhaps you can avoid this by sending the file name via `.input()` and change the shell command to read it from stdin?

martinpitt (changes requested):

src/components/vm/disks/diskAdd.jsx
```diff
@@ -719,7 +747,9 @@ export class AddDiskModalBody extends React.Component {
                            <Button id={`${idPrefix}-dialog-add`}
                                    variant='primary'
                                    isLoading={this.state.addDiskInProgress}
-                                   isDisabled={this.state.addDiskInProgress || dialogLoading || (storagePools.length == 0 && this.state.mode != CUSTOM_PATH)}
+                                   isDisabled={this.state.addDiskInProgress || dialogLoading ||
+                                               (storagePools.length == 0 && this.state.mode != CUSTOM_PATH) ||
+                                               (this.state.mode == CUSTOM_PATH && this.state.device === "disk" && !this.state.customDiskVerified)}
```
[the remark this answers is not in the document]
jelly: Should it? The text then would be "select a valid disk image or cdrom". Actually why did we decide on disabling the button again? Why can't we do the verification on submit hmmm?

test/check-machines-disks
jelly: Ok, that novell.qcow2 test needs fixing then, as it just touches a qcow2 file so that's wrong:

```
[jelle@t14s][~/projects/cockpit-machines]%touch /tmp/foo.qcow2
[jelle@t14s][~/projects/cockpit-machines]%qemu-img info /tmp/foo.qcow2
image: /tmp/foo.qcow2
file format: raw
virtual size: 0 B (0 bytes)
disk size: 0 B
```

jelly: About the bus type, not sure. The expected_target needs to be passed as the test asserts that in `verify_disk_added`.

src/components/vm/disks/diskAdd.jsx
martinpitt: Right, but you can use `xargs` to read a file name from stdin and call a program with it:

```
$ touch '/tmp/evil file"na'"'"'me'
$ stat -c %n /tmp/evil*
/tmp/evil file"na'me
$ echo -n '/tmp/evil file"na'"'"'me' | xargs -0n1 qemu-img info --output json
{
    "virtual-size": 0,
    "filename": "/tmp/evil file\"na'me",
    "format": "raw",
    "actual-size": 0,
    "dirty-flag": false
}
```

I know this is ugly, but as soon as you are in shell land, one has to resort to such measures. A better solution would be to teach the bridge to set ulimits on its own (via some cockpit.spawn() options), but I think let's not block on that.

jelly: Ah, that's the magic I was missing. Thanks!

src/components/vm/disks/diskAdd.jsx
```diff
@@ -509,11 +510,38 @@ export class AddDiskModalBody extends React.Component {
...
+
             if (value.endsWith(".iso")) {
                 // use onValueChange instead of setState in order to perform subsequent state change logic
                 this.onValueChanged("device", "cdrom");
             }
-            this.setState({ file: value });
+
+            if (value) {
+                // Limit process time to 2 seconds and ram to 100 MB
+                cockpit.spawn(["/bin/bash", "-c", "ulimit -t 2 -v 100000; echo -n " + value + " | xargs -0n1 qemu-img info --output json "], { err: "message" })
```
martinpitt, line 520: No, that doesn't help -- it just moves the shell escaping madness to a slightly different place in the command line. You really need to `.input(value)` here to feed the name through stdin.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -509,11 +510,39 @@ export class AddDiskModalBody extends React.Component {
...
+
             if (value.endsWith(".iso")) {
                 // use onValueChange instead of setState in order to perform subsequent state change logic
                 this.onValueChanged("device", "cdrom");
             }
-            this.setState({ file: value });
+
+            if (value) {
+                // Limit process time to 2 seconds and ram to 100 MB
+                cockpit.spawn(["/bin/bash", "-c", "ulimit -t 2 -v 100000; xargs -0n1 qemu-img info --output json "], { err: "message" })
```
martinpitt, line 520: Please add `"-ec"`, to ensure the ulimit call works.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -509,11 +510,39 @@ export class AddDiskModalBody extends React.Component {
...
+                                if (data['backing-filename']) {
+                                    this.dialogErrorSet(_("Importing an image with a backing file is unsupported"),
+                                                        cockpit.format(_("Backing file: $0"), data['backing-filename']));
+                                    this.setState({ customDiskVerified: false });
+                                } else {
+                                    this.setState({ customDiskVerified: true, format: data.format, dialogError: "" });
+                                }
+                            } catch (err) {
+                                this.setState({ customDiskVerified: false });
+                                console.error("qemu-img provided invalid JSON");
```
martinpitt, line 534: Let's show the output, for useful error messages.

test/check-machines-disks
martinpitt: Right, please fix that test below as well (separate commit), it's rather confusing. I suggest to just rename the file from `novell.qcow2` to `empty.img`.

martinpitt (changes requested):
Thanks! Some minor details still, but I'm happy with the format detection now. I'll leave the final review to @skobyda, as he still had some outstanding discussion threads.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -509,11 +508,39 @@ export class AddDiskModalBody extends React.Component {
...
+
             if (value.endsWith(".iso")) {
                 // use onValueChange instead of setState in order to perform subsequent state change logic
                 this.onValueChanged("device", "cdrom");
             }
-            this.setState({ file: value });
+
+            if (value) {
+                // Limit process time to 2 seconds and ram to 100 MB
+                cockpit.spawn(["/bin/bash", "-ec", "ulimit -t 2 -v 100000; xargs -0n1 qemu-img info --output json "], { err: "message" })
```
skobyda, line 520: This command has to use superuser permission if possible:
```suggestion
                cockpit.spawn(["/bin/bash", "-ec", "ulimit -t 2 -v 100000; xargs -0n1 qemu-img info --output json "], { err: "message", superuser: "try" })
```
Without it, if user with administrative access will try to attach a disk with a path, which requires superuser permission, e.g. `/var/lib/libvirt/images/someFile.qcow2`, the command will fail: 
`could not execute qemu-img qemu-img: Could not open '/var/lib/libvirt/images/fedora-36-beta.qcow2': Could not open '/var/lib/libvirt/images/fedora-36-beta.qcow2': Permission denied`

That makes me think, maybe there should be a test for such occassion too.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -509,11 +508,39 @@ export class AddDiskModalBody extends React.Component {
...
+                                    this.dialogErrorSet(_("Importing an image with a backing file is unsupported"),
+                                                        cockpit.format(_("Backing file: $0"), data['backing-filename']));
+                                    this.setState({ customDiskVerified: false });
+                                } else {
+                                    this.setState({ customDiskVerified: true, format: data.format, dialogError: "" });
+                                }
+                            } catch (err) {
+                                this.setState({ customDiskVerified: false });
+                                console.error("qemu-img provided invalid JSON", err);
+                            }
```
skobyda, line 535: Sorry, but I realized there might be one more issue. It seems `qemu-img info` requires an exclusive lock when reading a disk. This however is not consistent with the other feature we support: disk sharing between multiple VMs.
How to reproduce:

1. Have a "raw" disk attached to running VM1, and set its access to "Writeable and shared":
2. Try to attach the same disk to VM2.
3. `qemu-img info` will fail because it could not get lock:
```
could not execute qemu-img qemu-img: Could not open '/var/lib/libvirt/images/disk1.img': Failed to get shared "write" lock
Is another process using the image [/var/lib/libvirt/images/disk1.img]? 3 [diskAdd.jsx:535:36](webpack://machines/src/components/vm/disks/diskAdd.jsx)
```

skobyda: I'm thinking how to work around this and I think of 2 solutions:

1. When `qemu-img info` fails, we could just fallback to the old way and use "raw" format. This would be simple solution (disk sharing only works with "raw" format, so it would work for above mentioned case)
2. When `qemu-img info` fails, show format selector and allow user to specify format manually (this would be a nice fallback for other unforeseeable cases, not just disk sharing, but is more complex).

skobyda (changes requested):

src/components/vm/disks/diskAdd.jsx
martinpitt: Ugh, this is serious. I didn't expect `qemu-img info` to open the file in write mode, or to even lock the file. It still doesn't make sense to add a qcow file as raw or vice versa, so IMHO falling back to letting the user decide is unacceptable: The user has no freedom in this decision, the format of the file is what it is.

If there really is no sane way to use `qemu-img info`, then let's just read it ourselves after all. Let's start with reading the first 16 bytes (with `dd`) into a binary channel, check the qcow magic, and just reject an image with a backing file. With that, we don't have to parse the backing file name in the first version. This will at least fix self-contained qcow files, which are certainly a good portion of the use case here. We can then refine this later, by filing a qemu-img info bug report ("stop the crap and just do what we ask"), doing more parsing ourselves, or perhaps resorting to `file` after all.

martinpitt: @skobyda : How does it make sense for two *running* VMs to access the same qcow image? This is begging for data corruption. I thought this was only supported as long as at most one VM was running at a time? Then there shouldn't be any write lock contention. From that aspect, the current behaviour of failing image detection is even desirable. I still reject the idea of opening the file for writing just for format detection, though -- that is crazy.

skobyda: > @skobyda : How does it make sense for two _running_ VMs to access the same qcow image?

This issue is not  about qcow, as you cannot share qcow2 image. But it's a valid use case with `raw` images, which even have <shareable /> flag for that: https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms , which tells that VMs will not try to gain lock when writing so disk, allowing it to be written to concurrently by multiple VMs. 
1. One use case is for example building clusters. You would have to have cluster filesystem set up for that). (more about that [here](https://www.humblec.com/how-to-share-a-disk-to-more-than-one-vm-using-libvirt-or-what-sharable-flag-do-in-libvirt/) . c-machines supports that as you can set "shareable" flag on raw disks when you click on "Edit' button next to an already attached disk.
2. The other use case is that you have one VM writing data to a raw disk, and then you want to attach the raw disk to a other VM only as a read-only so it can read from it. (you can do that by attaching disk and then editing it's configuration to "readonly")

In both cases however, with this PR, when attaching the disk to the second VM (even if the second VM is shut off, and only first VM is running), qemu-img info will fail to gain lock, and thus user will be unable to attach the disk to the second VM altogether.

martinpitt: Interesting, then that would give us a path for the next step: Do a first coarse detection ourselves for qcow magic and existence of an overlay; and if that is present, call qemu-img info (as then we can't have shared images anyway). But still, I'd do that in a follow-up, given how long this has been blocked already.

jelly: Created an upstream issue for the write lock issue https://gitlab.com/qemu-project/qemu/-/issues/1287

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +514,49 @@ export class AddDiskModalBody extends React.Component {
             break;
         }
         case 'file': {
+            this.setState({ file: value });
```
martinpitt, line 517: Please add this to the `this.state =` initialization in the ctor, so that it has a complete schema there.

(I know this is just moving code, but while we are at it..)

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +514,49 @@ export class AddDiskModalBody extends React.Component {
...
+
             if (value.endsWith(".iso")) {
                 // use onValueChange instead of setState in order to perform subsequent state change logic
                 this.onValueChanged("device", "cdrom");
             }
-            this.setState({ file: value });
+
+            if (value) {
+                cockpit.spawn(["/bin/bash", "-c", "xargs -0n1 head --bytes=16 "], { binary: true, err: "message", superuser: "try" })
```
martinpitt, line 525: Why does this need a shell wrapper? I agree that we can't use cockpit.file() for this, but calling just head directly ought to be fine.

You can also eliminate the xargs by just calling `["head", "--bytes=16", value]`, which is much simpler.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +514,49 @@ export class AddDiskModalBody extends React.Component {
...
+                        .input(value)
+                        .then(output => {
+                            console.log(output);
```
martinpitt, line 528: debugging leftover

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +514,49 @@ export class AddDiskModalBody extends React.Component {
...
+                            let format = "";
+                            const file_identifier = Array.from(output).slice(0, 4)
```
martinpitt, line 530: Please save Array.from(output) in a const, as you use it multiple times. But is this actually necessary? You already get an `Uint8Array` from spawn, which seems good enough?

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +514,49 @@ export class AddDiskModalBody extends React.Component {
...
+                                    .map(function(i) { return '0x' + i.toString(16) })
```
martinpitt, line 531: Use a proper arrow here, i.e. `.map(i= > '0x' ...)`.

Moreover, do we really need to do this conversion? Can't you just directly compare the first 4 entries of `output` to `[81, 70, 73, 251]`? (one by one of course, as list comparisons don't work in JS). That would avoid a lot of back and forth, and it's also not clear how other future formats look like -- they may not have a similar magic in the first 4 bytes.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +514,49 @@ export class AddDiskModalBody extends React.Component {
...
+                                    .join(' ');
+
+                            switch (file_identifier) {
+                            case "0x51 0x46 0x49 0xfb": { // QCOW
+                                format = "qcow2";
+                                const file_backing_offset = Array.from(output).slice(8);
+                                // All zeros, no backing file offset
+                                if (file_backing_offset.every(bit => bit === 0)) {
```
martinpitt, line 539: These are bytes, not bits.

martinpitt (changes requested):
Really cool, thanks for doing this yak shaving! I have some code cleanup suggestions, and this is conflict-y, but it finally looks like we can get this in soon.

Also, totally release-note worthy IMHO.

src/components/vm/disks/diskAdd.jsx
jelly: I find it kinda annoying to match against, but I guess I can change that.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +515,41 @@ export class AddDiskModalBody extends React.Component {
...
                 // use onValueChange instead of setState in order to perform subsequent state change logic
                 this.onValueChanged("device", "cdrom");
             }
-            this.setState({ file: value });
+
+            if (value) {
+                cockpit.spawn(["head", "--bytes=16", value], { binary: true, err: "message", superuser: "try" })
+                        .then(output => {
+                            let format = "";
+                            const file_header = Array.from(output);
```
martinpitt, line 529: What does this do? `output` is already an array, so this smells like a no-op? I.e. you could just directly do `.then(file_header => ...)` instead?

martinpitt (approved):
Thanks! My concerns are addressed, there's just one further simplification to be done. Not bad enough to veto, but I'll happily re-approve if you want/need to force-push again.

I leave the remaining review to @skobyda , as he also had some outstanding concerns wrt. UI behaviour.

src/components/vm/disks/diskAdd.jsx
jelly: I think it's an UIntArray which is special, let me second check.

jelly: Turns out, I'm incorrect, we can slice!

src/components/vm/disks/diskAdd.jsx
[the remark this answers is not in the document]
jelly: So, with the current code iteration we still disable the button, and in one scenario we can and up that the verification failed, if `head --bytes=16` throws an error. So I'll fix that to fallback to raw?

There is not much a user can do when it fails, and it should never happen.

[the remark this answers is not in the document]
skobyda: The main issue I have with this approach is that when you open a dialog and click on "Custom path", the user is immediately greeted with a disabled "Add" button, because they did not specify a file path yet.
Screenshot from 2022-11-17 01-06-34

Whereas if they just choose "Create new", the "Add" button is not disabled, despite user not specifying volume name yet.
Screenshot from 2022-11-17 01-41-16


So this is kinda inconsistent.
And to me, this seems confusing from user's perspective as they do not know it's disabled because the "Custom path" input is empty, since there is no message there too. And they would not perhaps understand, why did "Add" button get disabled when they switched from "Create new" to "Custom path". Maybe they could even thing that the whole functionality is disabled for their VM?

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +515,40 @@ export class AddDiskModalBody extends React.Component {
...
+                            let format = "";
+
+                            if (file_header[0] == 81 && file_header[1] == 70 &&
+                                file_header[2] == 73 && file_header[3] == 251) {
+                                format = "qcow2";
+                                // All zeros, no backing file offset
+                                if (file_header.slice(8).every(bytes => bytes === 0)) {
+                                    this.setState({ customDiskVerified: true, format: format, dialogError: "" });
+                                } else {
+                                    this.dialogErrorSet(_("Importing an image with a backing file is unsupported"));
```
skobyda, line 537: Instead of setting the dialog error, I think you should set FormGroup's `helperTextInvalid`, similar to how it's done here: https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/disks/diskAdd.jsx#L147
This would result in the error being shown right underneath the Input field, where the cause of the error is at, something similar to this:
Screenshot from 2022-11-17 01-28-09

Instead of showing the error somewhere else in the dialog:
Screenshot from 2022-11-17 01-25-33

src/components/vm/disks/diskAdd.jsx
```diff
@@ -358,7 +358,9 @@ export class AddDiskModalBody extends React.Component {
         super(props);
         this.state = {
             validate: false,
-            dialogLoading: true
+            dialogLoading: true,
+            customDiskVerified: false,
```
skobyda, line 362: Yeah, as I described more in this comment, since customDiskVerified is set to false by default, the "Add" button is disabled immediately as you open the dialog.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -513,11 +515,40 @@ export class AddDiskModalBody extends React.Component {
...
+                                format = "raw";
+                                this.setState({ customDiskVerified: true, format: format, dialogError: "" });
+                            }
+                        })
+                        .catch(e => {
+                            this.setState({ customDiskVerified: true, format: "raw", dialogError: "" });
+                            console.warn("could not execute head --bytes=16", e.toString());
+                        });
+            } else {
+                this.setState({ customDiskVerified: false });
```
skobyda, line 550: Yeah as I described in this comment, this disables the "Add" button if "Custom path" field is empty, but doesn't provide any message why it's disabled.
Screenshot from 2022-11-17 01-41-21

We should maybe move this check into [validateParams()](https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/disks/diskAdd.jsx#L418) aim for the approach that we let user click on "Add" button, then [validateParams()](https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/disks/diskAdd.jsx#L418) is called, it checks that some compulsory input field is empty, and shows message, similar to this:
Screenshot from 2022-11-17 01-46-58

skobyda (changes requested):

src/components/vm/disks/diskAdd.jsx
garrett: Why is the button missing here? Is this due to dark mode, where I changed the disabled state? (And do I need to write a workaround for the workaround?)

image

skobyda: @garrett What button is missing here? Do you mean that the "Add" button is disabled by default imediatelly when user opens the dialog? Yeah that was what my issue comment was about. It was disabled because the "Custom path" was empty, which I believe was incorrect.
But Jelle fixed it now, and now it disables button only when user inputs a path to a file which we cannot attach as a disk:
Screenshot from 2022-11-22 14-51-12

But when user doesn't input anything, it doesn't disable the button as before:
Screenshot from 2022-11-22 14-51-22

Ideally we probably also want some form verification, which would check if the field is empty, and say someting like the Custom path should not be empty... But that can be done in different PR.

garrett: No, the button is missing. There is no button, just text. It might be a dark theme artifact that the button and the background are the same color. I don't know if that's our fault or not. I need something reproducible locally to test.

skobyda: Ahhh right, I didn't realize that the disabled button is not the inline button, so it should be the outline around the text even when it's disabled... Great catch @garrett

KKoukiou (approved):

src/components/vm/disks/diskAdd.jsx
```diff
@@ -358,7 +358,9 @@ export class AddDiskModalBody extends React.Component {
         super(props);
         this.state = {
             validate: false,
-            dialogLoading: true
+            dialogLoading: true,
+            customDiskVerified: false,
+            file: null,
```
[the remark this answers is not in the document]
KKoukiou: It's missed but also it does not hurt. So let's just get it in, we have so many prs open :)

skobyda (approved):
Tested it out locally with some backing file and different disk types, works like charm :). Thanks fo all the hard work here

PR #651, opened 2022-04-04
test: Fix race condition in disk validation

martinpitt:
Wait until the API actually detected the disk, so that the following
if/else has some actual data to work with. Otherwise it could happen
that neither `*-disks-*-device` is present, and the test prematurely
fails with "Unknown disk device".


martinpitt:
Broken out from PR #583. I'm :+1: on this, but I need @KKoukiou to formally approve as this is my PR with her commit now.

KKoukiou (approved):

PR #652, opened 2022-04-05
WiP: Redesign delete/remove dialog

skobyda:
As pointed out at https://github.com/cockpit-project/cockpit-machines/pull/640 , there are many delete/removal dialogs in c-machines which do not follow PF guidelines. Here is my initial try to adhert to the guidelines. It will likely need some rewording.

- I mostly moved identification of what we are removing from modal header, where it sometimes may not fit, into the modal body.
- Fixed some Upper Case in Network deletion modal.
- Implemented DescriptionList for more complex object which cannot be identified by just 1 value

### Before
Screenshot from 2022-04-04 13-58-28

### After
Screenshot from 2022-04-05 11-43-31

### Before
Screenshot from 2022-04-04 15-44-57

### After
Screenshot from 2022-04-04 15-38-53

### Before
Screenshot from 2022-04-05 12-03-16

### After
Screenshot from 2022-04-05 12-03-47

### Before
Screenshot from 2022-04-05 12-03-54

### After
Screenshot from 2022-04-05 12-03-59

### Before
Screenshot from 2022-04-05 12-09-09

### After
Screenshot from 2022-04-05 12-12-09


skobyda:
I'm also thinking how to redesign Storage Pool deletion dialog. As it (depending on whetever is active) contains additional message or checkbox:
Screenshot from 2022-04-05 12-15-25
Screenshot from 2022-04-05 12-13-50

garrett:
> I'm also thinking how to redesign Storage Pool deletion dialog. As it (depending on whetever is active) contains additional message or checkbox:
> Screenshot from 2022-04-05 12-27-09
> Screenshot from 2022-04-05 12-26-52

I don't understand. Is this what it currently looks like?

There shouldn't be a label here, except for the mandatory one for the checkbox. Which volumes would optionally be removed? It should show those.

For example, when I expand my default storage pool, I see this:

image

Can you even delete the storage pool without the volumes? :thinking: 

...What would happen to the files? And the VMs?

...And then what happens to the VMs when you delete the storage pool and the volumes?

KKoukiou (commented):
I won't manage to review this PR. Just from the screen shots, I noticed that the modal titles don't follow the PF guidelines. In specfiic, Confirmation modals, need to have titles with a question.
For example;

https://www.patternfly.org/v4/components/modal/design-guidelines/#confirmation-dialogs

skobyda:
> I don't understand. Is this what it currently looks like?

Yes, that's the current look.

> There shouldn't be a label here, except for the mandatory one for the checkbox. Which volumes would optionally be removed? It should show those.
> 
> For example, when I expand my default storage pool, I see this:
> 
> image
> 
> Can you even delete the storage pool without the volumes? thinking
>
> ...What would happen to the files? And the VMs?

You can delete the pool without volumes. In that case, the files (e.g. qcow2 file or raw file) are left untouched. For example, if you have a file directory storage pool, then deleting a storage pool just deletes the storage pool "abstraction" over that directory, but the directory with files is still left on the system. This can be useful if you want to move some of those files/disk elsewhere after storage pool deletion.

> ...And then what happens to the VMs when you delete the storage pool and the volumes?

the storage pool **OR** the volumes. Those are 2 different situations:

If some VM is using disk with reference as [storagePoolName]:[volumeName], then you cannot delete the storage pool:
_(Button "Delete" should be probably greyed out. This needs aria-disabled label!)_
Screenshot from 2022-04-07 14-13-00

If you reference the storage volume directly thru path or some other ID without referencing storage pool, then you can delete the storage pool:
_(Maybe those info text shoould be a tooltip of disabled Button)_
Screenshot from 2022-04-07 14-20-55

skobyda:
> I won't manage to review this PR. Just from the screen shots, I noticed that the modal titles don't follow the PF guidelines. In specfiic, Confirmation modals, need to have titles with a question. For example;
> 
> https://www.patternfly.org/v4/components/modal/design-guidelines/#confirmation-dialogs

Then maybe just rewording those titles should be enough:
"Remove network" ->"Remove network?"
"Remove disk from VM" -> "Remove disk from VM?"

garrett:
I've mentioned our headlines not conforming to PF titles before (regarding the "?"), but we never did it as it was a one-off and inconsistent with the rest of Cockpit.

As this is one of those PRs with sweeping changes, it makes sense to switch the titles to questions wherever it makes sense. :+1: 

---

If we're making these changes, we might want to also consider adding an icon for destruction too:

https://www.patternfly.org/v4/components/modal/design-guidelines/#confirm-a-destructive-action
image

> Use a [primary button](https://www.patternfly.org/v4/components/button/design-guidelines#primary-button) to confirm a destructive action. If the action carries serious consequences, then use a [danger button](https://www.patternfly.org/v4/components/button/design-guidelines#danger-button) instead.
> 
> If an action is **difficult or impossible to undo, add a warning icon** beside the headline.

They're optional... But it seems to strongly suggest them above under certain destructive circumstances.

https://www.patternfly.org/v4/components/modal/design-guidelines/#icon-use-in-modal-dialogs:

> <p class="ws-p">Icons are optional in modal dialogs. Use or omit them as your use case requires.</p>
> 
>  Use case(s) | Usage
>  -- | --
>   Warning: Cautions or warns the user of a permanent action, or that information will be deleted upon action completion | Add to confirmation dialogs or passive dialogs to indicate a higher level of urgency and importance.
>    Critical Warning: Indicates that an error has occured | Add to error dialogs.
>    Acknowledgement: Informs the user of an action or result | Add to confirmation or passive dialogs to indicate a lower level of urgency.
> 

(But this could easily go in another PR.)

garrett:
> If some VM is using disk without storage pool reference, but instead is referencing the storage volume directly thru file path or some other ID without referencing storage pool, then you can delete the storage pool, but not its volumes:
(Maybe that info text shoould be a tooltip of disabled Button?)
>
> image

Should that dialog have some sort of solution within the dialog itself?

That is, list the VMs it's in use by and perhaps have a way to detach them from the individual VMs from within the dialog?

(This probably should be addressed in an separate PR; keep this PR mainly related to reworking the modals as-is without really adding features — except for showing more information.)

garrett:
What happens with errors? Are they inline alerts? Are those at the bottom or top? (If inline alerts, they should be at the top, like what https://github.com/cockpit-project/cockpit/pull/17221 is now doing.)

I know that Machines does do a lot of error handling as alerts in the page, after a modal though, so perhaps it doesn't even need to be addressed. :grin:

skobyda:
> it makes sense to switch the titles to questions wherever it makes sense. +1

Done

> If we're making these changes, we might want to also consider adding an icon for destruction too

Done

Screenshot from 2022-04-20 21-45-25

> > If some VM is using disk without storage pool reference, but instead is referencing the storage volume directly thru file path or some other ID without referencing storage pool, then you can delete the storage pool, but not its volumes:
> > (Maybe that info text shoould be a tooltip of disabled Button?)
> > image
> 
> Should that dialog have some sort of solution within the dialog itself?
> 
> That is, list the VMs it's in use by and perhaps have a way to detach them from the individual VMs from within the dialog?
> 
> (This probably should be addressed in an separate PR; keep this PR mainly related to reworking the modals as-is without really adding features — except for showing more information.)

Yeah

> What happens with errors? Are they inline alerts? Are those at the bottom or top? (If inline alerts, they should be at the top, like what [cockpit-project/cockpit#17221](https://github.com/cockpit-project/cockpit/pull/17221) is now doing.)
> 
> I know that Machines does do a lot of error handling as alerts in the page, after a modal though, so perhaps it doesn't even need to be addressed. grin

They were inline errors in all the cases, except for disk removal, where there were alerts in page (but that was accidental, I tracked it down to my [old PR](https://github.com/cockpit-project/cockpit/pull/12661) where I forgot to replace page alert there with inline modal alert.

So now I made it consistent, all the delete/remove dialogs now use inline alerts, and fixed them to show alert at the top of the modal:

Screenshot from 2022-04-20 23-00-24

garrett (changes requested):
> > it makes sense to switch the titles to questions wherever it makes sense. +1
> 
> Done
> 
> > If we're making these changes, we might want to also consider adding an icon for destruction too
> 
> Done
> 

Nice! Thanks!

---

Anything we can do about this error message?

image

1. We should never use the title "Failure". It should be specific to the error type.
2. What happened to the space between the alert and the rest of the modal content? (There is none here; there should be some.)
3. A CD is not a "disk". A CD is a "disc". (It's a "disc" with a C if it's a circle you can see.)
4. Should "target" really be "device"?
   For example. GNOME Disks labels the it as a "device":
   image

---

> image
> ...
> Yeah. This functionality would be way more complex, so I would leave it outside of this PR.

Did you add the checkbox's label here? It shouldn't have that duplicate bold label to the left... It's a single opt-in checkbox. It also shouldn't have "the": "Delete volumes inside this pool". Additionally, the whole "c..." thing is odd. But I understand that part will be handled in a follow-up.

skobyda:
> Anything we can do about this error message?
> 
> image
> 
>     1. We should never use the title "Failure". It should be specific to the error type.
>
>     2. What happened to the space between the alert and the rest of the modal content? (There is none here; there should be some.)

There was no by default in ModalError component. I added patternfly's padding-top medium: `pf-u-pt-md` 

>     3. A CD is not a "disk". A CD is a "disc". (It's a "disc" with a C if it's a circle you can see.)

Fixed it in modal title and error message title. However it's still present in error message, since that error message comes from libvirt itself.

Screenshot from 2022-04-26 01-21-51


>     4. Should "target" really be "device"?
>        For example. GNOME Disks labels the it as a "device":
>        image

Term "target" is quite commonly used across libvirt, and there is reason for it, as "device" and "target" may refer to different things.

- "Device" is the identifier under which the disk is known on the host.
- "Target" specifies identifier which will be exposed to the VM. (i.e. how it's "targeted" to the VM)

Those 2 will be usually different. 
For example: you have "sda" device on your host, you attach it to the VM, but device will be exposed to the VM under identifier "vda"
I believe if we just call it "device" , we eventually end up confusing user, and they might think they have attached some "vdb" device from host to the VM.
 
> Did you add the checkbox's label here? It shouldn't have that duplicate bold label to the left... It's a single opt-in checkbox. It also shouldn't have "the": "Delete volumes inside this pool". Additionally, the whole "c..." thing is odd. But I understand that part will be handled in a follow-up.

No, the label was already there.

garrett:
> What do you mean? There was never a space in our AlertNotification component.

I meant the space in that screenshot. There should be space; alert widgets should never be slammed into something else. (And most other widgets shouldn't either.)

Disc: Nice! Thanks for fixing that (in Cockpit, where possible).

Target: Ok. That's fine.

skobyda:
Current state of this PR:
Screenshot from 2022-04-27 12-33-39
Screenshot from 2022-04-27 12-32-53
Screenshot from 2022-04-27 12-32-40
Screenshot from 2022-04-27 12-32-04
Screenshot from 2022-04-27 12-31-43
Screenshot from 2022-04-27 12-31-28
Screenshot from 2022-04-27 12-30-59
Screenshot from 2022-04-27 12-30-47
Screenshot from 2022-04-27 12-30-32

skobyda:
The last issues, which can be seen on screenshots above, is the deletion of storage pool, which will probably need more comprehensive redesign:
Screenshot from 2022-04-27 12-33-39

And deleteting VMs, which it seems doesn't mention the VM name anywhere in the dialog. Maybe it should. Is it okay to be mentioned in the title, or should the modal message be reworded to include the VM name?
Screenshot from 2022-04-27 12-30-32

garrett (changes requested):
Most of these changes are excellent! Thanks!

What's up with the `:` here in this dialog?

image


This still shouldn't have a label at the side, and it still doesn't mention what will be deleted. This is not great. :cry: 

image


What's up with the spacing? It should look more like this:

image

src/components/common/deleteResource.jsx
```diff
@@ -63,10 +64,12 @@ export class DeleteResourceModal extends React.Component {
         return (
             <Modal position="top" variant="small" isOpen onClose={onClose}
                    id="delete-resource-modal"
-                   title={title}
+                   title={<>
```
marusak, line 67: Modal has `titleIconVariant` property where you can specify either `warning` or specific icon, see https://www.patternfly.org/v4/components/modal

Cannot we use that instead?

skobyda: Good catch, thanks!

skobyda:
@garrett So here is my new iteration:
### Adjusted spacing in VM deletion dialog:
Screenshot from 2022-05-03 14-40-11

### Storage pool with volumes will mention which volumes will be deleted:
Screenshot from 2022-05-03 14-43-00

### Inactive storage pool:
Screenshot from 2022-05-03 14-43-53

### Storage pool without any volumes:
This however looks a bit strange, to have just an empty modal. Or not? Not sure about this one.
Screenshot from 2022-05-03 14-44-00

skobyda:
> image

User would want to delete pool without deleting volumes if they are migrating them to another volumes. E.g:
1. Have a directory pool at /var/lib/libvirt/myPool1 with volumes as file in it.
2. You want to move those volumes to another pool, let's say /var/lib/libvirt/myPool2
3. To do that, you would delete myPool1, and then just move it's leftover volumes to myPool2 directory

src/components/vm/disks/vmDiskColumns.jsx
```diff
@@ -46,15 +65,9 @@ export const DiskSourceCell = ({ diskSource, idPrefix }) => {
...
-    addOptional(chunks, diskSource.dev, "device", _("Device"));
-    addOptional(chunks, diskSource.protocol, "protocol", _("Protocol"));
-    addOptional(chunks, diskSource.pool, "pool", _("Pool"));
-    addOptional(chunks, diskSource.volume, "volume", _("Volume"));
-    addOptional(chunks, diskSource.host.name, "host", _("Host"));
-    addOptional(chunks, diskSource.host.port, "port", _("Port"));
+    DISK_SOURCE_LIST.forEach(entry => addOptional(chunks, getDiskSourceValue(diskSource, entry.name), entry.name, entry.label));
 
-    return <DescriptionList isHorizontal>{chunks}</DescriptionList>;
+    return <DescriptionList isHorizontal>{chunks.map(chunk => chunk)}</DescriptionList>;
```
martinpitt, line 70: This looks like a no-op.

src/components/vm/disks/vmDisksCard.jsx
```diff
@@ -267,12 +267,24 @@ export class VmDisksCard extends React.Component {
             };
 
             const deleteDialogProps = {
+                title: disk.device === "cdrom" ? _("Remove disc from VM?") : _("Remove disk from VM?"),
                 actionName: _("Remove"),
-                title: cockpit.format(_("Remove disk $0"), disk.target),
-                errorMessage: cockpit.format(_("Disk $0 could not be removed"), disk.target),
+                errorMessage: disk.device === "cdrom"
+                    ? cockpit.format(_("Disc $0 could not be removed"), disk.target)
+                    : cockpit.format(_("Disk $0 could not be removed"), disk.target),
```
martinpitt, line 274: [My dictionary](https://dict.leo.org/englisch-deutsch/disc) says that disc vs. disk are mostly just American vs. British English. Duplicating these strings seems a bit dubious and unnecessary, especially as they are all in the same "Disks" table on the VM details anyway.

src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -231,12 +231,32 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
...
+
                 const deleteNICAction = (
                     <DeleteResourceButton objectId={`${id}-hostdev-${hostdevId}`}
                                           actionName={_("Remove")}
                                           showDialog={() => setDeleteDialogProps({
-                                              title: _("Remove host device"),
+                                              title: _("Remove host device from VM?"),
                                               errorMessage: ("Host device could not be removed"),
+                                              actionDescription: cockpit.format(_("Host device will be removed from $0:"), vm.name),
+                                              objectDescription,
```
martinpitt, line 259: Please adjust the test to make sure the dialog contains the expected device description.

test/machineslib.py
```diff
@@ -212,7 +212,7 @@ def deleteIface(self, iface):
         b = self.browser
 
         b.click(f"#delete-vm-subVmTest1-iface-{iface}")
-        b.wait_in_text(".pf-c-modal-box .pf-c-modal-box__header .pf-c-modal-box__title", "Delete network interface")
+        b.wait_in_text(".pf-c-modal-box .pf-c-modal-box__header .pf-c-modal-box__title", "Delete network interface?")
```
martinpitt, line 215: Please add a check for the description details in the dialog.

martinpitt (changes requested):
Thanks! I have some minor comments and some language bikeshedding. The pixel tests need to be updated (the diff looks good). @garrett , can you please re-review the latest screenshots?

garrett (changes requested):
Overall great! 

I've made a few adjustments in the mockups; it'd be nice to get them in this PR, but I understand if some are hard to do or would otherwise be follow-ups:

1. We should probably make the storage pool "remove", not delete. Deletion is for deleting stuff on a disk, whereas removing is just removing a definition of something. So we're removing storage pools and deleting volumes.
2. If we could prioritze the name earlier in the title, it might be nice. (WDYT? Adjusted in the screenshots. Ideally, it'd be bold too. But however this is done, it should be translation-friendly.)
3. Volumes being deleted should probably be a bulleted list, not an inline comma separated list, so they have more gravitas (serious importance; visual weightiness here in the UI) and are easier to scan.
4. I've changed the text for the volumes being deleted, to state that _all_ of them in the list will be deleted.
5. When a storage pool is being removed and volumes are being deleted (where the option is checked), I have the button switch to remove & delete.
6. I added some empty pool text. It's there just to give a "peace of mind" (reassurance) that volumes aren't being deleted.
- I might be hungry. ("LunchTime" came to mind for some reason...)


vm-pool-remove-unchecked excalidraw

vm-pool-remove excalidraw

vm-pool-remove-empty excalidraw(1)

garrett:
After posting the above comment: I thought about it and storage pools are _also_ on disk, so perhaps just ignore my suggestion to change remove to delete. That'll make things easier for this PR to land. The other suggestions would still apply.

Here are modified mockups for that:

vm-pool-delete excalidraw

vm-pool-delete-empty excalidraw

skobyda:
Updated design:
Screenshot from 2022-05-09 12-53-06
Screenshot from 2022-05-09 12-52-54

The only thing I could not achieve was to make Storage pool name in the modal title bold. The title is formatted thru cockpit.format() and it does not accept string where storage pool name is a html <strong> element. @martinpitt any idea if it's possible to highlight a word in a translated+formatted string?

martinpitt:
@skobyda : HTML markup inside translated strings is best avoided. There are some tricks to make it work, but it's really brittle through translations, and unsafe too.

src/components/vm/disks/vmDisksCard.jsx
skobyda: @garrett mind giving input?

skobyda: Yeah, but it seems there are specific cases where it's used independed of variation of english. And that is optical discs:
1. `Optical disc image` or in short: `ISO`
It sems that even wikipedia calls it "disk" when it's a disk device in general, but "disc" when talking about ISO/CD: An optical **disc** image (or ISO image) is a **disk** image. [wiki](https://en.wikipedia.org/wiki/Optical_disc_image)
3. CD: `Compact disc read-only memory`
4. DVD: `Digital Versatile Disc` or `Digital Video Disc`

And when we are talking about ISO images in VMs, we are talking about emulations of Optical Disc image or CD.

martinpitt: Right, I'm aware of that technical distinction -- but it doesn't make much difference here in the dialog IMHO. The VM details page has them all under the "Disks" title.

skobyda: Alright, reverted

src/components/vm/hostdevs/hostDevCard.jsx
skobyda: Done

test/machineslib.py
skobyda: Virtual network static DHCP entries doesn't have any object description in the dialog, it only a sentence "network {networkMac} will be permanently deleted.". So the only thing which could go wrong is to have an incorrect network mac. But we test in other places that we have correct MAC for a network entry, so is it worth testing?

skobyda: Actually, on second thought, we can never now what can go wrong, so let's lest, it. Updated the PR

src/components/storagePools/storagePoolDelete.jsx
```diff
@@ -160,26 +162,34 @@ export class StoragePoolDelete extends React.Component {
...
-                                  isChecked={this.state.deleteVolumes}
-                                  label={_("Delete the volumes inside this pool")}
-                                  onChange={checked => this.onValueChanged('deleteVolumes', checked)} />
-                    </FormGroup>}
-                    { !storagePool.active && _("Deleting an inactive storage pool will only undefine the pool. Its content will not be deleted.")}
-                </Form>
-                { storagePool.active && showWarning() }
+                { storagePool.active
+                    ? (volumes.length > 0
+                        ? <>
```
martinpitt, line 167: No need for this extra `<>` wrapper.

test/check-machines-networks
```diff
@@ -868,6 +868,9 @@ class TestMachinesNetworks(VirtualMachinesCase):
 
             def cleanup(self):
                 b.click(f"#delete-network-{self.network_name}-{self.connection_name}-ipv4-dhcp-host-{self.entry_id}-button")
+                b.wait_in_text(".pf-c-modal-box__body .pf-c-description-list", self.ip)
+                b.wait_in_text("#delete-resource-modal-IP", self.ip)
+                b.wait_in_text("#delete-resource-modal-MAC", self.mac)
```
martinpitt, line 873: This would better be squashed into the "Redesign vNIC removal dialog" commit.

martinpitt (approved):
Thank you! Code-wise this now looks good enough. The two nitpicks are only when you need to force-push anyway for some other reason.

@garrett , are you happy with the design now?

martinpitt:
@skobyda: Can you please write a release note? Personally I find the effect on the "Remove host device" dialog the most obvious, so maybe take that new picture, clip it so that it has an uniform 2px border, and use it for the relnote?

garrett (changes requested):
Apologies for not catching these earlier. They're all minor things that should be able to be fixed quickly, I think?

---

Screenshot 2022-05-11 at 10-45-19 Virtual Machines - garrett@Bolt

Which VM? The modal doesn't say. (I guess this could be a follow-up.)

---

Screenshot 2022-05-11 at 10-46-44 Virtual Machines - garrett@Bolt

This should have a `:`

---

Screenshot 2022-05-11 at 10-46-55 Virtual Machines - garrett@Bolt

This should not be delete. It should be remove.

Screenshot 2022-05-11 at 10-47-14 Virtual Machines - garrett@Bolt

Same for the list of networks. It should be remove, not delete.

---

Screenshot 2022-05-11 at 10-47-58 Virtual Machines - garrett@Bolt

Missing "of":

All **of** its...

(Perhaps this should be a follow-up after the release, as it would affect translations, right?)

---

Screenshot 2022-05-11 at 10-48-28 Virtual Machines - garrett@Bolt

This doesn't say what it is inside the container. (I think this one's fine for a follow-up though.)

martinpitt:
> [network iface] This should not be delete. It should be remove.

Should it? I thought it literally ceases to exist when you del^Wrem^Wmake it go away?

martinpitt:
> Perhaps this should be a follow-up after the release, as it would affect translations, right?

This PR changes so many strings, it'd actually be better to not change them twice. I.e. better to change the string right away here.

garrett:
> Should it? I thought it literally ceases to exist when you del^Wrem^Wmake it go away?

Are files being deleted? It should be remove, not delete.

A network card that's mapped into the VM doesn't physically cease to exist when you remove it from the VM. (It can be mapped in directly or mapped in virtually, but it's still the network interface on a physical device.)

> Delete: (v.) Use to describe completely removing an object. Delete is a destructive action.
> Opposite: Create
> Notes: Erase and Remove are not recommended for this use case. See Remove for usage guidelines.

> Remove: (v.) Use to describe removing an item from a list, group, view, or other container element without completely deleting it. Also see Add and Delete.
> Opposite: Add
> Notes: If what you’re removing is not readily apparent from the context, consider adding a noun (like “Remove file”).

Also the verb pairs are create/delete, add/remove... they shouldn't really be mixed in most circumstances (as in: create/remove, add/delete).

---

> This PR changes so many strings, it'd actually be better to not change them twice. I.e. better to change the string right away here.

Yeah, that's why I was saying follow-up for many. We could just change all string changes to follow-up.

Delete to Remove technically changes a string, but we already have that translated. So: Change delete to remove for this PR, make a follow-up to land after release for the rest.

garrett:
Also, we need to add some text at the top for the release notes and use just one screenshot.

I _think_ I made the script work where we can just add a horizontal rule with three dashes `---` to split the content and have the release notes at top and keep the rest that's already there after it.

test/check-machines-networks
skobyda: This is a completely different dialog. Removing vNIC and DHCP static entries from virtual networks are different functionalities with different dialogs.

skobyda:
> This doesn't say what it is inside the container. (I think this one's fine for a follow-up though.)

What do you mean? How can we tell the user what is inside the mount?

garrett:
Well, if we want this to land in the upcoming release, then we can't change strings. (Changing delete to remove isn't a string change; it's done elsewhere.)

Otherwise, we can defer this to next release in two weeks.

skobyda:
This PR doesn't have to be int this release, so let's change the string now and leave it for the next release

skobyda:
Added my draft for release notes at the [first comment](https://github.com/cockpit-project/cockpit-machines/pull/652#issue-1192935992)

martinpitt:
Thanks @skobyda : I tweaked the language a little bit.

skobyda:
Ahh, more test failures

martinpitt:
[pixel test failure](https://logs.cockpit-project.org/logs/pull-652-20220513-131905-cc3b0593-fedora-35/log.html), it needs updating for Delete → Remove.

src/components/storagePools/storagePoolDelete.jsx
```diff
@@ -160,26 +162,32 @@ export class StoragePoolDelete extends React.Component {
...
-                    </FormGroup>}
-                    { !storagePool.active && _("Deleting an inactive storage pool will only undefine the pool. Its content will not be deleted.")}
-                </Form>
-                { storagePool.active && showWarning() }
+                { storagePool.active
+                    ? (volumes.length > 0
+                        ? <Checkbox id='storage-pool-delete-volumes'
+                                    isChecked={this.state.deleteVolumes}
+                                    label={<>
+                                        { _("Also delete all volumes inside this pool: ")}
```
martinpitt, line 170: Trailing space. This is important for i18n'ed strings, as it's basically guaranteed that this won't survive translations. If you need the space for layouting, add it *after* the gettext call.

martinpitt (changes requested):
One tiny issue, and the pixel test needs updating, otherwise LGTM. Thanks! Let's try and land this today, as it's very prone to bitrotting.

src/components/storagePools/storagePoolDelete.jsx
skobyda: Ah this was probably typo, as list is in a new line, there is no need for a space

martinpitt:
@skobyda : Thanks! Please rebase, to let packit tests run.

skobyda:
Tests green. I think it's close to landing. @garrett @martinpitt mind giving your final word?

garrett (approved):
Looks good! Let's merge this! :+1:

martinpitt (approved):
Dakujem!

PR #655, opened 2022-04-07
I18n and pycodestyle fixes, stronger static code tests

martinpitt:
I've seen this broken i18n pattern [here](https://github.com/cockpit-project/cockpit-machines/pull/560#discussion_r844875443) and [here](https://github.com/cockpit-project/cockpit-machines/pull/640#discussion_r843483851), and we already have two instances of that in our code. Let's fix it and make sure it doesn't come back.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,6 @@ WEBLATE_REPO=tmp/weblate-repo
 WEBLATE_REPO_URL=https://github.com/cockpit-project/cockpit-machines-weblate.git
 WEBLATE_REPO_BRANCH=main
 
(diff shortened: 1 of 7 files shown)


jelly:
Should we apply the same treatment to c-podman? https://github.com/cockpit-project/cockpit-podman/blob/e9ec2977a88b9dea19a85f7ea281ce3899c6fef7/Makefile#L143

martinpitt:
@jelly : Certainly can't hurt. I even wonder if we should move this to test/common/ in cockpit and then re-use it instead of copying it. But it's not such a big deal to copy it either.

jelly (approved):

PR #656, opened 2022-04-07
test: Add check for explicitly selecting "BIOS" firmware

martinpitt:
This triggers some logic, and PR #583 managed to break this.


jelly (approved):
I was wondering, do we also already test switching firmware between BIOS/UEFI?

martinpitt:
@jelly : Yes, `TestMachinesCreate.testConfigureBeforeInstall` does that.

PR #660, opened 2022-04-08
fmf: Run tests without QEMU QXL support

martinpitt:
We recently dropped the QXL dependency from our package, as it is so
expensive (see commit 1edd79909eb). Start testing without that in FMF.

Drop the QXL disabling hack from testCreateUrlSource(), we don't need
this any more -- all tests (including that one) now run without QXL in
FMF.

 - Builds on top of PR #622


skobyda (approved):
LGTM, thanks

PR #664, opened 2022-04-11
test: add a test about unchecking all boot options, and also some style modifications

yunmingyang:


test/check-machines-settings
```diff
@@ -323,6 +322,39 @@ class TestMachinesSettings(VirtualMachinesCase):
         # Check boot order has changed and no warning is shown
         b.wait_not_in_text("#vm-subVmTest1-boot-order", bootOrder)
 
+        # Uncheck the last boot option which is network
+        def uncheckTheLastBootOption(firstOptionType="disk"):
```
skobyda, line 326: Thank you, I like this test.
To understand this correctly, the motivation of this test is to check that after removing all the boot order from every device, the Libvirt fallbacks to using the main hard disk to obtain the boot image and we are still able to detect that from the UI?
If so, then I would perhaps reword the function name and comments to be more clear about that, something like:
```suggestion
        # After unchecking all the boot options, the VM should fall back to using the first hard disk as a boot image
        # Check that UI is able to detect this fallback
        def uncheckTheLastBootOption(firstOptionType="disk"):
```

skobyda (changes requested):

test/check-machines-settings
skobyda: Actually I can just do this change myself and force push to your branch

PR #665, opened 2022-04-11
Follow up of VM creation redesign

skobyda:
Split from: https://github.com/cockpit-project/cockpit-machines/pull/560

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -156,10 +156,28 @@ function getStorageDefaults() {
     return { storageSize, storageSizeUnit, minimumStorage };
 }
 
+function getVmName(connectionName, vms, os) {
+    let retName = os.shortId;
+
+    const date = new Date();
+    retName += '-' + date.getFullYear() + "-" + (date.getMonth() + 1) + "-" + date.getDate();
+
+    let tmpRetName = retName;
+    // VM with same name already exists, append a character at the end, starting with 'B'
+    for (let i = 66; vms.some(vm => vm.name === tmpRetName && connectionName == vm.connectionName) && i <= 91; i++) {
+        // Could not generate name which doesn't collide with any other VM name
+        if (i === 91)
+            return "";
+        tmpRetName = retName + '-' + String.fromCharCode(i);
+    }
+
+    return tmpRetName;
+}
+
 function validateParams(vmParams) {
     const validationFailed = {};
 
-    if (isEmpty(vmParams.vmName.trim()))
+    if (isEmpty(vmParams.vmName.trim()) && isEmpty(vmParams.suggestedVmName.trim()))
         validationFailed.vmName = _("Name must not be empty");
     else if (vmParams.vms.some(vm => vm.name === vmParams.vmName))
         validationFailed.vmName = cockpit.format(_("VM $0 already exists"), vmParams.vmName);
@@ -222,7 +240,7 @@ function validateParams(vmParams) {
     return validationFailed;
 }
 
-const NameRow = ({ vmName, onValueChanged, validationFailed }) => {
+const NameRow = ({ vmName, suggestedVmName, onValueChanged, validationFailed }) => {
     const validationStateName = validationFailed.vmName ? 'error' : 'default';
 
     return (
@@ -234,7 +252,7 @@ const NameRow = ({ vmName, onValueChanged, validationFailed }) => {
                        validated={validationStateName}
                        minLength={1}
                        value={vmName || ''}
-                       placeholder={_("Unique name")}
+                       placeholder={isEmpty(suggestedVmName.trim()) ? _("Unique name") : cockpit.format(_("Unique name, default: $0"), suggestedVmName)}
                        onChange={value => onValueChanged('vmName', value)} />
         </FormGroup>
     );
@@ -748,6 +766,7 @@ class CreateVmModal extends React.Component {
             activeTabKey: 0,
             validate: false,
             vmName: '',
+            suggestedVmName: '',
             connectionName: LIBVIRT_SYSTEM_CONNECTION,
             sourceType: defaultSourceType,
             source: '',
@@ -870,6 +889,12 @@ class CreateVmModal extends React.Component {
                 // storage pools are different for each connection, so we set storagePool value to default (newVolume)
                 this.setState({ storagePool: "NewVolume" });
             }
+
+            // For different connections the generated VM names might differ
+            // try to regenerate it
+            if (this.state.os)
+                this.setState({ suggestedVmName: getVmName(value, this.props.vms, this.state.os) });
+
             break;
         case 'os': {
             const stateDelta = { [key]: value };
@@ -904,6 +929,10 @@ class CreateVmModal extends React.Component {
             if (!value || !value.unattendedInstallable)
                 this.onValueChanged('unattendedInstallation', false);
 
+            // generate VM name based on OS if user selects an OS
+            // clear generated VM name if they unselect an OS
+            stateDelta.suggestedVmName = value ? getVmName(this.state.connectionName, this.props.vms, value) : "";
+
             this.setState(stateDelta);
             break;
         }
@@ -915,6 +944,7 @@ class CreateVmModal extends React.Component {
 
     onCreateClicked(startVm) {
         const { storagePools, close, onAddErrorNotification, osInfoList, nodeMaxMemory, vms } = this.props;
+        const vmName = isEmpty(this.state.vmName.trim()) ? this.state.suggestedVmName : this.state.vmName;
 
         const validation = validateParams({ ...this.state, osInfoList, nodeMaxMemory, vms: vms.filter(vm => vm.connectionName == this.state.connectionName) });
         if (Object.getOwnPropertyNames(validation).length > 0) {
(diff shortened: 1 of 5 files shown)


skobyda:
In previous PR, @martinpitt had few issues with naming. @garrett might wanna join discussion too:

martin wrote:
`It's redundant to include the connection name into the domain name. These are already namespaced, and it will make them even longer.`

Ah that's true. If we don't include them, we will just end up with VM list looking like this. I'm fine removing system/session from the name:
Screenshot from 2022-04-11 11-47-51


martin wrote:
```
"a name will be suggested" is too surprising. This would only appear until the user selects an OS, right? and that is mandatory? We need to avoid a situation where you click "create" with "a name will be suggested".

Still, I feel this is too magic/jumpy, as a helper text. Easier/less surprising behaviour proposal: set the name input to an autogenerated name once the user chooses an OS and the name is empty still, and then just keep the current "is empty" validation.

Also, grammar: If you keep this sentence, it's "If blank, name will be set.." (drop the "a"). But still, just do it, and avoid the indirection.

As this whole auto-naming business is quite hairy, I also recommend splitting this out into its own PR. I feel it's getting unnecessarily difficult to handle all these independent changes in a single PR.
```

skobyda:
> It's redundant to include the connection name into the domain name. These are already namespaced, and it will make them even longer.

Ah that's true. If we don't include them, we will just end up with VM list looking like this. I'm fine removing system/session from the name:
Screenshot from 2022-04-11 11-47-51

skobyda:
> Split from: #560

Screenshot from 2022-04-11 11-51-26
Screenshot from 2022-04-11 11-51-19

garrett:
> Ah that's true. If we don't include them, we will just end up with VM list looking like this. I'm fine removing system/session from the name:
> Screenshot from 2022-04-11 11-47-51

Shouldn't the list also be renamed to "User session", like the toggle in the create dialog? That'd help differentiate between "System" and "Session", which look all too similar here otherwise.

skobyda:
> Shouldn't the list also be renamed to "User session", like the toggle in the create dialog? That'd help differentiate between "System" and "Session", which look all too similar here otherwise.

I thank you, I didn't catch that. Fixed it:

Screenshot from 2022-04-11 18-37-00

garrett:
> "a name will be suggested" is too surprising. This would only appear until the user selects an OS, right?
> ...
> Easier/less surprising behaviour proposal: set the name input to an autogenerated name once the user chooses an OS and the name is empty still, and then just keep the current "is empty" validation.

Let's do this.

The generated names are really best as a fallback. People should be naming their VMs with some purpose, like "SQL server", "DNS", "NGINX", "NextCloud", etc. — which we cannot guess. The generated names as a fallback are good, however, as differentiators that do provide some context about the VM... but they're not as good as a descriptive name. They're definitely better than an error, especially for the cases where someone wants to make a quick VM for testing, so this PR is still a nice feature.

Removing the text and treating this just as a fallback where the name wasn't entered (before selecting configurations) will simplify the UI, would make this PR easier to land, and (as a bonus) make @martinpitt happy too. :wink:

skobyda:
Current state:

### When user opens the dialog, they do not see any message about autogeneration:
Screenshot from 2022-04-20 14-12-19

### Once they enter the Operating system, the message appears:
Screenshot from 2022-04-20 14-12-26

### If afterwards they write some VM name themself, the message disappears:
Screenshot from 2022-04-20 14-12-36

garrett:
I don't like that the form jumps down from above with a setting below the area it jumps.

What happened to this?

>>     Easier/less surprising behaviour proposal: set the name input to an autogenerated name once the user chooses an OS and the name is empty still, and then just keep the current "is empty" validation.
>
> Let's do this.

I guess I didn't _clearly_ specifically specify helper text or not, but I was assuming it didn't. I did say "removing the text", but I didn't make it clear that it was helper text:

> _Removing the text_ and treating this just as a fallback where the name wasn't entered (before selecting configurations) will simplify the UI, would make this PR easier to land, and (as a bonus) make @martinpitt happy too. 

We could, however, do a hybrid approach and use placeholder text, I suppose... as this would be a fallback.

https://www.patternfly.org/v4/components/form/design-guidelines#placeholder-text-0

We could have the placeholder text say "_Unique name_" by default, and I suppose "_Unique name, default: XYZ_" where XYZ is the generated string. (I had to modify the PF recommendation a little to make this work. But it's similar "in spirit" to what the guidelines suggest.)

Doing this would:

- Relegate the autonaming to a fallback
- Not cause a jump in the form and modal height
- Still expose the name that will be chosen if nothing else is specified
- Only show up when nothing else is typed as the name (which happens to be what we want, really, if this is just a fallback)

skobyda:
Ahh, sorry I misunderstood previous comments, I though only "If left blank, name will be autogenerated" message should removed, but "If left blank, name will be set to someName" should stay.
Anyway, I hope now I got it right:
### When user opens the dialog, everything is as before:
Screenshot from 2022-04-25 19-25-37

### If user first specifies VM name, and then selects an OS, no name autogeneration will take place:
Screenshot from 2022-04-25 18-56-23

### If user however doesn't specify VM name, and goes straight to specifying OS:
Screenshot from 2022-04-25 19-14-28

### If user doesn't specify either, and clicks on "Create...", validation still works as before:

Screenshot from 2022-04-25 19-15-01

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -784,9 +803,11 @@ class CreateVmModal extends React.Component {
 
     onValueChanged(key, value) {
         switch (key) {
-        case 'vmName':
-            this.setState({ [key]: value.split(" ").join("_") });
+        case 'vmName': {
+            const stateDelta = { [key]: value.split(" ").join("_") };
+            this.setState(stateDelta);
             break;
+        }
```
martinpitt, line 810: No-op change

test/check-machines-create
```diff
@@ -139,6 +140,49 @@ class TestMachinesCreate(VirtualMachinesCase):
                                                                          create_and_run=False),
                                              {"source-url": "Installation source must not be empty"})
 
+    def testCreateNameGeneration(self):
```
martinpitt, line 143: Please add this to the exclude list in test/browser/run-test.sh , as it tests internal logic only.

martinpitt (changes requested):
Nice, thank you! I like that workflow much more than the "surprise! :tada:" approach. One or two tiny things to clean up, then it's good for me. @garrett, are you d'accord with this workflow, too? From what I can see, it matches your specification exactly. Cheers!

garrett (changes requested):
> Ahh, sorry I misunderstood previous comments

That's fine. The interpretation of what I originally wrote wasn't clear and, reading it again later, I could see it being understood either way.

So, after your recent changes: It looks mostly great! 

~However, I did find an interaction bug:~

~Select download, then select an OS, then switch download to something else. The suggested name does not update. The suggested name should probably be recomputed (if the name is blank) after every form change (possibly without an autogenerated name, if the values don't support it), not just the applicable ones (which is what it seems to currently do).~

~Screenshot after I select download CentOS 7, then switched to cloud base image:~

~Screenshot 2022-04-26 at 14-53-30 Virtual Machines - garrett@Bolt~

Edit: Oh! I still have CentOS 7 selected. Whoops! This is fine.

garrett (dismissed):
:+1: Looks great!

martinpitt:
The [two](https://logs.cockpit-project.org/logs/pull-665-20220426-120759-34fcaffa-rhel-9-0/log.html) [RHEL 9.x](https://logs.cockpit-project.org/logs/pull-665-20220426-120759-34fcaffa-rhel-9-1/log.htmll failures look a bit curious, like there's some VM corruption going on? I've never seen that before, this is worth a closer look. Can you please rebase, to run the packit tests, and retry the ones on our CI?

martinpitt:
Otherwise LGTM, will approve once green. Thanks!

martinpitt:
Rebased (without other changes).

skobyda:
> Meh, there is [something about that test](https://logs.cockpit-project.org/logs/pull-665-20220426-155628-cb51d3a7-rhel-9-0/log.html#72) which makes it very prone to corrupting the VM.

That's very odd. It seems that VM fails even to get prepared. But there is nothing unusual with this test which would fail it

martinpitt:
@skobyda : How do you mean? It was running before, the test killed it. Hence the timeout and "Restarting global machine 6".

skobyda:
Ah I was looking at different test run. @martinpitt after I retriggerred test, the failure is different: https://logs.cockpit-project.org/logs/pull-665-20220427-090744-7c02b2de-rhel-9-0/log.html
It says it failed at bots/machine/machine_core/ssh_connection.py, but isn't that a VM preparation which runs before the test?

martinpitt:
No, VM prep happens in the "initialization" section. If you grep the [raw log](https://logs.cockpit-project.org/logs/pull-665-20220427-090744-7c02b2de-rhel-9-0/log) for "ND@3" you see all tests which run in that persistent VM. `TestMachinesCreate.testCreateNameGeneration` killed it, presumably due to some kernel crash or an out of memory condition. Possibly the created VMs are allocating too much memory? This needs reproducing locally and checking with `virsh console`.

skobyda:
Tests will fail until https://github.com/cockpit-project/cockpit-machines/pull/696 gets merged

jelly:
> Blocked, tests will fail until #696 gets merged

Reviewed and merged so this can be re-tested.

martinpitt (approved):
Thanks! Enough nitpicking, let's land this.

garrett:
If we're going to mention this in the release notes, it needs a very short summary of what the PR actually changes as the title and a bit more detailed summary as the text at the top, ideally with a screenshot. Thanks!

PR #666, opened 2022-04-11
test: Manually enable virtstoraged.socket on Fedora 35, add missing QEMU dependencies

martinpitt:
This works around a packaging bug when only
libvirt-daemon-driver-storage-core is installed, but not the
libvirt-daemon-driver-storage metapackage (as we try to do in [1]).

https://bugzilla.redhat.com/show_bug.cgi?id=2025644

[1] https://github.com/cockpit-project/bots/pull/3227


skobyda (approved):
LGTM

PR #668, opened 2022-04-13
workflows: Split npm-update to PF and non PF tasks

marusak:
Run PF update every Monday and anything else try to update on Tue, Thu and Sat.


martinpitt (approved):
Cheers! Are you planning to do ostree and starter-kit as well? (For these it actually hurts most, as they don't have pixel tests)

marusak:
> Are you planning to do ostree and starter-kit as well? 

Yes, I sent to all projects we maintain

skobyda (approved):
Thanks

PR #669, opened 2022-04-17
test: add the check for non-root user

yunmingyang:
- Requires https://github.com/cockpit-project/bots/pull/3297

### DIFF
--- a/test/check-machines-lifecycle
+++ b/test/check-machines-lifecycle
@@ -38,9 +38,8 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         self._testBasic()
 
     def createUser(self, user_group):
-        user_name = 'test_' + user_group + '_user'
-        self.machine.execute(
-            'useradd -G {0} {1} && echo "{1}:foobar" | chpasswd'.format(user_group, user_name))
+        user_name = f"test_{user_group if user_group else 'none'}_user"
+        self.machine.execute(f"useradd{' -G ' + user_group if user_group else '' } {user_name} && echo '{user_name}:foobar' | chpasswd")
         # user libvirtd instance tends to SIGABRT with "Failed to find user record for uid .." on shutdown during cleanup
         # so make sure that there are no leftover user processes that bleed into the next test
         self.addCleanup(self.machine.execute, "pkill -u {0}; while pgrep -u {0}; do sleep 0.5; done".format(user_name))
@@ -48,6 +47,10 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         self.allow_journal_messages('Process .*libvirtd.* of user 10.* dumped core.*')
 
         return user_name
+
+    def testBasicNonrootUserUnprivileged(self):
+        user = self.createUser(user_group=None)
+        self._testBasic(user, False, True)
 
     # FIXME remove this skipImage
     @skipImage('Fails with Rejected send message, 1 matched rules; type="method_call"',
--- a/test/check-machines-lifecycle
+++ b/test/check-machines-lifecycle
@@ -59,9 +59,19 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         user = self.createUser(user_group='libvirt')
         self._testBasic(user, False)
 
-    def testBasicWheelUserUnprivileged(self):
-        user = self.createUser(user_group='wheel')
+    def testBasicAdminUser(self):
+        # The group in debian based distro should be sudo
+        if "ubuntu" in self.machine.image or "debian" in self.machine.image:
+            user_group = "sudo"
+        else:
+            user_group = "wheel"
+        user = self.createUser(user_group=user_group)
+
+        # Check NON-PRIVILEGED user won't see any VMs even though they are in wheel group
         self._testBasic(user, False, True)
+        self.machine.execute("virsh destroy subVmTest1 && virsh undefine subVmTest1")
+        # Check wheel group PRIVILEGED user will see VMs and can do operations on them
+        self._testBasic(user, True, False)
 
     def _testBasic(self, user=None, superuser=True, expect_empty_list=False):
         b = self.browser
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -65,6 +65,7 @@ EXCLUDES="$EXCLUDES
           TestMachinesDisks.testDiskEdit
 
           TestMachinesLifecycle.testBasicLibvirtUserUnprivileged
+          TestMachinesLifecycle.testBasicNonrootUserUnprivileged
           TestMachinesLifecycle.testBasicWheelUserUnprivileged
           TestMachinesLifecycle.testDelete
           TestMachinesLifecycle.testLibvirt
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -64,9 +64,9 @@ EXCLUDES="$EXCLUDES
           TestMachinesDisks.testDetachDisk
           TestMachinesDisks.testDiskEdit
 
+          TestMachinesLifecycle.testBasicAdminUser
           TestMachinesLifecycle.testBasicLibvirtUserUnprivileged
           TestMachinesLifecycle.testBasicNonrootUserUnprivileged
-          TestMachinesLifecycle.testBasicWheelUserUnprivileged
           TestMachinesLifecycle.testDelete
           TestMachinesLifecycle.testLibvirt
 


test/check-machines-lifecycle
```diff
@@ -38,9 +38,9 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         self._testBasic()
 
     def createUser(self, user_group):
-        user_name = 'test_' + user_group + '_user'
-        self.machine.execute(
-            'useradd -G {0} {1} && echo "{1}:foobar" | chpasswd'.format(user_group, user_name))
+        user_name = 'test_{}_user'.format(user_group if user_group else "none")
```
martinpitt, line 41: Please use f-strings, not `.format()`.

test/check-machines-lifecycle
```diff
@@ -49,6 +49,10 @@ class TestMachinesLifecycle(VirtualMachinesCase):
 
         return user_name
 
+    def testBasicNonrootUserUnprivileged(self):
+        user = self.createUser(user_group=None)
+        self._testBasic(user, False, True)
```
martinpitt, line 54: I don't quite understand how this can behave exactly the same way as testBasicWheelUserUnprivileged(). The latter ought to get superuser privileges, this one shouldn't. There must be some measurable difference. Actually, *this* new test case makes sense, I don't understand the existing testBasicWheelUserUnprivileged(). @yunmingyang , @skobyda , any thoughts here?

martinpitt (changes requested):

test/check-machines-lifecycle
skobyda: So in order for wheel group user to have special privileges, you have to amend libvirt polkit policies. Without them, which is the case in our tests, wheel group user is basically any user without access to libvirt daemon. So I think testBasicNonrootUserUnprivileged is doing exactly the same as testBasicWheelUserUnprivileged.

What I think is missing in our tests instead is the extension for testBasicWheelUserUnprivileged which grants wheel group user priviliged access to qemu:///system by ammending polkit policies and see if it works properly. However there are currently som bugs with polkit, which need to be fixed first (e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1820493), so that would be outside of the scope of this PR.

skobyda: So in summary, I think testBasicNonrootUserUnprivileged doesn't add anything new compared to testBasicWheelUserUnprivileged. But I'm up for deleting testBasicWheelUserUnprivileged, since its name is kinda misleading, and reintroducing it back, with extended tests for privileged access for wheel group once the polkit issues are fixed.

skobyda (changes requested):
I like this change, but let's also delete testBasicWheelUserUnprivileged since it doesn't do anything different and has a misleading name.

test/check-machines-lifecycle
martinpitt: What I meant is that a `wheel` user can run `sudo` and thus should be able to connect to libvirt as root. A test *can* disable this with `login_and_go(..., superuser=False)` option, but this test doesn't. That's where my confusion comes from -- why does this test not simply connect as root?

skobyda: So I went back in git history and found that this seems to be rewritten from older test:

```
-    def testWheelUserOperationWithVm(self):
-        name = 'staticvm'
-        vm_args = self.create_vm(name, wait=True)
-
-        # A admin user in the wheel group is allowed to see the VM and
-        # can do operations on it, but only with escalated privileges.
-
-        self.non_root_user_operations(user_group='wheel', privilege=False,
-                                      operations=False)
-
-        self.non_root_user_operations(user_group='wheel', privilege=True,
-                                      operations=True, vm_args=vm_args)

```

skobyda: So yeah I was wrong, you don't need polkit setup for that. But if you login with user with wheel privileges, you right now won't see any VMs (reproducible with `[wheeluser@fedora cockpit]$virsh -c qemu:///system list --all`) but once you authorize, you will get access to qemu:///system (reporducible with `[wheeluser@fedora cockpit]$sudo virsh -c qemu:///system list --all`). So I think the tests should reflect this behaviour

skobyda:
@yunmingyang I added a [commit](https://github.com/cockpit-project/cockpit-machines/pull/669/commits/79b2fca15a3731f866fdd10e1db6ec9fbb0945ee) which extends wheel user test so it's tests more extensivily the expected behaviour for wheel user (as described in the [other comment](https://github.com/cockpit-project/cockpit-machines/pull/669#discussion_r853892904)):  What do you think about that?

test/check-machines-lifecycle
```diff
@@ -60,9 +60,14 @@ class TestMachinesLifecycle(VirtualMachinesCase):
...
 
-    def testBasicWheelUserUnprivileged(self):
+    def testBasicWheelUser(self):
         user = self.createUser(user_group='wheel')
+
+        # Check NON-PRIVILEGED user won't see any VMs even though he is in wheel group
         self._testBasic(user, False, True)
+        self.machine.execute("virsh destroy subVmTest1 && virsh undefine subVmTest1")
+        # Check wheel group PRIVILEGED user will see VMs and can do operations on them
+        self._testBasic(user, True, False)
```
martinpitt, line 70: That makes sense, thanks! Would be even nicer to name the boolean arguments, for easier interpretation. I.e. `superuser=True`

yunmingyang:
@skobyda Yeah, sure, I also re-push the commit as modifying the f-string, please review, thanks.

test/check-machines-lifecycle
yunmingyang: I found we use self.eval_js('window.localStorage.setItem("superuser:%s", "%s");' % (user, "any" if superuser else "none")) to implement "superuser" in login_and_go. But it is "sudo" by window.localStorage.getItem("superuser:test_wheel") if changing the privilege in UI. Should modify "any" to "sudo"?

skobyda: Ah I'm not really familiar with that code, or where and how it gets read by cockpit during authorization process. @martinpitt any idea?

martinpitt: I'm afraid I don't understand @yunmingyang 's question. The `superuser` flag in `login_and_go()` determines whether the test will do the equivalent of clicking the "Limited access" button in the UI, so that the cockpit session will start a root bridge (through sudo) and hence all `{ superuser: "try" }` or `{ superuser: "require" }` cockpit API calls will run as root. Being in the `wheel` group allows a user to become root through sudo.

In other words, if you log in as `test_wheel_user` with `superuser=False`, this will behave like a normal unprivileged user. With `superuser=True` you will be able to do any administrative action including libvirt (this is totally independent of polkit, it just connects to libvirtd as root).

yunmingyang: I mean, in the test, it seems to use "window.localStorage.setItem("superuser:%s", "%s")" in the browser console, before click "Log in", to do the equivalent of clicking the "Limited access" button in the UI. And if we use a wheel group user like test_wheel_user, to get "Administrative access", in test, it seems to use "window.localStorage.setItem("superuser:test_wheel_user", "any")". But if we get "Administrative access" by clicking ""Limited access", the value of "superuser:test_wheel_user" getting by window.localStorage.getItem("superuser:test_wheel_user") is "sudo", not "any". So my question is, is there some different when setting "superuser:test_wheel_user" to "sudo" or "any"?

martinpitt: @yunmingyang : not in a meaningful way in this context, no. There used to be a fallback for running `pkexec cockpit-bridge` instead of `sudo cockpit-bridge`, but in practice, `any` and `sudo` are equivalent for "connect with browser" mode.

yunmingyang:
@martin, @skobyda I check the failed job(debian-stable). It seems that /etc/sudoers in debian-stable lacks "%wheel  ALL=(ALL)       ALL", after adding that, the case could run continually. And I guess it is the same with debian-testing as their failed snapshots are the same. Should we update these two image by image-refresh? or I think we could also add a if in the test code.

Also, after the case run finished, I found there is a core dump by libvirtd and qemu

yunmingyang:
According to: https://github.com/cockpit-project/bots/issues/3275, change testBasicWheelUser to testBasicAdminUser, and use "sudo" group for debian based distro

martinpitt:
Thanks @yunmingyang ! I squashed the two topmost commits to get individually working commits in this PR series, and did a few grammar tweaks. I also rebased to current main so that the packit tests can run.

martinpitt (dismissed):
LGTM once green.

martinpitt:
Test failures on [debian-stable](https://logs.cockpit-project.org/logs/pull-669-20220425-101042-7a78907f-debian-stable/log.html) and [rhel-8-6](https://logs.cockpit-project.org/logs/pull-669-20220425-101054-7a78907f-rhel-8-6/log.html)

martinpitt:
This needs to adjust the Testing Farm exclusions as well, [c9s already timed out](https://artifacts.dev.testing-farm.io/87d78f47-07cf-44cb-9883-c2bb87f92a37/).

I can't reproduce the rhel-8-6 failure, that just looks like a timeout on our busy CI (we run on a fallback today). 

The debian-stable crashes are real, and have long been known as https://github.com/cockpit-project/bots/issues/2220 . The naughty patterns need to be adjusted to also match the renamed and new test case.

martinpitt:
I adjusted the pattern in https://github.com/cockpit-project/bots/pull/3297 , running test against that; also retrying rhel-8-6.

Once the bots PR is in, I'll rebase and adjust the TF tests, too.

martinpitt:
The [testDelete](https://logs.cockpit-project.org/logs/pull-669-20220425-111653-7a78907f-debian-stable-bots-3297/log.html#73) failure happened again, seems the different timing on our AWS instances exposes this quite a bit. But I can reproduce this locally as well, I'll have a quick look.

martinpitt (approved):
LGTM now. @skobyda , can you please re-review as well?

PR #671, opened 2022-04-19
cockpit transclusion simplification

allisonkarlitskaya:
Simplify and unify our support for including files from the [cockpit-project/cockpit](https://github.com/cockpit-project/cockpit) repository.

Use the increased flexibility to also import `tools/make-bots` and use it.

One drawback is that we now have to depend on a single version of the cockpit repository, and indeed this PR bumps our `pkg/lib` depend from `262` to `264-137-gd1a410cf9`.

 - [x] Land after PR #672
 - [x] Land after PR #673

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,8 @@ VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 NODE_MODULES_TEST=node_modules/.bin/webpack
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/manifest.json
-# one example file in src/lib to check if it was already checked out
-LIB_TEST=src/lib/cockpit-po-plugin.js
+# one example file in pkg/lib to check if it was already checked out
+LIB_TEST=pkg/lib/cockpit-po-plugin.js
 
 WEBLATE_REPO=tmp/weblate-repo
 WEBLATE_REPO_URL=https://github.com/cockpit-project/cockpit-machines-weblate.git
@@ -127,7 +127,7 @@ $(TARFILE): export NODE_ENV=production
 $(TARFILE): $(WEBPACK_TEST) $(SPEC) packaging/arch/PKGBUILD packaging/debian/changelog
 	tar --xz -cf $(TARFILE) --transform 's,^,cockpit-$(PACKAGE_NAME)/,' \
 		--exclude '*.in' --exclude test/reference \
-		$$(git ls-files) src/lib/ package-lock.json $(SPEC) packaging/arch/PKGBUILD packaging/debian/changelog dist/
+		$$(git ls-files) pkg/lib/ package-lock.json $(SPEC) packaging/arch/PKGBUILD packaging/debian/changelog dist/
 
 $(NODE_CACHE): $(NODE_MODULES_TEST)
 	tar --xz -cf $@ node_modules
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ NODE_MODULES_TEST=node_modules/.bin/webpack
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/manifest.json
 # one example file in pkg/lib to check if it was already checked out
-LIB_TEST=pkg/lib/cockpit-po-plugin.js
+COCKPIT_REPO_STAMP = pkg/lib/cockpit.js
 
 WEBLATE_REPO=tmp/weblate-repo
 WEBLATE_REPO_URL=https://github.com/cockpit-project/cockpit-machines-weblate.git
@@ -87,7 +87,7 @@ packaging/arch/PKGBUILD: packaging/arch/PKGBUILD.in
 packaging/debian/changelog: packaging/debian/changelog.in
 	sed 's/VERSION/$(VERSION)/' $< > $@
 
-$(WEBPACK_TEST): $(LIB_TEST) $(shell find src/ -type f) package.json webpack.config.js
+$(WEBPACK_TEST): $(COCKPIT_REPO_STAMP) $(shell find src/ -type f) package.json webpack.config.js
 	test/download-dist $${DOWNLOAD_DIST_OPTIONS:-} || \
 	    if [ -z "$$FORCE_DOWNLOAD_DIST" ]; then \
 		($(MAKE) $(NODE_MODULES_TEST) && NODE_ENV=$(NODE_ENV) node_modules/.bin/webpack); \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -200,3 +180,21 @@ $(NODE_MODULES_TEST): package.json
...
+# checkout common files from Cockpit repository required to build this project;
+# this has no API stability guarantee, so check out a stable tag when you start
+# a new project, use the latest release, and update it from time to time
+COCKPIT_REPO_FILES = \
+	pkg/lib \
+	test/common \
+	$(NULL)
+
+COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
+COCKPIT_REPO_SHA = 4f8ad7314676b375fcca5c98a543ef5078cc1af5 # 267
```
martinpitt, line 190: Bumping to ≥ 267 requires some adjustments first, see PR #672. Adding blocked for that.

allisonkarlitskaya: I dropped back down to 264-137-gd1a410cf9.  Is that okay?

Makefile
```diff
@@ -87,7 +85,7 @@ packaging/arch/PKGBUILD: packaging/arch/PKGBUILD.in
 packaging/debian/changelog: packaging/debian/changelog.in
 	sed 's/VERSION/$(VERSION)/' $< > $@
 
-$(WEBPACK_TEST): $(LIB_TEST) $(shell find src/ -type f) package.json webpack.config.js
+$(WEBPACK_TEST): pkg/lib $(shell find src/ -type f) package.json webpack.config.js
```
martinpitt, line 88: I deliberately used file dependencies here instead of directory; I had the impression/experience that make dependencies for directories are rather unreliable.  Also, there is no actual rule to produce `pkg/lib`, so it feels like this would just fail if pkg/lib is not present?

Was there something particularly wrong with `LIB_TEST=pkg/lib/cockpit-po-plugin.js`? You introduce `COCKPIT_REPO_STAMP` below with the same value, ISTM that the definition should move to the top and be used here as well.

Makefile
```diff
@@ -200,3 +182,21 @@ $(NODE_MODULES_TEST): package.json
...
+# a new project, use the latest release, and update it from time to time
+COCKPIT_REPO_FILES = \
+	pkg/lib \
+	test/common \
+	$(NULL)
+
+COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
+COCKPIT_REPO_SHA = d1a410cf9ccb33e983fbd38cad5a6fc95bc1c4b4 # 264-137-gd1a410cf9
+
+COCKPIT_REPO_STAMP = pkg/lib/cockpit-po-plugin.js
```
martinpitt, line 194: See above, let's move this up and re-use in the webpack rule.

Makefile
```diff
@@ -200,3 +182,21 @@ $(NODE_MODULES_TEST): package.json
...
+$(COCKPIT_REPO_FILES): $(COCKPIT_REPO_STAMP)
+$(COCKPIT_REPO_STAMP): Makefile
+	@git rev-list --quiet --objects '$(strip $(COCKPIT_REPO_SHA))^{tree}' -- 2>/dev/null || \
```
martinpitt, line 197: That strip looks weird, that's for cutting off the `#264 ...` comment? Let's just move the comment into a new line.

Makefile
allisonkarlitskaya: Ya.  I can put it back.  It's possible to imagine a situation where the `Makefile` has been modified to check out a different version, but `make` considers `pkg/lib` newer because some random editor temp file got left there...

It just feels a bit evil to randomly elect a file for this purpose.  I guess now that I use 267, `cockpit.js` is a good choice.  Although, ironically: it's the file that we *don't* use :)

Makefile
allisonkarlitskaya: ya ya ya

Makefile
allisonkarlitskaya: What's going on here is that `make` parses the line up to the `#` — everything after is ignored.  The problem is that the space before the `#` gets included in the variable.  That's what's being stripped.

I prefer the one line thing.  If you're strongly against it I can change it, but I think it's the best way to do this, and the `$(strip)` thing isn't bad — in fact, I also applied it in some places where it wasn't strictly necessary.

martinpitt: Not strongly, but it makes the already complicated code here even more unwieldy, plus it gets repeated a lot. Perhaps move it into a separate line/new variable?

allisonkarlitskaya: I added an extra variable and removed the 'extra' uses of `$(strip)` that were unnecessary (because the shell already absorbs the whitespace).

martinpitt (changes requested):
Thanks! One tiny net-zero change away from awesome :grin: 

I just realized that after that we can also re-use cockpit's test/static-code, instead of the current (reduced) copy.

Makefile
```diff
@@ -200,5 +197,5 @@ $(COCKPIT_REPO_FILES): $(COCKPIT_REPO_STAMP)
 COCKPIT_REPO_TREE = '$(strip $(COCKPIT_REPO_COMMIT))^{tree}'
 $(COCKPIT_REPO_STAMP): Makefile
 	@git rev-list --quiet --objects $(COCKPIT_REPO_TREE) -- 2>/dev/null || \
-	    git fetch --no-tags --no-write-fetch-head --depth=1 $(COCKPIT_REPO_URL) $(COCKPIT_REPO_SHA)
+	    git fetch --no-tags --no-write-fetch-head --depth=1 $(COCKPIT_REPO_URL) $(COCKPIT_REPO_COMMIT)
```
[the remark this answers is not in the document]
allisonkarlitskaya: oops.  thanks.

Makefile
```diff
@@ -200,3 +184,21 @@ $(NODE_MODULES_TEST): package.json
...
+
+COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
+COCKPIT_REPO_COMMIT = 4f8ad7314676b375fcca5c98a543ef5078cc1af5 # 267
+
+$(COCKPIT_REPO_FILES): $(COCKPIT_REPO_STAMP)
+COCKPIT_REPO_TREE = '$(strip $(COCKPIT_REPO_COMMIT))^{tree}'
+$(COCKPIT_REPO_STAMP): Makefile
+	@git rev-list --quiet --objects $(COCKPIT_REPO_TREE) -- 2>/dev/null || \
+	    git fetch --no-tags --no-write-fetch-head --depth=1 $(COCKPIT_REPO_URL) $(COCKPIT_REPO_COMMIT)
+	git archive $(COCKPIT_REPO_TREE) -- tools/git-utils.sh $(COCKPIT_REPO_FILES) | tar x
```
martinpitt, line 201: Why special-case git-utils.sh? Shouldn't that just go into COCKPIT_REPO_FILES? This commit doesn't even use it -- the next one does, but that *does* add git-utils.sh to the var list. This is possibly just a leftover?

martinpitt (approved):
Thanks! Please land as-is if you want to. I'm going to send a follow-up for test/static-code, and can clean up the redundant git-utils.sh along with that. Otherwise, I stand by for quick re-approval :grin:

Makefile
allisonkarlitskaya: Thanks for the catch.  Handled as a follow-up: https://github.com/cockpit-project/cockpit-machines/pull/677

PR #672, opened 2022-04-19
test: Bump Cockpit test API to 267, use --track-naughties option

martinpitt:
run-tests recently got a new `--track-naughties` option [1] to explicitly
enable updating known issues on GitHub. Enable it for CI runs, and leave
it disabled for local runs and PackIt, as in these cases we don't want
to try and talk to GitHub.

[1] https://github.com/cockpit-project/cockpit/commit/7ef88e80ce613c

Cherry-picked from starter-kit commit a937b82cd2dc.

### DIFF
--- a/test/run
+++ b/test/run
@@ -4,6 +4,8 @@
 
 set -eu
 
+export RUN_TESTS_OPTIONS=--track-naughties
+
 make codecheck
 make check
 make po/machines.pot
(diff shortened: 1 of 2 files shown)


test/run
```diff
@@ -4,6 +4,8 @@
 
 set -eu
 
+export RUN_TESTS_OPTIONS=--track-naughties
```
skobyda, line 7: Why do we export that option here instead `--track-naughties` directly at Makefile. Do we expect more logic to be added to `test/run` which will make RUN_TESTS_OPTIONS options more dynamic/variable?

martinpitt: Because the idea of that starter-kit commit was that running `make check` as a developer does *not* talk to github. For most people it doesn't work anyway, and it's somewhat unexpected even for us.

skobyda (approved):

PR #678, opened 2022-04-21
Better handle delete resources modal messages

marusak:
The current approach had a few issues:
- Combining strings with `+` can lead to untranslatable situations
- `objectType` was not localized
- The title was title case (but should be sentence case)
- The error message always said `deleted` even when action was `Remove`
- Even if we would translate `objectType` it could lead to situations
  when the final sentence would not be correct, or correctly capitalized
  (for example `The network could not be deleted` is Czech would start
   with `network` as we don't have articles and thus the sentence would
   start with lowercase letter).

### DIFF
--- a/src/components/vm/hostdevs/hostDevCard.jsx
+++ b/src/components/vm/hostdevs/hostDevCard.jsx
@@ -231,8 +231,8 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
                     <DeleteResourceButton objectId={`${id}-hostdev-${hostdevId}`}
                                           actionName={_("Remove")}
                                           showDialog={() => setDeleteDialogProps({
-                                              objectType: cockpit.format(_("$1 host device"), hostdev.type),
-                                              objectName: "",
+                                              title: _("Remove host device"),
+                                              errorMessage: ("Host device could not be removed"),
                                               onClose: () => setDeleteDialogProps(undefined),
                                               deleteHandler: () => {
                                                   // refresh nodeDevice since usb number may be changed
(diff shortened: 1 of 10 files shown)


src/components/vm/hostdevs/hostDevCard.jsx
```diff
@@ -231,8 +231,8 @@ export const VmHostDevCard = ({ vm, nodeDevices, config }) => {
                     <DeleteResourceButton objectId={`${id}-hostdev-${hostdevId}`}
                                           actionName={_("Remove")}
                                           showDialog={() => setDeleteDialogProps({
-                                              objectType: cockpit.format(_("$1 host device"), hostdev.type),
-                                              objectName: "",
+                                              title: _("Remove host device"),
```
marusak, line 234: Here I dropped the `hostdev.type`. Is is necessary or any useful? I think we maybe should add `actionDescription` and show what is going to be removed

skobyda: I'm attending exactly this issue right now at https://github.com/cockpit-project/cockpit-machines/pull/652 where I add actionDescription, so no need to be added at this PR

skobyda (approved):
I'm doing also some changes to titles at https://github.com/cockpit-project/cockpit-machines/pull/652
But it's true I didn't take into account the translatability of strings if they are formatted  like `objectType + " will be removed"`.
So I'm fine with merging this and then just ammending https://github.com/cockpit-project/cockpit-machines/pull/652 to accomodate for these changes.

skobyda (approved):

PR #679, opened 2022-04-21
test: Use test/static-code from Cockpit

martinpitt:
Using our new "pull in arbitrary files from cockpit.git" mechanism, drop
the (reduced) code copy of test/static-code. This was recently [1] fixed
to only allow line lengths up to 200, which is compatible enough with
the limit of 195 that we had in c-machines.

[1] https://github.com/cockpit-project/cockpit/pull/17268


martinpitt:
[c9s run](https://artifacts.dev.testing-farm.io/edae8df3-012f-475a-b0ee-4618fe6d0cbb/) timed out, retrying. (Obviously unrelated)

Makefile
```diff
@@ -186,12 +186,13 @@ $(NODE_MODULES_TEST): package.json
 COCKPIT_REPO_FILES = \
 	pkg/lib \
 	test/common \
+	test/static-code \
 	tools/git-utils.sh \
 	tools/make-bots \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
-COCKPIT_REPO_COMMIT = 4f8ad7314676b375fcca5c98a543ef5078cc1af5 # 267
```
allisonkarlitskaya, line 194: I was imagining we put version numbers like `267-21-gb19b07757` here, but I suppose this is also OK...

allisonkarlitskaya (approved):
btw: magic number here is 118 eventually, hopefully.  I hope to continue to move cockpit in this direction, ratcheting down the number in `test/static-code` as I go...

That means that we're increasing the number of "annoyances" that we might impose on ourselves when updating.

PR #682, opened 2022-04-22
gitignore: add gitignore

yunmingyang:
Add .idea and .vscode to toplevel
Add  test/static-code to subdirs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@
 /po/LINGUAS
 /test/common/
 /test/images/
+/test/static-code


.gitignore
```diff
@@ -22,6 +22,8 @@
 /pkg/
 /tmp/
 /tools/
+/.idea
+/.vscode
```
martinpitt, line 26: Hmm, this rather belongs into your global config, doesn't it? Adding every editor's temporary files to every project is too much clutter IMHO.

yunmingyang: OK, then I will remove them

.gitignore
```diff
@@ -30,3 +32,4 @@
 /po/LINGUAS
 /test/common/
 /test/images/
+/test/static-code
```
skobyda, line 35: Why would this be part of gitignore? It's cockpit's custom code there which we might want to update in the future for the whole repo.

.gitignore
skobyda: I agree that this should be part of each developer's global config. There then is much more IDEA and OS folders to be gitignored specific for each developer, such as .DS_Store, and I don't feel like getting into this rabbit hole

.gitignore
skobyda: Ah right, this should be mantained in cockpit's repo, and not c-machines. Then I agree that this can be gitignored.

martinpitt (approved):
Thanks!

PR #683, opened 2022-04-22
Fix various issue with VM memory and storage size autodetection

### LINKED ISSUE #612: Failed to import vm with Alpine Linux 3.14 (ERROR Error: --memory 137.5367431640625: invalid literal for int() with base 10: '137.5367431640625')
Cockpit version: 
cockpit-264-1.fc35.x86_64
cockpit-machines-263-1.fc35.noarch
OS:Fedora35
Page: Vitual Machines
Chrome:Version 99.0.4844.51 (Official Build) (64-bit)

Failed to import vm with Alpine Linux 3.14 (ERROR Error: --memory 137.5367431640625: invalid literal for int() with base 10: '137.5367431640625')

Steps to reproduce
1. Prepare a vm image like cirros
2. Click "Import VM", and input needed field and "Alpine Linux 3.14" as operation system
3. Click Import
Creation of VM cirros failed 
ERROR Error: --memory 137.5367431640625: invalid literal for int() with base 10: '137.5367431640625'

skobyda:
Fixes #612

1. Reset memory and storage size if it cannot be autodetected for OS
This fixes the following issue: When user selects an operating system for which we know minimum required memory (e.g. centos), and then switched it to OS for which we don't know minimum memory (e.g. openBSD), old defaults are left from previous operating system, and even incorrect warning message "The selected operating system has minimum memory requirement of N MiB" could be shown.
2. The memory of a new VM, specified in MiB, cannot be a float: virt-install doesn't allow memory, supplied in MiB, to be a float number. Check if number is a integer, and if not, round up (ceil).
3. Don't use default memory for alpinelinux and hyperbola: Hyperbola and some version of alpinelinux have incorrect minimum memory
Bug report: https://gitlab.com/libosinfo/osinfo-db/-/issues/95

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -817,6 +817,15 @@ class CreateVmModal extends React.Component {
         case 'storageVolume':
             this.setState({ [key]: value });
             break;
+        case 'memorySize': {
+            // virt-install doesn't allow memory, supplied in MiB, to be a float number
+            // check if number is a integer, and if not, round up (ceil)
+            const valueMiB = Math.ceil(convertToUnit(value, this.state.memorySizeUnit, units.MiB));
+            value = convertToUnit(valueMiB, units.MiB, this.state.memorySizeUnit);
+
+            this.setState({ [key]: value });
+            break;
+        }
         case 'memorySizeUnit':
             this.setState({ [key]: value });
             key = 'memorySize';
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -133,6 +133,29 @@ function getSpaceAvailable(storagePools, connectionName) {
     return space;
 }
 
+function getMemoryDefaults(nodeMaxMemory) {
+    // Use available memory on host as initial default value in modal
+    let memorySizeUnit = units.MiB.name;
+    let memorySize = nodeMaxMemory && Math.floor(convertToUnit(nodeMaxMemory, units.KiB, units.MiB));
+    // If available memory is higher than 1GiB, set 1GiB as default value in modal
+    if (memorySize > 1024) {
+        memorySizeUnit = units.GiB.name;
+        memorySize = 1;
+    }
+
+    const minimumMemory = 0;
+
+    return { memorySize, memorySizeUnit, minimumMemory };
+}
+
+function getStorageDefaults() {
+    const storageSize = convertToUnit(10 * 1024, units.MiB, units.GiB); // tied to Unit
+    const storageSizeUnit = units.GiB.name;
+    const minimumStorage = 0;
+
+    return { storageSize, storageSizeUnit, minimumStorage };
+}
+
 function validateParams(vmParams) {
     const validationFailed = {};
 
@@ -597,6 +620,7 @@ const MemoryRow = ({ memorySize, memorySizeUnit, nodeMaxMemory, minimumMemory, o
                                onChange={value => onValueChanged('memorySize', Number(value))} />
                     <FormSelect id="memory-size-unit-select"
                                 className="unit-select"
+                                data-value={memorySizeUnit}
                                 value={memorySizeUnit}
                                 onChange={value => onValueChanged('memorySizeUnit', value)}>
                         <FormSelectOption value={units.MiB.name} key={units.MiB.name}
@@ -728,15 +752,11 @@ class CreateVmModal extends React.Component {
             sourceType: defaultSourceType,
             source: '',
             os: undefined,
-            memorySize: props.nodeMaxMemory ? Math.min(1, convertToUnit(props.nodeMaxMemory, units.KiB, units.GiB)) : 1,
-            memorySizeUnit: units.GiB.name,
-            storageSize: convertToUnit(10 * 1024, units.MiB, units.GiB), // tied to Unit
-            storageSizeUnit: units.GiB.name,
+            ...getMemoryDefaults(props.nodeMaxMemory),
+            ...getStorageDefaults(),
             storagePool: 'NewVolume',
             storageVolume: '',
             startVm: true,
-            minimumMemory: 0,
-            minimumStorage: 0,
 
             // Unattended installation or cloud init options for cloud images
             profile: '',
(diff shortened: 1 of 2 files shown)


src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -848,7 +848,10 @@ class CreateVmModal extends React.Component {
             if (value && value.profiles)
                 stateDelta.profile = value.profiles.sort().reverse()[0];
 
-            if (value && value.minimumResources.ram) {
+            // Hyperbola and some version of alpinelinux have incorrect minimum memory
+            // https://gitlab.com/libosinfo/osinfo-db/-/issues/95
+            if (value && value.minimumResources.ram &&
+                !["alpinelinux3.15", "alpinelinux3.15", "hyperbola03"].includes(value.shortId)) {
```
martinpitt, line 877: What does it use instead? It seems to me that with the rounding fix this should be sufficiently unbroken. The user can still tweak it, and once it gets fixed in libosinfo, we don't have to revert the hack (especially as nothing will tell us when it gets fixed).

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -817,6 +817,17 @@ class CreateVmModal extends React.Component {
         case 'storageVolume':
             this.setState({ [key]: value });
             break;
+        case 'memorySize': {
+            // virt-install doesn't allow memory, supplied in MiB, to be a float number
+            // check if number is a integer, and if not, round up (ceil)
+            let valueMiB = convertToUnit(value, this.state.memorySizeUnit, units.MiB);
+            if (valueMiB % 1 !== 0)
+                valueMiB = valueMiB.ceil();
```
martinpitt, line 837: Comparing floating point numbers to exact integers is unreliable in general, even if it should work for the special case of 0. Let's just drop the `if()`, and always call ceil(), it doesn't hurt. Then you can also make `valueMiB` `const`.

Also, I've never seen this being used as a method of a number object, and this doesn't work for me in node:
```
 x = 3
3
> x.ceil()
Uncaught TypeError: x.ceil is not a function
> 
```

If that works in some browsers, let's not rely on that. Please use standard [Math.ceil() function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/ceil)

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -870,6 +882,8 @@ class CreateVmModal extends React.Component {
                 if (bestUnit.base1024Exponent <= 1) bestUnit = units.MiB;
                 const converted = convertToUnit(stateDelta.minimumMemory, units.B, bestUnit);
                 this.setState({ memorySizeUnit: bestUnit.name }, () => this.onValueChanged("memorySize", converted));
+            } else {
+                this.setState(getMemoryDefaults(this.props.nodeMaxMemory));
```
martinpitt, line 886: This makes sense, thanks! Can you please capture this in a test?

martinpitt (changes requested):
Thanks!

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: I added it here because now when its broken in libosinfo, even with rounding fix, it will provide user with incorrect value for minimum resource and it would default to an incorrect value (only rounded up)
In this screenshot, the correct value according to alpine linux docs should be 128MiB, but because of a typo in libosinfo, it says 137MiB:
Screenshot from 2022-04-25 13-17-27
And I think for alpine linux it's better to show no value then incorrect one:
Screenshot from 2022-04-25 13-21-19

skobyda: However it's true that nothing would tell us when it gets fixed. Maybe we can add a assert in test which will warn us in the future?

martinpitt: I still think that proposing 137 MB is not so bad that we have to work around it. It's not wildly off, and the user can still change it. Of course fixing the rounding is important.

skobyda: Alright, dropped the workaround commit for alpine linux, now there is just rounding fix and and fix for the case when storage and memory requirements are not known

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -817,6 +817,17 @@ class CreateVmModal extends React.Component {
         case 'storageVolume':
             this.setState({ [key]: value });
             break;
+        case 'memorySize': {
+            // virt-install doesn't allow memory, supplied in MiB, to be a float number
+            // check if number is a integer, and if not, round up (ceil)
+            let valueMiB = convertToUnit(value, this.state.memorySizeUnit, units.MiB);
+            if (valueMiB % 1 !== 0)
```
martinpitt, line 837: Please drop this `if`, just always wrap the converToUnit() into Math.ceil().

martinpitt (changes requested):
Thanks! Nice test case :clap: 

Can you please rebase this to current main, so that the packit tests can run?

martinpitt:
I just did a clean force-push to re-trigger tests (argh CI infra trouble).

[several](https://logs.cockpit-project.org/logs/pull-683-20220426-045516-a4a81be1-rhel-9-0/log.html#67) [relevant](https://logs.cockpit-project.org/logs/pull-683-20220426-045516-a4a81be1-rhel-8-6/log.html#67) failures -- the [screenshot](https://logs.cockpit-project.org/logs/pull-683-20220426-045516-a4a81be1-rhel-9-0/TestMachinesCreate-testCreateBasicValidation-rhel-9-0-127.0.0.2-2201-FAIL.png) shows invalid floating point numbers.

skobyda:
> I just did a clean force-push to re-trigger tests (argh CI infra trouble).
> 
> [several](https://logs.cockpit-project.org/logs/pull-683-20220426-045516-a4a81be1-rhel-9-0/log.html#67) [relevant](https://logs.cockpit-project.org/logs/pull-683-20220426-045516-a4a81be1-rhel-8-6/log.html#67) failures -- the [screenshot](https://logs.cockpit-project.org/logs/pull-683-20220426-045516-a4a81be1-rhel-9-0/TestMachinesCreate-testCreateBasicValidation-rhel-9-0-127.0.0.2-2201-FAIL.png) shows invalid floating point numbers.

Ah the test showed another bug, that on some systems maximum available memory on host can also be float MiB number.
So now maximum available memory is always floored when set as default for dialog. And adjusted tests for that.

martinpitt:
Pushed a clean rebase to run packit tests.

martinpitt:
Pushed again to fix the first commit message.

martinpitt (approved):
Dakujem!

PR #689, opened 2022-04-27
build: Introduce node_modules/ as submodule

allisonkarlitskaya:
Build and maintain the node_modules/ as a submodule, using the
tools/node-modules script as it's done in the cockpit repo.

Updating the node_modules/ directory after changing package.json now
involves running `tools/node-modules install` and then committing the
result, and pushing the submodule commit with `tools/node-modules push`.

Update the npm-update{,-pf}.yml workflows accordingly.


allisonkarlitskaya:
This works in local testing, and the `npm-update` workflows also seems to work:

Update: https://github.com/cockpit-project/cockpit-machines/runs/6192374821?check_suite_focus=true → https://github.com/cockpit-project/cockpit-machines/pull/690
No update: https://github.com/cockpit-project/cockpit-machines/runs/6192360959?check_suite_focus=true

Makefile
```diff
@@ -35,10 +34,10 @@ po/$(PACKAGE_NAME).js.pot:
 		--keyword=gettextCatalog.getString:1,3c --keyword=gettextCatalog.getPlural:2,3,4c \
 		--from-code=UTF-8 $$(find src/ -name '*.js' -o -name '*.jsx')
 
-po/$(PACKAGE_NAME).html.pot: $(NODE_MODULES_TEST) $(COCKPIT_REPO_STAMP)
+po/$(PACKAGE_NAME).html.pot: package-lock.json $(COCKPIT_REPO_STAMP)
```
martinpitt, line 37: I liked the `NODE_MODULES_TEST` constant to express the meaning of a dependency and to avoid repetition. But no strong opinion.

martinpitt (approved):
:partying_face: buh-bye, npm install!

This is surprisingly unintrusive, thanks!

Makefile
allisonkarlitskaya: I more or less copied the rule directly from the cockpit project, and it really absolutely needs to be this file for it to work with the existing script...

I feel like depending on `package-lock.json` is a pretty clear indication of what's going on here.

PR #692, opened 2022-04-28
Makefile: Fix tools/node-modules dependency

martinpitt:
It first needs to be checked out before we can call it. This fixes
`make po/machines.pot` in a clean checkout (as the weblate-sync-pot
workflow does).

---

Last night's [weblate-sync-pot workflow failed](https://github.com/cockpit-project/cockpit-machines/runs/6204918134?check_suite_focus=true). I [triggered a run against this branch](https://github.com/cockpit-project/cockpit-machines/runs/6205008238?check_suite_focus=true) and it succeeded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ test/reference: test/common
 # We want tools/node-modules to run every time package-lock.json is requested
 # See https://www.gnu.org/software/make/manual/html_node/Force-Targets.html
 FORCE:
-package-lock.json: FORCE
+package-lock.json: FORCE tools/node-modules
 	tools/node-modules make_package_lock_json
 
 .PHONY: all clean install devel-install dist rpm check vm


Makefile
```diff
@@ -144,7 +144,7 @@ test/reference: test/common
 # We want tools/node-modules to run every time package-lock.json is requested
 # See https://www.gnu.org/software/make/manual/html_node/Force-Targets.html
 FORCE:
-package-lock.json: FORCE
+package-lock.json: FORCE tools/node-modules
```
allisonkarlitskaya, line 147: This `FORCE` stuff is sufficiently magical that I wondered if this would still work, but the short answer is: it does.

I created an artificial rule trivially dependent on `package-lock.json` to test things, and in every case it performs correctly (and every `make` invocation runs the hook correctly).

allisonkarlitskaya (approved):
Thanks :+1:

PR #693, opened 2022-04-28
Clarify difference between VM creation and import

skobyda:
In https://github.com/cockpit-project/cockpit-machines/issues/612, which was recently closed because the initial bug which was reported was fixed, @garrett  mentioned few issues with Create VM vs Import VM buttons. Let's fixed them immediately so it doesn't get forgotten about:

> In other words, I think there are these issues:
>     0. Adjust Create vs. Import.
>        
>        * Swap Create and Import buttons, so Create is in the default position (on the right)

Done:
Screenshot from 2022-04-28 13-37-56

>        * Perhaps we should consider having some tooltips (when hovering the buttons) that talk about creation being an installation process (such as with an iso) whereas import is using a disk image as-is.

I took some text from [redhat docs](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_virtualization/getting-started-with-virtualization-in-rhel-8_configuring-and-managing-virtualization#creating-vms-using-the-rhel-8-web-console_creating-vms-and-installing-an-os-using-the-rhel-8-web-console) and adjusted them for our tooltip messages, but I'm of course open to discussion about them:
Screenshot from 2022-04-28 14-42-04
Screenshot from 2022-04-28 14-41-19

Tooltips positions are auto "auto", and there is not enough space on the screen for the tooltip for "Create VM" to do anything else than align to the left. Should we hard-code the other tooltip to also align to the left?

>        * Possibly think about "Install" instead of "Create"? (Not sure about this one; just throwing an idea out there)

I'm not sure about this, as "installation" is a different phase, which you can avoid (or postpone to be more precise) by clicking "Create and edit". So the button would be called "Install", bu it would open a dialog where user can choose to _not install_

We would end up with following flow:

1. There would be button "Install", which opens a dialog

Screenshot from 2022-04-28 14-47-57

2. You fill out the dialog, but decide not to install VM immediatelly, but instead click on "Create and edit"
3. You end up on a VM page and there is another "Install" button
Screenshot from 2022-04-28 14-47-37

Having "Install" twice in the same flow feels a bit weird to me, but that's just my subjective opinion.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -212,7 +212,7 @@ class AppActive extends React.Component {
         };
         const createVmAction = <CreateVmAction {...properties} mode='create' />;
         const importDiskAction = <CreateVmAction {...properties} mode='import' />;
-        const vmActions = <> {createVmAction} {importDiskAction} </>;
+        const vmActions = <> {importDiskAction} {createVmAction} </>;
         const pathVms = path.length == 0 || (path.length > 0 && path[0] == 'vms');
 
         let vmContent;
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -1217,7 +1217,7 @@ export class CreateVmAction extends React.Component {
         let createButton = (
             <Button isDisabled={testdata !== undefined}
                     testdata={testdata}
-                    id={this.props.mode == 'create' ? 'create-new-vm' : 'import-vm-disk'}
+                    id={this.props.mode == 'create' ? 'create-new-vm' : 'import-existing-vm'}
                     variant='secondary'
                     onClick={this.open}>
                 {this.props.mode == 'create' ? _("Create VM") : _("Import VM")}
@@ -1227,6 +1227,21 @@ export class CreateVmAction extends React.Component {
             createButton = (
                 <Tooltip id='virt-install-not-available-tooltip'
                          content={_("virt-install package needs to be installed on the system in order to create new VMs")}>
+                    <span>
+                        {createButton}
+                    </span>
+                </Tooltip>
+            );
+        else
+            createButton = (
+                <Tooltip id={this.props.mode + '-button-tooltip'}
+                         position={this.props.mode === "create" ? "top-end" : "top"}
+                         className={this.props.mode === "create" && "custom-arrow"}
+                         content={this.props.mode === "create"
+                             ? _("Create VM from local or network installation medium")
+                             : _("Create VM by importing a disk image of an existing VM installation")
+                         }
+                         isContentLeftAligned>
                     <span>
                         {createButton}
                     </span>
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -37,7 +37,7 @@ from machinesxmls import NETWORK_XML_PXE, PXE_SERVER_CFG  # noqa
 @nondestructive
 class TestMachinesCreate(VirtualMachinesCase):
 
-    # This test contains basic form validation of the Create VM dialog
+    # This test contains basic button and form validation of the Create VM dialog
     # None of the sub-tests will actually call virt-install
     def testCreateBasicValidation(self):
         runner = TestMachinesCreate.CreateVmRunner(self)
@@ -49,6 +49,9 @@ class TestMachinesCreate(VirtualMachinesCase):
 
         self.login_and_go("/machines")
         self.browser.wait_in_text("body", "Virtual machines")
+
+        # test create and import [VM] buttons show appropriate tooltips when hovered over
+        runner.createAndImportTooltipsTest()
 
         # test just the DIALOG CREATION and cancel
         print("    *\n    * validation errors and ui info/warn messages expected:\n    * ")
@@ -775,7 +778,7 @@ vnc_password= "{vnc_passwd}"
             b = self.browser
 
             if self.sourceType == 'disk_image':
-                b.click("#import-vm-disk")
+                b.click("#import-existing-vm")
             else:
                 b.click("#create-new-vm")
 
@@ -1544,6 +1547,18 @@ vnc_password= "{vnc_passwd}"
                 self._deleteVm(dialog)
                 if check_env_empty:
                     self.checkEnvIsEmpty(dialog)
+
+        def createAndImportTooltipsTest(self):
+            b = self.browser
+
+            b.wait_visible("#import-existing-vm:not(:disabled)")
+            b.mouse("#import-existing-vm", "mouseenter")
+            b.wait_in_text("#import-button-tooltip", "existing")
+            b.mouse("#import-existing-vm", "mouseleave")
+            b.wait_visible("#create-new-vm:not(:disabled)")
+            b.mouse("#create-new-vm", "mouseenter")
+            b.wait_in_text("#create-button-tooltip", "medium")
+            b.mouse("#create-new-vm", "mouseleave")
 
         def cancelDialogTest(self, dialog):
             dialog.open() \
--- a/src/components/create-vm-dialog/createVmDialog.scss
+++ b/src/components/create-vm-dialog/createVmDialog.scss
@@ -1,3 +1,7 @@
 .unit-select, .size-input {
     max-width: 5rem;
 }
+
+.pf-c-tooltip__arrow {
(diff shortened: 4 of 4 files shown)


skobyda:
Seems to be passing tests (apart from packit which ddn't finish yet)
@martinpitt @garrett mind giving it a review?

src/app.jsx
```diff
@@ -212,7 +212,7 @@ class AppActive extends React.Component {
         };
         const createVmAction = <CreateVmAction {...properties} mode='create' />;
         const importDiskAction = <CreateVmAction {...properties} mode='import' />;
-        const vmActions = <> {createVmAction} {importDiskAction} </>;
+        const vmActions = <> {importDiskAction} {createVmAction} </>;
```
martinpitt, line 215: FWIW, I have no strong opinion about this (i.e. no veto), but it feels strange to me. It breaks muscle memory, both are secondary, and one is not obviously "more important" than the other.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1210,6 +1210,18 @@ export class CreateVmAction extends React.Component {
                     </span>
                 </Tooltip>
             );
+        else
+            createButton = (
+                <Tooltip id={this.props.mode + '-button-tooltip'}
+                         content={this.props.mode === "create"
+                             ? _("Create virtual machine from local installation medium, URL, PXE network boot, cloud base image, or download an OS")
```
martinpitt, line 1217: My "bad grammar nerve" tickles here. "Create VM from .. or download an OS" reads weird; perhaps that needs a semicolon. But either way this feels rather long, these poor translators. Could we shorten this a bit to perhaps: "Create VM from local or network installation medium"?

Also, these tooltips are sentences and should end in a period, but there might be some different conventions here.

test/check-machines-create
```diff
@@ -1482,6 +1485,18 @@ vnc_password= "{vnc_passwd}"
                 self._deleteVm(dialog) \
                     .checkEnvIsEmpty(dialog)
 
+        def createAndImportTooltipsTest(self):
+            b = self.browser
+
+            b.wait_visible("#import-existing-vm:not(:disabled)")
+            b.mouse("#import-existing-vm", "mouseenter")
+            b.wait_in_text("#import-button-tooltip", "Create virtual machine by importing a disk image of an existing VM installation")
```
martinpitt, line 1493: Feels a bit excessive to repeat the whole string here -- checking for "existing" seems enough, and likewise for perhaps "medium" for the create-new one.

martinpitt (commented):
Thanks! No veto, just some suggestions. This feels more like a @garrett PR for the grammar/wording fine-tuning.

src/app.jsx
garrett: Default actions are more on the right.

You can create a VM without anything but Cockpit and Cockpit Machines. But you cannot import a VM unless you have a disk image somewhere, needing more from the user. Hence Create being on the right, as originally designed, and import being on the left. The fact that they were swapped in implementation is a bug (and a longstanding one).

src/components/create-vm-dialog/createVmDialog.jsx
garrett: Yeah, the strings are quite long. I'll try to shorten them. I like your suggestion here.

```suggestion
                             ? _("Create VM from local or network installation medium")
```

garrett: We spell out VM elsewhere; I wonder if this needs it or if VM is fine enough. Either way, here's the expanded form:

```suggestion
                             ? _("Create virtual machine from local or network installation medium")
```

test/check-machines-create
garrett: We need to decide whether we want to say "VM" or "virtual machine" in the create and import buttons and be consistent.

```suggestion
            b.wait_in_text("#import-button-tooltip", "Import disk image of an existing installation to create a new virtual machine")
```

garrett:
Ideally, the tooltips would both be on the top, not obstructing the buttons they refer to or the other button, and the pointer part would be able to shift, so we'd have

For the center-aligned tooltip for the button on the left:

⎺⎺⎺⎺v⎺⎺⎺⎺

and for the button on the right, it could be right-aligned:

⎺⎺⎺⎺⎺⎺⎺v⎺

I don't think PF supports moving the caret, however. (CSS would of course, so we could do an override and suggest something for upstream if PF indeed does not support this..)

src/components/create-vm-dialog/createVmDialog.jsx
martinpitt: I wanted to make it consistent with "existing VM installation", and reduce the string as well. But of course I'm fine with spelling it out.

test/check-machines-create
martinpitt: FTR, the string change needs to go into the test, not the code. That's why I said the test shouldn't repeat the long string, just check for a significant word/short phrase.

skobyda:
> For the center-aligned tooltip for the button on the left:
> 
> ⎺⎺⎺⎺v⎺⎺⎺⎺
> 
> and for the button on the right, it could be right-aligned:
> 
> ⎺⎺⎺⎺⎺⎺⎺v⎺

So I played around with it a bit, and PF allows us to set position to "top-end". So if we have position as "top" for left button, and "top-end" for right button, it would look like this:
Screenshot from 2022-05-09 11-25-22
Screenshot from 2022-05-09 11-25-03

Now, the position of the caret (PF calls it "popper", or just an "arrow") on the right button is not centered around the button. That's a bit tricky, as whole tooltip has `position: absolute`, and PF just calculates positions of tooltip based on position of button. However we can manually change the position of arrow with class `.pf-c-tooltip__arrow`.
My naive approach was to just change the `right` value to `50px`, as that is the half of length of the right button.
```
.pf-c-tooltip__arrow {
    right: 50px !important;
}
```
Which will create this look:
Screenshot from 2022-05-09 11-25-31

But I guess it's not nice to hardcode "50px" value. Maybe @garrett, any idea for a nicer solution on how to move the arrow?
Or mabe we can leave hardcoded value temporarily, ask PF to provide better solution to move the arrow, and then switch to their solution once they implement it?

garrett (changes requested):
Thanks for fixing the overlapping! It looks correct now:

Screenshot from 2022-05-11 14-17-30

Screenshot from 2022-05-11 14-17-36

Can you left align the text? You'd need to add in `isContentLeftAligned=true` to the tooltips.

Aside from that, and adjusting the tooltip arrow offset from the right for the Create VM button, I think this is almost ready to land.

garrett:
I've pushed up a small change that re-implements the arrow offset with a PatternFly unit and also by using margin-right instead of redefining right, so that we don't need to adjust the specificity or add `!important`.

skobyda:
Thanks. I squashed the commit since it's related to a previous one.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1232,6 +1232,21 @@ export class CreateVmAction extends React.Component {
                     </span>
                 </Tooltip>
             );
+        else
+            createButton = (
+                <Tooltip id={this.props.mode + '-button-tooltip'}
+                         position={this.props.mode === "create" ? "top-end" : "top"}
+                         className={this.props.mode === "create" && "custom-arrow"}
+                         content={this.props.mode === "create"
+                             ? _("Create VM machine from local or network installation medium")
```
martinpitt, line 1241: "VM machine" is redundant. Either "VM" or "virtual machine", but please keep it consistent with the other string below. (IMHO, "VM" in a tooltip is fine)

martinpitt (changes requested):
Thanks! One tiny nitpick, otherwise LGTM. Leaving the final review to @garrett , as this is mostly a design change.

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: Fixed

martinpitt:
@skobyda : Please rebase, to let the packit tests run.

garrett (dismissed):
LGTM! :+1:

martinpitt (approved):
Cheers! Assuming green, let's land.

PR #696, opened 2022-05-03
Elements visible in VM listing should have connection name in their element ids

skobyda:
In https://github.com/cockpit-project/cockpit-machines/pull/665, I wrote a test which has 2 VMs with same name, but on different connection.
This unearthed many places in code where we don't put connectionName to an element id.
In our tests, this creates cases when referring to elements in VM list is ambigious (as there may be multiple VMs with same id)

### DIFF
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -25,7 +25,7 @@ distrosWithMonolithicDaemon = ["rhel-8-6", "ubuntu-stable", "ubuntu-2204", "debi
 class VirtualMachinesCaseHelpers:
     created_pool = False
 
-    def performAction(self, vmName, action, checkExpectedState=True):
+    def performAction(self, vmName, action, checkExpectedState=True, connectionName="system"):
         b = self.browser
         b.click("#vm-{0}-action-kebab button".format(vmName))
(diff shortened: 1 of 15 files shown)


test/machineslib.py
```diff
@@ -25,21 +25,21 @@
...
 
         if action == "pause":
-            b.wait_in_text("#vm-{0}-state".format(vmName), "Paused")
+            b.wait_in_text("#vm-{0}-{1}-state".format(vmName, connectionName), "Paused")
         if action == "resume" or action == "run":
-            b.wait_in_text("#vm-{0}-state".format(vmName), "Running")
+            b.wait_in_text("#vm-{0}-{1}-state".format(vmName, connectionName), "Running")
         if action == "forceOff":
-            b.wait_in_text("#vm-{0}-state".format(vmName), "Shut off")
+            b.wait_in_text("#vm-{0}-{1}-state".format(vmName, connectionName), "Shut off")
```
jelly, line 42: In the future we want to use f-strings everywhere I think, but as it was already there approved.

jelly (approved):

PR #698, opened 2022-05-04
Small fixes

jelly:
Small accumulated fixes from https://github.com/cockpit-project/cockpit-machines/pull/646

### DIFF
--- a/src/components/vm/disks/diskAdd.jsx
+++ b/src/components/vm/disks/diskAdd.jsx
@@ -246,7 +246,6 @@ const CreateNewDisk = ({
     storagePoolName,
     unit,
     validationFailed,
-    vm,
     vmStoragePools,
     volumeName,
 }) => {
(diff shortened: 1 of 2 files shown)


src/components/storagePools/storageVolumeCreateBody.jsx
```diff
@@ -145,5 +145,5 @@ VolumeCreateBody.propTypes = {
     size: PropTypes.number.isRequired,
     storagePool: PropTypes.object.isRequired,
     unit: PropTypes.string.isRequired,
-    volumeName: PropTypes.string.isRequired,
+    volumeName: PropTypes.string,
```
skobyda, line 148: the Add disk dialog has volumeName undefined only when it is initialized. But the parameter should always be passed to `VolumeCreateBody`, as it's an input of a child component, and user always needs to specify volume name when creating disk/volume, otherwise they will be met with a form validation error `"Please enter new volume name"`.
So I think it should stay as required prop of `VolumeCreateBody` component.
Instead, to get rid of  the React warning, I propose changing initial value of volumeName from
`volumeName: undefined`
into 
`volumeName: ""`
[here](https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/disks/diskAdd.jsx#L384)

skobyda (changes requested):

src/components/storagePools/storageVolumeCreateBody.jsx
martinpitt: Ah, that would make it similar to the `file: ""` initialization, makes sense.

@jelly : The rest LGTM, so assume a +1 from me after the above is fixed.

skobyda (approved):
Thanks, LGTM

PR #699, opened 2022-05-06
Cleanup VM creation process

skobyda:
- Don't timeout when creating VM
- Don't wait for VM creation to resolve to close the dialog, since we use asynchronous Toaster notifications for error handling
- When the user selects "Create and edit", redirect them to a new VM immediately


martinpitt (changes requested):
Thanks! I like the cleanup and changes, but the tests need fixing.

skobyda:
All tests now green

martinpitt (approved):
Thanks!

PR #704, opened 2022-05-11
Downloading RHEL images for new VMs

skobyda:
TODO list (some of it may go to follow-up PRs):
 - [ ] Store offline token for 30 days, so user doesn't have to retype it everytime they want to create a new RHEL VM
 - [ ] Refresh available RHEL versions based on supplied token
 - [ ] Check for validity of  offline token immediately after user supplies it
 - [ ] Add an option to cancel the ongoing image download
 - [ ] Handle download interruption
 - [ ] Add option to remove downloaded the image
 - [x] Have concensus on image download location (consensus: libvirt default storage pool at /var/lib/libvirt/images)
 - [x] tests

## Machines: Download RHEL images

You can now create Red Hat Enterprise Linux VMs. The dialog guides you to the page where you can get a download token for your Red Hat subscription, and uses that to download the VM image.

download rhel images

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -17,7 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 import React, { useState, useEffect } from 'react';
-import { AlertGroup, Button } from '@patternfly/react-core';
+import { AlertGroup, Button, Progress, ProgressMeasureLocation } from '@patternfly/react-core';
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import { superuser } from "superuser.js";
 import cockpit from 'cockpit';
@@ -241,18 +241,19 @@ class AppActive extends React.Component {
                                      }
                                      icon={ExclamationCircleIcon} />
                 </>);
-            }
-
-            if (vm.createInProgress) {
+            } else if (vm.createInProgress) {
                 return (<>
                     {allNotifications}
-                    <EmptyStatePanel title={ cockpit.format(_("Creating VM $0"), cockpit.location.options.name) }
+                    <EmptyStatePanel title={cockpit.format(vm.downloadProgress ? _("Downloading image for VM $0") : _("Creating VM $0"), cockpit.location.options.name)}
                                      action={
                                          <Button variant="link"
                                                  onClick={() => cockpit.location.go(["vms"])}>
                                              {_("Go to VMs list")}
                                          </Button>
                                      }
+                                     paragraph={vm.downloadProgress && <Progress aria-label={_("Download progress")}
+                                                                                 value={Number(vm.downloadProgress)}
+                                                                                 measureLocation={ProgressMeasureLocation.outside} />}
                                      loading />
                 </>);
             }
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -28,10 +28,11 @@ import {
     Select as PFSelect, SelectOption, SelectVariant,
     Tabs, Tab, TabTitleText,
     TextInput,
-    Button, Tooltip
+    Button, Tooltip, TextArea
 } from '@patternfly/react-core';
-import { DialogsContext } from 'dialogs.jsx';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
+import { DialogsContext } from 'dialogs.jsx';
 import cockpit from 'cockpit';
 import store from "../../store.js";
 import { MachinesConnectionSelector } from '../common/machinesConnectionSelector.jsx';
@@ -69,6 +70,7 @@ import {
     correctSpecialCases,
     filterReleaseEolDates,
     getOSStringRepresentation,
+    needsRHToken,
 } from "./createVmDialogUtils.js";
 import { domainCreate } from '../../libvirtApi/domain.js';
 import { storagePoolRefresh } from '../../libvirtApi/storagePool.js';
@@ -257,7 +259,7 @@ const NameRow = ({ vmName, suggestedVmName, onValueChanged, validationFailed }) 
     );
 };
 
-const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, os, osInfoList, cloudInitSupported, downloadOSSupported, onValueChanged, validationFailed }) => {
+const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, os, osInfoList, offlineToken, cloudInitSupported, downloadOSSupported, onValueChanged, validationFailed }) => {
     let installationSource;
     let installationSourceId;
     let installationSourceWarning;
@@ -364,11 +366,18 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, 
                              validated={validationStateSource}>
                     {installationSource}
                 </FormGroup>
-                : <OSRow os={os}
-                         osInfoList={osInfoList.filter(os => os.treeInstallable)}
+                : <>
+                    <OSRow os={os}
+                         osInfoList={osInfoList.filter(os => os.treeInstallable || (needsRHToken(os.shortId) && !os.version.endsWith("unknown")))}
                          onValueChanged={onValueChanged}
                          isLoading={false}
-                         validationFailed={validationFailed} />}
+                         validationFailed={validationFailed} />
+                    {os && needsRHToken(os.shortId) &&
+                        <OfflineTokenRow
+                            offlineToken={offlineToken}
+                            onValueChanged={onValueChanged}
+                            validationFailed={validationFailed} />}
+                </>}
         </>
     );
 };
@@ -455,6 +464,30 @@ class OSRow extends React.Component {
         );
     }
 }
+
+const OfflineTokenRow = ({ offlineToken, onValueChanged, validationFailed }) => {
+    const validationStateToken = validationFailed.offlineToken ? 'error' : 'default';
+
+    return (
+        <FormGroup label={_("Offline token")} fieldId="offline-token"
+                   id="offline-token-group"
+                   helperTextInvalid={validationFailed.offlineToken}
+                   validated={validationStateToken}
+                   helperText={<span>
+                       <a href="https://access.redhat.com/management/api" target="_blank" rel="noopener noreferrer">
+                           <ExternalLinkAltIcon className="pf-u-mr-xs" />
+                           {_("Get a new RHSM token.")}
+                       </a>
+                       {" " + _(" Then copy the token and paste it above.")}
+                   </span>}>
+            <TextArea id='offline-token'
+                      validated={validationStateToken}
+                      minLength={1}
+                      value={offlineToken || ''}
+                      onChange={value => onValueChanged('offlineToken', value)} />
+        </FormGroup>
+    );
+};
 
 const UnattendedRow = ({
     onValueChanged,
@@ -790,6 +823,7 @@ class CreateVmModal extends React.Component {
             userPassword: '',
             rootPassword: '',
             userLogin: '',
+            offlineToken: '',
         };
         this.onCreateClicked = this.onCreateClicked.bind(this);
         this.onValueChanged = this.onValueChanged.bind(this);
@@ -953,7 +987,7 @@ class CreateVmModal extends React.Component {
 
     onCreateClicked(startVm) {
         const Dialogs = this.context;
-        const { onAddErrorNotification, osInfoList, nodeMaxMemory, vms } = this.props;
+        const { onAddErrorNotification, osInfoList, nodeMaxMemory, vms, loggedUser } = this.props;
         const { storagePools } = store.getState();
         const vmName = isEmpty(this.state.vmName.trim()) ? this.state.suggestedVmName : this.state.vmName;
 
@@ -973,6 +1007,7 @@ class CreateVmModal extends React.Component {
                 source: this.state.source,
                 sourceType: this.state.sourceType,
                 os: this.state.os ? this.state.os.shortId : 'auto',
+                osVersion: this.state.os && this.state.os.version,
                 profile: this.state.profile,
                 memorySize: convertToUnit(this.state.memorySize, this.state.memorySizeUnit, units.MiB),
                 storageSize: convertToUnit(this.state.storageSize, this.state.storageSizeUnit, units.GiB),
@@ -982,7 +1017,9 @@ class CreateVmModal extends React.Component {
                 userPassword: this.state.userPassword,
                 rootPassword: this.state.rootPassword,
                 userLogin: this.state.userLogin,
-                startVm
+                startVm,
+                offlineToken: this.state.offlineToken,
+                loggedUser
             };
 
             domainCreate(vmParams).then(() => {
@@ -992,6 +1029,7 @@ class CreateVmModal extends React.Component {
                         storagePoolRefresh({ connectionName: storagePool.connectionName, objPath: storagePool.id });
                 }
             }, (exception) => {
+                console.error(`spawn 'vm creation' returned error: "${JSON.stringify(exception)}"`);
                 onAddErrorNotification({
                     text: cockpit.format(_("Creation of VM $0 failed"), vmParams.vmName),
                     detail: exception.message.split(/Traceback(.+)/)[0],
@@ -1053,6 +1091,7 @@ class CreateVmModal extends React.Component {
                 sourceType={this.state.sourceType}
                 os={this.state.os}
                 osInfoList={this.props.osInfoList}
+                offlineToken={this.state.offlineToken}
                 cloudInitSupported={this.props.cloudInitSupported}
                 downloadOSSupported={this.props.downloadOSSupported}
                 onValueChanged={this.onValueChanged}
--- a/src/components/create-vm-dialog/createVmDialogUtils.js
+++ b/src/components/create-vm-dialog/createVmDialogUtils.js
@@ -141,3 +141,7 @@ export function correctSpecialCases(os) {
 
     return os;
 }
+
+export function needsRHToken(osName) {
+    return osName.startsWith("rhel");
+}
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -45,6 +45,7 @@ import {
     finishVmCreateInProgress,
     setVmCreateInProgress,
     setVmInstallInProgress,
+    updateImageDownloadProgress,
     clearVmUiState,
 } from '../components/create-vm-dialog/uiState.js';
 import {
@@ -53,6 +54,7 @@ import {
     fileDownload,
     getHostDevSourceObject,
     getNodeDevSource,
+    LIBVIRT_SYSTEM_CONNECTION,
     logDebug,
     units,
 } from '../helpers.js';
@@ -70,7 +72,9 @@ import {
 } from '../libvirt-xml-update.js';
 import { storagePoolRefresh } from './storagePool.js';
 import { snapshotGetAll } from './snapshot.js';
+import { downloadRhelImage, getAccessToken, getRhelImageUrl } from './rhel-images.js';
 import { call, Enum, timeout, resolveUiState } from './helpers.js';
+import { DOWNLOAD_AN_OS, LOCAL_INSTALL_MEDIA_SOURCE, needsRHToken } from '../components/create-vm-dialog/createVmDialogUtils.js';
 
 export const domainCanConsole = (vmState) => vmState == 'running';
 export const domainCanDelete = (vmState, vmId) => true;
@@ -250,6 +254,7 @@ export function domainCreate({
     connectionName,
     memorySize,
     os,
+    osVersion,
     profile,
     rootPassword,
     source,
@@ -262,6 +267,8 @@ export function domainCreate({
     userLogin,
     userPassword,
     vmName,
+    offlineToken,
+    loggedUser
 }) {
     // shows dummy vm  until we get vm from virsh (cleans up inProgress)
     setVmCreateInProgress(vmName, connectionName, { openConsoleTab: startVm });
@@ -274,7 +281,7 @@ export function domainCreate({
     if (connectionName === 'system')
         opts.superuser = 'try';
 
-    const args = JSON.stringify({
+    const args = {
         connectionName,
         memorySize,
         os,
@@ -291,21 +298,57 @@ export function domainCreate({
         userLogin,
         userPassword,
         vmName,
-    });
+    };
 
     logDebug(`CREATE_VM(${vmName}): install_machine.py '${args}'`);
 
-    return cockpit
-            .spawn([pythonPath, "--", "-", args], opts)
-            .input(installVmScript)
-            .then(() => {
-                finishVmCreateInProgress(vmName, connectionName);
-                clearVmUiState(vmName, connectionName);
-            })
-            .fail(ex => {
-                clearVmUiState(vmName, connectionName); // inProgress cleanup
-                console.info(`spawn 'vm creation' returned error: "${JSON.stringify(ex)}"`);
-            });
+    const tryDownloadRhelImage = () => {
+        if (sourceType == DOWNLOAD_AN_OS && needsRHToken(os)) {
+            let arch;
+            let accessToken;
+            const options = { err: "message" };
+            if (connectionName === "system")
+                options.superuser = "try";
+
+            return cockpit.spawn(['uname', '-m'], options)
+                    .then(out => {
+                        arch = out.trim();
+
+                        return getAccessToken(offlineToken);
+                    })
+                    .then(out => {
+                        accessToken = out.trim();
+                        return getRhelImageUrl(accessToken, osVersion, arch);
+                    })
+                    .then(out => {
+                        const isSystem = connectionName === LIBVIRT_SYSTEM_CONNECTION;
+                        const outObj = JSON.parse(out);
+                        const url = outObj.url;
+                        const filename = outObj.filename;
+                        const downloadDir = isSystem ? "/var/lib/libvirt/images/" : loggedUser.home + "/.local/share/libvirt/images/";
+                        args.sourceType = LOCAL_INSTALL_MEDIA_SOURCE;
+                        args.source = downloadDir + filename;
+
+                        let buffer = "";
+                        return downloadRhelImage(accessToken, url, filename, downloadDir, isSystem)
+                                .stream(progress => {
+                                    buffer += progress;
+                                    const chunks = buffer.split("\n");
+                                    buffer = chunks.pop();
+
+                                    if (chunks.length > 0)
+                                        updateImageDownloadProgress(vmName, connectionName, chunks.pop());
+                                });
+                    });
+        } else {
+            return Promise.resolve();
+        }
+    };
+
+    return tryDownloadRhelImage()
+            .then(() => cockpit.spawn([pythonPath, "--", "-", JSON.stringify(args)], opts).input(installVmScript))
+            .then(() => finishVmCreateInProgress(vmName, connectionName))
+            .finally(() => clearVmUiState(vmName, connectionName));
 }
 
 export function domainCreateFilesystem({ connectionName, objPath, vmName, source, target, xattr }) {
--- a/src/libvirtApi/rhel-images.js
+++ b/src/libvirtApi/rhel-images.js
@@ -1,0 +1,65 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import * as python from "python.js";
+
+import downloadRhelImageScript from 'raw-loader!../scripts/rhsm/download_file_and_report_progress.py';
+import getRhelImageUrlScript from 'raw-loader!../scripts/rhsm/get_rhel_image_url.py';
+import getAccessTokenScript from 'raw-loader!../scripts/rhsm/get_access_token.py';
+
+import {
+    logDebug,
+} from "../helpers.js";
+
+/*
+ * Provider for Red Hat Subscription Manage API.
+ * See https://access.redhat.com/management/api/rhsm
+ */
+
+export function downloadRhelImage(accessToken, url, fileName, downloadDir, isSystem) {
+    logDebug(`Download rhel image: ${url}, ${fileName}, ${downloadDir}, ${isSystem}`);
+
+    const args = JSON.stringify({
+        accessToken,
+        url,
+        fileName,
+        downloadDir
+    });
+
+    return python.spawn(downloadRhelImageScript, args, { err: "message", superuser: isSystem && "require", environ: ['LC_ALL=C.UTF-8'] });
+}
+
+export function getAccessToken(offlineToken) {
+    logDebug(`Get access token`);
+
+    const args = JSON.stringify({ offlineToken });
+    return python.spawn(getAccessTokenScript, args, { err: "message", environ: ['LC_ALL=C.UTF-8'] });
+}
+
+export function getRhelImageUrl(accessToken, rhelVersion, arch) {
+    logDebug(`Download rhel image, ${rhelVersion}, ${arch}`);
+
+    const args = JSON.stringify({
+        accessToken,
+        rhelVersion,
+        arch,
+    });
+
+    return python.spawn(getRhelImageUrlScript, args, { err: "message", environ: ['LC_ALL=C.UTF-8'] });
+}
--- a/src/scripts/rhsm/download_file_and_report_progress.py
+++ b/src/scripts/rhsm/download_file_and_report_progress.py
@@ -1,0 +1,32 @@
+import urllib.request
+import urllib.error
+import sys
+import json
+import os
+
+args = json.loads(sys.argv[1], strict=False)
+
+download_path = os.path.join(args['downloadDir'], args['fileName'])
+if not os.path.exists(download_path):
+    # Create download directory, don't complain when it already exists
+    os.makedirs(args['downloadDir'], exist_ok=True)
+
+    try:
+        req = urllib.request.Request(args['url'])
+        req.add_header("Authorization", f"Bearer {args['accessToken']}")
+        with urllib.request.urlopen(req) as s, open(download_path, "wb") as f:
+            size = int(s.headers["content-length"])
+            done = 0
+            percentage = -1
+            block = s.read(2 ** 20)
+            while block:
+                done += len(block)
+                new_percentage = round(done * 100 / size)
+                if (new_percentage > percentage):
+                    percentage = new_percentage
+                    print(f"{percentage}", flush=True)  # print adds newling by default, which serves nicely as a delimeter
+                f.write(block)
+                block = s.read(2 ** 20)
+
+    except urllib.error.URLError as error:
+        sys.exit(error)
--- a/src/scripts/rhsm/get_access_token.py
+++ b/src/scripts/rhsm/get_access_token.py
@@ -1,0 +1,27 @@
+import urllib.request
+import urllib.error
+import sys
+import json
+
+args = json.loads(sys.argv[1], strict=False)
+
+data = {
+    "grant_type": "refresh_token",
+    "client_id": "rhsm-api",
+    "refresh_token": args["offlineToken"]
+}
+try:
+    req = urllib.request.Request("https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
+                                 urllib.parse.urlencode(data).encode(),
+                                 headers={"Content-Type": "application/x-www-form-urlencoded"})
+    with urllib.request.urlopen(req) as s:
+        response_body = s.read().decode()
+        ret_obj = json.loads(response_body)
+
+        # Handle RHSM API failure
+        if "error" in ret_obj:
+            sys.exit(ret_obj["error"])
+
+    print(ret_obj["access_token"])
+except Exception as error:
+    sys.exit(error)
--- a/src/scripts/rhsm/get_rhel_image_url.py
+++ b/src/scripts/rhsm/get_rhel_image_url.py
@@ -1,0 +1,25 @@
+import urllib.request
+import urllib.error
+import sys
+import json
+
+args = json.loads(sys.argv[1], strict=False)
+
+try:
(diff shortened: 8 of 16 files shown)


skobyda:
User can now automatically download RHEL images. OS selections offer RHEL version, which are still supported (i.e. End of Life still hasn't passed).
Note that OS versions are sorted by their release date, not name, which is why RHEL may seem disordered (e.g. rhel 7.9 was release after rhel 8.0). Maybe we should sort by rhel version number? Or keep it by their release date?  

Screenshot from 2022-05-11 14-38-52

After user selects RHEL OS, Offline token input field appears:
(Btw is there any difference between using PF button as link or just using `<a href...>` ? Is PF button somehow more friendly towards color blind?)
Screenshot from 2022-05-11 14-38-52
Screenshot from 2022-05-11 14-39-40

After clicking on Create, the progress of download is shown:

Screenshot from 2022-05-11 14-39-00

Screenshot from 2022-05-11 14-39-49

garrett:
> Note that OS versions are sorted by their release date, not name, which is why RHEL may seem disordered (e.g. rhel 7.9 was release after rhel 8.2). Maybe we should sort by rhel version number? Or keep it by their release date?

Correct. Sort by version, not release date. Thanks!

---

>  I use <a href...> in tooltip. Btw is there any difference between using PF button as link or just using <a href...> ? Is PF button somehow more friendly towards color blind?

Perhaps we actually should have this as helpertext, with less text, instead? It's something that nearly everyone would need nearly every time. (The only ones who wouldn't either have it saved in their browser already or have it in a text file they copy and paste from.) Therefore, I think having a line of text under the entry box makes sense.

It could be something like this:

> [Get a new RHSM token](#), then copy the token and paste it above.

The link would use the `fa-external-link-alt` icon and open in a new tab (with ), of course. You'd use `target='_blank' rel='noopener'` on the link to force it into a new tab. (It's important to do this, else the page would replace Cockpit, which isn't the desired action when using a web app.)

---

At first, I thought the download progress was under the list! :laughing: Thankfully, I realized you have the download info on the list and the progress picture is separate, as a placeholder details page.

Can you cancel the download? It's fine if you can't for now.

Also are downloads cached? If you want to create more than 1 VM of the same RHEL image, what's the flow? Would it already be downloaded and re-used? If it is kept, then should we provide a means of removing it too?

We have storage pools. Is it downloaded and represented as a pool? That would allow management within Cockpit without having to do much of anything extra, I think, if it makes sense.

skobyda:
> Correct. Sort by version, not release date. Thanks!
Done

> It could be something like this:
> 
> > [Get a new RHSM token](#), then copy the token and paste it above.
> 
> The link would use the `fa-external-link-alt` icon and open in a new tab (with ), of course. You'd use `target='_blank' rel='noopener'` on the link to force it into a new tab. (It's important to do this, else the page would replace Cockpit, which isn't the desired action when using a web app.)

Done:
Screenshot from 2022-05-24 12-45-22


> Can you cancel the download? It's fine if you can't for now.

No, but It would be good if you can. I use simple python requests to download the file, and there are ways how to amend that python script to allow canceling the download and cleanup. Adding it to [the "TODO" list](https://github.com/cockpit-project/cockpit-machines/pull/704#issue-1232583154).

> Also are downloads cached? If you want to create more than 1 VM of the same RHEL image, what's the flow? Would it already be downloaded and re-used? If it is kept, then should we provide a means of removing it too?
>
> We have storage pools. Is it downloaded and represented as a pool? That would allow management within Cockpit without having to do much of anything extra, I think if it makes sense.

If the image exists, it would be re-used. 
The download is kept in `/var/tmp`, but it will be removed by systemd after user doesn't use for more than 30 days,
`/var/tmp` directory is watched by systemd service `systemd-tmpfiles`, which periodically cleans /var/tmp of files, which have not been accessed in more than 30 days. You can read about it at [systemd docs](https://systemd.io/TEMPORARY_DIRECTORIES/). 
And I kinda like that solution. If user creates rhel VM often, systemd-tmpfiles would keep the RHEL installation image in `/var/tmp`
If user doesn't use that download to create VMs anymore, systemd will remove the download after some period of inactivity, so it would not take up space on user's server for a long time.
And we don't have to configure anything, systemd should do it automatically.

Alternatively we can use ~/.cache, which most users are familiar and is one of the first places a sysadmin goes if they want to cleanup som space on their machine.

skobyda:
Other alternative could also be to use ~/Downloads (gnome Boxes does this). It would make the presence of the download more obious to the user.

I don't think we should save the download as a volume to some storage pool. Firstly, user can have no local filesystem-based storage pool present on their system (e.g. they run their default storage pool on a network).
So we would have to have a fallback where we create our own storage pool for the rhel downloads, and that seems kinda magical and could be suprising to the user, no?

garrett:
I actually have `~/Downloads/` as a storage pool in Cockpit-Machines, for whatever unknown reason. But it does want to remove all my files there, so... :laughing:  We could have `~/Downloads/VirtualMachines/` or something. But your suggestion about keeping it in that systemd-watched place in /var/tmp/ makes sense, as long as it is cleaned up over time and (possibly) with space pressure (as in if the disk is filling up, systemd could clean up the tmp area for older/larger files).

It does feel like this is something that should be documented, wherever it goes.

skobyda:
Tests seems to be passing. @martinpitt I think this can get a code review.



So there seems to be an issue that everytime I put a ISO into `.cache` and then use that iso as an installation source, libvirt creates a storage pool over that directory and qemu-kvm group takes ownership of it. That's not ideal.

> I actually have `~/Downloads/` as a storage pool in Cockpit-Machines, for whatever unknown reason. But it does want to remove all my files there, so... laughing We could have `~/Downloads/VirtualMachines/` or something. But your suggestion about keeping it in that `systemd-tmpfiles` place in `/var/tmp/` makes sense, as long as it is cleaned up over time and (possibly) with space pressure (as in if the disk is filling up, systemd could clean up the tmp area for older/larger files).

Yeah, Downloads could work. You probably have that storage pool because everytime you create a VM with some file as a disk, libvirt creates a storage pool over a directory with that file, so maybe you created some VM from an iso file from your Downloads.
But then maybe we could probably use libvirt's default storage pools, which is located at `/var/lib/libvirt/images` (or `~/.local/share/libvirt/images` for session connection), rather then create a new directory and a new storage pool at user's location.

skobyda:
fedora-36 test run is stuck, force-pushing empty commit to re-trying

skobyda:
Rebased

garrett:
> so maybe you created some VM from an iso file from your Downloads

GNOME Boxes uses libvirt and installs from ISOs in `~/Downloads/` (really: `xdg-downloads`, which is generally the `Downloads` folder in `$HOME`), so that's probably it.

The problem with adding a directory like `~/Downloads/` means _all_ downloaded files are shown in that location within Cockpit and removing it, Cockpit suggests deleting _all_ files there. Whoops. That's probably unintended. In reality, not many people aside from me will hit this. And even still, you'd have to dive into the details of storage pools. :wink:

> But then maybe we could probably use libvirt's default storage pools

Yeah, true.

I guess, stepping back, the questions really are:

- How temporary is the ISO?
- How important is accessing the ISO? (Mainly outside of the creation dialog.)

Answering these questions will probably point us to where the file should be stored. 

If we want it to be temporary, the standard storage pool would be weird, right? It'd disappear.

If someone wants to access the ISO directly, then having it in a temporary pool would be weird too, I guess. Unless it was easy to access and if it stayed around for a while.

Should we have a temporary pool? If so, should we show a countdown before it will be deleted? Can we even show that? Does that even make sense? This is probably overkill, especially if we don't expect people to directly access ISOs.

At least caching the ISO means we can skip interacting with the API and the token for creating the same exact version, right? (Although what about point releases?)

skobyda:
> I guess, stepping back, the questions really are:
> 
>     * How temporary is the ISO?
>
>     * How important is accessing the ISO? (Mainly outside of the creation dialog.)
> 

Ideally we want the ISO to present while while user still plans to create other VMs with the same rhel version in the future. Once they stop using it to create new VMs (e.g. they switched to a different version of rhel). That's of course hard to guess.
I'm also thinking about other cases, when user creates a RHEL VM, but doesn't install it. And they might want to keep downloaded, but un-installed RHEL VM for several days, maybe even weeks, until they need to deploy. That means that downloaded image also needs to survive reboot and keep persistent for undefined period of time.
More I think about it, more I'm leaning towards the "permanent" answer. as it's better to keep the downloaded ISO permanent at user's location.

Storing it permanently in libvirt's managed storage pool brings other advantages:
- It would also work quite nicely with our existing functionalities of storage pools. For example we tell the user which storage volumes are used by some VMs: Screenshot from 2022-06-09 14-38-41
- If we store it in libvirt's storage pool, we will keep the ISO under our control. For example we can add some metadata into the storage pool's definition, telling us to offer the user some notification/warning to remove it such ISO file if it's not used by any VM anymore or hasn't been accessed for some time.
- We will generally adhert to libvirt plans, which is also trying to migrate as much storage management into storage pools, instead of just directly attaching disk's to the VM.

Lastly I don't think anybody will try to access the downloaded ISO directly. We create it for a VM installation, and then keep the ISO file for the future VM creations thru cockpit UI.

test/machineslib.py
```diff
@@ -226,6 +226,46 @@ def get_next_mac(self, last_mac):
         new_mac = ':'.join(parts)
         return new_mac
 
+    def setup_mock_server(self, mock_server_filename, subj_names):
+        self.machine.upload(["files/min-openssl-config.cnf", f"files/{mock_server_filename}"], self.vm_tmpdir)
+
+        self.restore_file("/etc/hosts")
+        for i, sub_name in enumerate(subj_names):
+            self.machine.execute(f'echo "127.0.0.1 {sub_name}" >> /etc/hosts')
+            self.machine.execute(f'echo "DNS.{i + 2} = {sub_name}" >> {self.vm_tmpdir}/min-openssl-config.cnf')
```
martinpitt, line 235: `self.write_file(..., append=True)` avoids the `>>` and restores automatically. For the second one you can use `self.machine.write(path, content, append=True)`.

test/check-machines-create
```diff
@@ -789,8 +790,10 @@ vnc_password= "{vnc_passwd}"
                 b.wait_in_text(".pf-c-modal-box .pf-c-modal-box__header .pf-c-modal-box__title", "Create new virtual machine")
 
             if self.os_name is not None:
+                # remove os codename from os name, e.g.: 'bullseye' in 'Debian 11 (bullseye)'
+                query_result = re.sub(r" \(.*\)", "", self.os_name)
+                query_result = f'{query_result}'
```
martinpitt, line 800: This seems quite unrelated? Please add a "why" to the commit message. In particular, can I still search for the code names and get a result? The [release names](https://www.debian.org/releases/) are quite popular, and easy to google for.

test/check-machines-create
```diff
@@ -914,7 +914,8 @@ vnc_password= "{vnc_passwd}"
                 self.browser.click(".pf-c-modal-box__footer button:contains(Create and edit)")
             self.browser.wait_not_present("#create-vm-dialog")
 
-            self.goToVmPage(self.name)
+            if self.create_and_run:
+                self.goToVmPage(self.name)
```
martinpitt, line 923: Also a separate bug fix. Please split out into separate PR, with a more detailed explanation, and test if possible. What does the "not redirected" mean?

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -380,13 +380,20 @@ class OSRow extends React.Component {
                 .map(os => correctSpecialCases(os))
                 .filter(os => filterReleaseEolDates(os) && !IGNORE_VENDORS.find(vendor => vendor == os.vendor))
                 .sort((a, b) => {
-                    if (a.vendor == b.vendor)
-                        if (a.releaseDate || b.releaseDate)
-                            return compareDates(a.releaseDate, b.releaseDate, true) > 0;
-                        else
+                    if (a.vendor == b.vendor) {
```
martinpitt, line 390: Makes sense, but can also land separately (and fairly quickly)

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -363,11 +363,18 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices,
                              validated={validationStateSource}>
                     {installationSource}
                 </FormGroup>
-                : <OSRow os={os}
-                         osInfoList={osInfoList.filter(os => os.treeInstallable)}
+                : <>
+                    <OSRow os={os}
+                         osInfoList={osInfoList.filter(os => os.treeInstallable || (os.shortId.startsWith("rhel") && !os.version.endsWith("unknown")))}
```
martinpitt, line 368: `os.shortId.startsWith("rhel")` is done in three places, and this feels a little heuristic.. Doesn't osinfo-db tell this? ("needs token" flag or so). If not, please move this into a helper function like `needsRHToken(os)` or so, so that we can see more easily where it is used, and tweak it.

What's with the "unknown" test here?

src/components/create-vm-dialog/createVmDialogUtils.js
```diff
@@ -27,6 +27,16 @@ import autoDetectOSScript from 'raw-loader!./autoDetectOS.py';
...
 
+export const URL_SOURCE = 'url';
+export const LOCAL_INSTALL_MEDIA_SOURCE = 'file';
+export const CLOUD_IMAGE = 'cloud';
+export const DOWNLOAD_AN_OS = 'os';
+export const EXISTING_DISK_IMAGE_SOURCE = 'disk_image';
+export const PXE_SOURCE = 'pxe';
+export const NONE = 0;
+export const RUN = 1;
+export const EDIT = 2;
```
martinpitt, line 38: Introducing these constants is useful. It's also unrelated to RHEL downloads, can you please split this out into a separate PR?

src/libvirtApi/domain.js
```diff
@@ -287,21 +293,57 @@ export function domainCreate({
...
+                        args.source = downloadDir + filename;
+
+                        return downloadRhelImage(accessToken, url, filename, downloadDir, isSystem);
+                    });
+        } else {
+            return Promise.resolve();
+        }
+    };
+
+    return tryDownloadRhelImage().then(() => {
```
martinpitt, line 335: Needs to handle rejections from tryDownloadRhelImage().

src/libvirtApi/rhsm.js
```diff
@@ -0,0 +1,66 @@
+/*
```
martinpitt, line 1: "rhsm.js" is an odd name. That means "Red Hat Subscription Manager", but it's not talking to that. "rhel-images.js"?

test/check-machines-create
```diff
@@ -1788,6 +1817,41 @@ vnc_password= "{vnc_passwd}"
...
+                                                      create_and_run=True))
+
+        # If run on different connection, the downloaded iso should be stored in ~/.local/share/libvirt/images/
+        runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='os',
+                                                      os_name=config.RHEL_5_0,
+                                                      os_short_id=config.RHEL_5_0_SHORTID,
+                                                      offline_token="my_offline_token",
+                                                      create_and_run=True,
+                                                      connection="session"),
+                          check_env_empty=False)
```
martinpitt, line 1868: Can you please add a test what happens if you specify a wrong token, or the download fails? There should be an error message in the dialog, and the possibility to fix/retry.

martinpitt (changes requested):
Thanks, amazing work! I'm really happy to see this feature (and so will be our employer :grin: )

This is quite big, and a lot of changes can be broken out into separate PRs. Please do that as much as possible, so that we can land/review this quicker.

src/scripts/rhsm/download_file_and_report_progress.py
```diff
@@ -0,0 +1,36 @@
...
+import sys
+import json
+import os
+from pathlib import Path
+
+args = json.loads(sys.argv[1], strict=False)
+access_token = args["accessToken"]
+url = args["url"]
+filename = args["fileName"]
+download_dir = args["downloadDir"]
```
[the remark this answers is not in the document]
skobyda: Nothing really. Just my personal preference as it seems a bit cleaner to me, but I'm not keen on this at all

src/scripts/rhsm/download_file_and_report_progress.py
```diff
@@ -0,0 +1,36 @@
...
+        total_length = int(r.headers.get("content-length"))
+        done = -1
+        for data in r.iter_content(chunk_size=4096):
+            downloaded_length += len(data)
+            f.write(data)
+            new_done = int(downloaded_length / total_length * 100)
+            if (new_done > done):
+                done = new_done
+                sys.stdout.flush()
+                sys.stdout.write(f"{done}")
```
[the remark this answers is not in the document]
skobyda: Why would we need a space separators if we flush? This just prints a single string number as a percentage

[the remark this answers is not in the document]
martinpitt: I meant that you'd print a string like "15812" with this, instead of "1 5 8 12" or even "1\n5\n8\n12". flush() doesn't insert any spaces, it just makes sure an 'incomplete' line gets actually printed.

src/scripts/rhsm/download_file_and_report_progress.py
```diff
@@ -0,0 +1,36 @@
+import requests
```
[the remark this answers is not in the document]
skobyda: I want the download to report progress, and in that I need the request to stream=True in order to print out after specific-sized chunks. Otherwise the program would just wait on the `requests.get` and then print the progress on 100%.
I could not find ability to stream thru standart urllib library. But maybe I'm just missing something, correct me if I'm wrong.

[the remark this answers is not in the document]
martinpitt: This is fairly straightforward:
```py
import urllib.request

with urllib.request.urlopen("https://github.com/cockpit-project/cockpit/releases/download/272/cockpit-272.tar.xz") as f:
    size = int(f.headers["content-length"])
    read = 0
    while block := f.read(2 ** 20):
        read += len(block)
        print(f"got {read} out of {size} bytes")
```

Of course you need to actually do something with the read blocks, like open a destination file and writing them there.

src/scripts/rhsm/get_access_token.py
```diff
@@ -0,0 +1,26 @@
...
+    r.raise_for_status()
+except requests.exceptions.RequestException as error:
+    sys.exit(error)
+
+# Handle RHSM API failure
+ret_obj = r.json()
+if "error" in ret_obj:
+    sys.exit(ret_obj["error_description"])
+
+print(ret_obj["access_token"])
```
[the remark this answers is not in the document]
skobyda: Hmm. I think it would end up being uncessarily more complex. Firstly, if there is a runtime error in python, that is eruttuner as aa objet { problem: null, exit_status: 1, exit_signal: null, message: "some message" }. Os if I return a JSOn, then message would contain another JSON with some object in it, instead of a message.
Also request exceptions `requests.exceptions.RequestException`, which I'm also trying to handle, are not serializable, so to make it work that's additional complexity.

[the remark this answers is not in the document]
martinpitt: OK, makes sense. Then please just fix the `error` vs. `error_description` mismatch, as you test for one and then assume the other.

test/check-machines-create
skobyda: I think the first test which encountered this problem is for downloading RHEL images.
Also it's not a bug in the UI, just this function not being sufficient for test case `testCreateDownloadRhel`. So maybe it could be merged with other commit, like this one https://github.com/cockpit-project/cockpit-machines/pull/704/commits/765e5593fa1af9744d182ad74c6945432b7caa0f

src/libvirtApi/rhel-images.js
```diff
@@ -0,0 +1,83 @@
...
+
+    const headers = {
+        grant_type: "refresh_token",
+        client_id: "rhsm-api",
+        refresh_token: offlineToken,
+    };
+
+    return http.post("/auth/realms/redhat-external/protocol/openid-connect/token", undefined, [headers]);\
+    */
+
```
skobyda, line 67: @marusak

test/check-machines-create
skobyda: added to the commit message.
```
In the UI, we try to create a full OS name by combining a distro,
version and a release name, e.g. Red Hat Enterprise Linux 9.0 (Plow).
However osinfo-query output doesn't contain release name, therefore
we need ommit it in our tests when matching osinfo-query output with
an OS list shown in the UI.
```
So basically, in the UI, we show OS with a string like "Red Hat Enterprise Linux 9.0 (Plow)", and then check if it is also present in the `$osinfo-query os` command output. But `osinfo-query os` only contains string "Red Hat Enterprise Linux 9.0", without the release name. So here I just get rid of the release name, and then try to match the OS name from the UI with an `osinfo-query os` command output.

src/scripts/rhsm/get_rhel_image_url.py
```diff
@@ -0,0 +1,28 @@
...
+import json
+
+args = json.loads(sys.argv[1], strict=False)
+access_token = args["accessToken"]
+rhel_version = args["rhelVersion"]
+arch = args["arch"]
+
+# Handle HTTP request failure
+try:
+    r = requests.get(f"https://api.access.redhat.com/management/v1/images/rhel/{rhel_version}/{arch}", headers={"Authorization": f"Bearer {access_token}"})
```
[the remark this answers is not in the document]
skobyda: Somehow, I could not make cockpit.http() work. It seems it has a problem with https, I set `tls : { validated : True }`, set `"Content-Type": "application/x-www-form-urlencoded" `, yet that leads to a bad requests. I guess I forget to set some other header of the request, but cannot figure out which one: https://github.com/cockpit-project/cockpit-machines/pull/704/files#diff-824224f14df7f3f8af7b2e139909421218d97fe6667c9a111ba9fc665f0e397bR67

src/scripts/rhsm/download_file_and_report_progress.py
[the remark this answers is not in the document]
skobyda: Yeah but the output this script produces isn't meant to be looked at once, since rather check the progess of the download with [process.stream()](https://cockpit-project.org/guide/latest/cockpit-spawn.html#cockpit-spawn-stream), which calls a handler for each single write to a stdout. So it's not really important if I add any whitespaces after percentage, no?

src/libvirtApi/rhel-images.js
skobyda: I think all the issues above are fixed, apart from this one. I can't seem to figure out how to make cockpit.http work with an https. Anyway, here, commented out, is how I tried it. I believe the issue is with incorrect headers of the https request.
Without Content-Type set, POST request fails with: { status: 415, reason: "Unsupported Media Type", message: "Unsupported Media Type", problem: null }
With Content-Type set, POST request fails with: { status: 400, reason: "Bad Request", message: "Bad Request" }

@martinpitt have you encountered any issues with cockpit.http using https before?

src/scripts/rhsm/download_file_and_report_progress.py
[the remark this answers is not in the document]
martinpitt: It is because you can't guarantee that every write() will have exactly one corresponding read(). The client may be busy with something else for a bit, and then read() the contents from two write()s. Or the other way round, and only read half of the pipe contents. You really need to read(), split on a separator (space or `\n`), and show the last *complete* data set.

With this "block" wise read/write, you need a SEQPACKET socket, not a TCP one or a standard Unix pipe.

src/libvirtApi/rhel-images.js
martinpitt: No, I didn't, and the http"s" part is fine -- otherwise you'd get certificate verification errors and not even get as far as getting a HTTP status. Supposedly it sends some different headers. What I would do is to point this against http://localhost:12345 instead of https://sso.redhat.com/, run `nc -l 12345`, run the request with both methods (urllib.request() and cockpit.http.post()), and compare the raw netcat output of the request. This will show the http version, headers, body etc. in the finest possible detail.

Note that this was mostly a suggestion for simplification. I find it ugly and unwieldy to invoke an external Python script, but as long as that works with just "batteries included" python, I won't veto it. What I do veto is to introduce a python-requests dependency, but you already rewrote everything in terms of urllib (thanks!).

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -454,6 +462,23 @@ class OSRow extends React.Component {
...
 
+const OfflineTokenRow = ({ offlineToken, onValueChanged, validationFailed }) => {
+    const validationStateToken = validationFailed.offlineToken ? 'error' : 'default';
+
+    return (
+        <FormGroup label={_("Offline Token")} fieldId="offline-token"
+                   id="offline-token-group"
+                   helperTextInvalid={validationFailed.offlineToken}
+                   validated={validationStateToken}>
+            <TextArea id='offline-token'
```
martinpitt, line 480: So, are offline tokens really multi-line things? Are these PEM certificates?

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -985,6 +1014,7 @@ class CreateVmModal extends React.Component {
                         storagePoolRefresh({ connectionName: storagePool.connectionName, objPath: storagePool.id });
                 }
             }, (exception) => {
+                console.info(`spawn 'vm creation' returned error: "${JSON.stringify(exception)}"`);
```
martinpitt, line 1024: Unrelated to this commit. Please make this either a `logDebug` (if the error already appears in the UI), or drop it if the debug log already contains it (which I'm rather sure of), or turn into a console.error().

src/libvirtApi/rhel-images.js
```diff
@@ -0,0 +1,84 @@
...
+    logDebug,
+} from "../helpers.js";
+
+/*
+ * Provider for Red Hat Subscription Manage API.
+ * See https://access.redhat.com/management/api/rhsm
+ */
+
+export function downloadRhelImage(accessToken, url, fileName, downloadDir, isSystem) {
+    logDebug(`Download rhel image`);
```
martinpitt, line 36: Print the parameters (except accessToken) as well?

test/check-machines-create
```diff
@@ -148,6 +148,14 @@ class TestMachinesCreate(VirtualMachinesCase):
                                                                          create_and_run=False),
                                              {"source-url": "Installation source must not be empty"})
 
+        # When switching from PXE mode to anything else make sure that the source input is empty
+        runner.checkDialogFormValidationTest(TestMachinesCreate.VmDialog(self, storage_size=1,
```
martinpitt, line 152: This seems useful, but should be an independent commit.

src/app.jsx
```diff
@@ -17,7 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 import React, { useState, useEffect } from 'react';
-import { AlertGroup, Button } from '@patternfly/react-core';
+import { AlertGroup, Button, Progress, ProgressMeasureLocation } from '@patternfly/react-core';
```
martinpitt, line 20: I think "show download progess" might better be squashed into the previous commit. If you want to keep it separate, please clarify the commit message, like "Show RHEL image download progress".

src/libvirtApi/domain.js
```diff
@@ -328,7 +329,11 @@ export function domainCreate({
                         args.sourceType = LOCAL_INSTALL_MEDIA_SOURCE;
                         args.source = downloadDir + filename;
 
-                        return downloadRhelImage(accessToken, url, filename, downloadDir, isSystem);
+                        return downloadRhelImage(accessToken, url, filename, downloadDir, isSystem)
+                                .stream(progress => {
+                                    const percentages = progress.trim().split('\n');
+                                    updateImageDownloadProgress(vmName, connectionName, percentages[percentages.length - 1]);
```
martinpitt, line 335: Of course complete precision does not matter here, but reading from a stream is a generic skill, so for the sake of teaching: It can happen that some packet gets transfered with a boundary that is in the middle of a number. The clean way of reading a line-separated stream is to *not* do the trim(), but:

 - add the currently read block to a string buffer.
 - split the buffer on `\n`; keep the last part in the buffer
 - process the first n-1 parts (i.e. the ones that got terminated); of course for your case you only need the most recent complete percentage, i.e. parts[n-2]

Have a look at [c-podman's monitor](https://github.com/cockpit-project/cockpit-podman/blob/18fb5a2f9e702e3447dceacec8bdb951b36351f4/src/rest.js#L25) how to do this correctly and elegantly.

test/check-machines-create
```diff
@@ -1359,17 +1360,31 @@ vnc_password= "{vnc_passwd}"
...
-            # STARTED IMMEDIATELLY as that causes "creating VM" stage to take longer time.
-            if dialog.create_and_run:
-                self._assertVmStates(dialog.name, dialog.connection, init_state, second_state)
+
+            final_state = "Running" if dialog.create_and_run else "Shut off"
+            # Test dowloading RHEL image shows the progress of the download
+            if dialog.os_name == TestMachinesCreate.TestCreateConfig.RHEL_5_0:
+                # Check download percantage is shown as a whole number 0-100 followed by '%'
+                if dialog.create_and_run:
+                    b.wait_in_text(f"#vm-{dialog.name}-{dialog.connection}-state", "Downloading")
```
martinpitt, line 1369: This only downloads a tiny fake image from a local test http server, right? So this seems like a race condition, it could be too fast to catch. Perhaps you can introduce some slowness into the mock server, or do a trick similar to what we do with package installations in cockpit's tests: They stop in the middle and wait for some `/tmp/finish-pk` file to appear. The test can then check the progress state, `m.execute("touch /tmp/finish-pk")` and thus go on in a controlled fashion. That way, you can even pin down the percentage more precisely, if you stop at a well-defined 40% or so.

If you find this is too much work, just make sure the download takes at least 3s?

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -466,10 +466,17 @@ const OfflineTokenRow = ({ offlineToken, onValueChanged, validationFailed }) =>
     const validationStateToken = validationFailed.offlineToken ? 'error' : 'default';
 
     return (
-        <FormGroup label={_("Offline Token")} fieldId="offline-token"
+        <FormGroup label={_("Offline token")} fieldId="offline-token"
```
martinpitt, line 469: Please squash "Add a helper text for offline token" commit.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -466,10 +466,17 @@ const OfflineTokenRow = ({ offlineToken, onValueChanged, validationFailed }) =>
...
                    id="offline-token-group"
                    helperTextInvalid={validationFailed.offlineToken}
-                   validated={validationStateToken}>
+                   validated={validationStateToken}
+                   helperText={<span>
+                       <a href="https://access.redhat.com/management/api" target="_blank" rel="noopener noreferrer">
+                           <ExternalLinkAltIcon className="pf-u-mr-xs" />
+                           {_("Get a new RHSM token")}
+                       </a>
+                       {_(", then copy the token and paste it above.")}
```
martinpitt, line 478: This is another case of stringing together two half-sentences and making assumptions about English. For simplicity, let's just make this two separate sentences.

martinpitt (changes requested):
Thank you! There's still a bunch of smaller issues which I'd like to see fixed before landing this. But I've tried to keep the bigger issues as follow-ups :grin:

src/scripts/rhsm/download_file_and_report_progress.py
```diff
@@ -0,0 +1,33 @@
+import urllib.request
+import urllib.error
+import sys
+import json
+import os
+
+args = json.loads(sys.argv[1], strict=False)
+
+download_path = os.path.join(args['downloadDir'], args['fileName'])
+if not os.path.exists(download_path):
```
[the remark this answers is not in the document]
skobyda: Added in the TODO list at the beginning of this PR/

src/scripts/rhsm/get_rhel_image_url.py
```diff
@@ -0,0 +1,25 @@
...
+    with urllib.request.urlopen(req) as s:
+        ret_obj = json.loads(s.read())
+except urllib.error.URLError as error:
+    sys.exit(error)
+
+if "error" in ret_obj:
+    sys.exit(ret_obj["error"])
+
+for downloadable_content in ret_obj["body"]:
+    if downloadable_content["imageName"].endswith("Boot ISO"):
```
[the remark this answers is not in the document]
skobyda: We can search for a files ending with "boot.iso". There are other ISO files, like "Driver update" files in it, i.e. dd-mt7921-4.18.0_363_dup8.5-2.el8_5.iso , a full Binary DVD, i.e. `rhel-8.5-x86_64-dvd.iso`, suplementary DVD, virtio-win DVD iso... Here is an example of JSON object: http://pastebin.test.redhat.com/1065440

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: They are quite huge. Their length varies, but in most cases it's 600+ characters long. You can try to generate yours here: https://access.redhat.com/management/api

test/check-machines-create
skobyda: That test is identical to the one above that. I guess I accidentally doubled it when rewriting some stuff and it got into my commit.

src/components/create-vm-dialog/createVmDialog.jsx
martinpitt: Urgh! Well, nothing we can change on our side about that :cry:

test/check-machines-create
skobyda: That think that approach in package installations is done with some sleep in post install scripts, so the similar approach would be to have some sleep before/after the download?

One way to make the download slower to set some throttle for http.server, e.g. to make the download file limited to only 100kb/s or something like that. But looking at the python's standard http.server, I could not find any option for that.

However, alternatively, to copy the file to wfile stream, I'm using `shutil.copyfileobj` which copies the file by chunks, and allows us to set buffer size. So what if I create a big file, and set small buffer size, to slow down the process of downloading?

src/scripts/rhsm/download_file_and_report_progress.py
```diff
@@ -0,0 +1,33 @@
...
+            done = 0
+            percentage = -1
+            block = s.read(2 ** 20)
+            while block:
+                done += len(block)
+                new_percentage = round(done * 100 / size)
+                if (new_percentage > percentage):
+                    percentage = new_percentage
+                    sys.stdout.flush()
+                    sys.stdout.write(f"{percentage}\n")
```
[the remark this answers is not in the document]
skobyda: Hmm, it seems that unless I flush.

test/check-machines-create
martinpitt: Right, http.server does not have any fancy options for rate limiting. I recently added a [packagekit cancel test](https://github.com/cockpit-project/cockpit/commit/76db56ea6440ec8) which blocked the download indefinitely, but this trick does not work here -- we can't read from pipes or char devices, as the web server needs to compute/send Content-Length, so this requires plain files. However, I didn't realize that this re-uses the 500 MB test images. So these are far from "tiny", and should take some time to download. @skobyda mentioned ~ 7s, which is more than enough. I was just concerned it would take 0.1s or so. So I'm happy enough with this.

martinpitt: As discussed on IRC, let's change the mock buffer size to 1 kB to ensure this is slow enough, then it should be fine.

martinpitt: I see that you now have a sleep(3) (which by itself would already be sufficient) *and* a small buffer size. This is almost too much, but if we care we can drop it again in the future.

test/files/mock-rhsm-rest
```diff
@@ -0,0 +1,62 @@
...
+            filepath = "/var/lib/libvirt/images/example.img"
+            with open(filepath, 'rb') as f:
+                self.send_response(200)
+                self.send_header("Content-Type", 'application/octet-stream')
+                self.send_header("Content-Disposition", f"attachment; filename='{os.path.basename(filepath)}'")
+                fs = os.fstat(f.fileno())
+                self.send_header("Content-Length", str(fs.st_size))
+                self.end_headers()
+                time.sleep(3)  # Simulate network delay
+                shutil.copyfileobj(f, self.wfile, 1024)  # Specify 1KiB buffer size, otherwise the download might be too fast on some platforms
```
martinpitt, line 52: This now might be too slow -- as long as this takes at least 2 seconds, it's fine for the tests. But as long as it's < 10s, it's fine (i.e. we shouldn't make our tests unnecessarily slow)

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -456,6 +465,30 @@ class OSRow extends React.Component {
...
+        <FormGroup label={_("Offline token")} fieldId="offline-token"
+                   id="offline-token-group"
+                   helperTextInvalid={validationFailed.offlineToken}
+                   validated={validationStateToken}
+                   helperText={<span>
+                       <a href="https://access.redhat.com/management/api" target="_blank" rel="noopener noreferrer">
+                           <ExternalLinkAltIcon className="pf-u-mr-xs" />
+                           {_("Get a new RHSM token.")}
+                       </a>
+                       {"" + _(" Then copy the token and paste it above.")}
```
martinpitt, line 481: `"" +` is a no-op. You meant `" " +`, i.e. add a space.

src/libvirtApi/domain.js
```diff
@@ -291,21 +298,56 @@ export function domainCreate({
...
+                        args.source = downloadDir + filename;
+
+                        let buffer = "";
+                        return downloadRhelImage(accessToken, url, filename, downloadDir, isSystem)
+                                .stream(progress => {
+                                    buffer += progress;
+                                    const chunks = buffer.split("\n");
+                                    buffer = chunks.pop();
+
+                                    updateImageDownloadProgress(vmName, connectionName, chunks.pop());
```
martinpitt, line 339: You need to check the length of `chunks`  here. It can happen that after pop'ing off the "rest", there is no completed chunk left over, and you get `undefined` here.

martinpitt (changes requested):

test/files/mock-rhsm-rest
skobyda: Lowered, now it's around 2-3 seconds

martinpitt:
@skobyda : This somehow breaks `TestMachinesCreate.testCreateUrlSource` on RHEL and CentOS 9.

skobyda:
> @skobyda : This somehow breaks `TestMachinesCreate.testCreateUrlSource` on RHEL and CentOS 9.

Ahhh, once again, I have removed [this console.info](https://github.com/cockpit-project/cockpit-machines/pull/704#discussion_r921791595) which broke the naughty. Reintroduced it back as console.error()

martinpitt:
I tested this interactively again, and this works quite nicely. The only oddity was that the list of downloadable OSes does not include RHEL 8.6, which was released some time ago already. Installing 8.5 worked fine. 

I'll add an initial release note.

Thank you, awesome work! :clap:

martinpitt (approved):

garrett:
Hey, so I was waiting for a video to re-review this. Any chance of making a video of it in action still? 

It would be nice, in case there's anything that needs to be done that could be accomplished in a quick follow-up — or if it's fine, it'd make me feel better about it. (Worst case scenario is backing out the PR if something is quite wrong, but I doubt it'd come to this.)

skobyda:
Hey @garrett , I'm really sorry for forgetting about the video, It's my fault. Here it is:

https://user-images.githubusercontent.com/42733240/179701612-7aed02a3-59a0-4d5a-bad1-6aca2afc68de.mp4

PR #706, opened 2022-05-16
fmf: qemu-virtiofsd does not exist on RHEL 9

martinpitt:
The functionality is merged into the qemu-kvm package there. So only
install it for Fedora ≥ 36.

---

This broke [gating on RHEL 9.1](https://dashboard.osci.redhat.com/#/artifact/brew-build/aid/45292119) (only visible inside Red Hat, sorry).


skobyda:
Re-trying fedora-rawhide

skobyda (approved):
LGTM

PR #708, opened 2022-05-16
test: Adjust for rhel-8-7

martinpitt:


test/check-machines-migrate
```diff
@@ -163,7 +163,7 @@ class TestMachinesMigration(VirtualMachinesCase):
 
         b.set_checked("#temporary", temporary)
 
-        if self.machine.image in ["rhel-8-6", "rhel-9-0", "rhel-9-1", "centos-9-stream"]:
+        if self.machine.image.startswith("rhel") or self.machine.image.startswith("centos"):
```
martinpitt, line 166: This looked intuitively correct -- why would it be disabled on rhel 8 and centos 9, but work on centos 8? But that's [exactly what happens](https://logs.cockpit-project.org/logs/pull-708-20220516-132615-c6bcc969-centos-8-stream/log.html#61)... So, fixing this to be "centos-9" again.

martinpitt:
[This](https://logs.cockpit-project.org/logs/pull-708-20220516-134627-fe2ed248-rhel-8-7/log.html#27-2) just seems to be an overly picky test. Added another commit to adjust.

martinpitt:
Wohoo, [success](https://logs.cockpit-project.org/logs/pull-708-20220517-081150-b846f3b6-rhel-8-7/log.html)! Re-running with full tests.

skobyda:
[centos-stream-9-x86_64](https://artifacts.dev.testing-farm.io/35b19672-5708-4219-94b2-e8aae4f7ca13/) failed, seems like flakes, re-running

test/check-machines-disks
```diff
@@ -235,7 +235,7 @@ class TestMachinesDisks(VirtualMachinesCase):
         # Test domstats
         self.wait_for_disk_stats("subVmTest1", "vda")
         if b.is_present("#vm-subVmTest1-disks-vda-used"):
-            b.wait_in_text("#vm-subVmTest1-disks-vda-used", "0.0")
+            b.wait_in_text("#vm-subVmTest1-disks-vda-used", "0")
```
skobyda, line 238: I think this needs more specific fix, and now values such as 0.1 GiB, 1.01 GiB... would pass test, although they are incorrect. What about using "0 GiB" for the test?

skobyda (changes requested):

test/check-machines-disks
martinpitt: So I suppose it needs to accept either "0 GiB" or "0.0 GiB"? I'll do a test run to see how that works.

martinpitt: On F36 it is [0.01 GiB](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-708-20220517-112858-3f1fc2c0-fedora-36/log.html#48-2). Will adjust. So I figure the point of this test is to check "almost zero"?

skobyda: Test [failed](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-708-20220517-112858-3f1fc2c0-fedora-36/log.html#48-2), as it seems that the value is also sometimes 0.01GiB. That seems like a race condition, perhaps stats about disk usage weren't updated yet, and test caught it before it could?

It seems we update disks stats every 5000ms: https://github.com/cockpit-project/cockpit-machines/blob/main/src/libvirtApi/common.js#L165
So maybe we should put a timeout before this test condition for 5000ms?

skobyda: So I tried running it locally, and put some print of "virsh domblkinfo" to see the disk allocation, and it seems that allocation value of disk is not fixed.  For me, it started with 17121280 bytes, then went to 18235392 bytes, all of which are roughly 0.01GiB, so that value of 0.01GiB is also correct.

If we want that test to not be so unpredictable, we could do something like:
```
            allocation = m.execute("virsh domblkinfo --domain subVmTest1 --device vda | grep 'Allocation:' | awk '{print $2}'").strip()
            b.wait_in_text("#vm-subVmTest1-disks-vda-used", math.floor(int(allocation)/1024/1024/1024 * 100) / 100.0)
```)
But I think that's maybe unnecessary and overkill, and would probably introduce other race conditions, since as I mentioned, the value of disk allocation fluctuates all the time on a running VM.

PR #721, opened 2022-05-23
test: Add cases for storage pool deletion

yunmingyang:


test/check-machines-storage-pools
```diff
@@ -748,6 +748,238 @@ class TestMachinesStoragePools(VirtualMachinesCase):
             remove=True,
         ).execute()
 
+    def testStoragePooslDeletion(self):
```
skobyda, line 751:
```suggestion
    def testStoragePoolsDeletion(self):
```

skobyda (changes requested):
This is amazing work, thanks!

There are also some checks for storage-pool deletion in [testStoragePool](https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-storage-pools#L45) (you can just CTRL+F for the string "delete" to see all those checks)
Some of them are a duplicate of some of your tests (e.g. [deletion of inactive pool](https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-storage-pools#L194)), so that test should be deleted from testStoragePool, so we don't test the same thing twice.
Other checks (e.g. d[eletion of transient pool](https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-storage-pools#L250)) should maybe be included in testStoragePoolsDeletion too?

yunmingyang:
Sure, I will check them later and re-push, thanks. And I also check the failed jobs, it seems that 
"[0523/065157.086428:ERROR:cert_verify_proc_builtin.cc(681)] CertVerifyProcBuiltin for 127.0.0.2 failed:
----- Certificate i=0 (CN=fedora-36-127-0-0-2-2201,O=bcc8400d49b44879b9e69d2a326a8378,C=US) -----
ERROR: No matching issuer found" is the reason why the case failed, " shows the reason why the cases failed. Do you have any ideas about that?

test/check-machines-storage-pools
```diff
@@ -748,6 +748,238 @@ class TestMachinesStoragePools(VirtualMachinesCase):
...
+
+                # Check if deletion dialog is shown
+                b.wait_visible("div.pf-c-modal-box")
+
+                if self.activate:
+                    if self.vol_name or self.pool_type == "iscsi":
+                        b.set_checked("#storage-pool-delete-volumes", self.remove_volume)
+                else:
+                    # Check the content of inactive pool deletion dialog
+                    b.wait_text(".pf-c-form.pf-m-horizontal", "Deleting an inactive storage pool will only undefine the pool. Its content will not be deleted.")
```
skobyda, line 865: > Sure, I will check them later and re-push, thanks. And I also check the failed jobs, it seems that
"[0523/065157.086428:ERROR:cert_verify_proc_builtin.cc(681)] CertVerifyProcBuiltin for 127.0.0.2 failed:
----- Certificate i=0 (CN=fedora-36-127-0-0-2-2201,O=bcc8400d49b44879b9e69d2a326a8378,C=US) -----
ERROR: No matching issuer found" is the reason why the case failed, " shows the reason why the cases failed. Do you have any ideas about that?

Where do you see that error? I only see tests failing with the error:
`wait_js_cond(ph_is_present(".pf-c-form.pf-m-horizontal")): Uncaught (in promise) Error: condition did not become true`
Which should be fixed by this change:
```suggestion
                    b.wait_text(".pf-c-form .pf-m-horizontal", "Deleting an inactive storage pool will only undefine the pool. Its content will not be deleted.")
```

skobyda (changes requested):

test/check-machines-storage-pools
yunmingyang: The error could be found in results of debian-stable, fedora 35, fedora 36 and etc. It seems that results of rhel-8.7 and rhel-9.1 doesn't contain this error. 

Also, thanks to point the selector problem, I have modify it locally, and will push it later

yunmingyang:
After some checks, I move the cases which are "inactive pool deletion", "active pool deletion" and "transient pool deletion", to testStoragePoolsDeletion. As these cases are about "dir" type pool deletion. Before moving them, the cases in testStoragePoolsDeletion are about other type pool deletion(netfs, disk and iscsi). So, maybe we should keep the cases about "dir" type pool deletion, and move them from testStoragePools to testStoragePoolsDeletion.

skobyda:
Rebasing

yunmingyang:
Re-push as there are several fixes in the previous commit

yunmingyang:
Re-push, since arch has no iscsid and add a skip for that

martinpitt:
Rebased, as the branch was 20 commits behind, so that packit tests can run. Triggering our CI as well. Thanks! I'll leave the final review to @skobyda when he comes back next week, but I'll do an initial one.

test/check-machines-storage-pools
```diff
@@ -748,6 +748,239 @@ class TestMachinesStoragePools(VirtualMachinesCase):
...
+        b = self.browser
+        m = self.machine
+
+        # Prepare some storage pools dependencies
+        # NFS
+        m.execute("if selinuxenabled 2>/dev/null; then setsebool -P virt_use_nfs 1; fi")
+        self.restore_file("/etc/exports")
+        nfs_pool = os.path.join(self.vm_tmpdir, "nfs_pool")
+        mnt_exports = os.path.join(self.vm_tmpdir, "mnt_exports")
+        m.execute(f"mkdir {nfs_pool} {mnt_exports} && echo '{mnt_exports} 127.0.0.1/24(rw,sync,no_root_squash,no_subtree_check,fsid=0)' > /etc/exports")
```
martinpitt, line 761: The restore_file and that echo can be done more elegantly (and robustly) with

    self.write_file("/etc/exports", "...content...", append=True, post_restore_action="systemctl restart nfs-server")

This will also restart nfs-server after restoring.

martinpitt (commented):
Thanks! LGTM, just one possible little improvement. Let's see what CI says.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/3545 to address the [C9S](https://artifacts.dev.testing-farm.io/887a8c47-3853-412a-bb4e-934bf1c3762a/) and [rawhide](https://artifacts.dev.testing-farm.io/14e2072c-2204-4cdf-9f06-daca0057eb11/) failures.

test/check-machines-storage-pools
yunmingyang: OK, re-push

test/check-machines-storage-pools
```diff
@@ -748,6 +748,242 @@ class TestMachinesStoragePools(VirtualMachinesCase):
...
+                        elif self.pool_type == "dir":
+                            vol_path = os.path.join(self.target, self.vol_name)
+                            wait(lambda: m.execute(f'test -f {vol_path}'))
+                        elif self.pool_type == "disk":
+                            wait(lambda: self.vol_name in m.execute("lsblk"))
+
+                # general check
+                self.test_obj.waitPoolRow(self.name, present=False)
+                wait(lambda: self.name not in m.execute("virsh pool-list --all"))
+
```
skobyda, line 914: How about also testing a deletion of dir storage pools? That way, we can remove a lot of code from testStoragePools, making this file a bit shorter :)

test/check-machines-storage-pools
```diff
@@ -748,6 +748,242 @@ class TestMachinesStoragePools(VirtualMachinesCase):
...
+                            wait(lambda: self.vol_name not in m.execute("lsblk"))
+                    elif (not self.remove_volume) and self.vol_name:
+                        if self.pool_type == "netfs":
+                            vol_path = os.path.join(self.source_path, self.vol_name)
+                            wait(lambda: m.execute(f'test -f {vol_path}'))
+                        elif self.pool_type == "dir":
+                            vol_path = os.path.join(self.target, self.vol_name)
+                            wait(lambda: m.execute(f'test -f {vol_path}'))
+                        elif self.pool_type == "disk":
+                            wait(lambda: self.vol_name in m.execute("lsblk"))
```
skobyda, line 909: I think this nested conditions can be simplified a bit a believe:

```suggestion
                if self.activate and self.vol_name:
                    if self.remove_volume:
                        if self.pool_type == "netfs":
                            wait(lambda: self.vol_name not in m.execute(f'ls -A {self.source_path}'))
                        elif self.pool_type == "dir":
                            wait(lambda: self.vol_name not in m.execute(f'ls -A {self.target}'))
                        elif self.pool_type == "disk":
                            wait(lambda: self.vol_name not in m.execute("lsblk"))
                    else:
                        if self.pool_type == "netfs":
                            vol_path = os.path.join(self.source_path, self.vol_name)
                            wait(lambda: m.execute(f'test -f {vol_path}'))
                        elif self.pool_type == "dir":
                            vol_path = os.path.join(self.target, self.vol_name)
                            wait(lambda: m.execute(f'test -f {vol_path}'))
                        elif self.pool_type == "disk":
                            wait(lambda: self.vol_name in m.execute("lsblk"))
```

skobyda (changes requested):

skobyda:
I pushed a fixup commits which should fix my above mentioned comments. @yunmingyang  Can you check if you are fine with the changes?

yunmingyang:
@skobyda LGTM for these changes. Many thanks.

skobyda:
All green, apart from fedora-rawide which fails with AttributeError: '_DebugOutcome' object has no attribute 'errors'. That one unrelated flake.

martinpitt:
The _DebugOutcome thing got fixed recently in PR #748.

I pushed a rebase and squashed the commits. @jelly , @allisonkarlitskaya , can you please have a look at the [protection checks failure](https://github.com/cockpit-project/cockpit-machines/runs/7160367566?check_suite_focus=true) here?

martinpitt (approved):

PR #725, opened 2022-05-31
Bump Cockpit API to 270 + Firefox test fix, packit cleanup

martinpitt:


martinpitt:
We need to bump this further to pick up https://github.com/cockpit-project/cockpit/pull/17429 . So this is release critical now, as Firefox 101 now fails tests.

martinpitt:
[rawhide](https://artifacts.dev.testing-farm.io/18b3632b-165d-4953-9cd3-450ddb0052cc/) now has an entirely new set of problems, but it's not due to Firefox 101 any more. So I guess let's land this, and then look at these other ones?

marusak (approved):

PR #728, opened 2022-05-31
Fix and refactor alert notifications

skobyda:
This fixes 2 issues:
- Set resourceId of notification during error handling of VNC console:  'resourceId' not being set to vm.id would result in a notification not being shown on the related VM page.
- Show notifications also on VM empty state page: This fixes a bug where if an error happens during VM creation, no notification is visible on the currently open (VM empty state) page.

It also refactors a lot of code related to rendering these error AlertGroup notifications.


src/app.jsx
```diff
@@ -112,10 +112,9 @@ class AppActive extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            notifications: {},
+            notifications: [],
```
martinpitt, line 115: The commit message is a bit confusing: "There is no reason for notifications list to be an array". This commit changes it from a map to a list, and "list" and "array" are pretty much the same.

src/app.jsx
```diff
@@ -171,39 +170,34 @@ class AppActive extends React.Component {
...
-            },
-            notificationIdCnt: this.state.notificationIdCnt + 1,
+            notifications: this.state.notifications.concat([notification]), // append new notification to the end of array
             resourceHasError,
         });
     }
 
     /*
-     * Removes the notification with index notificationIndex from the notifications Object.
+     * Removes the notification with index notificationIndex from the notifications arrat.
```
martinpitt, line 180: Hmmm, arrats! :wink:

src/components/vm/consoles/vnc.jsx
```diff
@@ -128,7 +128,8 @@ class Vnc extends React.Component {
                         return domainSendKey({ connectionName, id: vmId, keyCodes: [Enum.KEY_LEFTCTRL, Enum.KEY_LEFTALT, Enum[cockpit.format("KEY_$0", keyName.toUpperCase())]] })
                                 .catch(ex => onAddErrorNotification({
                                     text: cockpit.format(_("Failed to send key Ctrl+Alt+$0 to VM $1"), keyName, vmName),
-                                    detail: ex.message
+                                    detail: ex.message,
+                                    resourceId: vmId
```
martinpitt, line 132: Tiniest thing: This is why we usually use trailing commas for the last element. It avoids annoying extra deltas like for ex.message when adding a new element. (Feel free to ignore)

src/app.jsx
skobyda: Nice catch, I meant to say There is no reason for notifications list to be an object

martinpitt:
Ah, they don't rebase, so nevermind.

PR #730, opened 2022-06-01
Show "Remove" button only for usb and pci device

skobyda:


src/helpers.js
```diff
@@ -819,7 +819,7 @@ export function getNodeDevSource(dev) {
 
         source = `${bus}.${device}`;
     } else {
-        throw new Error(`getNodeDevSource: unsupport device type '${dev.type}'`);
+        console.warn(`getNodeDevSource: unsupport device type '${dev.type}'`);
```
martinpitt, line 822: Fair enough, but then you actually need to handle this in the caller:
```
    devices.forEach(dev => {
        args.push("--add-device", "--hostdev", getNodeDevSource(dev));
    });
```

This is going to crash on an invalid command line (or even data type).

I suppose the following commit tries to avoid that by restricting the supported device types -- but code should be more defensive, as the lookup could fail for usb/pci as well.

src/helpers.js
```diff
@@ -846,7 +846,7 @@ export function getHostDevSourceObject(dev) {
 
         source = { vendor, product, bus, device };
     } else {
-        throw new Error(`getHostDevSourceObject: unsupport device type '${dev.type}'`);
+        console.warn(`getHostDevSourceObject: unsupport device type '${dev.type}'`);
```
martinpitt, line 849: This is also going to crash. `getHostDevElemBySource()` does not get along with `source === undefined `, you can't call `"x" in undefined`.

martinpitt (changes requested):
Thanks!

skobyda:
All green, apart from centos-stream-9 which fails with `AttributeError: '_DebugOutcome' object has no attribute 'errors'`. That one unrelated flake.

martinpitt:
Hmm, github is playing tricks on me. I got an email notification about "All green, apart from centos-stream-9 which fails with `AttributeError: '_DebugOutcome' object has no attribute 'errors'`. That one unrelated flake.", but I don't see any recent pushes or comments here. This needs some fixes and a rebase. The _DebugOutcome thing got fixed recently in PR #748

src/helpers.js
skobyda: So I fixed it up. We use those 2 functions for 2 use cases basically, and here is what can happen in both situations:
1. Show device's source to in the UI. This is not critical, so I think here it's sufficient that console.warn() just prints the message and that's it.
2. Use device's source for attach/detach operations. This is critical, and should fail with a "throw" error. This is in both cases caught buy standard .then().catch(), and then dealth with by a caller (e.g. showing the error in the modal),

martinpitt:
Tests [fail on C9s](https://artifacts.dev.testing-farm.io/6baf9c7a-f4c2-4370-827a-eb990e615169/) with
```
error: Failed to attach device from /tmp/scsihost_hostdevxml
error: unsupported configuration: This QEMU doesn't support vhost-scsi devices
```

@skobyda : This doesn't look related to this PR, but needs to be addressed. This just smells like a missing test dependency in test/browser/main.fmf (or possibly test/browser/browser.sh if it's dependent on `$TEST_OS`)

martinpitt:
C9S succeeded 7 hours ago in PR #755, so it's not obvious that it's unrelated.

test/check-machines-hostdevs
```diff
@@ -105,6 +105,17 @@ class TestMachinesHostDevs(VirtualMachinesCase):
 
         b.wait_in_text("#vm-subVmTest1-hostdev-1-source #1-slot", "0000:00:0f.0")
 
+        # QEMU version on rhel and centos-8-stream doesn't support scsi-host devices yet
+        if not m.image.startswith("rhel-8") and not m.image.startswith("rhel-9") and m.image != "centos-8-stream":
```
martinpitt, line 109: Ah, here -- you also need to cover "centos-9-stream".

martinpitt (changes requested):
Thanks! Looks mostly good, but needs fixing tests.

martinpitt (approved):
Dakujem!

PR #734, opened 2022-06-07
Drop dist cache

martinpitt:
This was introduced for packit SRPM builds when that still happened in
sandcastle, as that did not have enough RAM to run webpack. But since we
moved to COPR for building the SRPMS, this does not apply any more.

Building the webpack is not a big enough burden for local developers to
justify the overhead and extra dist git repo setup, so drop it. This
also unbreaks packit tests for PRs whose source branch is not completely
up to date with origin/main.

With this, c-machines' build system is now very similar to c-podman's.

----

 - [x] Builds on top of PR #725 
 - [ ] After this lands, drop https://github.com/cockpit-project/cockpit-machines-dist/


martinpitt:
The [SRPM build log](https://copr-be.cloud.fedoraproject.org/results/packit/cockpit-project-cockpit-machines-734/srpm-builds/04503622/builder-live.log.gz) shows that runnnig webpack takes pretty much exactly one minute. This is offset, and even over-compensated by not having to wait for the github "build" workflow (that also took > 1 minute, plus the overhead of upload/download/polling)

Makefile
```diff
@@ -62,12 +62,7 @@ packaging/debian/changelog: packaging/debian/changelog.in
 	sed 's/VERSION/$(VERSION)/' $< > $@
 
 $(WEBPACK_TEST): $(COCKPIT_REPO_STAMP) $(shell find src/ -type f) package.json webpack.config.js
-	test/download-dist $${DOWNLOAD_DIST_OPTIONS:-} || \
```
skobyda, line 65: Can we also delete `test/download-dist` file, since nothing is now triggering it?

skobyda (changes requested):

Makefile
martinpitt: Thanks for spotting! Removed.

skobyda (approved):
LGTM (as long as tests pass)

PR #736, opened 2022-06-08
Fetch tools/node-modules in release workflow

skobyda:
tools/node-modules are not defautly present on cockpit-machines. Instead
they should be fetched from cockpit repository.

This broke release of 270: https://github.com/cockpit-project/cockpit-machines/runs/6792784262


martinpitt (approved):
Argh argh, thanks!

PR #737, opened 2022-06-09
Re-drop node_modules from release tarballs, Include Cockpit repo files in release tarballs

martinpitt:
Commit 523df295a had the unintended side effect of including
node_modules/ into the dist tarball, as `git ls-files` now includes
`node_modules. This made the tarball balloon from 1.2 to 74 MB, and make
`tar` take ages to compress.

Explicitly ignore the directory.

----

Compare the tar size of [270](https://github.com/cockpit-project/cockpit-machines/releases/tag/270) and https://github.com/cockpit-project/cockpit-machines/releases/tag/269. Like for https://github.com/cockpit-project/cockpit-podman/pull/1013 this breaks Debian releases, so release-blocker and I'd like to tag a 270.1 after this lands.


martinpitt:
I added another fix for https://src.fedoraproject.org/rpms/cockpit-machines/pull-request/4# (as they both touch the same place)

Makefile
```diff
@@ -18,6 +18,28 @@ COCKPIT_REPO_STAMP = pkg/lib/cockpit.js
 
 all: $(WEBPACK_TEST)
 
+# checkout common files from Cockpit repository required to build this project;
+# this has no API stability guarantee, so check out a stable tag when you start
+# a new project, use the latest release, and update it from time to time
```
martinpitt, line 23: I moved this code to the top of the file, as variables must be declared before use.

marusak (approved):
Thanks!

PR #739, opened 2022-06-12
packit: Drop FMF hack, enable downstream job notifications

martinpitt:
tmt 1.14.0 can finally discover tests from the dist-git source, instead
of having them to checkout from upstream git.

https://src.fedoraproject.org/rpms/cockpit-machines/pull-request/4
dropped the hardcoded `ref:`, so we can drop all that hackery.


allisonkarlitskaya (approved):
OK.

PR #742, opened 2022-06-17
test: allow targetctl failing to pass on Fedora

jelly:
Work around targetcli erroring out but cleaning up the test ramdisk due
to https://bugzilla.redhat.com/show_bug.cgi?id=2093976


marusak (approved):
Thank you!

PR #743, opened 2022-06-21
.github: add reposchutz workflow

jelly:
Verify the commits in the node_modules repository using the similiar
workflow as cockpit's with a small adjustment to checkout the
tools/node-modules script.


martinpitt (approved):
Thanks! srpm build failure is some infra problem, unrelated.

PR #744, opened 2022-06-23
Robustify OS name sorting in VM creation dialog

skobyda:
Split from https://github.com/cockpit-project/cockpit-machines/pull/704

- OS version now has higher priority in sorting than release date
  (It solves the problem when OS of lower version, e.g. rhel-8.11,
   is released later than OS of higher version, e.g. rhel-9.0)
- If neither os version or release date are present, or both are the
  same, sort by by OS name shown in the UI
  (e.g. centos-8 and centos-8-stream have the same version and release
   date, so the only sensible thing left to do is to sort by their name)

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -379,13 +379,20 @@ class OSRow extends React.Component {
                 .map(os => correctSpecialCases(os))
                 .filter(os => filterReleaseEolDates(os) && !IGNORE_VENDORS.find(vendor => vendor == os.vendor))
                 .sort((a, b) => {
-                    if (a.vendor == b.vendor)
-                        if (a.releaseDate || b.releaseDate)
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thanks! LGTM when green.

garrett (changes requested):
This doesn't work as expected.

In a lot of places, numbers descend.

image
image

But look at Debian...!

image

...Is this because 9 > 1?

Don't we have a version field to sort by using a natural number-based sort?

skobyda:
Fixed the sort to use natural sorting.

martinpitt (dismissed):
Thanks!

garrett (approved):
Looks great now! Thanks!

martinpitt (approved):
Thanks!

skobyda:
fedora-rawhide failed with: `AttributeError: '_DebugOutcome' object has no attribute 'errors'`.
I've seen this error in other PRs, so unrelatede to this one

PR #748, opened 2022-06-30
Makefile: Sync with starter-kit; bring in Python 3.11 test fix

martinpitt:


martinpitt:
So this works in general, just some [iSCSI tests are broken](https://artifacts.dev.testing-farm.io/278549d8-c609-44d7-a85f-5907bb987e86/) due to `module 'inspect' has no attribute 'getargspec'`.

That still needs to be investigated, it happened while we couldn't run tests on rawhide at all. There is [one minor pixel test failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-748-20220630-072450-5f5f748e-fedora-35/log.html) due to updating the new cockpit pkg/lib, I'll adjusts it.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/3575 to naughty the iscsi crash, let's run again to pick this up.

martinpitt:
Empty force-push to trigger tests, github had a hiccup before and didn't run any.

jelly:
Damn, now we ran into a maximum timeout:

```
Maximum test time '50m' exceeded.
Adjust the test 'duration' attribute if necessary.
```

Is the failing test waiting too long?

martinpitt:
Hmm, none of the iSCSI tests take particularly long, but we also have some other naughties -- in total, 6 tests are broken in Rawhide. These are exactly the ones which I'd like to *keep*, because they are the ones which would have held back broken packages if we would do proper testing in Fedora.

However, c-machines has been very close to the 50 minute limit for quite some time already, we need to drop some more tests. I'll try a re-run ([previous log](https://artifacts.dev.testing-farm.io/61b6eabd-d500-4fcb-85ec-bc70818efd54/)), but in the meantime, I'm looking at `grep ': ok ' /tmp/log.txt`. None of them take an absurd amount of time, the biggest ones are `TestMachinesDisks.testAddDiskNFS` (3:40), `TestMachinesFilesystems.testBasic{System,Session}Connection` (~ 2:20), `TestMachinesHostDevs.testHostDevAddSessionConnection` (2:17), `TestMachinesNICs.testNICAdd` (2:20).

So I think we should better cut down the ones which we don't desperately need for reverse dependency testing, instead of the ones taking the longest time. On that, I'd drop these:

  - `TestMachinesFilesystems.testBasicSessionConnection` (we already test system connection, and session is more difficult to get right due to different owners and SELInux)
  - `TestMachinesLifecycle.testCloneSystemConnection` (we already test session connection, and it feels the API is very similar to system)

Let's try with these two for now?

jelly (approved):

martinpitt:
Down to "ok 6", so still 5 more tests to go. But I propose to land this as incremental improvement.

PR #751, opened 2022-07-05
Rewire dialogs

mvollmer:


martinpitt:
Converting to draft due to the remaining two test failures. This looks nice, thanks!

mvollmer:
> I'll have a think about whether to let the dialogs keep reacting to external events...

Yes, we have to do this.  At the same time, we might be able to clean up the code significantly by reading directly out of `store.getState()` in each component, instead of handing down the state via a dozen properties.

martinpitt (dismissed):
Amazing work, thank you! I'm happy with this. Leaving the last word to @skobyda , though.

Long version: I mentioned yesterday that I don't particularly like redux, and this change binds us even more tightly to it: more things access the store directly now, instead of getting the info passed as props through the component hierarchy (React style). But I understood you that maintaining the state is expensive at least for some operations, so we want to make sure we don't do system API queries (i.e. libvirt-dbus or file system scans and such) more than once. So using the redux store as some kind of "mirror model" of the system, keeping it up to date in libvirtApi/, and building everything to just listen to the model makes some sense. Getting rid of the "passing down props" overhead is at least quite a win in terms of making components more independent from one another! :+1:

src/app.jsx
```diff
@@ -305,10 +305,7 @@ class AppActive extends React.Component {
                 }
                 {path.length > 0 && path[0] == 'networks' &&
                 <NetworkList networks={networks}
-                    resourceHasError={this.state.resourceHasError}
-                    vms={vms}
-                    nodeDevices={nodeDevices}
-                    interfaces={interfaces} />
+                             resourceHasError={this.state.resourceHasError} />
```
skobyda, line 308: Just a question, but why hasn't similar cleanup been done for other components? e.g. storagePools, networks, nodeDevice are also passed down to <VmDetailsPage>, <HostVmsList>, and vms array is passed to <StoragePoolList>, etc... 
I'm however fine with doing that in a followup, just asking if there is any problem which prevents similar cleanups for other components.

mvollmer: > Just a question, but why hasn't similar cleanup been done for other components?

Because it would have been more work, and wasn't necessary to keep the dialog behavior unchanged.  If you all are happy with this direction, I can work more on this.

src/components/storagePools/storagePoolDelete.jsx
```diff
@@ -207,10 +212,11 @@ export class StoragePoolDelete extends React.Component {
 StoragePoolDelete.propTypes = {
     storagePool: PropTypes.object.isRequired,
     vms: PropTypes.array.isRequired,
-    close: PropTypes.func.isRequired,
 };
 
 export const StoragePoolDeleteAction = ({ storagePool, open, vms }) => {
```
skobyda, line 217: This `open` parameter is not used anywhere in this function anymore.

src/components/networks/createNetworkDialog.jsx
```diff
@@ -420,41 +424,16 @@ class CreateNetworkModal extends React.Component {
     }
 }
 CreateNetworkModal.propTypes = {
-    close: PropTypes.func.isRequired,
-    devices: PropTypes.array.isRequired,
 };
```
skobyda, line 427: This empty proptypes object doesn't make much sense, I guess it was left here by accident, and should be deleted:
```
CreateNetworkModal.propTypes = {
};
```

src/components/storagePools/storagePool.jsx
```diff
@@ -206,10 +205,6 @@ class StoragePoolActions extends React.Component {
                           isOpen={this.state.isActionOpen}
                           position='right'
                           dropdownItems={dropdownItems} />
-                {this.state.showDeleteModal &&
-                <StoragePoolDelete close={() => this.setState({ showDeleteModal: false })}
-                                   storagePool={storagePool}
-                                   vms={vms} />}
```
skobyda, line 212: `StoragePoolActions` component still has `showDeleteModal` property initialized in its state. It should be removed from there too.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -417,7 +421,7 @@ export class AddDiskModalBody extends React.Component {
             if (poolTypesNotSupportingVolumeCreation.includes(this.state.storagePoolType)) {
                 validationFailed.storagePool = cockpit.format(_("Pool type $0 does not support volume creation"), this.state.storagePoolType);
             }
-            const poolCapacity = parseFloat(convertToUnit(this.props.storagePools.find(pool => pool.name == this.state.storagePoolName).capacity, units.B, this.state.unit));
+            const poolCapacity = parseFloat(convertToUnit(this.state.storagePools.find(pool => pool.name == this.state.storagePoolName).capacity, units.B, this.state.unit));
```
skobyda, line 424: It seems no test caught this, but this looks like a bug. `this.state` has no `storagePools` paremeter. I guess you meant:
```suggestion
            const poolCapacity = parseFloat(convertToUnit(storagePools.find(pool => pool.name == this.state.storagePoolName).capacity, units.B, this.state.unit));
```

src/components/storagePools/storagePoolList.jsx
```diff
@@ -45,7 +46,8 @@ export class StoragePoolList extends React.Component {
         const actions = (<CreateStoragePoolAction loggedUser={loggedUser} libvirtVersion={libvirtVersion} />);
 
         return (
-            <Page groupProps={{ sticky: 'top' }}
+            <WithDialogs>
+                <Page groupProps={{ sticky: 'top' }}
                   isBreadcrumbGrouped
                   breadcrumb={
```
skobyda, line 52: Just a nitpick, but this now looks a bit misintended. Strange that compiler doesn't throw an error "Expected intendation of A, but found B"

src/components/networks/createNetworkDialog.jsx
```diff
@@ -386,7 +390,7 @@ class CreateNetworkModal extends React.Component {
                                        onValueChanged={this.onValueChanged} />
                 { (this.state.forwardMode === "nat" || this.state.forwardMode === "route") &&
                 <NetworkDeviceRow dialogValues={this.state}
-                                  devices={this.props.devices}
+                                  devices={getNetworkDevices()}
```
skobyda, line 393: Anoter nitpick, but since we are now trying to avoid unnecessary passing of arguments, maybe we can call `getNetworkDevices()` directly in `NetworkDeviceRow` component, and get rid of "devices" argument.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1254,36 +1260,12 @@ export class CreateVmAction extends React.Component {
...
-            </>
-        );
+        return createButton;
     }
 }
 
 CreateVmAction.propTypes = {
     mode: PropTypes.string.isRequired,
-    networks: PropTypes.array.isRequired,
-    nodeDevices: PropTypes.array.isRequired,
```
skobyda, line 1286: These 2 parameters are still passed CreateVmAction component in src/app.jsx

skobyda (changes requested):

src/components/storagePools/storagePoolDelete.jsx
mvollmer: Good catch, thanks!  Fixed

src/components/networks/createNetworkDialog.jsx
mvollmer: Done.

src/components/storagePools/storagePool.jsx
mvollmer: Yes, thansk! Fixed.

src/components/storagePools/storagePoolList.jsx
mvollmer: Yeah... I have jus added the `WithDialogs` wrapper and then let eslint take care of the indentation... it doesn't always seem to do a great job... Fixed now.

src/components/create-vm-dialog/createVmDialog.jsx
mvollmer: Yes, thanks! Fixed.

src/components/vm/disks/diskAdd.jsx
mvollmer: `this.state.storagePools` is set by componentDidMount, via `get initialState()`.

src/components/networks/createNetworkDialog.jsx
mvollmer: Yes, good idea.

skobyda:
Packit tests `fedora-rawhide` fails with:
`wait_js_cond(!ph_is_present("#vm-subVmTest1-add-iface-dialog")): timeout`
That seems a bit related, no?

src/components/vm/disks/diskAdd.jsx
skobyda: Ah, didn't notice that, thanks

mvollmer:
> Packit tests `fedora-rawhide` fails with: `wait_js_cond(!ph_is_present("#vm-subVmTest1-add-iface-dialog")): timeout` That seems a bit related, no?

This is the screenshot:

image

The content of the dialog looks ok, no?

src/app.jsx
skobyda: I'm fine with doing that in the future, as that can be split to a different PR

mvollmer:
> Both centos and feora-rawhide packit test fail on timeout. It was recently changed to 60 minutes: #753 So I guess rebase should make packit tests green

Done. I have seen the 50m timeout a lot as well.  In our CI, tests finish in 15 minutes or so, no?

martinpitt:
@mvollmer : Yes, but our CI runs 8 tests in parallel, while on TF we can't have parallelism.

skobyda (approved):
Amazing work. Looks good and all tests greens, thanks

garrett:
Alerts should not be at the bottom of modals.

Referring to this screenshot:

image

Please open a follow-up PR to fix this.

PR #753, opened 2022-07-05
test: increase test timeout

jelly:
Fedora Rawhide timeouts now on 50 minutes, we already skipped two tests
in 7972eb8ae4985bede952 which did not improve the situation. This makes
the timeout the same as cockpit-podman.

### DIFF
--- a/test/browser/main.fmf
+++ b/test/browser/main.fmf
@@ -18,4 +18,4 @@ require:
   - libvirt-daemon-driver-storage-logical
   - targetcli
 test: ./browser.sh
-duration: 50m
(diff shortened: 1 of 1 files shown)


jelly:
After 60 minutes rawhide is still not happy, so we need to adjusts the tests really.

jelly:
> After 60 minutes rawhide is still not happy, so we need to adjusts the tests really.

And now it is! That's really strange but would unblock https://github.com/cockpit-project/cockpit-machines/pull/755

martinpitt (approved):
Thanks!

PR #756, opened 2022-07-11
Show volume's usage in "Add disk" dialog as a Form helper text

### LINKED ISSUE #754: Add disk inline alert should be a contextual form error
This is a contextual, targeted error, but it's styled as an inline alert.

image

Inline alerts should only be used when the error is generic and at the top of a modal. It's best to have an error that's specific and located next to where the problem comes from (which this actually does) and styled as a contextual error (which this is not currently).

See: https://www.patternfly.org/v4/components/form/design-guidelines/#errors-and-validation

skobyda:
Instead of an inline dialog, an information about volume being used a
disk by another VM should be shown as a form helper text.

Fixes #754

### DIFF
--- a/src/components/vm/disks/diskAdd.jsx
+++ b/src/components/vm/disks/diskAdd.jsx
@@ -18,7 +18,7 @@
  */
 import React, { useState, useEffect } from 'react';
 import {
-    Alert, Bullseye, Button, Checkbox,
+    Bullseye, Button, Checkbox,
     ExpandableSection, Form, FormGroup,
     FormSelect, FormSelectOption,
     Grid,
@@ -64,7 +64,23 @@ function getFilteredVolumes(vmStoragePool, disks) {
     return filteredVolumesSorted;
 }
 
-const SelectExistingVolume = ({ idPrefix, storagePoolName, existingVolumeName, onValueChanged, vmStoragePools, vmDisks }) => {
+function getDiskUsageMessage(vms, storagePool, volumeName) {
+    const isVolumeUsed = getStorageVolumesUsage(vms, storagePool);
+
+    if (!isVolumeUsed[volumeName] || (isVolumeUsed[volumeName].length === 0))
+        return null;
+
+    const vmsUsing = isVolumeUsed[volumeName].join(', ');
+    const volume = storagePool.volumes.find(vol => vol.name === volumeName);
+
+    let message = cockpit.format(_("This volume is already used by $0."), vmsUsing);
+    if (volume.format === "raw")
+        message += " " + _("Attaching it will make this disk shareable for every VM using it.");
+
+    return message;
+}
+
+const SelectExistingVolume = ({ idPrefix, storagePoolName, existingVolumeName, onValueChanged, vmStoragePools, vmDisks, vms }) => {
     const vmStoragePool = vmStoragePools.find(pool => pool.name == storagePoolName);
     const filteredVolumes = getFilteredVolumes(vmStoragePool, vmDisks);
 
@@ -86,11 +102,16 @@ const SelectExistingVolume = ({ idPrefix, storagePoolName, existingVolumeName, o
         initiallySelected = "empty";
     }
 
+    const diskUsageMessage = getDiskUsageMessage(vms, vmStoragePools.find(pool => pool.name === storagePoolName), existingVolumeName);
     return (
-        <FormGroup fieldId={`${idPrefix}-select-volume`} label={_("Volume")}>
+        <FormGroup fieldId={`${idPrefix}-select-volume`}
+                   label={_("Volume")}
+                   helperText={diskUsageMessage}
+                   validated="warning">
             <FormSelect id={`${idPrefix}-select-volume`}
                         onChange={value => onValueChanged('existingVolumeName', value)}
                         value={initiallySelected}
+                        validated={diskUsageMessage && "warning"}
                         isDisabled={!filteredVolumes.length}>
                 {content}
             </FormSelect>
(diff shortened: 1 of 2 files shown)


skobyda:
Alternatively we also discussed showing this as an error or a text variant. As there are use cases (e.g. disk sharing) where user might want to attach a VM used by another VM, therefore it should not be an error. I gave it some thought and I agree with what was mentioned on standup, that it's probably the best fitting as a warning variant, as disk sharing is quite a corner case.

Screenshot from 2022-07-12 02-00-33

jelly:
Please rebase for the packit tests, the timeout is now 60 minutes.

garrett (approved):
Thanks! That's better.

garrett (changes requested):
Ah, wait. Why isn't the select showing an underline and icon?

https://www.patternfly.org/v4/components/form/design-guidelines/#errors-and-validation

I think it's supposed to look like this:

image

https://www.patternfly.org/v4/components/select/html#warning

For React, set `validated` to `warning`.

skobyda:
Fixed:
Screenshot from 2022-07-13 01-08-16

skobyda:
Tests getting quite green, @martinpitt @jelly @marusak Mind giving it a code review?

garrett (dismissed):
Looks good! Thanks!

Minor nitpick on the string; it shouldn't use a colon and period. Please remove both. Once this is done, it's good to go from my end.

image

It'd look like this:

image

skobyda:
@garrett FYI There is one more situation where the warning contains 2 sentences. If the disk is raw, it's being automatically transfprmed to "shareable" disk, and the warning looks like this:
Screenshot from 2022-07-13 12-03-13

skobyda:
Updated the PR.

### For non-raw disks, the warning looks like this:
Screenshot from 2022-07-13 12-07-16

### For raw disks, the warning looks like this:
Screenshot from 2022-07-13 12-06-53

### If there is multiple VMs using the disk, they are separated by comas:
Screenshot from 2022-07-13 12-09-01

src/components/vm/disks/diskAdd.jsx
```diff
@@ -63,7 +63,22 @@ function getFilteredVolumes(vmStoragePool, disks) {
...
 
-const SelectExistingVolume = ({ idPrefix, storagePoolName, existingVolumeName, onValueChanged, vmStoragePools, vmDisks }) => {
+function getDiskUsageMessage(vms, storagePool, volumeName) {
+    const isVolumeUsed = getStorageVolumesUsage(vms, storagePool);
+
+    if (!isVolumeUsed[volumeName] || (isVolumeUsed[volumeName].length === 0))
+        return null;
+
+    const vmsUsing = isVolumeUsed[volumeName].join(', ');
+    let text = _("This volume is already used by ") + vmsUsing;
```
martinpitt, line 73: This doesn't work -- not every language can put the object at the very end. Please use "... by $0" and cockpit.format().

src/components/vm/disks/diskAdd.jsx
```diff
@@ -63,7 +63,22 @@ function getFilteredVolumes(vmStoragePool, disks) {
...
+    const volume = storagePool.volumes.find(vol => vol.name === volumeName);
+    if (volume.format === "raw")
+        text += ". " + _("Attaching it will make this disk shareable for every VM using it.");
```
martinpitt, line 76: This kind of addition may also be problematic. It's not a given that every language uses a period. Honestly, I'd rather compromise a little and always add the period to the "already used by ..." message. Otherwise you need *two* string templates which just differ by ".", and that's a little rude towards translators. But this doesn't map grammar-wise very well. E.g. in German you would always end a sentence with a period. (Honestly it's rather confusing that English doesn't? or is that just some PF guideline? If so, it's very English centric.)

Alternatively, force a line break and drop the period from both sentences? Then there's at least some consistency in the UI.

martinpitt (changes requested):
Dakujem!

garrett:
What does "Attaching it will make this disk shareable for every VM using it" mean? :thinking:

skobyda:
> What does "Attaching it will make this disk shareable for every VM using it" mean? thinking

It will make it "Writeable and shareable"
Screenshot from 2022-07-14 12-34-30
https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.1/html/administration_guide/shareable_disks_in_red_hat_enterprise_virtualization
The whole discussion why we do this can be found here: https://github.com/cockpit-project/cockpit/pull/12314#issuecomment-529871971

src/components/vm/disks/diskAdd.jsx
skobyda: I like the line break idea. We use something similar for snapshots:

skobyda: On the second thought, it seems that pattenrfly forces one-liner text for Form helperText by ignoring `\n`. We could do some workaround, but maybe it's just simpler to use sentences with dots on the same line.

martinpitt: +1 on just using proper sentences, yes.

skobyda: Reverted back to using sentences ending with a comma. But let's wait if @garrett agrees.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -64,7 +64,23 @@ function getFilteredVolumes(vmStoragePool, disks) {
...
+function getDiskUsageMessage(vms, storagePool, volumeName) {
+    const isVolumeUsed = getStorageVolumesUsage(vms, storagePool);
+
+    if (!isVolumeUsed[volumeName] || (isVolumeUsed[volumeName].length === 0))
+        return null;
+
+    const vmsUsing = isVolumeUsed[volumeName].join(', ');
+    const volume = storagePool.volumes.find(vol => vol.name === volumeName);
+
+    let message = cockpit.format(_("This volume is already used by $0. "), vmsUsing);
```
martinpitt, line 76: Please no leading or trailing spaces, they are almost guaranteed to get lost in translation. Please add the " " separately (and not translated)

martinpitt (changes requested):

martinpitt (approved):
Dakujem!

garrett:
Oh, this is another one that was prematurely merged.

As I pointed out, the string "Attaching it will make this disk shareable for every VM using it" doesn't make sense. I couldn't suggest anything else as I didn't know what it meant (nor would our users), so I asked what it meant.

I still don't know:
- What "it" is
- If this will attach it if it's selected
- What "attach" means in this context of the "add" dialog (adding the disk will "attach" it right?)
- If this is a bad thing

This needs an immediate fixup to land before the release or it should be reverted.

src/components/vm/disks/diskAdd.jsx
garrett: You didn't wait before merging.

garrett:
Additionally, "Previously taken snapshots allow you to revert to an earlier state if something goes wrong"?!?!?

But there are no snapshots that have been previously taken. It's an empty state!

This also needs rewording.

martinpitt:
> Additionally, "Previously taken snapshots allow you to revert to an earlier state if something goes wrong"?!?!?

This PR does not touch snapshots, so this needs to be changed independently. I'll split it out into an issue.

PR #758, opened 2022-07-12
Fix link to the guide

bLuka:
Looks like the guide URL changed.
Fix the address and add it as a link within the README.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,4 +1,4 @@
-Please refer guide/html/feature-virtualmachines.html to general information and how to set up nested virtualization.
+Please refer [to the following guide](https://cockpit-project.org/guide/195/feature-virtualmachines.html) for general information and how to set up nested virtualization.
 
 With nested virtualization enabled, Vagrant can be used to try cockpit-machines:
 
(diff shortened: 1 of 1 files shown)


src/README.md
```diff
@@ -1,4 +1,4 @@
-Please refer guide/html/feature-virtualmachines.html to general information and how to set up nested virtualization.
+Please refer [to the following guide](https://cockpit-project.org/guide/195/feature-virtualmachines.html) for general information and how to set up nested virtualization.
```
martinpitt, line 1: This is an ancient version. This feature page simply does not exist any more. How about linking to https://www.linux-kvm.org/page/Nested_Guests for this aspect, and drop the "general information"? Or maybe link to the [libvirt documentation](https://libvirt.org/docs.html), although at first sight I don't see an user-facing "how to set this up" doc.

martinpitt (changes requested):
Thanks for spotting this!

src/README.md
bLuka: Yup, that's better this way :+1:

I find the KVM doc looks perfect for the purpose.

martinpitt (approved):
Thank you!

PR #759, opened 2022-07-13
Show toast notication if VM's storage deletion fails

skobyda:
Show toast notication if VM's storage deletion fails

If VM deletion fails user is shown an error message in the VM deletion
dialog. But if VM deletion passes, but subsequent deletion of its
storage fails, there is no error message shown, as the VM along with
its dialog is already deleted and a VM deletion dialog unmounted from
the UI.
Fix this by adding a warning message about unsuccessful storage deletion
to the toast alerts.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2105984

### DIFF
--- a/src/components/vm/deleteDialog.jsx
+++ b/src/components/vm/deleteDialog.jsx
@@ -94,10 +94,10 @@ const DeleteDialogBody = ({ disks, vmName, destroy, onChange }) => {
         );
 
     return (
-        <div className="modal-body">
+        <>
             {alert}
             {disksBody}
-        </div>
+        </>
     );
 };
 
--- a/src/components/vm/deleteDialog.jsx
+++ b/src/components/vm/deleteDialog.jsx
@@ -23,6 +23,7 @@ import {
     Button,
     DataList, DataListItem, DataListItemRow, DataListCheck, DataListItemCells, DataListCell,
     Form,
+    FormGroup,
     Modal
 } from '@patternfly/react-core';
 import { ExclamationTriangleIcon } from '@patternfly/react-icons';
@@ -78,27 +79,17 @@ const DeleteDialogBody = ({ disks, vmName, destroy, onChange }) => {
         );
     }
 
-    let alert = null;
-    if (destroy)
-        alert = <p>{cockpit.format(_("The VM $0 is running and will be forced off before deletion."), vmName)}</p>;
-
-    let disksBody = null;
-    if (disks.length > 0)
-        disksBody = (
-            <Form onSubmit={e => e.preventDefault()}>
-                <p>{_("Delete associated storage files:")}</p>
+    return (<Form onSubmit={e => e.preventDefault()}>
+        <FormGroup>
+            {destroy && <p>{cockpit.format(_("The VM $0 is running and will be forced off before deletion."), vmName)}</p>}
+            {disks.length > 0 && <>
+                <p className="pf-u-mb-sm">{_("Delete associated storage files:")}</p>
                 <DataList isCompact>
                     { disks.map(disk_row) }
                 </DataList>
-            </Form>
-        );
-
-    return (
-        <>
-            {alert}
-            {disksBody}
-        </>
-    );
+            </>}
+        </FormGroup>
+    </Form>);
 };
 
 export class DeleteDialog extends React.Component {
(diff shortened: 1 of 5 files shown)


skobyda:
If Vm deletion fails:
Screenshot from 2022-07-13 10-55-49

If VM's storage deletion fails:
Screenshot from 2022-07-13 11-47-43

garrett:
What happened with the spacing? The spacing should be like this:

image

(More vertical spacing is needed nearly everywhere except between the list and its declaration.)

Here's the difference with an "onion skin" approach (to let you see through to the original):

image

The text needs work too.
"Could not delete almalinux8-2022-7-13-G"

Additionally, what can someone do here in this modal to make deletion work? I don't see anything that can be done different. Therefore, it should be a notification after the deletion attempt, not within the modal itself.

garrett:
Why is this notification so far down the page? It's overlapping the action buttons for the VM list.

Screenshot from 2022-07-13 11-56-15

Additionally, it should probably say something like "Could not delete storage volume for almalinux8-2022-7-13-G"

martinpitt (changes requested):
Thank you! This approach is good, just some details to work out. Plus of course the spacing issues from Garrett.

skobyda:
> Additionally, it should probably say something like "Could not delete storage volume for almalinux8-2022-7-13-G"

Not every disk attached to the VM has to be a storage volume. Sometimes it can be a direct block device, or a file attached directly to the VM. So how about "Could not delete disk for. .." or just "Could not delete storage for..."?
Or do you not mean storage volume not in a virtualization sense, but in a broader sense as volume == `any identifiable unit of data storage`?

> Additionally, what can someone do here in this modal to make deletion work? I don't see anything that can be done different. Therefore, it should be a notification after the deletion attempt, not within the modal itself.

That could also work, but I see it for another PR, as we do this kind of thing all over c-machines, and currently, Toast Alerts in VM are shown only on the main page, not on the VM page, so it would require bigger refactoring.

src/components/vm/deleteDialog.jsx
```diff
@@ -137,17 +137,41 @@ export class DeleteDialog extends React.Component {
...
+                .then(() => {
+                    return domainDelete({
+                        name: vm.name,
+                        id: vm.id,
+                        connectionName: vm.connectionName,
+                        live: this.props.vm.state != 'shut off',
+                        storagePools
+                    })
+                            .catch(exc => {
+                                console.warn(cockpit.format(_("VM $0 failed to get deleted: $1"), vm.name, exc.message));
```
martinpitt: Please drop the translation of this. The message id is different than the error message below, it's redundant work for translators, and in bug reports we actually want the English messages.

Edit: the following commit fixes that.

martinpitt (approved):
Could be squashed a bit, but it's good enough. Thank you! I just retried a flake.

skobyda:
Current screenshot:

Fixed spacing of in-modal error:
Screenshot from 2022-07-19 14-09-25

Warning message if storage deletion fails:
Screenshot from 2022-07-19 14-08-55

As described on IRC, the issue with a spacing for Toast Alerts is unrelated to this PR, as it was already existing on the main branch:
Screenshot from 2022-07-19 14-08-55
And therefore should be fixed in a different PR.

skobyda:
FYI The notification is overlapping with actions only if you click on "Show more". If you don't, it's not overlapping with anything: 
Screenshot from 2022-07-19 14-14-19

garrett:
We definitely should fix that error so it's not in the modal dialog. Can you work on a follow-up for this? Thanks!

> FYI The notification is overlapping with actions only if you click on "Show more". And "Show more" can have very long text in some cases, so even if we move notification a bit up, it may still end up covering action buttons if you click on "Show more". If you don't click on "show more", it's not overlapping with anything

It's still not in the right place (if it's at the top-right of the page, it should be at the top-right of the page with a little bit of margin, not nudged down this much), but it's less bad in the collapsed state. Thanks for pointing that out!

skobyda:
> We definitely should fix that error so it's not in the modal dialog. Can you work on a follow-up for this? Thanks!

As I mentioned, we do this all across c-machines, so I created an [issue](https://github.com/cockpit-project/cockpit-machines/issues/763) and plan to fix it in multiple places at once.


> > FYI The notification is overlapping with actions only if you click on "Show more". And "Show more" can have very long text in some cases, so even if we move notification a bit up, it may still end up covering action buttons if you click on "Show more". If you don't click on "show more", it's not overlapping with anything
> 
> It's still not in the right place (if it's at the top-right of the page, it should be at the top-right of the page with a little bit of margin, not nudged down this much), but it's less bad in the collapsed state. Thanks for pointing that out!

PR #767, opened 2022-07-19
Always reset disk's device type to 'disk' unless it's an iso file

skobyda:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2107247


martinpitt (changes requested):
Thank you! The fix looks fine, but the test fails everywhere.

martinpitt (approved):
Thank you!

PR #770, opened 2022-07-20
Fix leftover stylelint issues

jelly:
Stylelint's glob needs to be quoted for it to also lint nested folders.

https://github.com/stylelint/stylelint/issues/1970

### DIFF
--- a/src/components/vm/hostdevs/hostDevAdd.scss
+++ b/src/components/vm/hostdevs/hostDevAdd.scss
@@ -1,4 +1,4 @@
-@import "_global-variables.scss";
+@use "_global-variables.scss";
 
 .pf-c-table.vm-device-table {
   /* Set overflow with a limiter on the list, so it scrolls instead of the dialog */
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "build": "webpack",
     "eslint": "eslint --ext .jsx --ext .js src/",
     "eslint:fix": "eslint --fix --ext .jsx --ext .js src/",
-    "stylelint": "stylelint src/*/*{.css,scss}",
-    "stylelint:fix": "stylelint --fix src/*/*{.css,scss}"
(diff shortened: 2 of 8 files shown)


jelly:
Forgot to push... node-modules.

martinpitt (approved):
Thanks! Do we need a similar scripts fix in our other projects?

jelly:
> Thanks! Do we need a similar scripts fix in our other projects?

Not for c-podman, but would be safe to do so:

```
[jelle@t14s][~/projects/cockpit-podman]%fd css src
src/ContainerTerminal.css
src/Containers.scss
src/ImageRunModal.scss
src/ImageSearchModal.css
src/Images.css
src/podman.scss
```

src/components/vm/hostdevs/hostDevAdd.scss
```diff
@@ -1,4 +1,4 @@
-@import "_global-variables.scss";
+@use "_global-variables.scss";
```
garrett, line 1: - Switching to use breaks the point of the import.
- However, I don't see anything using variables in this file, so the line itself is probably incorrect and should probably be removed anyway? 
- Also: Imports should only exist at toplevel in most cases, as these sub-files that are also included in a page will pull in duplicate CSS. (Of course, anything that relies on SASS variables (not CSS custom properties) should be imported. This is the exception. And anything that needs to pull in SASS variables should not have CSS in the file that it is importing.)

jelly: This was originally introduced in `818cb5d10b718c3933246c93dde64617c701a3b6` I think for `--pf-c-table--m-compact-th--PaddingTop`? But no, it compiles with out it.

So I guess `--pf-c-table--m-compact-th--PaddingTop` is just implicitly available through one of our existing imports, in that case I hope we have a pixel test for that dialog.

PR #774, opened 2022-07-22
Downloading rhel images: Impromevents of offline token management

skobyda:
- Check validity of offline token immediatelly after supplied by a user
- Add saving and loading offline tokens from localStorage

---
## Machines: Improvements of offline token management 

Offline token, which needs to be specified to download a RHEL OS, is now validated on the fly. 
demo1

Valid tokens are then saved by Cockpit and re-used on the next occasion.
demo2

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -594,7 +594,8 @@ const OfflineTokenRow = ({ offlineToken, onValueChanged, formValidationFailed })
                       disabled={disabled}
                       minLength={1}
                       value={offlineToken || ""}
-                      onChange={setOfflineTokenHelper} />
+                      onChange={setOfflineTokenHelper}
+                      rows="4" />
         </FormGroup>
     );
 };
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -21,6 +21,7 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { debounce } from 'throttle-debounce';
 import {
+    Flex, FlexItem,
     Form, FormGroup,
     FormSelect, FormSelectOption, FormSelectOptionGroup,
     InputGroup,
@@ -28,7 +29,8 @@ import {
     Select as PFSelect, SelectOption, SelectVariant,
     Tabs, Tab, TabTitleText,
     TextInput,
-    Button, Tooltip, TextArea
+    Button, Tooltip, TextArea,
+    Spinner,
 } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
@@ -75,6 +77,7 @@ import {
 } from "./createVmDialogUtils.js";
 import { domainCreate } from '../../libvirtApi/domain.js';
 import { storagePoolRefresh } from '../../libvirtApi/storagePool.js';
+import { getAccessToken } from '../../libvirtApi/rhel-images.js';
 import { PasswordFormFields, password_quality } from 'cockpit-components-password.jsx';
 
 import './createVmDialog.scss';
@@ -216,6 +219,9 @@ function validateParams(vmParams) {
             validationFailed.source = _("Installation source must not be empty");
     }
 
+    if (vmParams.os && needsRHToken(vmParams.os.shortId) && isEmpty(vmParams.offlineToken))
+        validationFailed.offlineToken = _("Offline token must not be empty");
+
     if (vmParams.memorySize === 0) {
         validationFailed.memory = _("Memory must not be 0");
     }
@@ -260,7 +266,7 @@ const NameRow = ({ vmName, suggestedVmName, onValueChanged, validationFailed }) 
     );
 };
 
-const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, os, osInfoList, offlineToken, cloudInitSupported, downloadOSSupported, onValueChanged, validationFailed }) => {
+const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, os, osInfoList, cloudInitSupported, downloadOSSupported, offlineToken, onValueChanged, validationFailed }) => {
     let installationSource;
     let installationSourceId;
     let installationSourceWarning;
@@ -377,7 +383,7 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, 
                         <OfflineTokenRow
                             offlineToken={offlineToken}
                             onValueChanged={onValueChanged}
-                            validationFailed={validationFailed} />}
+                            formValidationFailed={validationFailed} />}
                 </>}
         </>
     );
@@ -466,26 +472,85 @@ class OSRow extends React.Component {
     }
 }
 
-const OfflineTokenRow = ({ offlineToken, onValueChanged, validationFailed }) => {
-    const validationStateToken = validationFailed.offlineToken ? 'error' : 'default';
+// This method needs to be outside of component as re-render would create a new instance of debounce
+// Debounce will trigger "getAccessToken" only if >500ms has passed since user changed offlineToken
+// Since "getAccessToken" basically triggers HTTP request, this prevents user from triggering dozens of HTTP requests
+// while typing an offline token
+const getAccessTokenDebounce = debounce(500, (offlineToken, onValueChanged, setValidationState, validationStates) => {
+    getAccessToken(offlineToken)
+            .then(out => {
+                const accessToken = out.trim();
+                onValueChanged("accessToken", accessToken);
+                setValidationState(validationStates.SUCCESS);
+            })
+            .catch(ex => {
+                console.error(`Offline token validation failed: "${JSON.stringify(ex)}"`);
+                onValueChanged("accessToken", "");
+                setValidationState(validationStates.FAILED);
+            });
+});
+
+const OfflineTokenRow = ({ offlineToken, onValueChanged, formValidationFailed }) => {
+    const link = <a href="https://access.redhat.com/management/api" target="_blank" rel="noopener noreferrer">
+        <ExternalLinkAltIcon className="pf-u-mr-xs" />
+        {_("Get a new RHSM token.")}
+    </a>;
+    const validationStates = {
+        DEFAULT: {
+            option: "default",
+            message: <span id="token-helper-message">
+                { link }
+                {" " + _("Then copy and paste it above.")}
+            </span>,
+        },
+        INPROGRESS: {
+            option: "default",
+            message: <span id="token-helper-message"><Spinner isSVG size="md" /> {_("Checking token validity...")}</span>,
+        },
+        FAILED: {
+            option: "error",
+            message: <Flex id="token-helper-message">
+                <FlexItem className="invalid-token-helper" grow={{ default: 'grow' }}>{_("Error checking token") + " "}</FlexItem>
+                <FlexItem>
+                    { link }
+                    {" " + _("Then copy and paste it above.")}
+                </FlexItem>
+            </Flex>,
+        },
+        SUCCESS: {
+            option: "success",
+            message: _("Valid token"),
+        },
+    };
+
+    const [validationState, setValidationState] = useState(validationStates.DEFAULT);
+
+    const setOfflineTokenHelper = (offlineToken) => {
+        onValueChanged("offlineToken", offlineToken);
+        // Reset accessToken to prevent race conditions where state could still have access token paired with an old offline token
+        // e.g. user inputs offline token, we obtain an access token, then they change offline token and quickly click on "Create" before new access token can be obtained
+        onValueChanged("accessToken", "");
+
+        if (isEmpty(offlineToken)) {
+            setValidationState(validationStates.DEFAULT);
+            onValueChanged("accessToken", "");
+        } else {
+            setValidationState(validationStates.INPROGRESS);
+            getAccessTokenDebounce(offlineToken, onValueChanged, setValidationState, validationStates);
+        }
+    };
 
     return (
         <FormGroup label={_("Offline token")} fieldId="offline-token"
                    id="offline-token-group"
-                   helperTextInvalid={validationFailed.offlineToken}
-                   validated={validationStateToken}
-                   helperText={<span>
-                       <a href="https://access.redhat.com/management/api" target="_blank" rel="noopener noreferrer">
-                           <ExternalLinkAltIcon className="pf-u-mr-xs" />
-                           {_("Get a new RHSM token.")}
-                       </a>
-                       {" " + _(" Then copy the token and paste it above.")}
-                   </span>}>
-            <TextArea id='offline-token'
-                      validated={validationStateToken}
+                   validated={formValidationFailed.offlineToken ? "error" : validationState.option}
+                   helperText={validationState.option !== "error" && validationState.message}
+                   helperTextInvalid={formValidationFailed.offlineToken || (validationState.option === "error" && validationState.message)}>
+            <TextArea id="offline-token"
+                      validated={formValidationFailed.offlineToken ? "error" : validationState.option}
                       minLength={1}
-                      value={offlineToken || ''}
-                      onChange={value => onValueChanged('offlineToken', value)} />
+                      value={offlineToken || ""}
+                      onChange={setOfflineTokenHelper} />
         </FormGroup>
     );
 };
@@ -825,6 +890,7 @@ class CreateVmModal extends React.Component {
             userPassword: '',
             rootPassword: '',
             userLogin: '',
+            accessToken: '',
             offlineToken: '',
         };
         this.onCreateClicked = this.onCreateClicked.bind(this);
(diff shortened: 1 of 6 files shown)


skobyda:
From now on, the validity of an offline token is validated immediately once user supplies it:

https://user-images.githubusercontent.com/42733240/180436040-474d74ea-1aef-4cb9-b44d-7757347246bb.mp4

Once user supplies a valid token, it is saved into a localStorage, and loaded again into the UI when user tries to Download a RHEL VM:

https://user-images.githubusercontent.com/42733240/180436025-67aa89f5-7c28-445c-9af7-0380ed19b045.mp4

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -465,26 +465,72 @@ class OSRow extends React.Component {
...
+            getAccessToken(offlineToken)
+                    .then(out => {
+                        const accessToken = out.trim();
+                        onValueChanged("accessToken", accessToken);
+                        setValidationState(validationStates.SUCCESS);
+                    })
+                    .catch(ex => {
+                        console.error(`Offline token validation failed: "${JSON.stringify(ex)}"`);
+                        onValueChanged("accessToken", "");
+                        setValidationState(validationStates.FAILED);
```
martinpitt, line 549: Can you please check the "success" and "failure" code paths in tests? I suggest to split the PR to do the "check token" part first, as that otherwise looks fine.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -498,6 +504,26 @@ const OfflineTokenRow = ({ onValueChanged, validationFailed }) => {
     const [validationState, setValidationState] = useState(validationStates.DEFAULT);
     const [offlineToken, setOfflineToken] = useState("");
 
+    useEffect(() => {
+        loadOfflineToken((token) => {
+            if (token) {
+                setValidationState(validationStates.LOADING);
+                getAccessToken(token)
+                        .then(out => {
+                            console.log(out);
```
martinpitt, line 513: Debugging leftover. This is a secret, let's not log these.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -482,6 +484,10 @@ const OfflineTokenRow = ({ onValueChanged, validationFailed }) => {
             option: "default",
             message: _("Validating..."),
         },
+        LOADING: {
+            option: "default",
+            message: _("Checking for saved token..."),
```
martinpitt, line 489: This doesn't involve the network or any heavy computation, right? Just checking localStorage. Is it really worth introducing a state and translatable string for this? How long do you actually see this? This should be a small fraction of a second.

If it does take longer than 0.5s for some reason, maybe a spinner is enough?

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -508,9 +534,14 @@ const OfflineTokenRow = ({ onValueChanged, validationFailed }) => {
             setValidationState(validationStates.INPROGRESS);
             getAccessToken(offlineToken)
                     .then(out => {
+                        console.log(out);
                         const accessToken = out.trim();
                         onValueChanged("accessToken", accessToken);
                         setValidationState(validationStates.SUCCESS);
+
+                        // TODO encrypt offline token
```
martinpitt, line 542: But with what? I don't want c-machines to become a password manager, and the user having to set and remember some random password here. Either we accept the risk of stealing local storage secrets from a browser (which should be manageable, as localStorage is per origin), or we don't, and then we should instead use sessionStorage and leave longer-term secrets management to the user.

My gut feeling is that localStorage is okay, with some fixes (see other threads).

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -508,9 +534,14 @@ const OfflineTokenRow = ({ onValueChanged, validationFailed }) => {
...
+                        return saveOfflineToken(offlineToken)
+                                .catch(ex => console.warn(`Could not save offline token to a storage: "${JSON.stringify(ex)}"`));
                     })
                     .catch(ex => {
                         console.error(`Offline token validation failed: "${JSON.stringify(ex)}"`);
```
martinpitt, line 547: If the token is invalid, it should be cleared from localStorage. I suppose the most common reason is that is lifetime is over?

In general, we need to take care to clean up the tokens somehow. Doing that here should be enough.

src/components/create-vm-dialog/createVmDialogUtils.js
```diff
@@ -145,3 +146,11 @@ export function correctSpecialCases(os) {
 export function needsRHToken(osName) {
     return osName.startsWith("rhel");
 }
+
+export function saveOfflineToken(offlineToken) {
+    return cockpit.user().then(user => localStorage.setItem(user.id + "-offline-token", offlineToken));
```
martinpitt, line 151: Thinking aloud.. this uses the same name for all remote hosts. But that should be fine, the token is not dependent on a particular machine, and the browser user can use the same token for all of them. But the same is true for `cockpit.user()` -- if I log into a remote machine as two different Linux users, the human browser user can still use the same token. Hence I think this should drop the per-user storage here.

This also avoids cluttering the storage with an unbounded number of tokens, and this remains self-cleaning.

src/components/create-vm-dialog/createVmDialogUtils.js
```diff
@@ -145,3 +146,11 @@ export function correctSpecialCases(os) {
...
+}
+
+export function loadOfflineToken(setToken) {
+    return cockpit.user().then(user => setToken(localStorage.getItem(user.id + "-offline-token")));
```
martinpitt, line 155: This needs a test. AFAIK the current tests already create a fake server and token, thus this functionality should be testable (log out and back in)

martinpitt (changes requested):
Thank you! This isn't half as scary as I thought, *phew* :grin:

garrett (changes requested):
Looks great, overall! I have a few small changes to the text and icons.

image

Checking for offline token sure seems to take a long time to query LocalStorage. This most certainly must be a bug. If it's combining loading from LocalStorage and checking the validity, please load and show the token immediately and then say something like "Checking token validity..." (possibly with a spinner in front of the text).

image

The token text is truncated. Is the entry tall enough? It should probably be a little taller, right?

The icon is also getting overlapped by the scrollbar, so a third of it is not even visible. I guess fixing the height so it isn't scrolled will fix this too.

There is an auto-resizing textarea variant @ https://www.patternfly.org/v4/components/text-area#auto-resizing — this _might_ help, but I'm not sure.

image

It should probably be "[Get a new RHSM token](#), then copy the token and paste it above." — basically changing the first period to a comma.

image

Please add a spinner during validation process. Also, what is it validating? (You and I know what it's doing. This is a rhetorical question.) The text should probably be "Checking token validity..." ((mentioned above) as it is not actually validating the token, but checking to see if the token is currently valid.

garrett:
Here's how I thought the saved token state would happen:

vm-rhel-token-saved excalidraw

(That dotted circle graphic in checking validity should be a spinner.)

And if there's an error, like a network error or something, we could do this:

vm-rhel-token-saved-error excalidraw

(Although if there is a network error, then perhaps keep the token there? Or not?

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: How about using some information about the server which the user is connected to as a secret? E.g. its UUID?

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: It does invole network delay of checking validity of the token thru RHSM REST API. But the text will be changed to "Checking validity of saved token" with a spinner next to it as shown in the new mockup: https://github.com/cockpit-project/cockpit-machines/pull/774#issuecomment-1210577729

src/components/create-vm-dialog/createVmDialogUtils.js
skobyda: I know its a niche scenario, but if multiple different users log thru the same computer, one will get an offline token of the other user.
Since the benefit of the cockpit is that you manage your server from anywhere, let's imagine scenario of 2 colleagues, or a scenario in some lab where user logs into cockpit with some public/shared client.
I know it's a niche scenario, but offline token is a very sensitive information, since with that you can manage users[ whole Red Hat subscription](https://access.redhat.com/management/api/rhsm), e.g. manage its systems, allocations, errata... And not just subscriptions, but also manage [accounts](https://access.redhat.com/management/api/account) and [cases](https://access.redhat.com/management/api/case_management)

src/components/create-vm-dialog/createVmDialog.jsx
martinpitt: You mean the /etc/machine-id? But that's not secret at all. We are not really concerned about offline attacks here -- people should encrypt their home directories. In a running browser, anyone can read any of the public OS and the user's files. The only secret which we have generally available is the user's login password (but that's not accessible at all from the browser) and the browser's general password keyring (but there's a like a million implementations of that, and websites can't generally control these).

src/components/create-vm-dialog/createVmDialogUtils.js
martinpitt: Different users on the same computer have separate user accounts and sessions, and thus also separate localStorages. If two humans literally use the same linux account, then (1) security-wise they *are* the same identity and share these secrets anyway, and (2) cockpit.user() will then be the same, so storing the tokens per-user name will not change anything here.

My point is, an offline token is *not* attached to a particular linux account on a particular machine. It's attached to the *browser* user. So it doesn't make sense to include the remote account name here.

skobyda:
Hey, I believe this might be ready for another design review @garrett :

Screenshot from 2022-08-17 02-32-48
Screenshot from 2022-08-17 02-27-40
Screenshot from 2022-08-17 02-27-18

skobyda:
@garrett Btw I had a problem with TextArea's Autoresize, because in certain cases, when TextArea's value does not come from user unput but from the cockpit (e.g. autofilling save offline token) it won't resize. There is however a github issue for that: patterfly-react#7834 .
Until then, what about using [this workaround](https://github.com/cockpit-project/cockpit-machines/pull/774/files#diff-1a31234c32b66be1f72784830b534b55932c9a00cca58b40ee792abe3042c83a) and changing the position of an icon manually::
Screenshot from 2022-08-17 10-45-55

skobyda:
All tests green apart from the TF: 
```
Error: Cannot find module 'chrome-remote-interface'
RuntimeError: CDP broken
```

garrett:
Are tokens always going to be that length? We could at least increase the minimum height if they're going to be 3 lines.

What about textarea's `rows="3"` attribute?

skobyda:
> rows="3"

Token have a varied length. But they are all quite long, and would take around 10 - 12 rows:
(Ignore the black rectangle, that's just me lazily censoring my persnal offline token):
Screenshot from 2022-08-17 02-27-40

marusak:
Needs rebase

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -465,26 +471,85 @@ class OSRow extends React.Component {
...
+    const link = <a href="https://access.redhat.com/management/api" target="_blank" rel="noopener noreferrer">
+        <ExternalLinkAltIcon className="pf-u-mr-xs" />
+        {_("Get a new RHSM token.")}
+    </a>;
+    const validationStates = {
+        DEFAULT: {
+            option: "default",
+            message: <span>
+                { link }
+                {" " + _(" Then copy the token and paste it above.")}
```
marusak, line 507: Why is there space concatenated with string starting with empty string? Translators are likely to break the leading space.

src/components/create-vm-dialog/createVmDialogUtils.js
marusak: Agree with Martin. Also this avoids accessing of user name which I always found a bit awkward.

martinpitt (changes requested):
Thanks! There's a few outstanding review comments from the last round. landing this is blocked until after the next release anyway (due to RHEL translation update).

OOI, how did you get the TF tests to succeed? Did you do anything to your fork? (Could it have something to do with your fork having an outdated main branch? I wouldn't see how that'd break it, but I have no other idea..)

marusak:
This is not blocked anymore. Can we drive this to completion?

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -517,13 +522,48 @@ const OfflineTokenRow = ({ offlineToken, onValueChanged, formValidationFailed })
...
+                getAccessToken(token)
+                        .then(out => {
+                            const accessToken = out.trim();
+                            setDisabled(false);
+                            onValueChanged("accessToken", accessToken);
+                            setValidationState(validationStates.SUCCESS);
+                        })
+                        .catch(ex => {
+                            setValidationState(validationStates.EXPIRED);
+                            onValueChanged("offlineToken", "");
```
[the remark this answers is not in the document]
skobyda: It comes from a previous design discussion: https://github.com/cockpit-project/cockpit-machines/pull/774#issuecomment-1210577729

If we just go directly to a DEFAULT state, then the process would look like this: User opens a dialog, there would be a greyed-out area with a old token visible for a few seconds while verification is happening (it turns out it's expired), then UI would empty the offline token field. I believe in that case, it might be confusing, and user might thing "what just happend to an old token I had saved there?". It's not obvious knowledge that tokens expire after 30 days of inactivity.

src/components/create-vm-dialog/createVmDialogUtils.js
```diff
@@ -153,3 +153,15 @@ export function isDownloadableOs(os) {
         // RHSM API supports only rhel versions >= 8: https://access.redhat.com/management/api/rhsm#/images/listImageDownloadsByVersionArch
         (os.version.localeCompare("8.0", undefined, { numeric: true, sensitivity: 'base' }) >= 0));
 }
+
+export function saveOfflineToken(offlineToken) {
+    return localStorage.setItem("offline-token", offlineToken);
```
martinpitt, line 158: This localStorage key is shared between all cockpit pages, right? I wonder if we should namespace this a bit. I'd call it at least `rhsm-offline-token` (as tokens are being used in a lot of places) to avoid naming collisions. It's also fine to spell this out as `redhat-subscription-offline-token`.

Please move this to a constant instead of repeating it in the three functions here.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -517,13 +521,52 @@ const OfflineTokenRow = ({ offlineToken, onValueChanged, formValidationFailed })
...
             option: "success",
             message: _("Valid token"),
         },
     };
 
     const [validationState, setValidationState] = useState(validationStates.DEFAULT);
+    const [disabled, setDisabled] = useState(false);
+
+    useEffect(() => {
+        loadOfflineToken((token) => {
```
martinpitt, line 544: This functionality needs a test. Please log out, back in, and start installing a RHEL image. The dialog should get a valid token immediately. You should also be able to test expiry -- either by dropping it from the fake rhsm (cleaner), or by changing the value in the browser with a direct `localStorage.setItem()` from the tests (quick&dirty, but may be easier).

martinpitt (changes requested):
Thanks, looking really good now!

garrett:
Gentle reminder that I'd like to see a video of this feature in the PR as it is currently, because I'm curious how it turned out after the modifications... and... just in case there's something minor, we could be improve it.

But it's probably fine. I liked what I saw so far and we have integrated changes since.

skobyda:
User gets a new token from Red Hat page:

success

User already has a valid token saved:
valid

Token is expired:
expired

There was an issue trying to reach Red Hat API to validate the token (e.g. network issues):
error

garrett (changes requested):
When you paste in the entry, it gets quite large and the other values scroll out of visible range:

image

I talked about this in [my comment from 18 Aug](https://github.com/cockpit-project/cockpit-machines/pull/774#issuecomment-1219601781):

> I didn't realize they could be that large. I guess having a bit more space is good, but having 12 rows total is quite excessive. I guess this means it will always scroll, which means that your workaround of moving the indicator is correct.
> 
> I still think it needs to be more than it was, but not 10 - 12 rows high.
> 
> 4 or 5 rows tall is probably about right?
> 
> Other than that, I think this PR is about ready and we can handle other things in a follow-up? I don't want to block this for half a month while I'm away, when it's so close.
> 
> So... please:
> 
> -  make the input about 4 or 5 rows tall

Can you please make that form have a max height?

That's my only complaint. It looks great otherwise!

skobyda:
> Can you please make that form have a max height?
> 
> That's my only complaint. It looks great otherwise!

Ah yeah, the max height was already fixed, but I forgot to remove the buggy "Autoresize" option. Now it should work correctly:
fixed_autoresize

garrett (approved):
Thanks for the fix! LGTMN :+1: :rocket:

martinpitt (approved):
Thank you! This should get a release note, can you please add one with a recent video?

PR #784, opened 2022-08-01
Ship npm integration test dependencies in dist tarball

martinpitt:
This avoids having to run `npm`, and thus a potentially brittle internet
access, during FMF test runs. It also avoids the npm test dependency,
which tends to be uninstallable in Rawhide.

We have successfully used this approach in cockpit for a long time.

----

This avoids [this failure](https://artifacts.dev.testing-farm.io/2d44eeb8-bb89-42a8-bd9d-6764c37bed60/), which has been going on for a week [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2110750). This already hit us in the past.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,19 +119,28 @@ devel-uninstall:
 print-version:
 	@echo "$(VERSION)"
 
+# required for running integration tests; commander and ws are deps of chrome-remote-interface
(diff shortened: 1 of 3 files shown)


marusak (approved):
Thanks!

PR #786, opened 2022-08-03
Drop unrequired @use import

jelly:
The global variables are not required any variables in this file.

Follow up from https://github.com/cockpit-project/cockpit-machines/pull/770#discussion_r926308798

After this change the table headers are still aligned, that's why we remove the padding.
image


garrett (approved):
I can confirm. This SCSS file does not use SASS variables, so it shouldn't need our special breakpoints (in the global variables file).

PR #787, opened 2022-08-05
test: Check install phase after shutdown

martinpitt:
Extend testCreateThenInstall() to validate what happens after the
initial "Install" button action.

If the install was successful, we expect `has_install_phase` to be false
after the initial boot and shutdown. It should then show the "Run"
button and not the "Install" one.

If the install failed, `has_install_phase` stays at true, and the VM
continues to offer the "Install" action, but not "Run".

---

This was discussed in [this thread](https://github.com/cockpit-project/cockpit-machines/pull/583#pullrequestreview-932972125). This checks the status quo, but is useful to ensure that PR #583 will not regress this behaviour.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1426,6 +1426,10 @@ vnc_password= "{vnc_passwd}"
                 b.wait_in_text(f"#vm-{name}-{connection}-state", "Shut off")
                 b.wait_visible(f"#vm-{name}-{connection}-install")
                 b.wait_in_text(".pf-c-alert", "failed to get installed")
+                # can reattempt installation
+                b.wait_visible(f"#vm-{name}-{connection}-install")
+                # can't run, as it is still in install phase
+                b.wait_not_present(f"#vm-{name}-{connection}-run")
             else:
                 b.wait_in_text(f"#vm-{name}-{connection}-state", "Running")
                 self._assertCorrectConfiguration(dialog)
@@ -1436,6 +1440,9 @@ vnc_password= "{vnc_passwd}"
                 # unfinished install script runs indefinitely, so we need to force it off
                 self.test_obj.performAction(name, "forceOff", checkExpectedState=False)
                 b.wait_in_text(f"#vm-{name}-{connection}-state", "Shut off")
+                # install phase got cleared
+                b.wait_visible(f"#vm-{name}-{connection}-run")
+                b.wait_not_present(f"#vm-{name}-{connection}-install")
 
             return self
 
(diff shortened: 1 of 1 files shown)


martinpitt:
[pxe test failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-787-20220805-100642-0c85102c-ubuntu-2204/log.html#75), [same thing](https://cockpit-logs.us-east-1.linodeobjects.com/pull-787-20220805-094800-0c85102c-ubuntu-stable/log.html#75). Unrelated, we've been seeing this for some time. Retrying here, and investigating in a separate PR, as it's not quite obvious what fails here.

test/check-machines-create
```diff
@@ -31,7 +31,7 @@ sys.path.append(os.path.join(TEST_DIR, "common"))
 sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 
 from machineslib import VirtualMachinesCase  # noqa
-from testlib import nondestructive, skipImage, test_main, wait, Error  # noqa
+from testlib import nondestructive, skipImage, test_main, wait, Error # noqa
```
skobyda, line 34: Is there any reason for this, or just a typo?
Also I remember we had a E261 rule, but I guess we dropped it.

martinpitt: Argh, fallout from temporarily adding "sit" there. Fixed.

The static code tests are indeed broken, I fixed them in PR #788.

skobyda: I don't think that fixed this specific static-code check. I tried with the latest main checkout, and make codecheck will not catch the E261.

martinpitt: @skobyda : Do you have "flake8" installed? That's what I tried to make more prominent with commit c68e54886315. 

Now that the fix landed, I'll continue on https://github.com/cockpit-project/cockpituous/pull/517 so that flake8 will actually run in CI.

skobyda: I do. I have problem with eslint and python3-vulture, but flake8 works for me just fine, and mentiones ntohign about E261:
```
[skobyda@fedora machines]$ make codecheck
test/static-code --tap
1..8
not ok 1 /static-code/test-eslint
# 
# WARNING (@babel/preset-env): We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.
# 
# You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:
# 
#   npm install --save core-js@2    npm install --save core-js@3
#   yarn add core-js@2              yarn add core-js@3
# 
# More info about useBuiltIns: https://babeljs.io/docs/en/babel-preset-env#usebuiltins
# More info about core-js: https://babeljs.io/docs/en/babel-preset-env#corejs
ok 2 /static-code/test-flake8
ok 3 /static-code/test-include-config-h
ok 4 /static-code/test-js-translatable-strings
ok 5 /static-code/test-json-verify
ok 6 /static-code/test-no-translatable-attr
ok 7 /static-code/test-pyvulture # SKIP no python3-vulture
ok 8 /static-code/test-unsafe-security-policy

```

skobyda (approved):
Thanks, LGTM

PR #788, opened 2022-08-05
test fixes

martinpitt:
Running `a && b` is wrong for our intent here, as it will silently
succeed if `a` fails, and never run `b`. This can often cause follow-up
failures which are harder to investigate.

Use `set -e` where appropriate. In some cases, run the commands separately so
that we will know which one failed exactly.

----

This was triggered by [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-787-20220805-100642-0c85102c-ubuntu-2204/log.html#75), which happens fairly often. The [log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-787-20220805-100642-0c85102c-ubuntu-2204/TestMachinesCreate-testCreatePXE-ubuntu-2204-127.0.0.2-2701-FAIL.log.gz) and the [screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-787-20220805-100642-0c85102c-ubuntu-2204/TestMachinesCreate-testCreatePXE-ubuntu-2204-127.0.0.2-2701-FAIL.png) both confirm that the destroy actually worked, but the undefine either failed, never ran, or did not generate a libvirt-dbus event, or the UI was missing it. Let's find out!

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -471,7 +471,7 @@ vnc_password= "{vnc_passwd}"
             print(self.machine.execute("; ".join(cmd).format(target_iqn)))
             wait(lambda: "unit:0:0:0" in self.machine.execute("virsh pool-refresh iscsi-pool && virsh vol-list iscsi-pool"), delay=3)
 
-            self.addCleanup(self.machine.execute, "virsh pool-destroy iscsi-pool; virsh pool-delete iscsi-pool; virsh pool-undefine iscsi-pool")
+            self.addCleanup(self.machine.execute, "virsh pool-destroy iscsi-pool; virsh pool-undefine iscsi-pool")
 
             self.browser.reload()
             self.browser.enter_page('/machines')
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -296,23 +296,20 @@ class TestMachinesCreate(VirtualMachinesCase):
                                                                            storage_pool="NoStorage",
                                                                            connection="session"))
 
-        self.machine.execute("virsh net-destroy default && virsh net-undefine default")
+        self.machine.execute("set -e; virsh net-destroy default; virsh net-undefine default")
 
         # Set up the PXE server configuration files
         cmds = [
+            "set -e",
             "mkdir -p /var/lib/libvirt/pxe-config",
             f"echo \"{PXE_SERVER_CFG}\" > /var/lib/libvirt/pxe-config/pxe.cfg",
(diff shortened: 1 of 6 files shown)


martinpitt:
Surprisingly, most tests look good here, -create got a good workout (3x retry). Only [c8s](https://cockpit-logs.us-east-1.linodeobjects.com/pull-788-20220805-110015-0a8f6284-centos-8-stream/log.html). I've seen the console errors for quite some time, I'll have a look. 

Moreover, my `--tap` change proved my suspicion that we stopped running these:

    ok 2 /static-code/test-flake8 # SKIP no flake8

We have tons of violations right now, most of them due to line length. I'll check how that was done in cockpit, fix it, and then we can put flake8 into the tasks container.

test/check-machines-create
```diff
@@ -296,23 +296,20 @@ class TestMachinesCreate(VirtualMachinesCase):
                                                                            storage_pool="NoStorage",
                                                                            connection="session"))
 
-        self.machine.execute("virsh net-destroy default && virsh net-undefine default")
+        self.machine.execute("set -e; virsh net-destroy default; virsh net-undefine default")
 
         # Set up the PXE server configuration files
         cmds = [
+            "set -e",
```
skobyda, line 303: Since `set -e` was already called, isn't this one unecessary?

martinpitt: Where was it called? Each self.machine.execute() call runs a new shell. I really wish that API would call the shell with `-e` by default.. It's really always what you want. I have it on my list to do that and fix the fallout in our projects, but let's not block on that.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/3706 to enable `set -e` in Machine.execute() by default. So we can clean up the extra `set -e` again soon -- but I'd like to keep them in for now as it will still take some time to land that PR. But that needs some additional fixes to commands that are *supposed* or at least allowed to fail, see [this log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3706-20220808-044722-9d7a8a86-fedora-36-firefox-cockpit-project-cockpit-machines/log.html#37).

So I added two more commits to prepare c-machines for that change. I run some extra tests against that bots PR.

test/check-machines-create
skobyda: Ahh, I forgot it's called with a new shell, nevermind then

skobyda (approved):
Thanks. LGTM

PR #791, opened 2022-08-09
VM Creation: Do not pass --unattended option to script with Cloud Image

skobyda:
--unattended and --cloud-init options are conflicting when used at the
same time during VM creation. This results in cloud-init user data not
being set for a new VM.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2096200

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -1000,7 +1000,7 @@ class CreateVmModal extends React.Component {
             // then show errors in the notification area
             this.setState({ createMode: startVm ? RUN : EDIT, validate: false });
 
-            const unattendedInstallation = this.state.rootPassword || this.state.userLogin || this.state.userPassword;
+            const unattendedInstallation = !(this.state.sourceType === CLOUD_IMAGE) && (this.state.rootPassword || this.state.userLogin || this.state.userPassword);
             const vmParams = {
                 connectionName: this.state.connectionName,
                 vmName,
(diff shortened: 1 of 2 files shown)


src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1000,7 +1000,7 @@ class CreateVmModal extends React.Component {
             // then show errors in the notification area
             this.setState({ createMode: startVm ? RUN : EDIT, validate: false });
 
-            const unattendedInstallation = this.state.rootPassword || this.state.userLogin || this.state.userPassword;
+            const unattendedInstallation = !(this.state.sourceType === CLOUD_IMAGE) && (this.state.rootPassword || this.state.userLogin || this.state.userPassword);
```
martinpitt, line 1003: This is easier to read with `this.state.sourceType !== CLOUD_IMAGE`, as it otherwise isn't clear at first sight whether or not this applies to the whole condition.

martinpitt (approved):
LGTM, thanks! I retried the test failures, the TF failures are just something really stupid again (there's something still unexplained about having an old/broken node_modules/ there)

One remark in case you have to push again for other reasons.

martinpitt:
This keeps failing on the TF. I looked at such failures a lot already, but I have not the slightest idea what's wrong there, and why this is so specific. I sent multiple c-m PRs recently and they are fine. Someone™ needs to debug this..

skobyda:
I get same failures on [other PR of mine](https://github.com/cockpit-project/cockpit-machines/pull/774/). Maybe it's specific only to PRs which are not created through your account @martinpitt . During the TF tests, is there something specified which is specific yto your account? like your email or nick...

src/components/create-vm-dialog/createVmDialog.jsx
marusak: So you have decided not to change this?

marusak (approved):
Thanks! Re-running tests

PR #793, opened 2022-08-12
test: Clean up after Machine.execute() does `set -e` by default, test fixes

martinpitt:
https://github.com/cockpit-project/bots/pull/3706 allows us to clean up and robustify our tests a lot.

### DIFF
--- a/test/check-machines-consoles
+++ b/test/check-machines-consoles
@@ -74,7 +74,7 @@ class TestMachinesConsoles(VirtualMachinesCase):
         args = self.createVm("subVmTest1", "vnc")
 
         if urlroot != "":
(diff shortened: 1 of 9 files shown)


martinpitt:
[This failure](https://artifacts.dev.testing-farm.io/427a01e2-29ed-4af2-8670-493eb8f48ecf/) needs fixing. It's not directly related to the commit here, but to previous commit 91afb06e56a627. Thus fixing separately.

[This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-793-20220812-061421-91afb06e-fedora-35/log.html#33-2) is also interesting, investigating. This is *exactly* what I want to expose: Silently ignored commands due to failing a previous one.

jelly (approved):

PR #794, opened 2022-08-18
Offer downloading RHEL OS only for RHEL >= 8

skobyda:
RHSM listImageDownloadsByVersionArch API[1] only works for RHEL images
with a version >= 8. Therefore, we should only provide 'Download OS'
option for images with that version

[1] https://access.redhat.com/management/api/rhsm#/images/listImageDownloadsByVersionArch

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2118236

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -71,6 +71,7 @@ import {
     filterReleaseEolDates,
     getOSStringRepresentation,
     needsRHToken,
+    isDownloadableOs,
 } from "./createVmDialogUtils.js";
 import { domainCreate } from '../../libvirtApi/domain.js';
 import { storagePoolRefresh } from '../../libvirtApi/storagePool.js';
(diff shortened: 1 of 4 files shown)


src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -368,7 +369,7 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices,
                 </FormGroup>
                 : <>
                     <OSRow os={os}
-                         osInfoList={osInfoList.filter(os => os.treeInstallable || (needsRHToken(os.shortId) && !os.version.endsWith("unknown")))}
+                         osInfoList={osInfoList.filter(os => isDownloadableOs(os))}
```
marusak, line 372: you can call this just as `osInfoList={osInfoList.filter(isDownloadableOs)}`

marusak:
Nice, thanks! Just one small code style comment, but tests are rather unhappy. I updated BZ as well.

marusak (approved):
Thanks!

PR #795, opened 2022-08-22
test: add checks about shut off button of the VM

yunmingyang:


test/machineslib.py
```diff
@@ -40,6 +40,8 @@ def performAction(self, vmName, action, checkExpectedState=True, connectionName=
             b.wait_in_text("#vm-{0}-{1}-state".format(vmName, connectionName), "Running")
         if action == "forceOff":
             b.wait_in_text("#vm-{0}-{1}-state".format(vmName, connectionName), "Shut off")
+        if action == "off":
+            b.wait_in_text("#vm-{0}-{1}-state".format(vmName, connectionName), "Shut off")
```
marusak, line 45: You defined just new name for the same action, I don't think we need to do that. But more curiously, why force is represented just as 'Shut off' in those tests?

test/machineslib.py
```diff
@@ -40,6 +40,8 @@ def performAction(self, vmName, action, checkExpectedState=True, connectionName=
             b.wait_in_text("#vm-{0}-{1}-state".format(vmName, connectionName), "Running")
         if action == "forceOff":
```
marusak, line 41: why not just doing `if action == "forceOff" or action == "off"` the same way as on line 39?

yunmingyang: Sure

marusak:
rhel-8-7 failed 3 times in a row always on `wait(lambda: "login as 'cirros' user." in self.machine.execute(f"cat {args['logfile']}"), delay=3)`. It is something that needs to be looked at.

yunmingyang:
Hi @marusak, I could not download this image by bots/images-download, so I couldn't reproduce it locally. Could get the args['logfile'] of the VM or add --trace when running the job?

marusak (approved):
Thank you, so the test was just flaky, locally it works as well

PR #799, opened 2022-08-29
test: fix some possible race conditions

yunmingyang:
If the VM boot process finishes too fast,  'wait' will be never successful. Thus, remove one 'log clear' which is not necessary, and change one 'log clear' before VM running


skobyda (approved):
Thanks, LGTM. Let's wait if tests pass.

PR #801, opened 2022-08-31
test: Drop copies of {net,storage}lib.py

martinpitt:
https://github.com/cockpit-project/cockpit/pull/17646 moved these to
test/common, so that we get them through our partial git import now.

Cockpit's storagelib.py add_ram_disk() does not currently accept a
non-default machine argument. Until it does, add a little variable
dance.


marusak (approved):
danke!

Makefile
```diff
@@ -35,7 +35,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
-COCKPIT_REPO_COMMIT = 49a7122df205ab434bab884eb3a7be94d1a8e255 # 273 + PR#17577
+COCKPIT_REPO_COMMIT = 1e5f74b9f5e486fca7aa87670983254a2288f37a # 275 + test/common/{net,storage}lib.py
```
marusak, line 38: don;t we use more than just there files from it though?

martinpitt: yes of course, but adding these files to test/common is the reason that we pull in a main snapshot instead of the latest release.

PR #803, opened 2022-09-01
Use initial password for password fields in VM create dialog

skobyda:
Password fields are different component instances when you switch
between Cloud base image and Download OS. This causes a bug:
When user sets user/root password, and then changes Installation type,
it will reset the passwords, since a different component instance of the
password field rendered. This is however not ideal behaviour.
Using the initial password value would fix this.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2109986

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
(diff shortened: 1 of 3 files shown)


skobyda:
The new test will always fail until https://github.com/cockpit-project/cockpit/pull/17699 is merged

martinpitt (approved):
Cheers! I'm going to squash the commits, as they are the wrong way around.

PR #804, opened 2022-09-05
Immediatelly show password quality message

skobyda:
Usually we used to show message password quality message only after user
clicks on "Create" as part of form validation. Perhaps this should be
shown immediatelly as the user types the password.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2122444

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -678,7 +678,9 @@ const UsersConfigurationRow = ({
                         setRootPasswordMessage('');
                     });
         } else {
+            setRootPasswordErrors({});
             setRootPasswordStrength('');
+            setRootPasswordMessage('');
         }
     }, [rootPassword, validationFailed]);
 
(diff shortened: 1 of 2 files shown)


skobyda:
Before we've only shown the red bar underneath the password. User had to click on "Create" to get the validation message.

Screenshot from 2022-09-05 14-36-11

Now, the message is present as suer type the password:

Screenshot from 2022-09-05 14-35-28

cxhlunar:
https://bugzilla.redhat.com/show_bug.cgi?id=2122444 verification failed and introduced new problem, I add comment in it.
Any by the way, are we going to block the installation if the password check failed ? I remember in the past, we just remind that it's weak password , but it didn't block user to install the vm.

garrett:
> this should be shown immediately as the user types the password

Yes, this feedback is supposed to be immediate, not on submit. Thanks for fixing.

> Any by the way, are we going to block the installation if the password check failed ? 

That would be pretty annoying for someone testing something on a "throwaway" disposable VM from within Cockpit, which is one of the intended uses of Cockpit Machines.

KKoukiou (dismissed):

skobyda:
Ok it seems all comments has been fixed, comment at https://bugzilla.redhat.com/show_bug.cgi?id=2122444#c4 is no longer reproducable, and tests has been fixed for like past 2 weeks.
Why did not I request re-review?

Anyway, @KKoukiou I think this is ready for a re-review.

KKoukiou (approved):

PR #806, opened 2022-09-08
Fix incorrect setting of shareable flag

skobyda:
Because of mistaking boolean value with an empty array, we would sometimes incorrectly set disk's access as "shareable". Fix this and also add test for checking correct disk's access.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1933966


martinpitt:
Hmm, tests here [look really sad](https://cockpit-logs.us-east-1.linodeobjects.com/pull-806-20220908-120336-0ccc5291-fedora-37/log.html#27-2).. related, or is our infra having a hissy fit again?

skobyda:
Seems like it was infra is healthy again, so far everything passes

PR #808, opened 2022-09-14
Fix undefined dialogs at console page

skobyda:
Without WithDialogs, trying to delete VM from console page would cause
a runtime error: "Uncaught TypeError: Dialogs is undefined".

Also, since I'm already re-writing test for console page, 
I added a test which checks console page with shut-off VM.


martinpitt (approved):
LGTM, thank you! Retrying flaked test.

@mvollmer, want to have a look as well?

skobyda:
All tests green

PR #811, opened 2022-09-26
Add connection information to the VM detail page

yunmingyang:


src/components/vm/overview/vmOverviewCard.jsx
```diff
@@ -180,6 +180,11 @@ class VmOverviewCard extends React.Component {
                             {_("General")}
                         </Text>
 
+                        <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Connection")}</DescriptionListTerm>
+                            <DescriptionListDescription id={`${idPrefix}-connection`}>{vm.connectionName}</DescriptionListDescription>
```
skobyda, line 185: Thanks fo this PR. This needs to be translated.
Also To show some libvirt values in the UI, we use [`rephraseUI` function in the from src/helpers.js](https://github.com/cockpit-project/cockpit-machines/blob/main/src/helpers.js#L281). What it does is that it will turn it into more human friendly strings, and also **translate them**. E.g. "session" becomes "User session" (plus translation into the user's language)
```suggestion
                            <DescriptionListDescription id={`${idPrefix}-connection`}>{rephraseUI('connections', vm.connectionName)}</DescriptionListDescription>
```

skobyda (changes requested):

skobyda (changes requested):
This also needs to be tested. I think the best place to test this is at [`_assertCorrectConfiguration()` at test/check-machines-create](https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-create#L1540).
The `connection` variable is already defined there, so all we have to do is check if it's in the text of `#vm-{name}-connection` element.

skobyda:
Also since this changes adds a new value into the design, maybe we need a nod from @garrett . It's true that there is no information about the connection IN the VM's detail page, and I think it's best at the Overview tab.
### Before this PR
Screenshot from 2022-09-29 12-29-38

### After this PR
Screenshot from 2022-09-29 12-29-44

src/components/vm/overview/vmOverviewCard.jsx
yunmingyang: Thanks, will re-push it.

yunmingyang:
Re-push, add translated and tests related

test/check-machines-create
```diff
@@ -1545,6 +1545,14 @@ vnc_password= "{vnc_passwd}"
...
                 self.test_obj.goToVmPage(name, dialog.connection or "system")
 
+            # Check connection information in the details page
+            if dialog.connection == "system":
+                wait(lambda: "System" == b.text(f"#vm-{dialog.name}-connection"))
+            elif dialog.connection == "session":
+                wait(lambda: "User session" == b.text(f"#vm-{dialog.name}-connection"))
+            else:
+                raise Exception("unsupported connection")
+
```
skobyda, line 1555: We have a function `wait_in_text()` which does the same as `wait(lambda: [string] == b.text(...)`. And with conditional expression resolution this can be a one liner:
```suggestion
            # Check connection information in the details page
            `wait_in_text(f"#vm-{dialog.name}-connection", "System" if dialog.connection == "system" else "User session")`
```
I don't think it's necessary to test for other types of connection, as libvirt only supports those 2, either you run the VM instance as privileged user, or as unprivileged user.

skobyda (changes requested):

test/check-machines-create
yunmingyang: OK

skobyda:
Thanks. I just fixed the formatting of the test code a bit and trigerred tests. This change in the UI will probably change a lot of sceenshots in pixel test (I can't tell from my head which tests will be affected), but we will soon see at [fedora-36](https://cockpit-logs.us-east-1.linodeobjects.com/pull-811-20220930-075147-c3f00238-fedora-36/log.html)

skobyda (approved):

skobyda:
Ahhhhhhhhh, I accidentally clicked on "Rebase and merge" instead of "Squash", so the commit with an ugly message "squash! fix formatting" didn't get squashed with a previous commit.

PR #819, opened 2022-09-29
Disable "Add" button when attachement is in progress

skobyda:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2129845


martinpitt (approved):
Looks great, thanks!

martinpitt:
@skobyda : When you post a PR, can you please update the bz status like I did in https://bugzilla.redhat.com/show_bug.cgi?id=2129845 ? I.e. set target release, devel_ack+, internal target release, internal target milestone, and status POST. Otherwise we'll have a really hard time with finding the correct bugs when doing the downstream releases. Thanks!

martinpitt (changes requested):
There are pixel test failures: due to adding the `isLoading` button property, they now became wider.

marusak (approved):
Thank you!

PR #824, opened 2022-10-07
Use Promise.reject() instead of throw()

skobyda:
Both of these approaches are almost identical, except for one situation: If exception happens in a promise resolving chain, throw() will not be caught by Promise.catch() handler, and will be left unresolved, causing 'Ooops'.
Promise.reject() is caight by Promise.catch() and can be therefore handled with our own exception handlers.


src/libvirt-xml-update.js
```diff
@@ -56,7 +56,7 @@ export function updateBootOrder(domXml, devices) {
     const doc = getDoc(domXml);
     const domainElem = doc.firstElementChild;
     if (!domainElem)
-        throw new Error("updateBootOrder: domXML has no domain element");
+        Promise.reject(new Error("updateBootOrder: domXML has no domain element"));
```
jelly, line 59: This and the function above isn't a promise so this looks wrong to me? So I think `domainChangeBootOrder` should instead handle a possible exception from `updateBootOrder`

skobyda: It is handled one layer above, at  [BootOrderModal](https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/overview/bootOrder.jsx#L255), which calls `domainChangeBootOrder`. This way, this exception bubbles all the way to the UI and is shown as an error in the dialog.

jelly: Yes, but it feels kinda weird to me, if you want to use `updateBootOrder` somewhere where there is no promise it simply won't work.

skobyda: From what I have read and tried locally, `throw` and `Promise.reject` have the same behaviour when outside of promise chain. Meaning in both cases, they will cause Uncaught exception.
The only difference is that if you re inside a promise chain, `Promise.reject` can be caught by `Promise.catch`, while `throw` can't.

jelly (approved):
Not sure what the deal is with rawhide, too many test failures so it exceeds the run timeout?

PR #825, opened 2022-10-10
test: Add a case about adding multiple serial consoles

yunmingyang:


test/check-machines-consoles
```diff
@@ -145,6 +145,19 @@ class TestMachinesConsoles(VirtualMachinesCase):
...
+            "virt-xml --remove-device vmWithSerialConsole --console all "
+            "&& virt-xml vmWithSerialConsole --add-device --console pty,target.type=serial"
+            "&& for i in {1..5};do virt-xml vmWithSerialConsole --add-device --console pty,target.type=virtio;done")
+        m.execute("virsh start vmWithSerialConsole")
+        for i in range(0, 6):
+            b.click("#pf-c-console__type-selector")
+            b.wait_visible("#pf-c-console__type-selector + .pf-c-select__menu")
+            b.click("li:contains('Serial console ({}{})') button".format("serial" if i == 0 else "console", i))
+            b.wait(lambda: m.execute("ps aux | grep 'virsh -c qemu:///system console vmWithSerialConsole {}{}'".format("serial" if i == 0 else "console", i)))
+
```
skobyda, line 160: We try to use [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) now, so do you mind changing it to something like this
```suggestion
        for i in range(0, 6):
            console_type = "serial" if i == 0 else "console"
            b.click("#pf-c-console__type-selector")
            b.wait_visible("#pf-c-console__type-selector + .pf-c-select__menu")
            b.click(f"li:contains('Serial console ({console_type}{i})') button")
            b.wait(lambda: m.execute(f"ps aux | grep 'virsh -c qemu:///system console vmWithSerialConsole {console_type}{i}'"))

```

skobyda (changes requested):
Thanks for new tests

test/check-machines-consoles
yunmingyang: Also separate the command group of virt-xml, for easy debugging in the future. And it seems that expressions could be used in f-strings, so no new variables there.

skobyda (approved):
LGTM, thanks

PR #828, opened 2022-10-12
Update Github workflow actions

jelly:
See cockpit-project/cockpit#17811 for more details and test results.

I wonder if like podman this doesn't have reposchutz mandatory.


marusak (approved):
Thank you!

PR #839, opened 2022-10-27
Allow unmap/trim requests by default for newly added disks

skobyda:
Allow unmap/trim requests by default for newly added disks
    
    Libvirt and QEMU ignore unmap/trim requests by default, unless specified
    otherwise and stated they would not change this in order to not break
    back-compatibility. [1][2]
    Most distros such as Ubuntu and Fedora already do periodic trim by
    default. Virt-install also uses unmap/trimp for default disk of newly
    created VMs. [3]
    There is also discussion if we should expose the discard option in the
    UI, since there are corner cases when user might want the unmap/trim
    requests to be ignored.
    But either way, allowing unmap/trim requests is the correct thing to do,
    regardless whether it will get exposed in the UI or not.
    
    [1] https://listman.redhat.com/archives/libvirt-users/2018-June/msg00007.html
    [2] https://listman.redhat.com/archives/libvir-list/2020-August/msg00341.html
    [3] https://github.com/virt-manager/virt-manager/commit/381aa4050c1ca4d3320fdabb579a128b5855d022


src/libvirt-xml-create.js
```diff
@@ -10,6 +10,7 @@ export function getDiskXML(type, file, device, poolName, volumeName, format, tar
     if (format && ['qcow2', 'raw'].includes(format))
         driverElem.setAttribute('type', format);
     driverElem.setAttribute('cache', cacheMode);
+    driverElem.setAttribute('discard', 'unmap');
```
jelly, line 13: The virt-manager commit only enables unmap on two conditions, we seems to enable it unconditionally? (It sounds logical to me to run discard for block storage but not for /var/lib/images/foo.qcow2. Although virt-manager does that under some circumstances https://github.com/virt-manager/virt-manager/commit/381aa4050c1ca4d3320fdabb579a128b5855d022#diff-4616797a3c857c365c4005588f94c019f5eb809b198225b526f483ef664c3f11R1003

skobyda: Why not also for qcow2? Qcow2 is a thin-provisioned (sparse) storage device, so running `fstrim` from inside the VM does make an effect and sense so you will save up some actual size. (You can check virtual vs actual size of a disk with `qemu-img info`).

So from that virt-manage code, that code, I can see that they set unmap for all disks they create themself and for block devices.
So basically they do not set disk to unmap only for non-block disks which were supplied to them by the user, since user can have that disk setup in some advanced way (read as corner case) where they might not want disk trimming to be allowed by default.

**However**, as mentioned in the call, would users who have these kind of advanced, corner cases, use cockpit?
**Futhermore**, they still allow user to specify unmap manually. In cockpit, user is not allowed to do so. If we set unmap only for some cases, we should offer user to manually set it for the other cases, and we want to avoid since again, users would do that only in some corner cases.

jelly: Ok, thanks for the explanation!

jelly (approved):

jelly:
This should be a release-note right?

PR #842, opened 2022-11-03
Bump cockpit lib to dark theme support

jelly:
Include the dark theme patternfly improvements for a consistent experience.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1366,6 +1366,7 @@ vnc_password= "{vnc_passwd}"
             root = ET.fromstring(rhel_8_1_xml)
             root.find('os').find('resources').find('minimum').find('ram').text = '134217728'
             root.find('os').find('resources').find('minimum').find('storage').text = '134217728'
+            root.find('os').find('release-date').text = datetime.today().strftime('%Y-%m-%d')
             new_rhel_8_1_xml = ET.tostring(root)
             self.machine.execute(f"echo '{str(new_rhel_8_1_xml, 'utf-8')}' > {self.test_obj.vm_tmpdir}/rhel-8.1.xml")
             self.machine.execute(f"mount -o bind  {self.test_obj.vm_tmpdir}/rhel-8.1.xml /usr/share/osinfo/os/redhat.com/rhel-8.1.xml")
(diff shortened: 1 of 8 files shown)


jelly:
image

marusak:
We need to figure those check icons, that is definitely a regression.

skobyda:
Popover has a weird border:

skobyda (dismissed):
Looks like both of my above mentioned issues are cockpit-wide, not only machines specific, so I'm fine with merging this PR as long as we will eventually add some overrides for that into cockpits patternfly overrides

test/check-machines-create
```diff
@@ -1366,6 +1366,7 @@ vnc_password= "{vnc_passwd}"
             root = ET.fromstring(rhel_8_1_xml)
             root.find('os').find('resources').find('minimum').find('ram').text = '134217728'
             root.find('os').find('resources').find('minimum').find('storage').text = '134217728'
+            root.find('os').find('release-date').text = datetime.today().strftime('%Y-%m-%d')
```
skobyda, line 1369: I think same fix should be added for thel RHEL_8_2, which is being faked just few lines below. Otherwise we will run into the same issue in few months.

skobyda (changes requested):

jelly:
TestMachinesSettings.testDomainMemorySettings flakes on Ubuntu/Debian, needs a separate investigation

martinpitt:
Fallout from new test API perhaps? I checked a couple of recent PRs, and this test wasn't flaky at all.

jelly:
> Fallout from new test API perhaps? I checked a couple of recent PRs, and this test wasn't flaky at all.

Sounds weird, the issue is clearly the VM using more ram then what this memory settings thing allows:

```
root@ubuntu:~# virsh dominfo subVmTest1
Id:             1
Name:           subVmTest1
UUID:           ac696d88-c91d-4561-ba23-07ae53526105
OS Type:        hvm
State:          running
CPU(s):         1
CPU time:       12.9s
Max memory:     262144 KiB
Used memory:    262144 KiB
```

jelly:
For testing, changed memory to 512 and now I get:

image

So something is off with the memory calculation?

martinpitt:
This actually makes perfect sense. debian-stable and ubuntu 22.04 OS images have cockpit-bridge < 266. The tests do
```
       if self.system_before(266):
            b.wait_in_text("#vm-subVmTest1-memory-count", f"{current_memory - 10} MiB")
        else:
            b.wait_in_text("#vm-subVmTest1-memory-count", f"{mib_to_mb(current_memory - 10)} MB")
```
thus the tests expect base-2 units still, as the c-bridge *package* is old. But bundling cockpit.js means the "base 10" number formatting API change now applies.

So TL/DR: Drop all the ` if self.system_before(266):` special-cases in the "bundle cockpit" commit.

marusak (approved):
Nice, thank you!
The last commit could be squashed with the bundle one, but that is not important :)

test/check-machines-create
jelly: Done!

PR #843, opened 2022-11-03
fmf: Run storage and network tests in separate test plans

martinpitt:
On the Testing Farm we can run up to 5 parallel test plans. So split the tests into three plans "basic", "network", and "storage".

This mitigates the utterly slow Fedora Rawhide tests due to a kernel bug [1], and also should generally speed up results.

We previously skipped all TestMachinesConsoles tests, so drop that whole block and just not run TestConsoles in any plan. Once that is reconsidered, let's re-evaluate why the tests fail.

Similarly, TestMachinesMigration cannot be run in TMT, so drop the "skip all tests" block and just don't run it in any plan.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2139658

----

This should fix the [rawhide timeouts](https://artifacts.dev.testing-farm.io/88a56cd5-c943-4d7c-b34e-dfecd9797b49/), and also speed up results on the sane Fedora releases. Same approach as in https://github.com/cockpit-project/cockpit/pull/17873

### DIFF
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -1,6 +1,0 @@
-summary:
-    Run all tests
-discover:
-    how: fmf
-execute:
-    how: tmt
--- a/plans/basic.fmf
+++ b/plans/basic.fmf
@@ -1,0 +1,7 @@
+summary:
+    Run basic tests (creation and lifetime)
+discover:
+    how: fmf
+    test: /test/browser/basic
+execute:
+    how: tmt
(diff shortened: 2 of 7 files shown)


martinpitt:
Comparison: the single-plan run like [this](https://artifacts.dev.testing-farm.io/712ca462-0660-4155-be89-62dfaaec1258/) took 46 mins overall. This [f37 run](https://artifacts.dev.testing-farm.io/12a380ba-9cea-4375-ad0c-863d4a56a642/) took 31 mins. So it pays off even on non-broken kernels.

marusak (approved):
Nice! Thank you \o/

PR #844, opened 2022-11-04
enable eslint key-spacing rule

jelly:


skobyda (approved):
Nice cleaning

jelly:
Packit test fails, this should be fixed in main. I'm however not going to rebase as the other distros succeeded and nothing failed so far in rawhide.

PR #846, opened 2022-11-07
fmf: Factorize and merge plan files

martinpitt:
FMF plans can inherit common properties [1]. Go back to a single plans/all.fmf (like it was until commit cf0fec7e71), put the common properties at the top, and enumerate the three plans with their specific properties.

This will make it easier to keep the Fedora downstream dist-git in sync.

[1] https://tmt.readthedocs.io/en/latest/examples.html#inherit-plans

### DIFF
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -1,0 +1,19 @@
+discover:
+    how: fmf
+execute:
+    how: tmt
+
+/basic:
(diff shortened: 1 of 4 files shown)


jelly:
createDownloadRhel is a known issue, fixed in the dark theme PR

jelly (approved):

PR #852, opened 2022-11-09
Check offline token only when the source type is 'Download an OS'

yunmingyang:
For now, all source type will check offline token when OS is RHEL, that will cause 'Create and run'/'Create and edit' button disabled unexpectedly, add a filter to check offline token only when the source type is 'Download an OS'


skobyda (approved):
LGTM, thanks

skobyda:
On second thought, I think this bug should be tested, so I added a test for that

skobyda (approved):
All tests green

PR #853, opened 2022-11-09
Enable object property newline rule

jelly:
this is more readable and automatically applied using `eslint --fix`.


skobyda (approved):
Thanks

PR #856, opened 2022-11-14
Limit disk source column to 25% of the whole 'Disks' table width

### LINKED ISSUE #813: Headers are truncated on the Disks section
Headers are oddly truncated on Disks:

image

There's an additional section that's empty too. It should probably also not be shown unless there's something to show.

KKoukiou:
The disk source text is sometimes expressed as file path, making that column take too much space compared to the rest.

Fixes #813

### DIFF
--- a/src/components/vm/disks/vmDisksCard.jsx
+++ b/src/components/vm/disks/vmDisksCard.jsx
@@ -218,7 +218,7 @@ export const VmDisksCard = ({ vm, vms, disks, renderCapacity, supportedDiskBusTy
             columns.push({ title: access });
         }
 
-        columns.push({ title: <DiskSourceCell diskSource={disk.source} idPrefix={idPrefixRow} /> });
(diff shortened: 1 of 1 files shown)


garrett (approved):
I've found so many issues reviewing this PR, but not related to this PR. :laughing: 

- Breakpoint happens way too quickly (PR @ https://github.com/cockpit-project/cockpit-machines/pull/864)
- All the delete and remove buttons in the tables are primary and should be secondary (it's super noticeable in dark mode)
- Dark mode has contrast issues in varios places (the disabled text — not sure if this is our issue or PF... it's probably PF, but we'll have to have a local fix regardless)

But, the important thing is that this PR does the job, and works in a smaller breakpoint too. (It's grid-xl now, but should be grid-md (which has problems within a couple pixels of the breakpoint) or at least grid-lg, which is still lots better than xl. Anyway, breakpoints would be a follow-up. :wink:)

I'll file some issues and/or send some PRs for the list above.

For this PR: :sparkles:  :+1: :sparkles:

skobyda (approved):
LGTM

Issue #816, opened 2022-09-26
Unable to remove detached virtual disk files


shodanshok:
Detaching a vdisk from a virtual machine does not remove the physical file. While this is the right thing to do, no other method exists to remove the vdisk file other than opening a shell and execute the corresponding `rm` or `virsh` command.

`virt-manager` copes with the same issue by giving the user a view of the files stored on pools, from where the user can delete the detached files.

`cockpit-machines` lacks such pool-wide file selection and removal capability, so user is forced to open a shell to remove the just-detached file. A quick solution would be to ask the user if he want to remove the vdev file upon detaching it, similarly on what happens when an user delete an entire virtual machine.

See also https://bugzilla.redhat.com/show_bug.cgi?id=2129524

PR #857, opened 2022-11-14
Offer an option to delete storage when detaching disk

skobyda:
Fixes https://github.com/cockpit-project/cockpit-machines/issues/816

In this draft, deleting disk file is checked by default (but that's up for discussion)
Screenshot from 2022-11-14 18-51-27
Screenshot from 2022-11-14 18-51-22

### If storage file fails to get deleted, the disk will still get detached from the VM, but we show a warning:
Screenshot from 2022-11-14 18-48-54

### DIFF
--- a/src/components/common/deleteResource.jsx
+++ b/src/components/common/deleteResource.jsx
@@ -46,11 +46,10 @@ export class DeleteResourceModal extends React.Component {
         this.dialogErrorSet = this.dialogErrorSet.bind(this);
     }
 
-    delete(primaryAction = true) {
+    delete(deleteHandler) {
         const Dialogs = this.context;
         this.setState({ inProgress: true });
-        const handler = primaryAction ? this.props.deleteHandler : this.props.deleteHandlerSecondary;
-        handler()
+        deleteHandler()
                 .then(Dialogs.close, exc => {
                     this.setState({ inProgress: false });
                     this.dialogErrorSet(this.props.errorMessage, exc.message);
@@ -72,11 +71,20 @@ export class DeleteResourceModal extends React.Component {
                    titleIconVariant="warning"
                    footer={
                        <>
-                           <Button variant='danger' id="delete-resource-modal-primary" isLoading={this.state.inProgress} isDisabled={this.state.inProgress} onClick={this.delete}>
+                           <Button variant='danger'
+                               id="delete-resource-modal-primary"
+                               isLoading={this.state.inProgress}
+                               isDisabled={this.state.inProgress}
+                               onClick={() => this.delete(this.props.deleteHandler)}>
                                {actionName || _("Delete")}
                            </Button>
                            {actionNameSecondary &&
-                           <Button variant='secondary' id="delete-resource-modal-secondary" isLoading={this.state.inProgress} isDisabled={this.state.inProgress} onClick={() => this.delete(false)} isDanger>
+                           <Button variant='secondary'
+                               id="delete-resource-modal-secondary"
+                               isLoading={this.state.inProgress}
+                               isDisabled={this.state.inProgress}
+                               onClick={() => this.delete(this.props.deleteHandlerSecondary)}
+                               isDanger>
                                {actionNameSecondary}
                            </Button>}
                            <Button variant='link' className='btn-cancel' onClick={Dialogs.close}>
--- a/src/components/common/deleteResource.jsx
+++ b/src/components/common/deleteResource.jsx
@@ -46,10 +46,11 @@ export class DeleteResourceModal extends React.Component {
         this.dialogErrorSet = this.dialogErrorSet.bind(this);
     }
 
-    delete() {
+    delete(primaryAction = true) {
         const Dialogs = this.context;
         this.setState({ inProgress: true });
-        this.props.deleteHandler()
+        const handler = primaryAction ? this.props.deleteHandler : this.props.deleteHandlerSecondary;
+        handler()
                 .then(Dialogs.close, exc => {
                     this.setState({ inProgress: false });
                     this.dialogErrorSet(this.props.errorMessage, exc.message);
@@ -62,7 +63,7 @@ export class DeleteResourceModal extends React.Component {
 
     render() {
         const Dialogs = this.context;
-        const { title, objectDescription, actionName, actionDescription } = this.props;
+        const { title, objectDescription, actionName, actionDescription, actionNameSecondary } = this.props;
 
         return (
             <Modal position="top" variant="small" isOpen onClose={Dialogs.close}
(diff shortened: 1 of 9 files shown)


garrett:
I thought we had support for this already?

(Or are we just doing something similar elsewhere? Perhaps I'm thinking of optionally deleting the disk when removing a VM? If so, we should make sure the UI is consistent.)

I'm not a fan of phrases starting with "Also" in a UI (it sounds really odd), but I'll have to think of something better.

As far as deleting by default: It does make sense as it's no longer used, but it's probably too easy to delete it accidentally, so we probably want to do something about that. Ideas (probably one or more of the following):
- Make it opt-in instead of opt out
- Change the button's text based on the checkbox
- Drop the checkbox and make it a dual-action dialog... in this case, probably "Remove" (default action), "Remove and delete file" (secondary), and "Cancel" (of course)

I think the third option probably makes the most sense? And then we should make sure other places that have a remove & delete are similar.

skobyda:
Updated, new look:
Screenshot from 2022-11-15 17-56-40

Additionally, when disks ave no deletable file (e.g. file doesn't exist, or the disk is a block device), we offer no option to delete file:
Screenshot from 2022-11-15 17-58-48

skobyda:
@garrett Mind giving it another review?

shodanshok:
@skobyda first, thanks for fixing the reported issue.

I would say that I like your first UI design more, but with a checkbox that the user must explicitly mark (ie: unselected by default). The reason is that with such a design, a careless user will do no permanent damage (the disk is removed but not deleted by default), while the second proposed UI (or the first one with the checkbox selected by default) is prone to accidental deletion by simply pressing the wrong button.

skobyda:
Hmmm, tests began to fail, I'll investigate

garrett (approved):
Looks good.

---

The default, secondary, and cancel pattern is a pattern we use elsewhere in Cockpit.

This is already a warning confirmation prompt, which is already a step where someone must confirm that they want to remove a disk. And this is not a default action with a button shown by default; someone has to click the "..." button, then click "Remove", and then the modal shows up.

Additionally, there is already some additional "friction" in the UI of this warning modal dialog, as a person must move the mouse cursor to a non-default location (that is: it's not over on the extreme left or right of the dialog, and it's not as attention-getting as the default item).

KKoukiou (commented):
@skobyda this will just work for file based disks.  

domainDeleteStorage in fact has a switch case that ignores alls disks that are not of type 'file' or 'volume'.

Should be maybe not offer this options when such disks are present.?

skobyda:
> @skobyda this will just work for file based disks.
> 
> domainDeleteStorage in fact has a switch case that ignores alls disks that are not of type 'file' or 'volume'.
> 
> Should be maybe not offer this options when such disks are present.?

Yeah that's exactly what it does here: https://github.com/cockpit-project/cockpit-machines/pull/857/files#diff-49ba0c85360693b8e89ae0e0428284a55db749f08f2b9993d26910def646e02f
and here:  https://github.com/cockpit-project/cockpit-machines/pull/857/files#diff-540e826f3f2e25ce426ab9c870f09cbd8fea0a604d4bbc6ff309f5611082fe32R164

martinpitt:
@skobyda: Wrt. [downstream bug planning](https://bugzilla.redhat.com/show_bug.cgi?id=2129524), do you have a gut feeling how much work remains here? Where on the "no way ←→ maybe ←→ sure!" scale is it to land in the next 3 weeks before the freeze? Thanks!

skobyda:
So this PR fails with flake
`wait_js_cond(ph_is_present("#delete-vm-subVmTest1-disks-sda")): Uncaught (in promise) Error: condition did not become true`
That is however a flake which is present in other PRs too, and is not related to this PR. It's just that this PR affects the test where the flake happens, so it's much harder to pass since affected tests need to pass 3-times.
So I think this PR should not be blocked by that, since again, it's pre-existing flake which was not created by this PR.

skobyda:
Huh, this now has a bunch of seemingly new race conditions related to disk detachement

src/components/common/deleteResource.jsx
```diff
@@ -71,9 +72,13 @@ export class DeleteResourceModal extends React.Component {
                    titleIconVariant="warning"
                    footer={
                        <>
-                           <Button variant='danger' isLoading={this.state.inProgress} isDisabled={this.state.inProgress} onClick={this.delete}>
+                           <Button variant='danger' id="delete-resource-modal-primary" isLoading={this.state.inProgress} isDisabled={this.state.inProgress} onClick={this.delete}>
                                {actionName || _("Delete")}
                            </Button>
+                           {actionNameSecondary &&
+                           <Button variant='secondary' id="delete-resource-modal-secondary" isLoading={this.state.inProgress} isDisabled={this.state.inProgress} onClick={() => this.delete(false)} isDanger>
```
martinpitt, line 79: I find this hard to read. Boolean arguments aren't nice. How about passing it the actual handler instead? I think `this.delete(this.props.deleteHandlerSecondary)` would be much cleaner?

test/check-machines-disks
```diff
@@ -1501,6 +1501,7 @@ class TestMachinesDisks(VirtualMachinesCase):
         b.click("#delete-vm-subVmTest1-disks-vdc")
         b.wait_visible(".pf-c-modal-box")
         b.click(".pf-c-modal-box__footer button:contains(Remove)")
+        wait(lambda: "vdc" not in m.execute("virsh domblklist subVmTest1"), delay=1)
```
martinpitt, line 1556: How was that a race condition previously? It's already a `wait_*`. Could it be that it simply takes longer than 15s sometimes, and thus it needs a wait_timeout()?

That said, this extra check is actually nice -- it just has the wrong justification.

test/check-machines-disks
```diff
@@ -1408,23 +1408,24 @@ class TestMachinesDisks(VirtualMachinesCase):
...
+        b.click("#delete-resource-modal-primary")
+        b.wait_not_present(".pf-c-modal-box")
         # When live-detaching disks the guest OS needs to cooperate so that we can
         # see the disk getting detached in the UI.
         # Wait until we see the login prompt before attempting operation that need
         # the OS for fully respond
         with b.wait_timeout(180):
             b.wait_not_present("#vm-subVmTest1-disks-vdc-device")
-        b.wait_not_present(".pf-c-modal-box")
+            wait(lambda: "vdc" not in m.execute("virsh domblklist subVmTest1"), delay=1)
```
martinpitt, line 1428: This is confusing. wait() is not a Browser class function, and thus not affected by `b.wait_timeout()`. Please move it outside of the context.

Also: Normally, it's the *delete* operation (i.e. clicking the button in the modal) which may be slow, not updating the UI after the dialog closed. So are you sure that the `b.wait_not_present(".pf-c-modal-box")` in line 1421 is fine with the default timeout?

The comment tries to justify the long timeout for updating the UI. But it doesn't match any more, as nothing actually waits for the login promt any more here. Can you please clean this up?

martinpitt (changes requested):
Thanks! By and large LGTM, just needs some cleanup.

test/check-machines-disks
skobyda: Yeah, since this is a hot unplug, unplugging device from running VM might take longer than 15s since it needs the cooperation of the underlying VM OS.

And libvirt has a timeout on its hotunplug API of 5 seconds, and if it will not get a response from OS about hotunplug in 5seconds, it will just finish the API call, which results in the dialog being closed.

And thus `b.wait_not_present(".pf-c-modal-box")` alone is not enough

test/check-machines-disks
martinpitt: Thanks for the update. I see you completely dropped the wait_timeout(), if that works, that's fine.

martinpitt (approved):
Assuming green tests, I'm happy with this, thanks!

Garrett previously approved this -- not sure if you changed any designs since then. If you did, please re-ask for his review, otherwise good to land.

But please add a light theme screenshot and a release note. Cheers!

skobyda:
All tests green, apart from pybridge flake

PR #860, opened 2022-11-15
Implement watchdog device configuration

skobyda:
### Things to do after this is merged:

 * [ ] Create a libvirt issue requiring all watchdog actions to have relevant VM events (it seems that currently only watchdog reset and watchdog force-off have events)
 * [ ] Create a libvirt issue asking for API to check VM's history of watchdog events (this would allow us to see if VM is stuck in a reboot loop)
 * [ ] Create an issue to implement a warning if VM is crashed
 * [ ] Redesign VM's Overview tab since it's getting crowded

### DIFF
--- a/src/components/vm/overview/watchdog.jsx
+++ b/src/components/vm/overview/watchdog.jsx
@@ -1,0 +1,158 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import cockpit from 'cockpit';
+import {
+    Button,
+    Flex, FlexItem,
+    Form, FormGroup,
+    Modal,
+    Radio,
+} from '@patternfly/react-core';
+import { useDialogs } from 'dialogs.jsx';
+
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { domainRemoveWatchdog, domainSetWatchdog } from "../../../libvirtApi/domain.js";
+import { rephraseUI } from "../../../helpers.js";
+import { WarningInactiveAlert, WarningInactiveTooltip } from "../../common/warningInactive.jsx";
+
+const _ = cockpit.gettext;
+
+export const WatchdogModal = ({ vm, isWatchdogAttached, idPrefix }) => {
+    const [dialogError, setDialogError] = useState();
+    const [watchdogAction, setWatchdogAction] = useState(vm.watchdog.action || "no-device");
+    const [inProgress, setInProgress] = useState(false);
+
+    const Dialogs = useDialogs();
+
+    const save = () => {
+        setInProgress(true);
+        return domainSetWatchdog({
+            connectionName: vm.connectionName,
+            vmName: vm.name,
+            hotplug: vm.state === "running",
+            permanent: vm.persistent,
+            action: watchdogAction,
+            isWatchdogAttached,
+        })
+                .then(Dialogs.close, exc => setDialogError({ text: _("Failed to configure watchdog"), detail: exc.message }))
+                .finally(() => setInProgress(false));
+    };
+
+    const detach = () => {
+        setInProgress(true);
+        return domainRemoveWatchdog({
+            connectionName: vm.connectionName,
+            vmName: vm.name,
+            hotplug: vm.state === "running",
+            permanent: vm.persistent,
+            model: vm.watchdog.model,
+        })
+                .then(Dialogs.close, exc => setDialogError({ text: _("Failed to detach watchdog"), detail: exc.message }))
+                .finally(() => setInProgress(false));
+    };
+
+    const supportedActions = ["reset", "poweroff", "inject-nmi", "pause"];
+
+    const showWarning = () => {
+        if (vm.state === 'running' && isWatchdogAttached && vm.watchdog.action !== watchdogAction)
+            return <WarningInactiveAlert idPrefix={idPrefix} />;
+    };
+
+    return (
+        <Modal id={`${idPrefix}-watchdog-modal`}
+               position="top"
+               variant="small"
+               onClose={Dialogs.close}
+               title={isWatchdogAttached ? _("Edit watchdog device type") : _("Add watchdog device type")}
+               description={_("Watchdogs act when systems stop responding. To use this virtual watchdog device, the guest system also needs to have an additional driver and a running watchdog service.")}
+               isOpen
+               footer={
+                   <>
+                       <Button variant='primary'
+                               id="watchdog-dialog-apply"
+                               onClick={save}
+                               isLoading={inProgress}
+                               isDisabled={inProgress}>
+                           {isWatchdogAttached ? _("Save") : _("Add")}
+                       </Button>
+                       {isWatchdogAttached &&
+                       <Button variant='secondary'
+                               id="watchdog-dialog-detach"
+                               onClick={detach}
+                               isLoading={inProgress}
+                               isDisabled={inProgress}>
+                           {_("Remove")}
+                       </Button>}
+                       <Button variant='link' onClick={Dialogs.close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }>
+            {showWarning()}
+            {dialogError && <ModalError dialogError={dialogError.text} dialogErrorDetail={dialogError.detail} />}
+            <Form onSubmit={e => e.preventDefault()} isHorizontal>
+                <FormGroup role="radiogroup"
+                           label={_("Action")}
+                           fieldId="watchdog-action"
+                           hasNoPaddingTop
+                           isStack>
+                    {supportedActions.map(action => <Radio label={rephraseUI("watchdogAction", action)}
+                                                           key={action}
+                                                           id={action}
+                                                           isChecked={watchdogAction === action}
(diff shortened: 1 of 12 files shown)


skobyda:
I created this PR so we can move some discussion about watchdog and it's design here.
This is my first naive take at how the design could look.
These are the things I would like to discuss:
- Perhaps instead of "Watchdog" we should say "Watchdog action", since that's the only thing user configures
- After user attached watchdog, they still need to setup a watchdog daemon and install a watchdog driver inside the guest. This can be a message at the bottom of the modal? What should the message say?
- 

### Watchdog device is not attached:
Screenshot from 2022-11-15 18-02-43

### User tries to attach watchdog device:
Screenshot from 2022-11-15 18-19-54


### Watchdog device is attached and has some action set:
Screenshot from 2022-11-15 18-02-58

### User can change action of existing watchdog device or detach a watchdog:
Screenshot from 2022-11-15 18-20-01

garrett:
This looks generally good!

- Top screenshot should be: Watchdog none edit
- You shouldn't "Attach" a watchdog, you should "Add" it
- Labels and their values must be properly aligned to the baseline of the text; the modal is not properly aligned
- "Detatch" should be "Remove"
- We might want some text on the modal explaining what a watchdog is? Perhaps: "A watchdog device can reboot a system when it stops responding. The guest system needs to have a watchdog service enabled and configured to use the virtual watchdog device."
- We need to have text explaining that this is the device and the system needs to have watchdog support enabled and active. And then we should tell them how to do that (in a very basic way). I suggested something in the previous point that adds this and the previous point together. What do you think?

skobyda:
So I updated the PR, can you re-review @garrett ?
Also maybe I like the hint, it generally says everything needed. I don't think we can be any more specific about how to setup a watchdog service, since that can vary a lot depending on a linux distro (or if user has windows VM, setting up a watchdog might be completely different process), so I'm fine with just telling them that they need to configure it.

Maybe only a little thing. It's a little bit weird that we offer a variety of watchdog actions, apart from reboot, but then in the tooltip we say "watchdog device can **reboot**". So maybe instead of a `A watchdog device can reboot a system` we could say `A watchdog device can reboot a system`?

Screenshot from 2022-12-08 14-36-17
Screenshot from 2022-12-08 14-18-57
Screenshot from 2022-12-08 14-18-36

garrett (dismissed):
> A watchdog device can recover a system

Recover is an overloaded term.
- Things aren't going well and you're rebooting to a snapshot? Recover.
- Big crash of an app and you can restart a process? Recover. 
- Hard drive crash? Recover.
- Accidentally wiped data from your drive and you need to delve into backups? Recover.
- Have a blanket on your computer sometimes, but it's not on the computer now, but you need to keep the computer warm through cold nights? :cold_face:  Recover. (This one's a joke. :wink:)

---
In this image, does the text change based on the option? It should be indented if so.

If it is talking about everything, then it needs to be at the top.

The order of UI is (in most languages) is left to right, top to bottom. Things that come earlier describe (or modify) things that come later.

(It's a basic z-flow, part of the visual hierarchy of design. [Adobe has a bunch of visual hierarchy principles listed with some examples on a blog post](https://xd.adobe.com/ideas/process/information-architecture/visual-hierarchy-principles-examples/), if you'd like to know more. These rules basically stem from [Gestalt psychology](https://en.wikipedia.org/wiki/Gestalt_psychology) in Germany and Austria from the early 1900s and were put to use in the [Bauhaus](https://en.wikipedia.org/wiki/Bauhaus) movement starting in 1919, originating at the art school of the same name. They're part of all design and many educational fields worldwide these days. Before this, these kinds of hierarchy were something humans did subconsciously; the movements codified these principles, however.)

image

For PatternFly, you want to have the description be at the top, above the content, and use a description block as the example @ https://www.patternfly.org/v4/components/modal#with-description shows:

```jsx
      <Modal
        title="Modal with description"
        isOpen={isModalOpen}
        onClose={handleModalToggle}
        description="A description is used when you want to provide more info about the modal than the title is able to describe. The content in the description is static and will not scroll with the rest of the modal body."
        actions={[
          <Button key="confirm" variant="primary" onClick={handleModalToggle}>
            Confirm
          </Button>,
          <Button key="cancel" variant="link" onClick={handleModalToggle}>
            Cancel
          </Button>
        ]}
      >
        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore
        magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
        consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
        pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est
        laborum.
      </Modal>
```

However, descriptions should be very short, as they are not part of the content and therefore they do not scroll.

We might want to change the title of the modal to be a "Watchdog device type" instead of "Watchdog configuration".

What about this for the description?

> Watchdogs act when systems stop responding. To use this virtual watchdog device, the guest system also needs to have a running watchdog service.

skobyda:
> In this image, does the text change based on the option? It should be indented if so.
> 
> If it is talking about everything, then it needs to be at the top.
> 
> The order of UI is (in most languages) is left to right, top to bottom. Things that come earlier describe (or modify) things that come later.

Right, so if the text is dynamic and changes based on what watchdog action user chooses, the text should be at the bottom of the modal. Otherwise the radio buttons would change the text which is above it, which is wrong.
But the text also contain general information about what a watchdog device is. And user probably needs to read that first, so then they can make an informed decision about chooding watchdog action. So considering that, the text should be also at top of the modal, right?

So with contradictions , I think it's better to have the static text, which doesn't change, and which describes the situation universally for all watchdog actions. So I like your proposed text `Watchdogs act when systems stop responding. To use this virtual watchdog device, the guest system also needs to have a running watchdog service.`

skobyda:
So I updated the PR, here is how it looks now:
Screenshot from 2022-12-15 15-57-42

garrett:
That's better. We could also split the text up and have the first sentence as a description and the second as helper text as an info (with an icon) or even possibly warning (although there's nothing wrong, so info might be better?) at the bottom, as you suggest. 

This would make sense too, and it's possibly even better, as people aren't bombarded and they know that they have to do something after installing this.

But I'm happy with this now as-is too.

src/components/vm/overview/watchdog.jsx
```diff
@@ -0,0 +1,138 @@
...
+        if (vm.state === 'running' && isWatchdogAttached && vm.watchdog.action !== watchdogAction)
+            return <WarningInactiveAlert idPrefix={idPrefix} />;
+    };
+
+    return (
+        <Modal id={`${idPrefix}-watchdog-modal`}
+               position="top"
+               variant="small"
+               onClose={Dialogs.close}
+               title={_("Watchdog device type")}
```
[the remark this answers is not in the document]
KKoukiou: How about: To use tihs virtual watchdog device an additional driver and management daemon in required the guest virtual machine.

src/helpers.js
```diff
@@ -276,6 +276,15 @@ const transform = {
...
         down: _("down"),
     },
+    watchdogAction: {
+        reset: _("Reset"),
+        shutdown: _("Gracefully shutdown"),
+        poweroff: _("Power off"),
+        pause: _("Pause"),
+        none: _("Do nothing"),
+        dump: _("Dump"),
+        "inject-nmi": _("Non-maskable interupt"),
```
KKoukiou, line 286: How about use the libivrt docs wording? 

https://libvirt.org/formatdomain.html#watchdog-device

'Inject a non-maskable interrupt into the guest'

src/components/vm/overview/watchdog.jsx
KKoukiou, line 75: Let's configure the title accordingly. When adding watchdig maybe "Add a watchdog device"or when editing/removing "Edit watchdog device"

src/components/vm/overview/watchdog.jsx
```diff
@@ -0,0 +1,138 @@
...
+               description={_("Watchdogs act when systems stop responding. To use this virtual watchdog device, the guest system also needs to have a running watchdog service.")}
+               isOpen
+               footer={
+                   <>
+                       {dialogError && <ModalError dialogError={dialogError.text} dialogErrorDetail={dialogError.detail} />}
+                       <Button variant='primary' id="watchdog-dialog-apply" onClick={save}>
```
KKoukiou, line 81: Can you please disable and add an isLoading while the operation is in progress? on my system when the VM is running it takes ~2 seconds for the dialog to close.

src/components/vm/overview/watchdog.jsx
```diff
@@ -0,0 +1,138 @@
...
+                       {isWatchdogAttached && <Button variant='secondary' id="watchdog-dialog-detach" onClick={detach}>
+                           {_("Remove")}
+                       </Button>}
+                       <Button variant='link' onClick={Dialogs.close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }>
+            {showWarning()}
+            <Form isHorizontal>
```
KKoukiou, line 93: Add also onSubmit={e => e.preventDefault()}  to avoid page reloading when hitting enter.

src/components/vm/overview/watchdog.jsx
```diff
@@ -0,0 +1,138 @@
...
+                <FormGroup role="radiogroup"
+                           label={_("Action")}
+                           fieldId="watchdog-action"
+                           hasNoPaddingTop
+                           isStack>
+                    {supportedActions.map(action => <Radio label={rephraseUI("watchdogAction", action)}
```
KKoukiou, line 99: You probably are missing a key here, I guess there are warnings in console.

KKoukiou (commented):
Just some Patternfly usage related comments! Great job!

src/helpers.js
skobyda: Seems a bit too long. and looks too long in the UI compared to the others:
Screenshot from 2023-01-09 23-58-09
How about just saying
"Inject a non-maskable interrupt"? I guess that the interrupt happens on the VMs level can be safely assumed.

src/components/vm/overview/watchdog.jsx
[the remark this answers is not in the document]
skobyda: > To use tihs virtual watchdog device an additional driver and management daemon in required the guest virtual machine.

The word order looks seems a bit harder to read, at least to me, no? How about:
`To use tihs virtual watchdog device an additional driver and management daemon in required the guest virtual machine.`

src/components/vm/overview/watchdog.jsx
skobyda: Wait, what? the key is just below that:
`key={action}`

KKoukiou (approved):

PR #864, opened 2022-11-17
Reduce breakpoint width for device cards

garrett:
The breakpoint width was collapsing far too soon. Ideally, we'd want it at `md`, not `lg`, but there's a problem around the first few pixels. I think it's a side effect of PF's breakpoints being wonky (and us still not having a proper solution for that).

Meanwhile, `lg` is a big improvement over `xl` (which is what it's currently set at).

### DIFF
--- a/src/components/vm/disks/vmDisksCard.jsx
+++ b/src/components/vm/disks/vmDisksCard.jsx
@@ -244,7 +244,7 @@ export const VmDisksCard = ({ vm, vms, disks, renderCapacity, supportedDiskBusTy
     return (
         <>
             <ListingTable variant='compact'
-                gridBreakPoint='grid-xl'
+                gridBreakPoint='grid-lg'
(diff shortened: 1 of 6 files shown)


skobyda (approved):
This looks good. Update pixel test reference, if it passes, let's merge it.

skobyda:
Looks like the pixel tests are breaking constantly, but that because in each test MAC address is different in each test run, meaning the text itself is of different length, which affects the width of the columns with the MAC address and its neighbouring columns:
https://cockpit-logs.us-east-1.linodeobjects.com/pull-864-20221123-101417-37e55b92-fedora-36/pixeldiff.html#TestMachinesNICs-testVmNICs-vm-details-nics-card-medium

I will add new `ignore` elements into that pixel tests so it also ignores the width of the column with MAC address and its neighbouring column.

skobyda (approved):
Tests all green, including the pixel tests

PR #865, opened 2022-11-17
Limit use of primary buttons; switch to secondary

garrett:
As stated @ https://www.patternfly.org/v4/components/button/design-guidelines#primary-button:

> A primary button is the most prominent button on a page, used for the most important call to action on a page. Try to limit primary buttons to one per page.

These were basically glowing at you in dark mode, where they are even more prominent visually.

### DIFF
--- a/src/components/common/deleteResource.jsx
+++ b/src/components/common/deleteResource.jsx
@@ -103,8 +103,14 @@ DeleteResourceModal.propTypes = {
     deleteHandler: PropTypes.func.isRequired
 };
 
-export const DeleteResourceButton = ({ objectId, disabled, overlayText, actionName, dialogProps, isLink, isInline, className, isDropdownItem }) => {
+export const DeleteResourceButton = ({ objectId, disabled, overlayText, actionName, dialogProps, isLink, isInline, isSecondary, className, isDropdownItem }) => {
     const Dialogs = useDialogs();
+
+    let variant = "danger";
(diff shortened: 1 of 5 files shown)


garrett:
Also: I tried variant, but couldn't get that working, hence the className, which works. If you know of a better way, please update this PR with it. Otherwise, it's probably fine? :grin:

skobyda:
This changed the "remove" button for DHCP entries, which are however already inline links, so adding secondary made t it look weird:
TestMachinesNetworks-testNetworkState-networks-page-pixels

What is your intended change for the original buttons @garrett ?
TestMachinesNetworks-testNetworkState-networks-page-reference

skobyda (dismissed):
Even in the dark mode, it looks incorrect:
Screenshot from 2022-11-22 15-35-22

But the original looks fine, it doesn't really glow/stick out as you described in your original comment, so maybe changing it there is unecessary?
Screenshot from 2022-11-22 15-36-08

garrett:
> This changed the "remove" button for DHCP entries, which are however already inline links, so adding secondary made t it look weird

No, that was not intended. Thanks for the catch, @skobyda. Could you fix it in this PR please? (You know this code better than I do.)

KKoukiou:
@skobyda I will deal with it now so that we get it out of thelist.

KKoukiou (approved):

PR #866, opened 2022-11-17
Use PF text utility for disabled text

garrett:
The PF utility uses a standard color & has better contrast in dark mode.


skobyda (approved):
This looks good. I just updated the pixel test reference, and if that passes, let's merge it.

skobyda:
Tests are all green

PR #867, opened 2022-11-23
test: loop to press Enter in serial console to make sure the console has received it

yunmingyang:


KKoukiou (commented):
Can you please explain what you 're trying to achieve here :) ?

yunmingyang:
> Can you please explain what you 're trying to achieve here :) ?

Sure. Sometimes, I found that pressing Enter doesn't take effect("cirros login" is not shown). I am not sure the reason of the behavior. But I guess that, it is because, the VM boot too fast and pressing Enter is also too fast, so that the serial console doesn't get the signal of pressing Enter, then no "cirros login" shown. That will cause the case fail. So I add this loop to press Enter circularly, to make sure we can see "cirros login" in the console

KKoukiou:
@yunmingyang so this is essentially a hack to cover a timing issue in the test execution. I don't like introducing hacks in the code/test code without having a proof that it actually happens. I checked on grafana CI statistics and I don't see this test being flaky https://grafana-frontdoor.apps.ocp.ci.centos.org/d/ci/cockpit-ci?orgId=1

How did you notice that this test is being flaky?

yunmingyang:
The test being flaky in some manual automation runs, it is not 100 percent to be flaky, but I saw it several times. So I use this hack in some branch for downstream run, and the flaky seems disappeared.

test/check-machines-consoles
```diff
@@ -118,12 +119,16 @@ class TestMachinesConsoles(VirtualMachinesCase):
...
+        # In case the OS already finished booting, press Enter into the console to re-trigger the login prompt
+        # Sometimes, pressing Enter one time doesn't take effect, so loop to press Enter to make sure the console has accepted it
+        for _ in range(0, 60):
+            b.focus(f"#{name}-terminal .xterm-accessibility-tree")
+            b.key_press("\r")
+            if "cirros login" in b.text(f"#{name}-terminal .xterm-accessibility-tree"):
+                break
+            time.sleep(1)
+        # Make sure the content of console is expected
+        wait(lambda: "cirros login" in b.text(f"#{name}-terminal .xterm-accessibility-tree"))
```
KKoukiou, line 131: This is not needed, as its' already checked in line 127.

KKoukiou (approved):

PR #868, opened 2022-11-23
Change the default source mode of direct attachment NIC to bridge

yunmingyang:
As description in this Bugzilla, VEPA mode needs switch supplement. Also, for MacVTap NIC, it seems that virt-manager use bridge mode by default, and virt-manager seems that doesn't support mode options for MacVTap unless editing the XML. So changing default mode of "Direct attachment" NIC to bridge.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2095967


KKoukiou (approved):
Thanks a lot. This will enable by default host -> guest communication. Great improvement.

skobyda (approved):
LGTM

PR #874, opened 2022-12-01
Move dialog errors on the top of the body

KKoukiou:
Not all dialogs but some representative.

border-Screen Shot 2022-12-01 at 19 20 13
border-Screen Shot 2022-12-01 at 19 20 29
border-Screen Shot 2022-12-01 at 19 20 57
border-Screen Shot 2022-12-01 at 19 22 55
border-Screen Shot 2022-12-01 at 19 23 33
border-Screen Shot 2022-12-01 at 19 24 27
border-Screen Shot 2022-12-01 at 19 24 47


skobyda:
VCPU dialog will have no space between the error and the dialog body:
Screenshot from 2022-12-02 16-12-57

But this is already pre-existing issue, since the same problem is also with warnings:
Screenshot from 2022-12-02 16-32-53

So this should be probably fixed in separate PR.

KKoukiou:
> So this should be probably fixed in separate PR.

Happy to fix in this PR, not urgent to merge.

jelly (dismissed):

skobyda (approved):
This looks good, thanks

Issue #792, opened 2022-08-10
Explain "system" vs. "user session"


garrett:
We should probably explain the difference between running as a system VM versus running in a user session, in a popover connected to a (?) icon.

It should probably simply state the benefits and drawbacks of each approach too.

garrett:
(This is the same as https://github.com/cockpit-project/cockpit-podman/issues/1067, but for virtual machines.)

GregJohnStewart:
So have to ask, since the change hasn't made it into the version of Cockpit on Fedora 37 yet.... what's the difference?

(Google search is pretty sparse)

marusak:
You can see how it looks like in https://github.com/cockpit-project/cockpit-machines/pull/875

It was released in version cockpit-machines-281 which got into f-37 stable a day ago, see https://bodhi.fedoraproject.org/updates/FEDORA-2023-49b3964047

PR #875, opened 2022-12-02
Explain the difference between user and system connection

skobyda:
Fixes https://github.com/cockpit-project/cockpit-machines/issues/792

## Clarify the difference between user and system connection

**System:**
- Ideal for server VMs
- VM will launch with root permissions
- Ideal networking support
- Permissions denied for disk images in home directories

**User session:**
- Good choice for desktop virtualization
- VM launched with unprivileged limited access, with the process and PTY owned by your user account
- Restrictions in networking (SLIRP-based emulation) and PCI device assignment
- Disk images can be stored in user home directory

Screenshot from 2023-01-03 14-43-36

### DIFF
--- a/src/components/common/machinesConnectionSelector.jsx
+++ b/src/components/common/machinesConnectionSelector.jsx
@@ -20,17 +20,66 @@
 import React from 'react';
 
 import { LIBVIRT_SYSTEM_CONNECTION, LIBVIRT_SESSION_CONNECTION, rephraseUI } from '../../helpers.js';
-import { FormGroup, Radio } from '@patternfly/react-core';
+import { Button, FormGroup, Radio, Popover, Text, TextContent, TextList, TextListItem, TextVariants } from '@patternfly/react-core';
+import { HelpIcon } from "@patternfly/react-icons";
 import cockpit from 'cockpit';
+import './machinesConnectionSelector.scss';
 
 const _ = cockpit.gettext;
 
-export const MachinesConnectionSelector = ({ onValueChanged, loggedUser, connectionName, id }) => {
+export const MachinesConnectionSelector = ({ onValueChanged, loggedUser, connectionName, id, showInfoHelper }) => {
     if (loggedUser.id == 0)
         return null;
 
     return (
-        <FormGroup label={_("Connection")} isInline hasNoPaddingTop id={id} className="machines-connection-selector">
+        <FormGroup label={_("Connection")}
+                   isInline hasNoPaddingTop
+                   id={id}
+                   className="machines-connection-selector"
+                   labelIcon={
+                       showInfoHelper && <Popover id="machines-connection-selector-popover"
+                           bodyContent={<>
+                               <TextContent>
+                                   <Text component={TextVariants.h4}>{rephraseUI("connections", LIBVIRT_SYSTEM_CONNECTION)}</Text>
+                                   <TextList>
+                                       <TextListItem>
+                                           {_("Ideal for server VMs")}
+                                       </TextListItem>
+                                       <TextListItem>
+                                           {_("VM will launch with root permissions")}
+                                       </TextListItem>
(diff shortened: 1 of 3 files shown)


KKoukiou (dismissed):
I like this but I am sure @garrett will have some comments. Let's wait for his feedback

garrett (dismissed):
This is a great start! Thanks for writing the text! My initial impression is that it's quite verbose. But this makes it easier to edit down, so it's wonderful for a first pass. :+1: 

Should we use [lists](https://www.patternfly.org/v4/components/list)? It would be quicker to scan and understand, and if we don't conform to full sentences, we could use fragments to get to the point more quickly. (That is: Lists should be consistent; either full sentences for all points on the screen at the same time, or just fragments. Both are correct, as long as it's consistent overall. [See PF's guidelines for more information](https://www.patternfly.org/v4/ux-writing/writing-design-guidelines-for-patternfly/#lists). The sentence/fragment rule is implied in PF's guidelines, but it's a general style everywhere, at least in English. There's a good reference in [Microsoft's style guide](https://learn.microsoft.com/en-us/style-guide/scannable-content/lists#punctuation), for example.)

What do you think of the following? I tried to distill it down to main points for each.

## System

- Ideal for server VMs
- VM will launch with root permissions
- Ideal networking support
- Permissions denied for disk images in home directories

## User session

- Good choice for desktop virtualization
- VM launched with unprivileged limited access, with the process and PTY owned by your user account
- Restrictions in networking (SLIRP-based emulation) and PCI device assignment
- Disk images are stored in user home directory

skobyda:
Updated the PR. The list and the suggested text looks good. 
Maybe just one little thing, we could say `Disk images can be stored in user home directory` instead of `Disk images are stored in user home directory`? Because The difference is that System VM cannot access disks inside users home, but user can still configure VM to access disk which is outside of home. Then there are also network disks and block disks, which might might also be outside of home directory.
Anyway, here is a new version:

Screenshot from 2022-12-08 15-03-46

garrett:
> Maybe just one little thing, we could say Disk images can be stored in user home directory instead of Disk images are stored in user home directory? Because The difference is that System VM cannot access disks inside users home, but user can still configure VM to access disk which is outside of home. Then there are also network disks and block disks, which might might also be outside of home directory.
Anyway, here is a new version:

Sure! Thanks for the suggestion.

Also, is there a way to reduce the left padding (or margin?) of the lists? It seems rather excessive... is it nested in any other component? If not, then perhaps [using a spacing utility for margin or padding](https://www.patternfly.org/v4/utilities/spacing) (depending on what the lists use) might be needed.

garrett:
The lists have both padding and margin. Reducing padding on the left to 0 seems to do the trick. And I also changed the width to 100%, as it seemed constricted on the right side too.

image

The direction of the popover is also questionable, but I guess this is OK?

And it seems we have a regression for the alignment somehow...?!? Not sure if it's from this PR or in main like this.

image

skobyda:
> The lists have both padding and margin. Reducing padding on the left to 0 seems to do the trick. And I also changed the width to 100%, as it seemed constricted on the right side too.

Fixed
Screenshot from 2023-01-02 13-39-28


> The direction of the popover is also questionable, but I guess this is OK?

The direction is chosen automatically by patternfly based on available space. If we hardcode it to place the popover to the right, it would not entirely fit in the page (see screenshot below). I think letting patternfly choose (in our case, it places popover to the bottom) is okay.
Screenshot from 2023-01-02 13-39-55


> And it seems we have a regression for the alignment somehow...?!? Not sure if it's from this PR or in main like this.

Fixed:
Screenshot from 2023-01-02 13-39-11

martinpitt (approved):
LGTM, thank you! This may be release-notes material, if you'd like to see it mentioned, can you please update the description's screenshot to the current design and light mode?

PR #888, opened 2022-12-13
Improve warning wording for attaching raw disks to more than one VM

### LINKED ISSUE #765: Clarify attaching an already used disk
Oh, this is another one that was prematurely merged.

As I pointed out, the string "Attaching it will make this disk shareable for every VM using it" doesn't make sense. I couldn't suggest anything else as I didn't know what it meant (nor would our users), so I asked what it meant.

I still don't know:
- What "it" is
- If this will attach it if it's selected
- What "attach" means in this context of the "add" dialog (adding the disk will "attach" it right?)
- If this is a bad thing

This needs an immediate fixup to land before the release or it should be reverted.

If this is actually a bad thing that will cause data loss, then this needs to be a (!) danger not a <!> warning. Otherwise, suggestion:

"Adding this disk will change its mode to writable and shared."

_Originally posted by @garrett in https://github.com/cockpit-project/cockpit-machines/issues/756#issuecomment-1189183471_

In this dialog:

dialog

KKoukiou:
All hard disks are setup for exclusive access by default. When using raw disks, we can override the default behavior with the <shareable> libvirt flag. In cockpit-machines we add this flag automatically when the user tries to add to a VM a disk that is already used by a VM.

Technical note:
The shareable flag causes the SELinux/AppArmor policy to allow shared access, and tells the libvirt & QEMU lock managers to use shared locks.

Fixes #765

### DIFF
--- a/src/components/vm/disks/diskAdd.jsx
+++ b/src/components/vm/disks/diskAdd.jsx
@@ -74,7 +74,7 @@ function getDiskUsageMessage(vms, storagePool, volumeName) {
 
     let message = cockpit.format(_("This volume is already used by $0."), vmsUsing);
     if (volume.format === "raw")
-        message += " " + _("Attaching it will make this disk shareable for every VM using it.");
+        message += " " + _("Adding this disk will change its access mode to shared.");
 
     return message;
 }


src/components/vm/disks/diskAdd.jsx
```diff
@@ -74,7 +74,7 @@ function getDiskUsageMessage(vms, storagePool, volumeName) {
 
     let message = cockpit.format(_("This volume is already used by $0."), vmsUsing);
     if (volume.format === "raw")
-        message += " " + _("Attaching it will make this disk shareable for every VM using it.");
+        message += " " + _("Adding this disk will change its mode to shared.");
```
skobyda, line 77: Thanks

- I like `s/Attach/Add` and `s/shareble/shared` , since those are the terms we use in the UI.
- Maybe we should use "access" instead of "mode" to be consistent with the term we use in the UI? Screenshot from 2022-12-15 14-22-49
Or maybe combine it to "access mode" ? 
- I would still mention that the mode will be changed to share **for every VM using it**. Since it will set access mode to shared for every VM, not just the VM where the disk is being added to. The new text you suggest kinda feels like it will set the mode only for the disk added to the VM2, doesn't it? But maybe that's just my feeling.
Anyway, what about message
``suggestion
        message += " " + _("Adding this disk will change its access mode to shared or every VM using it.");
```

skobyda: Ah, I just realized that the message will always be preceded by `This volume is already used by [VM NAME]`. So probably "for every VM using it" is unnecessary, since it's implied by the context of the preceding sentence? But I'm not an English native speaker so I'm not 100% confident ¯\_(ツ)_/¯

garrett: Yeah, I think it's implied that it's for the VMs using it (which are mentioned already).

It's not changing its "access" to shared, but it is changing its "access mode" to shared. Good call for consistency and clarity. "Adding this disk will change its access mode to shared."

You're right on both points. :+1:

garrett: :1st_place_medal: This gets an "OK, let's merge!" from me after this change.

(If you need official approval after the change, you can check with someone else on the team within the next 2 weeks. Happy holidays!)

martinpitt: The PR uses "access mode" now, so resolving this thread and assuming Garrett's approval as above.

martinpitt (approved):

PR #893, opened 2022-12-16
Fix resolution issues with VNC console

skobyda:
### Use minimum height for VNC console

Sometimes when machine is booted up, VNC console is resized to the
small screen resolution of bootloader. Once the VM is booted, the
resolution is retained. This resolution is however too small for any
meaningfull interaction (e.g. it doesn't even allow you to click on
"Settings" at GNOME since the panel is cropped off)

**Before**
Screenshot from 2022-12-20 02-18-40

**After**
Screenshot from 2022-12-20 02-19-05

### Add minimal height for expanded VNC

Expanding VNC console sometimes affects only the width of the console,
not its height. One would expect that expanded VNC console should try to
stretch to the whole available height of the screen.

**Before**
Screenshot from 2022-12-20 13-15-38

**After**
Screenshot from 2022-12-20 13-43-20

### Don't force session resize when in expanded VNC dialog

When the resizeSession is turned on, it will automatically resize the VM
resolution based on the container element in which VNC is located. This
is however counter-productive, as it will in effect prevent user from
setting resolution inside the VM and always resize the back to the
container element.

**Before**
(Resolution change is not successful and always scales back to the previous value)

[Screencast from 2022-12-20 14-30-36.webm](https://user-images.githubusercontent.com/42733240/208682524-318c38ad-34c6-4faa-b88c-19ba23fc0a16.webm)

**After**

[Screencast from 2022-12-20 14-31-44.webm](https://user-images.githubusercontent.com/42733240/208682259-9f893f59-3efe-4bb2-9710-64225cdc16d1.webm)

### DIFF
--- a/src/components/vm/consoles/consoles.css
+++ b/src/components/vm/consoles/consoles.css
@@ -6,6 +6,10 @@
     height: auto !important;
 }
 
+#vnc-display-container {
+    min-height: 480px;
+}
+
 /* Hide send key button - there is not way to do that from the JS
  * https://github.com/patternfly/patternfly-react/issues/3689
  */
--- a/src/components/vm/consoles/consoles.css
+++ b/src/components/vm/consoles/consoles.css
@@ -6,8 +6,12 @@
     height: auto !important;
 }
 
-#vnc-display-container {
+#vnc-display-container-minimized {
     min-height: 480px;
+}
+
+#vnc-display-container-expanded {
+    min-height: 68vh;
 }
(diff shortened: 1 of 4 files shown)


KKoukiou:
Some pixel tests failing. Otherwise, no objection here.

skobyda:
All tests green. @KKoukiou mind giving it another review?

KKoukiou (approved):

Issue #885, opened 2022-12-10
Virtual machine search should be case insensitive


shodanshok:
Virtual machine filter/search should be case insensitive by default - otherwise, one has to try various combination of upper/lower case when searching virtual machines.

For example, I have a virtual machine called "Hubble". However, searching for "hubble" shows no results:
![Cattura](https://user-images.githubusercontent.com/13137622/206841583-4913170a-4297-4f6f-bc5b-2c7011768a8f.JPG)

garrett:
I agree :100:

PR #896, opened 2022-12-20
Make VM search case insensitive

skobyda:
Fixes https://github.com/cockpit-project/cockpit-machines/issues/885


jelly (approved):
Thanks!

jelly:
Seems the rawhide test failure also occurs in https://artifacts.dev.testing-farm.io/6e224570-80e6-45f8-a16c-82db6c26e134/

skobyda:
All green, apart from that fedora-rawhide failure, which can be also found in other PRs, so it's not related to this

Issue #29, opened 2021-03-30
Mouse clicks does not work


brunomgrocha:
Hi,

I'm using cockpit with my KVM server, but since last update i can´t install anything using the web console because my clicks on the mouse doesn't reflect on the screen. Do you know how to fix this?
Thank you

KKoukiou:
Hi @brunomgrocha you seem to be  facing this issue, https://bugzilla.redhat.com/show_bug.cgi?id=1913548 

Try to expand the Vm's console which will open it in a new window.

brunomgrocha:
> Hi @brunomgrocha you seem to be facing this issue, https://bugzilla.redhat.com/show_bug.cgi?id=1913548
> 
> Try to expand the Vm's console which will open it in a new window.

Hi 
I can only expand in the same windows, can't expand to a new windows. 
Can you please show me how to do that?
thank you

KKoukiou:
@brunomgrocha Which version of cockpit-machines are you using? 
You need at least cockpit-machines-231 for the console expand feature. See second screenshot in https://cockpit-project.org/blog/cockpit-231.html

brunomgrocha:
> @brunomgrocha Which version of cockpit-machines are you using?
> You need at least cockpit-machines-231 for the console expand feature. See second screenshot in https://cockpit-project.org/blog/cockpit-231.html

Hi, I'm using version 237

brunomgrocha:
> > @brunomgrocha Which version of cockpit-machines are you using?
> > You need at least cockpit-machines-231 for the console expand feature. See second screenshot in https://cockpit-project.org/blog/cockpit-231.html
> 
> Hi, I'm using version 237

If you want i can upload a screenshot or a small movie of the issue.

KKoukiou:
@brunomgrocha Yes that would be helpful :) any of the two, maybe I misunderstand the issue you have.

KKoukiou:
This is probably the same issue as discussed here https://bugzilla.redhat.com/show_bug.cgi?id=1913548

1ras:
Just to note, this also occurs in expanded view, depending on guest resolution and browser window size (whenever scaling takes place).

garrett:
Yes, the VNC library we use doesn't map mouse clicks 1:1 when the virtual display of the machine is larger than the region it is in. (In other words, if it is scaled.)

The only things you can do right now are:

- Make sure the resolution of the VM is smaller than the expanded version of the screen.
[...]

PR #900, opened 2022-12-22
Use viewport options for VNC console

skobyda:
There is an issue where if virtual machine system's resolution is set to something larger than VNC console element rendered in the browser, the view will be scaled down, but there will be an offset for mouse clicks:

[false.webm](https://user-images.githubusercontent.com/42733240/209139339-e08ce7fa-3fcb-45f5-a89b-e7a5e0ed103d.webm)

This should fix it:

[true-(2).webm](https://user-images.githubusercontent.com/42733240/209139322-ebb84f0a-7810-44f0-960f-df9d8bc0a1bc.webm)

Fixes https://github.com/cockpit-project/cockpit-machines/issues/29
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1913548
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2030836

### DIFF
--- a/src/components/vm/consoles/vnc.jsx
+++ b/src/components/vm/consoles/vnc.jsx
@@ -169,7 +169,8 @@ class Vnc extends React.Component {
                         textDisconnected={_("Disconnected")}
                         textDisconnect={_("Disconnect")}
                         consoleContainerId={isExpanded ? "vnc-display-container-expanded" : "vnc-display-container-minimized"}
-                        resizeSession={!isExpanded}
+                        resizeSession
+                        scaleViewport
             />
         );
     }
--- a/src/components/vm/consoles/consoles.css
+++ b/src/components/vm/consoles/consoles.css
@@ -1,17 +1,27 @@
 @import "@patternfly/react-styles/css/components/Consoles/SerialConsole.css";
 
-/* Make the VNC canvas scale down to available width */
-.pf-c-console__vnc canvas {
-    max-width: 100%;
-    height: auto !important;
+.pf-c-console,
+.pf-c-console__vnc,
+.pf-c-console__vnc > div,
+.pf-c-console__vnc > div > div {
+    width: 100%;
+    height: 100%;
 }
 
-#vnc-display-container-minimized {
-    min-height: 480px;
+.pf-c-console__actions {
+    max-width: 25ch;
 }
 
-#vnc-display-container-expanded {
-    min-height: 68vh;
+.pf-c-console {
+     /* auto minimum width, causing problem */
+     grid-template-columns: 1fr;
+     grid-template-rows: min-content 1fr;
+}
+
+.consoles-page-expanded .pf-c-page__main {
+    .actions-pagesection {
+        padding-bottom: 0;
+    }
 }
 
 /* Hide send key button - there is not way to do that from the JS
@@ -19,9 +29,4 @@
  */
 #pf-c-console__send-shortcut {
     display: none;
-}
-
-/* Add spacing between action buttons */
-.pf-c-console__actions-vnc {
-    gap: var(--pf-global--spacer--sm);
 }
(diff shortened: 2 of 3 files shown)


KKoukiou:
This is now unblocked :raised_hands:

src/components/vm/consoles/vnc.jsx
```diff
@@ -171,6 +171,9 @@ class Vnc extends React.Component {
                         textDisconnect={_("Disconnect")}
                         consoleContainerId={isExpanded ? "vnc-display-container-expanded" : "vnc-display-container-minimized"}
                         resizeSession={!isExpanded}
+                        clipViewport
+                        dragViewport
+                        scaleViewport
```
KKoukiou, line 176: Looks like scalling and clipping don't work well together? https://github.com/novnc/noVNC/blob/master/core/rfb.js#L344

skobyda:
This Improved the scaling of VNC singificantly, but still breaks with higher resolution. 

Basically before this PR, there was offset between mouse position shown on screen and mouse position recorded by the system basically for every resolution, as shown [here](https://github.com/cockpit-project/cockpit-machines/pull/900#issue-1507838265), as shown here.

Now however the VNC scales properly for small resolution (e.g. 800x600), high resolution (e.g. 1080p), but scaling still breaks for ultra high resolution (3840x2160).

To show what I mean, here is a demo:

[presentation-(2).webm](https://user-images.githubusercontent.com/42733240/219353663-3a8e5949-ccfa-4dc9-b22f-03d88a0517b2.webm)

skobyda:
But still, this improves usability of our VNC singificantly, so I'm very happy to merge this as it is.

skobyda:
But hey, we are passing the "scaleViewport" and "resizeSession" flags to noVnc, so I guess that it doesn't scale and resize properly for ultra-high resolution is probably now a bug for [noVnc](https://github.com/novnc/noVNC)?

KKoukiou:
> But hey, we are passing the "scaleViewport" and "resizeSession" flags to noVnc, so I guess that it doesn't scale and resize properly for ultra-high resolution is probably now a bug for [noVnc](https://github.com/novnc/noVNC)?

Yes, looks like that. I will check if exissts otherwise report. I will update pixel tests here, to get it in as partial fix.

KKoukiou:
The debian installer is being heavily clipped. It does not seem to respond to the resize requests.

garrett:
> The debian installer is being heavily clipped. It does not seem to respond to the resize requests.

Possibly silly suggestion: Is there a way to set a fallback/default resolution — like to 1024x768 — which is what would be used unless a size request works? If so, then Debian (and others) would then work, but those supporting size requests should have whatever resolution they need too.

KKoukiou:
> > The debian installer is being heavily clipped. It does not seem to respond to the resize requests.
> 
> Possibly silly suggestion: Is there a way to set a fallback/default resolution — like to 1024x768 — which is what would be used unless a size request works? If so, then Debian (and others) would then work, but those supporting size requests should have whatever resolution they need too.

Unfortunately the API does not make this posisble.

garrett:
Still a problem with Debian Testing, apparently:

image


Expanding works though:

image


Expanded, very narrow:

image


But if I resize so it's a little less vertical, it zooms in and crops:

image

skobyda:
Ubuntu seems to work correctly:
Screenshot from 2023-02-21 11-52-05

The issue on ultra-high resolution for fedora persists:
Screenshot from 2023-02-21 11-51-05

skobyda:
For some reason, debian-testing is not so bad for me. The bottom of the screen seems to be slightly cut off, but nothing as bad as seen in garret's screenshot:
Screenshot from 2023-02-21 11-57-05
Screenshot from 2023-02-21 11-56-56

garrett:
> But still, this improves usability of our VNC singificantly, so I'm very happy to merge this as it is.

How ready is this? I know it's not perfect, but it's much better than we have, at least from when I looked at it. It would be nice to get it into this week's release.

skobyda:
I'm personally happy with how it is now, as it improves stuff by a lot. Let's merge it! There are corner case issues, let's create a github issue to track it after this gets merged.

@KKoukiou  since I'm author of this PR, I can't approve it, so do you mind approving (if you are also happy with the current state) and merging. From my side it's LGTM

KKoukiou (approved):

PR #902, opened 2023-01-02
test: Map "fedora-rawhide" to "fedora-38"

martinpitt:
This will properly apply naughties.

### DIFF
--- a/test/run
+++ b/test/run
@@ -6,6 +6,11 @@ set -eu
 
 export RUN_TESTS_OPTIONS=--track-naughties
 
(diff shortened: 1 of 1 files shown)


martinpitt:
See https://github.com/cockpit-project/bots/pull/4230

martinpitt:
Bah, no, that [doesn't work](https://cockpit-logs.us-east-1.linodeobjects.com/pull-902-20230102-143536-100e3730-fedora-rawhide-bots-4230/log)

PR #905, opened 2023-01-03
 workflows: Pacify git's permission check in release workflow

KKoukiou:
Similar to https://github.com/cockpit-project/bots/pull/4224


martinpitt (approved):
Thanks! Triggering required d-testing check.

PR #907, opened 2023-01-04
Remove some unused CSS

KKoukiou:


KKoukiou:
So this CSS is used, and it making the labels text alligning with the column headers. I am not sure if that's correct or not though.

PR #917, opened 2023-01-19
Snapshot: Offer force revert option if regural revert fails

skobyda:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1977789

## Machines: Option to forcefully revert a snapshot

There are situations where a snapshot revert involves extra risk and the user should either consider not reverting snapshot, or force it. The few most common cases are:
- Snapshot that lacks full domain information for reverting configuration. In this case, the user should be sure that snapshot is compatible with the current configuration (if it is not, the domain will likely fail to run).
- Reverting from a running domain to an active state where a new hypervisor has to be created rather than reusing the existing hypervisor.
- Snapshot of inactive VM which also has a managed saved state. This requires forceful revert because the snapshot does not contain a memory state and will therefore not replace the existing memory state. This ends up switching a disk underneath a running system and can cause filesystem corruption or crashes due to swap content mismatches when run.

Screenshot from 2023-01-19 13-28-57

[Screencast from 2023-01-19 13-29-42.webm](https://user-images.githubusercontent.com/42733240/213447922-47cab8cd-8dd3-4afb-873a-ba3ec8cebcdf.webm)


skobyda:
The approach is consistent forceful disk ejection:

194514524-174861d2-cdb3-475c-8010-e124fe657e5e

https://user-images.githubusercontent.com/42733240/213448516-1de30b7b-fc6d-40a8-aa8e-315f3bdd290d.mp4

martinpitt:
TestMachinesSnapshots.testSnapshotRevert [fails on the Testing Farm](https://artifacts.dev.testing-farm.io/b1fa873d-0fe3-49b1-bfbe-798656c6a73b/)

martinpitt:
Looks good to me otherwise. Can you please re-do the screenshot in light theme, so that we get consistent release notes? Thank you!

skobyda:
Fixed the snapshot test flake. But [Fedora-rawhide](https://artifacts.dev.testing-farm.io/d07bca74-c97c-4af2-9deb-6667f40aa7c1/) constatnly keeps failing on disk and storage pool tests. I guess those are unrelated flakes?

martinpitt:
Rawhide [was fine](https://artifacts.dev.testing-farm.io/76eb5705-9bd3-4297-becb-887005d218c6/) in #915 4 days ago. `TestMachinesDisks.testInsertDiscCDROM` failed here, and also in #916, so it seems unrelated indeed.

martinpitt (approved):
Thank you! Please still update the screenshot for light theme.

martinpitt:
Oops, maybe I shouldn't have merged this yet, you asked for @garrett 's review as well. @garrett , if you have something to change, I'll send a revert PR.

PR #924, opened 2023-01-25
Bump cockpit repo to slightly before 284

martinpitt:
This picks the fix for the "ugly zero" font fix from tabular numbers.
Adjust pixel tests accordingly.

Disable the "RTL" pixel scenario, as it's still causing unstable
reference images, and we don't yet know what exactly to do with that in
Cockpit.  Unfortunately we can't use test/browser-layouts.json as we
already have the "dark" scenario, want to retain it, but it does not 
work on Firefox.  This gets computed dynamically in testlib, but a
static browser-layouts.json prevents this. Thus, explicitly skip the 
"rtl" scenario everywhere for now.

Don't update to full 284 yet, as [1] breaks testStoragePoolsCreate. This
works interactively, but not in the tests. This is not trivial to work
around, so do this in a separate commit that isn't blocking the release.

[1] https://github.com/cockpit-project/cockpit/commit/a3fad8f6e71037539

----

[pixel test review](https://github.com/cockpit-project/pixel-test-reference/compare/27c3efa8dc7fa03e184569c9464c04aab6b56f6d..7355d3b47607fb21ccc911d0e6e32bddddf57857)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
-COCKPIT_REPO_COMMIT = 69aac32ebe3e23384a00321b9570f64f4ac07583 # 282.1
+COCKPIT_REPO_COMMIT = 736e2369118015ff1da352a790d1a92168439c7a # 284 minus "Disable file watching in FileAutoComplete"
 
 $(COCKPIT_REPO_FILES): $(COCKPIT_REPO_STAMP)
 COCKPIT_REPO_TREE = '$(strip $(COCKPIT_REPO_COMMIT))^{tree}'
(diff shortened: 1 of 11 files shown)


martinpitt:
react-console pulls in xterm 4.19.0, but cockpit's shared terminal widget now requires xterm 5 and xterm-addon-canvas. I interactively tested the serial console with xterm 5, and it works just fine:

image

So let's bump this along.

martinpitt:
I threw about an hour on [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-924-20230125-130123-0ef5bce8-fedora-37/log.html#78-2), and tried various workarounds in the test -- but I need some more time to investigate that more closely. Let's not block the release unnecessarily and drag this out even more. I reverted to an earlier cockpit snapshot, still good enough to pick up the Ø fix (which was my main motivation)

jelly (approved):

PR #925, opened 2023-01-25
Transient domains should not be redefined when editing their network devices

KKoukiou:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=2030948


skobyda (changes requested):
Thanks, great fix. Maybe we could also add some tests so it doesn't break in the future again?

skobyda (approved):
LGTM, thanks

Issue #920, opened 2023-01-22
Error when setting user password during installation of guest OS


GregJohnStewart:
![image](https://user-images.githubusercontent.com/7083701/213945399-b104be6d-3506-4f23-80f6-ec0273709b26.png)

Context:

- System mode
- Installing guest of Fedora 37
- Set user account and admin password
- Happened during the install, after the install settings screen, download, etc

KKoukiou:
Hi! I guess you set the password and user name from the anaconda installer UI, right? And not from the 'Automation tab' in cockpit-machines' UI.
Which ISO or installation media did you use in that case?


----->>>> Here is the UI I talk about:
![Screen Shot 2023-01-23 at 12 02 06](https://user-images.githubusercontent.com/14921356/214023998-8b74ffe6-d4ed-4efa-bb2d-4efccaa428b7.png)

GregJohnStewart:
The password/username was set from the ui you pictured, I was doing an automated install on an automatically downloaded Fedora 37 image. I selected "Workstation"

GregJohnStewart:
Just got this as well:
![image](https://user-images.githubusercontent.com/7083701/215239572-dcc5b327-0e92-4ff9-94f2-7a1ef6354128.png)

![image](https://user-images.githubusercontent.com/7083701/215239585-dd05b7c6-b875-4667-8797-3db9af91c88c.png)

With no user specified, just root password....

If I had to guess, maybe the password characters aren't escaped in the script? the password I gave had a single quote `'`, and an opening square bracket `[`. I used the same admin password both times.

GregJohnStewart:
Confirmed that removing the single quote fixed it....
Probably want to sanitize/escape passwords, and test to ensure works as intended.

jelly:
Interesting, [virt-install](https://man.archlinux.org/man/virt-install.1#profile=) does not mention this is an requirements so it sounds like we should certainly escape quotes buy also report a bug upstream.

skobyda:
Hey @GregJohnStewart , since we are hasing passwords now: https://github.com/cockpit-project/cockpit-machines/pull/932 , this might have been already fixed. Can you check if the bug still exists?

PR #932, opened 2023-01-31
Hash passwords provided by the user for cloud base image

skobyda:
This will allow us to safely store user-data file in the system, and thus allow "Cloud base image" + "Create and edit" options together.

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -66,7 +66,7 @@ import { storagePoolRefresh } from './storagePool.js';
 import { snapshotGetAll } from './snapshot.js';
 import { downloadRhelImage, getRhelImageUrl } from './rhel-images.js';
 import { call, Enum, timeout, resolveUiState } from './helpers.js';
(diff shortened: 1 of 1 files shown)


jelly:
Unrelated, but can this approach also resolve https://github.com/cockpit-project/cockpit-machines/issues/920

KKoukiou:
mkpasswd does not exist on ach, centos-8, debians, rhel and ubuntu. Essentially it's just available on fedora.

skobyda:
Those two failing packing tests seems like unrelated flakes. @KKoukiou Mind giving a re-review? 
In the end, I decided to use openssl CLI tool to generate hash instead of python's crypt/haslib. The reason is it was more straighforward, uses its own defaults for length and generating random salt (in python we would have to specify all of tha tourself, generate our own random salt number of corrent length...) and already produces the hash in a format of /etc/shadow: `$6$[salt]$[hash]`. In python, we would have to format it ouserls.

KKoukiou (approved):

skobyda:
Tests all green, thanks for review

PR #933, opened 2023-02-01
test: export TEST_BROWSER for the Firefox scenario

jelly:
To actually test on Firefox we need to export TEST_BROWSER also print out the run commands so we can verify firefox is started.


martinpitt (approved):

martinpitt:
https://artifacts.dev.testing-farm.io/10d58214-9b49-42af-97a5-58d21280555a/ is a naughty, fixed in https://github.com/cockpit-project/bots/pull/4348

PR #950, opened 2023-02-14
Makefile: set TEST_OS to Fedora-37

jelly:
Fedora 37 is the last supported Fedora release.


jelly:
Added `no-testing` so we don't have to run all the tests, testing `debian-testing` should be good enough

martinpitt (approved):

jelly:
Unrelated packit failures, merging.

PR #955, opened 2023-02-17
build: Move to a webpack module, generalize bundler language/variable names

martinpitt:
Cockpit recently changed to an ESM build system [1]. Bump COCKPIT_REPO_COMMIT to that and follow suit.

This does not work with old node.js 10 any more which is still the default in RHEL 8. Install the newer version 16 instead.

The updated cockpit lib causes some subtle noise in all pixel tests, refresh them.

[1] https://github.com/cockpit-project/cockpit/pull/18366

-----

Pretty much identical to https://github.com/cockpit-project/starter-kit/pull/625 except for the node_modules submodule refresh and some extra import fixes.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -20,11 +20,11 @@ your browser.
 
 You can also use
 [watch mode](https://webpack.js.org/guides/development/#using-watch-mode) to
-automatically update the webpack on every code change with
+automatically update the bundle on every code change with
 
     $ make watch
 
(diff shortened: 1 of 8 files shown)


martinpitt:
Whoops, this [subtly changed pixel tests](https://cockpit-logs.us-east-1.linodeobjects.com/pull-955-20230217-200932-56fdc4db-fedora-37/log.html#20)...

KKoukiou (approved):

PR #956, opened 2023-02-20
Test: Add a check when no host devices selected

yunmingyang:


KKoukiou (commented):
@yunmingyang this breaks testHostDevAddSessionConnection

yunmingyang:
@KKoukiou After checking, there is a issue in the createVM as the log path is fixed, so, if using a non-root user and the VM will be run, there will be an error which is "Permission denied" for the log path. Add another commit for mapping the log path with corresponding users

skobyda (approved):
Looks good to me. There is a test failure:
```
testlib.Error: FAIL: Test completed, but found unexpected journal messages:
asyncio-ERROR: Task was destroyed but it is pending!
```
,but it's also present in other PRs, so no related to me. So this looks good to me.

PR #960, opened 2023-02-22
spec: Work around spice bug in Fedora 38

martinpitt:
Until this is fixed, unbreak spice consoles by pulling in qemu-audio-spice if the other spice packages are installed. This has no additional dependencies and is just 20 kB, and splitting the packages is pointless anyway.

See https://bugzilla.redhat.com/show_bug.cgi?id=2170110

----

This should eventually fix [testExternalConsole](https://cockpit-logs.us-east-1.linodeobjects.com/pull-955-20230219-100124-53a496a0-fedora-38/log.html), which gets currently [naughtied](https://github.com/cockpit-project/bots/issues/4415). This should fail in the first run on fedora-38 as the qemu-audio-spice package does not exist on our fedora-38 image. I'll add it as the next step, but I want to make sure the dependency works as expected.

 - https://github.com/cockpit-project/bots/pull/4443

### DIFF
--- a/packaging/cockpit-machines.spec.in
+++ b/packaging/cockpit-machines.spec.in
@@ -51,6 +51,10 @@ Recommends: qemu-block-curl
 Recommends: qemu-char-spice
 Recommends: qemu-device-usb-host
 Recommends: qemu-device-usb-redirect
+# HACK: https://bugzilla.redhat.com/show_bug.cgi?id=2170110
+%if 0%{?fedora} >= 38
+Requires: (qemu-audio-spice if qemu-char-spice)
+%endif
 %endif
 %endif
 Requires: libvirt-client


martinpitt:
This works as expected - [fedora-37](https://cockpit-logs.us-east-1.linodeobjects.com/pull-960-20230222-064432-46e10ce1-fedora-37/log.html) ignores the new dependency, and [fedora-38 package installation fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-960-20230222-064436-46e10ce1-fedora-38/log)  due to

> error: Failed dependencies:
>     (qemu-audio-spice if qemu-ui-spice-core) is needed by cockpit-machines-283.12.g46e10ce-1.fc38.noarch

So, let's add the dependency to the image, then this should be okay.

mvollmer (approved):
LGTM

mvollmer:
Green enough!

PR #961, opened 2023-02-22
Allow creating a VM without immediatelly starting with Cloud base image

### LINKED ISSUE #880: Allow 'Create and edit' when create vm from Cloud base image
Considering rhel guest image supports boot from uefi and legacy bios mode, but cockpit-machines keep this button grey and cannot click to change boot mode between uefi and bios. Could cockpit-machine add this support? Thanks

skobyda:
The reason why we didn't allow creation of VM without immediatelly
starting it when when root password and login password were supplied
was that we didn't want to store sensitive information anywhere on
the system.
Ever since #932 was merged, this is no longer the issue

Fixes #880 

# Machines: Create a cloud image VM without immediately starting it

A cloud base image VM can now be created without immediately starting it by clicking "Create and edit", which was disabled until now. This allows users to configure additional VM properties (e.g. bootloader, boot order, disks, networking, etc.) before the VM is installed.

Screenshot from 2023-03-21 14-07-46
Screenshot from 2023-03-21 14-07-58

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -1320,18 +1320,17 @@ class CreateVmModal extends React.Component {
                     key="secondary-button"
                     id="create-and-edit"
                     isLoading={this.state.createMode === EDIT}
-                    isAriaDisabled={
+                    isAriaDisabled={!!(
                         this.state.createMode === EDIT ||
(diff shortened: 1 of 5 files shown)


skobyda:
Apart from 1 flake, all green

KKoukiou (changes requested):

KKoukiou (approved):

PR #964, opened 2023-02-23
Don't load translations via globbing

jelly:


martinpitt (approved):
Ugh, so sad that we missed that here, we fixed it in most (all?) other projects.

jelly:
> 

> Ugh, so sad that we missed that here, we fixed it in most (all?) other projects.

Yes :( I want to search the rest of Github for more occurrences of this.

PR #972, opened 2023-03-02
test: add a check about the connection description

yunmingyang:


test/check-machines-lifecycle
```diff
@@ -89,6 +89,21 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         self.waitPageInit()
         self.waitVmRow("subVmTest1")
 
+        # Check connection interpretation
+        b.click("#import-existing-vm")
+        b.wait_visible("#create-vm-dialog")
+        b.wait_visible('#create-vm-dialog button[aria-label="more info"]')
```
martinpitt, line 95: Would it make sense to actually check what it does, i.e. that hovering over it creates a popup with some significant word or two in it?

martinpitt (commented):
Thanks! I triggered tests. I leave the main review to @KKoukiou or @skobyda , but one suggestion.

test/check-machines-lifecycle
yunmingyang: Make sense, I will complement the case, thanks

martinpitt (approved):
Thanks! @skobyda , any other comments?

skobyda (approved):
LGTM. Once the tests pass (apart from that fedora-rawhide which is underlated failure), let's merge this

PR #975, opened 2023-03-08
Don't start virtproxyd.socket

skobyda:
Unit virtproxyd.socket not found started to show up on fedora-rawhide.
    Let's see if we actually need it for our CI.


martinpitt (approved):
Hah, nice -- seems to work! :+1: Not finished yet, but so far tests look  good.

martinpitt:
@skobyda: It still [fails](https://artifacts.dev.testing-farm.io/f34f82c9-3ec0-441c-a161-bf7d3f167702/) two storage tests, but I think that's still better than failing completely right at the start. So my +1 stands.

PR #980, opened 2023-03-14
fix: check if virtualization is disabled and show warning

### LINKED ISSUE #891: Check if machine can run VMs (as some BIOS/EFI settings need to be enabled)
We had someone report in #879 that they could not run a VM. We diagnosed the issue as a result of the BIOS/EFI virtualization setting not being enabled.

Many laptops and desktops ship with virtualization off. It's possible some servers could too.

As people do also use Cockpit for staging systems on laptops and desktops — and there's a possibility that some server machines could also ship with this toggle — I think we need to check that a VM can run on the system and show an appropriate error message, with a hint that they might need to reboot into BIOS/EFI and enable virtualization.

I've added a question to the FAQ @ https://github.com/cockpit-project/cockpit-project.github.io/pull/594 — but even when we check in `cockpit-machines`, there will be older versions of Cockpit in use in Long Term Enterprise distros.

Hopefully we can resolve it rather soon for `cockpit-machines`, however.

Without this detection, it looks like `cockpit-machines` does not work, so it's both a bug an enhancement.

subhoghoshX:
Fixes #891.

Check if virtualization is disabled in BIOS/EFI and show an appropriate warning outlining how to fix it.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -109,7 +109,6 @@ export const App = () => {
     useEffect(() => {
         async function checkVirtualization() {
             const hardwareVirtCheck = await cockpit.script("virt-host-validate | grep 'Checking for hardware virtualization'");
-            console.log('cpuinfo => ', hardwareVirtCheck.includes('PASS'));
 
             setVirtualizationEnabled(hardwareVirtCheck.includes('PASS'));
         }
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -19,8 +19,11 @@
 import React, { useState, useEffect } from 'react';
 import { AlertGroup } from "@patternfly/react-core/dist/esm/components/AlertGroup/index.js";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
+import { EmptyState, EmptyStateIcon, EmptyStateBody, EmptyStatePrimary } from "@patternfly/react-core/dist/esm/components/EmptyState/index.js";
 import { Progress, ProgressMeasureLocation } from "@patternfly/react-core/dist/esm/components/Progress/index.js";
-import { ExclamationCircleIcon } from '@patternfly/react-icons';
+import { Text } from "@patternfly/react-core/dist/esm/components/Text/index.js";
+import { Title } from "@patternfly/react-core/dist/esm/components/Title/index.js";
+import { ExclamationCircleIcon, WrenchIcon } from '@patternfly/react-icons';
 import { superuser } from "superuser.js";
 import cockpit from 'cockpit';
 
@@ -69,6 +72,8 @@ export const App = () => {
     const [loadingResources, setLoadingResources] = useState(false);
     const [error, setError] = useState('');
     const [systemSocketInactive, setSystemSocketInactive] = useState(false);
+    const [virtualizationEnabled, setVirtualizationEnabled] = useState(true);
+    const [emptyStateIgnored, setEmptyStateIgnored] = useState(false);
 
     useEvent(superuser, "changed");
     useEffect(() => {
@@ -101,7 +106,38 @@ export const App = () => {
                 .finally(() => setLoadingResources(false));
     }, []);
 
-    if ((superuser.allowed && systemSocketInactive) || loadingResources) {
+    useEffect(() => {
+        async function checkVirtualization() {
+            const hardwareVirtCheck = await cockpit.script("virt-host-validate | grep 'Checking for hardware virtualization'");
+            console.log('cpuinfo => ', hardwareVirtCheck.includes('PASS'));
+
+            setVirtualizationEnabled(hardwareVirtCheck.includes('PASS'));
+        }
+
+        checkVirtualization();
+    }, []);
+
+    if (!virtualizationEnabled && !emptyStateIgnored) {
+        return (
+            <EmptyState>
+                <EmptyStateIcon icon={WrenchIcon} />
+                <Title headingLevel="h4" size="lg">
+                    Virtualization support is not available
+                </Title>
+                <EmptyStateBody>
+                    <Text>Enable virtualization support in BIOS/EFI settings.</Text>
+                    <Text>
+                        Instructions vary per manufacturer, but usually involves pressing a hot key such as ESC, F1, F12, or Del during boot.
+                        In the settings screen, look for "virtualization", "VM", "VMX', "SVM", "VTX", "VTD". It will be different on every computer.
+                        Enable any of these options. Consult your computer's manual for details.
+                    </Text>
+                </EmptyStateBody>
+                <EmptyStatePrimary>
+                    <Button variant="link" onClick={() => setEmptyStateIgnored(true)}>Ignore</Button>
+                </EmptyStatePrimary>
+            </EmptyState>
+        );
+    } else if ((superuser.allowed && systemSocketInactive) || loadingResources) {
         return (
             <LibvirtSlate loadingResources={loadingResources} />
         );
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -139,7 +139,7 @@ export const App = () => {
                     </Text>
                 </EmptyStateBody>
                 <EmptyStatePrimary>
-                    <Button variant="link" onClick={() => setEmptyStateIgnored(true)}>{_("Ignore")}</Button>
+                    <Button id="ignore-hw-virtualization-disabled-btn" variant="link" onClick={() => setEmptyStateIgnored(true)}>{_("Ignore")}</Button>
                 </EmptyStatePrimary>
             </EmptyState>
         );
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -21,7 +21,7 @@ import { AlertGroup } from "@patternfly/react-core/dist/esm/components/AlertGrou
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
 import { EmptyState, EmptyStateIcon, EmptyStateBody, EmptyStatePrimary } from "@patternfly/react-core/dist/esm/components/EmptyState/index.js";
 import { Progress, ProgressMeasureLocation } from "@patternfly/react-core/dist/esm/components/Progress/index.js";
-import { Text } from "@patternfly/react-core/dist/esm/components/Text/index.js";
+import { Text, TextContent } from "@patternfly/react-core/dist/esm/components/Text/index.js";
 import { Title } from "@patternfly/react-core/dist/esm/components/Title/index.js";
 import { ExclamationCircleIcon, WrenchIcon } from '@patternfly/react-icons';
 import { superuser } from "superuser.js";
@@ -127,16 +127,18 @@ export const App = () => {
 
     if (!virtualizationEnabled && !emptyStateIgnored) {
         return (
-            <EmptyState>
+            <EmptyState className="virtualization-disabled-empty-state">
                 <EmptyStateIcon icon={WrenchIcon} />
                 <Title headingLevel="h4" size="lg">
                     {_("Hardware virtualization is disabled")}
                 </Title>
                 <EmptyStateBody>
-                    <Text>{_("Enable virtualization support in BIOS/EFI settings.")}</Text>
-                    <Text>
-                        {_("Changing BIOS/EFI settings is specific to each manufacturer. It involves pressing a hotkey during boot (ESC, F1, F12, Del). Enable a setting called \"virtualization\", \"VM\", \"VMX\", \"SVM\", \"VTX\", \"VTD\". Consult your computer's manual for details.")}
-                    </Text>
+                    <TextContent>
+                        <Text>{_("Enable virtualization support in BIOS/EFI settings.")}</Text>
+                        <Text>
+                            {_("Changing BIOS/EFI settings is specific to each manufacturer. It involves pressing a hotkey during boot (ESC, F1, F12, Del). Enable a setting called \"virtualization\", \"VM\", \"VMX\", \"SVM\", \"VTX\", \"VTD\". Consult your computer's manual for details.")}
+                        </Text>
+                    </TextContent>
                 </EmptyStateBody>
                 <EmptyStatePrimary>
                     <Button id="ignore-hw-virtualization-disabled-btn" variant="secondary" onClick={() => setEmptyStateIgnored(true)}>{_("Ignore")}</Button>
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -122,18 +122,18 @@ export const App = () => {
             <EmptyState>
                 <EmptyStateIcon icon={WrenchIcon} />
                 <Title headingLevel="h4" size="lg">
-                    Virtualization support is not available
+                    {_("Virtualization support is not available")}
                 </Title>
                 <EmptyStateBody>
-                    <Text>Enable virtualization support in BIOS/EFI settings.</Text>
+                    <Text>{_("Enable virtualization support in BIOS/EFI settings.")}</Text>
                     <Text>
-                        Instructions vary per manufacturer, but usually involves pressing a hot key such as ESC, F1, F12, or Del during boot.
+                        {_(`Instructions vary per manufacturer, but usually involves pressing a hot key such as ESC, F1, F12, or Del during boot.
                         In the settings screen, look for "virtualization", "VM", "VMX', "SVM", "VTX", "VTD". It will be different on every computer.
-                        Enable any of these options. Consult your computer's manual for details.
+                        Enable any of these options. Consult your computer's manual for details.`)}
                     </Text>
                 </EmptyStateBody>
                 <EmptyStatePrimary>
-                    <Button variant="link" onClick={() => setEmptyStateIgnored(true)}>Ignore</Button>
+                    <Button variant="link" onClick={() => setEmptyStateIgnored(true)}>{_("Ignore")}</Button>
                 </EmptyStatePrimary>
             </EmptyState>
         );
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -139,7 +139,7 @@ export const App = () => {
                     </Text>
                 </EmptyStateBody>
                 <EmptyStatePrimary>
-                    <Button id="ignore-hw-virtualization-disabled-btn" variant="link" onClick={() => setEmptyStateIgnored(true)}>{_("Ignore")}</Button>
+                    <Button id="ignore-hw-virtualization-disabled-btn" variant="secondary" onClick={() => setEmptyStateIgnored(true)}>{_("Ignore")}</Button>
                 </EmptyStatePrimary>
             </EmptyState>
         );
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -135,9 +135,7 @@ export const App = () => {
                 <EmptyStateBody>
                     <Text>{_("Enable virtualization support in BIOS/EFI settings.")}</Text>
                     <Text>
-                        {_(`Changing BIOS/EFI settings is specific to each manufacturer. It involves pressing a hotkey during boot (ESC, F1, F12, Del).
-                        Enable a setting called "virtualization", "VM", "VMX', "SVM", "VTX", "VTD".
-                        Consult your computer's manual for details.`)}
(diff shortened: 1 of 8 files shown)


src/components/vms/hostvmslist.jsx
```diff
@@ -42,6 +42,7 @@ import { AggregateStatusCards } from "../aggregateStatusCards.jsx";
 import store from "../../store.js";
 
 import "./hostvmslist.scss";
+import { Alert, List, ListComponent, ListItem, OrderType } from '@patternfly/react-core';
```
KKoukiou, line 45: please move a few lines up and change to use full path import (@patternfly/react-core/dist/esm/components .*), as we import other Patternfly components. There is a reason for that, check the git blame to figure out why if you want :)

jelly (changes requested):
Thanks for the PR, I didn't test it yet but here are some general improvements which can be made.

src/components/vms/hostvmslist.jsx
```diff
@@ -126,6 +141,13 @@ const HostVmsList = ({ vms, config, ui, storagePools, actions, networks, onAddEr
             <Page>
                 <PageSection>
                     <Gallery className="ct-cards-grid" hasGutter>
+                        {!virtualizationEnabled && <Alert variant="warning" title="Warning: Virtualization is disabled in the firmware" component="h5">
```
KKoukiou, line 144: replace title with: _("Hardware virtualization extensions are disabled in BIOS")

I am not sure if we want to add the steps for fixing this inline in cockpit. @garrett @martinpitt what do you think? 
It's quite of an edge scenario, as all new servers will come with virtualization enabled. And it's a lot of text to translate.

src/components/vms/hostvmslist.jsx
```diff
@@ -63,9 +64,23 @@ const _ = cockpit.gettext;
...
 const HostVmsList = ({ vms, config, ui, storagePools, actions, networks, onAddErrorNotification }) => {
+    const [virtualizationEnabled, setVirtualizationEnabled] = useState(true);
     const [statusSelected, setStatusSelected] = useState({ value: _("All"), toString: function() { return this.value } });
     const [currentTextFilter, setCurrentTextFilter] = useState("");
     const [statusIsExpanded, setStatusIsExpanded] = useState(false);
+
+    useEffect(() => {
+        async function checkVirtualization() {
+            try {
+                await cockpit.script("grep -E 'vmx|svm|0xc0f' --color=auto /proc/cpuinfo");
```
KKoukiou, line 75: `grep -E "vmx|svm" /proc/cpuinfo` is enough. Color just highlights the text when users read it.

However this will not trow an exception, if will just return an empty ouput. 

So you need something like: 
````
cockpit.script("grep -E 'vmx|svm'  /proc/cpuinfo").then(output => setVirtualiationEnabled(!!output.trim()))
```

subhoghoshX: Thanks for looking into it.

It throws an error at least in my case.

Also in grep manual (`man grep`):

image

I think it's better to not rely on exit codes or whether `grep` return an empty string. It might not even be consistent across systems.

Using `await cockpit.file("/proc/cpuinfo").read()` instead, suggested by @martinpitt, seems better to me.

What do you think?

KKoukiou: martins suggestion looks good!

src/components/vms/hostvmslist.jsx
subhoghoshX: Thanks for the feedbacks.

Linking it to the FAQ sounds better to me because it avoids repeating instructions already present in the FAQ.

I tested it by rebooting and toggling the settings in BIOS a few times. It was not ideal. I'll try `rmmod kvm_intel` . Thanks.

src/components/vms/hostvmslist.jsx
```diff
@@ -126,6 +141,13 @@ const HostVmsList = ({ vms, config, ui, storagePools, actions, networks, onAddEr
             <Page>
                 <PageSection>
                     <Gallery className="ct-cards-grid" hasGutter>
+                        {!virtualizationEnabled && <Alert variant="warning" title="Warning: Virtualization is disabled in the firmware" component="h5">
+                            <List component={ListComponent.ol} type={OrderType.number}>
+                                <ListItem>Restart the computer running Cockpit and enter the BIOS/EFI settings by pressing the appropriate key.</ListItem>
+                                <ListItem>In the BIOS / EFI settings, enable virtualization and save changes before exiting.</ListItem>
+                                <ListItem>Finally, restart the computer and try starting a VM in Cockpit.</ListItem>
```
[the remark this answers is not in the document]
subhoghoshX: Used `_("")`.

src/components/vms/hostvmslist.jsx
```diff
@@ -63,9 +64,23 @@ const _ = cockpit.gettext;
...
     const [currentTextFilter, setCurrentTextFilter] = useState("");
     const [statusIsExpanded, setStatusIsExpanded] = useState(false);
+
+    useEffect(() => {
+        async function checkVirtualization() {
+            try {
+                await cockpit.script("grep -E 'vmx|svm|0xc0f' --color=auto /proc/cpuinfo");
+            } catch (e) {
+                setVirtualizationEnabled(false);
+            }
```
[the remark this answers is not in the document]
subhoghoshX: Used `await cockpit.file("/proc/cpuinfo").read()`.

@jelly I found 0xc0f to be for ARM processors.

subhoghoshX:
Now it looks like this:

image

src/components/vms/hostvmslist.jsx
subhoghoshX: Implemented the suggested changes.

@martinpitt `rmmod kvm_intel` doesn't seem to work. At least, it doesn't have an immediate effect.

src/components/vms/hostvmslist.jsx
subhoghoshX: Got it.

I wonder how reliable the bundler's tree shaking capabilities are.

src/components/vms/hostvmslist.jsx
jelly: `rmmod kvm_intel` would't help as the kernel populates `/proc/cpuinfo` and `kvm_intel` is just a way to use the virtualisation extensions of the CPU, it could as well be Xen or Virtualbox.

src/components/vms/hostvmslist.jsx
```diff
@@ -126,6 +140,14 @@ const HostVmsList = ({ vms, config, ui, storagePools, actions, networks, onAddEr
             <Page>
                 <PageSection>
                     <Gallery className="ct-cards-grid" hasGutter>
+                        {!virtualizationEnabled &&
+                        (<Alert variant="warning" title={_("Hardware virtualization extensions are disabled in BIOS")} component="h5">
+                            <Text>
```
KKoukiou, line 145: Garrett can you help a bit with wording here?

src/components/vms/hostvmslist.jsx
KKoukiou: This is the exact reason we moved to using tihs import style. Because importing from react-core would leave some imported but unused CSS in place.

martinpitt (changes requested):
Thanks!

src/components/vms/hostvmslist.jsx
garrett: Something like this?

---

**Virtualization support is not available**

Enable virtualization support in BIOS/EFI settings.

Instructions vary per manufacturer, but usually involves pressing a hot key such as ESC, F1, F12, or Del during boot. In the settings screen, look for "virtualization", "VM", "VMX', "SVM", "VTX", "VTD". It will be different on every computer. Enable any of these options. Consult your computer's manual for details.

garrett: However, I thought this would be a blocker for using Cockpit-Machines, so it would be an empty state pattern, not an alert at the top of the page?

I understand if you do have already-installed VMs if you're doing development, but someone not doing development and using Cockpit-Machines won't have this experience. If virtualization support is off in their BIOS/EFI settings, then they wouldn't have been able to create and install VMs. Therefore, it really blocks functionality.

garrett: Oh, I suggested an alert at the top of the page on the issue. :sweat_smile: https://github.com/cockpit-project/cockpit-machines/issues/891#issuecomment-1353364015

> Since a machine may have multiple VMs set up (due to the toggle changing, which happens during BIOS/EFI updates sometimes, or multiple attempts at installing VMs), I think the right place might be to have a big error at the top of the page (in an alert or banner) and still show the VMs otherwise. It should provide details about how to fix the error (rebooting, hitting a key, changing a value).

An alternate approach is to have an empty state, so it's in your face and have a bypass button, such as "Ignore" to be able to get to the machines to delete it.

WDYT?

garrett (changes requested):
I suggested test @ https://github.com/cockpit-project/cockpit-machines/pull/980#discussion_r1144438122

(Happy to have feedback on that.)

And I suggested changing this to an empty state pattern instead of an alert, despite what I said previously. https://github.com/cockpit-project/cockpit-machines/pull/980#discussion_r1144444026

I don't think most people would have made VMs prior, especially if they come across this warning first (and fix things)... Unless it's possible to actually use Cockpit-Machines without VM extensions (but then it would be exceedingly slow, right?)

If we're going to keep this as an alert (if Cockpit-Machines is actually useful without VM being enabled in the BIOS/EFI), it would be better to do so with an inline style, as it's technically an inline alert that's embedded in the page, not an alert that's popping up above something (and then going away). Making it inline is not only more proper, but it would also draw more attention to it due to the background color. It would need to have the card shadow around it to be consistent, however. This may require some custom CSS, such as adding a class to the element or using `--pf-c-card--BoxShadow: var(--pf-global--BoxShadow--sm);` (which is what `pf-c-card` uses).

But I think in almost every typical case, the empty state pattern (with a "Dismiss" or "Ignore" action) would probably be better here.

test/check-machines-virtualization
```diff
@@ -0,0 +1,70 @@
...
+
+
+@nondestructive
+class TestMachinesVirtualization(VirtualMachinesCase):
+
+    def testVirtualizationDisabled(self):
+        b = self.browser
+        m = self.machine
+
+        m.execute("touch /tmp/cpuinfo.txt && mount --bind /tmp/cpuinfo.txt /proc/cpuinfo")
```
jelly, line 39: Our tests do not run in a guaranteed order so this tests mount's something and the test below umounts which will only work if they are run after each other.

We should not rely on this behaviour as it's random. Our test library has a special `self.addCleanup` function which can call umount after the test is done.

So please add an `self.addCleanup(m.execute, "umount /proc/cpuinfo || true")`  the `|| true` is required in case the test fails to mount (should never happen imo)

jelly (changes requested):

subhoghoshX:
> All packit tests seem to fail, can you retry to rebase your branch and re-push?

It appears the tests failed again :(

test/check-machines-virtualization
```diff
@@ -0,0 +1,68 @@
...
+TEST_DIR = os.path.dirname(__file__)
+sys.path.append(os.path.join(TEST_DIR, "common"))
+sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
+
+from machineslib import VirtualMachinesCase  # noqa
+from testlib import nondestructive, test_main, wait  # noqa
+
+
+@nondestructive
+class TestMachinesVirtualization(VirtualMachinesCase):
```
KKoukiou, line 33: If we keep this seperate file for this check, we can also move here 'testLibvirt from test/check-machines-lifecycle'
as it's also doing checks for showing empty states for different reason.

Can be also followup.

src/app.jsx
```diff
@@ -101,7 +106,37 @@ export const App = () => {
...
+
+        checkVirtualization();
+    }, []);
+
+    if (!virtualizationEnabled && !emptyStateIgnored) {
+        return (
+            <EmptyState>
+                <EmptyStateIcon icon={WrenchIcon} />
+                <Title headingLevel="h4" size="lg">
+                    {_("Virtualization support is not available")}
```
KKoukiou, line 124: 'Hardware virtualization is disabled' is more correct. is not available sounds like you can't undo it. THis was the suggestion from Martin and i like it.

src/app.jsx
```diff
@@ -101,7 +106,37 @@ export const App = () => {
...
+                </Title>
+                <EmptyStateBody>
+                    <Text>{_("Enable virtualization support in BIOS/EFI settings.")}</Text>
+                    <Text>
+                        {_(`Instructions vary per manufacturer, but usually involves pressing a hot key such as ESC, F1, F12, or Del during boot.
```
KKoukiou, line 129: The ` will not let the string be translated. It also works with ".

If you need to use nested ", then you need to escape these like \".

KKoukiou, line 129: @garrett @martinpitt I feel that the suggestion text starting from 'Instructions' is too wordy, causing a lot of work for translations without even giving a very clear guidance.
I would suggest to skip it and just stick with the 'Enable virtualization support...`

src/app.jsx
```diff
@@ -101,7 +106,37 @@ export const App = () => {
...
+                        In the settings screen, look for "virtualization", "VM", "VMX', "SVM", "VTX", "VTD". It will be different on every computer.
+                        Enable any of these options. Consult your computer's manual for details.`)}
+                    </Text>
+                </EmptyStateBody>
+                <EmptyStatePrimary>
+                    <Button variant="link" onClick={() => setEmptyStateIgnored(true)}>{_("Ignore")}</Button>
```
KKoukiou, line 135: Since we don't use any cookie and we store that in react state that 'ignore' wil last only till the user refreshes the page. I think it's fine as first pass, but we probably want to either use a cookie for remembering the option or reconsider the design and not hide the whole app behind this empty state. But fine for follow up.

test/check-machines-virtualization
```diff
@@ -0,0 +1,68 @@
...
+        b.wait_visible("#card-pf-storage-pools")
+        b.wait_visible("#card-pf-networks")
+        b.wait_visible("#virtual-machines-listing")
+
+        b.wait_visible(".pf-c-page .pf-c-card__title:contains('Virtual machines')")
+
+
+    def testVirtualizationEnabled(self):
+        b = self.browser
+
```
KKoukiou, line 57: No need for this test, all our other tests do indirectly test this :)

KKoukiou (changes requested):

src/app.jsx
subhoghoshX: Yes, it looks better. Changed it.

src/app.jsx
subhoghoshX: We could also use localStorage.

test/check-machines-virtualization
subhoghoshX: Okay :)

src/app.jsx
garrett: > Hardware virtualization is disabled

Good call. I like that too.

src/app.jsx
```diff
@@ -101,7 +106,37 @@ export const App = () => {
...
+                <EmptyStateIcon icon={WrenchIcon} />
+                <Title headingLevel="h4" size="lg">
+                    {_("Virtualization support is not available")}
+                </Title>
+                <EmptyStateBody>
+                    <Text>{_("Enable virtualization support in BIOS/EFI settings.")}</Text>
+                    <Text>
+                        {_(`Instructions vary per manufacturer, but usually involves pressing a hot key such as ESC, F1, F12, or Del during boot.
+                        In the settings screen, look for "virtualization", "VM", "VMX', "SVM", "VTX", "VTD". It will be different on every computer.
+                        Enable any of these options. Consult your computer's manual for details.`)}
```
garrett, line 131:
```suggestion
                        {_(`Changing BIOS/EFI settings is specific to each manufacturer. It involves pressing a hotkey during boot (ESC, F1, F12, Del).
                        Enable a setting called "virtualization", "VM", "VMX', "SVM", "VTX", "VTD".
                        Consult your computer's manual for details.`)}
```

I've tried to trim it down a bit. How's this?

garrett:
I'm not sure how to test this without rebooting and messing with my EFI settings (and then rebooting back). Is there a better way?

Also, could you post an updated screenshot and/or small screencast video (audio wouldn't be necessary)?

jelly:
> I'm not sure how to test this without rebooting and messing with my EFI settings (and then rebooting back). Is there a better way?
> 
> Also, could you post an updated screenshot and/or small screencast video (audio wouldn't be necessary)?

This PR should have a test which mounts over /proc/cpuinfo and thus hiding the virtualization extensions. You should be able to do the same locally.

```
touch /tmp/cpuinfo.txt && mount --bind /tmp/cpuinfo.txt /proc/cpuinfo
umount /proc/cpuinfo
```

src/app.jsx
subhoghoshX: Thanks for the feedback. I implemented saving the preference in localStorage.

KKoukiou (commented):
Unfortunately packit test environment seems to not have virtualization enabled in bios? All test fail on the first page load where now this new screen appears. I suggest to just click ignore for all tests if the selector exists:

https://github.com/cockpit-project/cockpit-machines/blob/main/test/machineslib.py#L28

So write something like;

if b.is_visible("h4:contains('Hardware virtualization is disabled')"):
    b.click('button:contains(Ignore)')

garrett (changes requested):
Three things:

1. Ignore should be a button. I suggest a secondary button (even though it's the primary action... I'd suggest the implied actual primary action is to fix the BIOS, making this secondary).
   - We should _never_ use links or link styled buttons, _except_ when they go somewhere (then it should be a link, not a button), are to downplay importance in an otherwise cluttered interface (such as in table rows, and they should be a link-styled button, unless they go somewhere as they'd be links instead), or are specifically the "Cancel" (link-styled) button on modals. ("Close" and other buttons should still look like buttons.) Everything else should be a button that looks like a button (either primary, secondary, warning, or danger).

2. For some odd reason, there's not enough vertical spacing between paragraphs when there are multiple paragraphs in the empty state. Are we using it incorrectly here, or is this a PF4 bug that we should workaround? (Perhaps something like `.pf-c-empty-state p + p { margin-top: var(--pf-global--spacer--sm) }`?

3. The text is far too wide.

image

This text should be narrower... ideally a around 60 characters or so, except on mobile. We can do that without breakpoints.

We'd want something like this:

```scss
.pf-c-empty-state.ct--empty-state--constrain {
  // Use a max width of 60 0-characters across and let it size for mobile too
   --pf-c-empty-state__content--MaxWidth: min(60ch, 100%);
}
```

(I'm not sure why there isn't a constraint by default, and even mystified that it doesn't properly auto-center by default. :shrug:)

And we'd want to add that at a top-level for all of Cockpit in a file that other pages can use and other projects import, probably in `lib/` somewhere.

The reason why I'm saying we should make this a custom ct-* class is that https://github.com/cockpit-project/cockpit/pull/18592 is struggling with the same thing right now. We probably want this across all the empty states.

We _could_ even use PF overrides and make this the default for Cockpit. If that's the case, we'd want to drop the class:

```scss
.pf-c-empty-state {
  // Use a max width of 60 0-characters across and let it size for mobile too
   --pf-c-empty-state__content--MaxWidth: min(60ch, 100%);
}
```

With this bit of CSS, it'd look like this:

image

We probably always want to constrain the width (when there's width that's greater than the width shown above, so I'm in favor of it if @KKoukiou agrees with that too.

If that's the case, we might want to make it a separate PR and rebase this and the other PR on top of it. Or keep it as a separate commit and cherry-pick it from one PR to the other (and have it first in each PRs... so whatever lands first wins and it would fall out from the other, yet both PRs would work as expected... practically the same thing, really).

When all the changes I outlined above are implemented, it would look like this:

image

Additional point:

4. We might also want to consider changing the icon from a wrench as well. Perhaps `pf-icon-virtual-machine` instead? For reference, it looks like this:

pf-icon-virtual-machine

KKoukiou (approved):

KKoukiou (approved):

garrett:
Please update the screenshot; it's still using the old wrench icon and doesn't have a summary yet. Thanks!

subhoghoshX:
> Please update the screenshot; it's still using the old wrench icon and doesn't have a summary yet. Thanks!

Oops, my bad. I missed that point. I hope I'm not too late to make the change.

garrett:
@subhoghoshX: No worries. It's easy to miss things like this sometimes. We've all done it from time to time. It's why we plan to have a day or two to get everything in and have follow-up quick fixes if we notice anything small like this.

PR #984, opened 2023-03-20
Use id instead of shortId to workaround osinfo-db issue

KKoukiou:
https://gitlab.com/libosinfo/osinfo-db/-/commit/61c86b242bf3bb8be18f8970670bad3dbadee952

This can be reverted once the above fix is released.


martinpitt (approved):
This is for `key=` only, I see. That seems fine!

PR #985, opened 2023-03-21
test: General checks for storage pool deletion

yunmingyang:


test/check-machines-storage-pools
```diff
@@ -695,6 +695,59 @@ class TestMachinesStoragePools(VirtualMachinesCase):
             remove=True,
         ).execute()
 
+    def testGeneralCheckForPoolsDeletion(self):
```
KKoukiou, line 698: There is already a test for pools deletion. It's called testStoragePoolsDeletion. So introducing this with this name is confusing.

I see this test does not actually perform deletion, it's always canceling or erroring. So I guess I can accept it as a seperate test.

test/check-machines-storage-pools
```diff
@@ -695,6 +695,59 @@ class TestMachinesStoragePools(VirtualMachinesCase):
...
+        b.click(".pf-c-backdrop footer button:contains(Cancel)")
+        b.wait_not_present(".pf-c-backdrop")
+
+        # Check "X" button of the dialog
+        b.click("#pool-images-system-action-kebab button")
+        b.click("#pool-images-system-action-kebab a:contains(Delete)")
+        b.click(".pf-c-backdrop button[aria-label=Close]")
+        b.wait_not_present(".pf-c-backdrop")
+
+        # Check that the error shown after the title, not the footer
```
KKoukiou, line 730: How about adding instead a pixel test for figuring out where the error is shown??

KKoukiou: If we add the pixel test you can also remove https://github.com/cockpit-project/cockpit-machines/pull/985/commits/0a482e73086e8705184f8143cd3c48493ac907ec as it's covered.

test/check-machines-storage-pools
```diff
@@ -695,6 +695,59 @@ class TestMachinesStoragePools(VirtualMachinesCase):
...
+        b.click("#card-pf-storage-pools .pf-c-card__header button")
+        b.wait_visible("#storage-pools-listing")
+
+        m.execute(f'chattr +a {args["image"]}')
+        self.addCleanup(m.execute, f'chattr -a {args["image"]}')
+        b.click("#pool-images-system-action-kebab button")
+        b.click("#pool-images-system-action-kebab a:contains(Delete)")
+        b.set_checked("#storage-pool-delete-volumes", True)
+        b.click(".pf-c-backdrop footer button[aria-disabled=false]:contains(Delete)")
+        b.wait_in_text(".pf-c-backdrop header + div .pf-c-alert__description", "cannot unlink file")
```
KKoukiou, line 748: CHecking that the error appears in the dialog is already done in https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-storage-pools#L842 and in specific in this test https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-storage-pools#L954

If there is some extra check missing please extend that one instead.

test/check-machines-storage-pools
KKoukiou, line 698: The biggest part of this test can be removed as fix for my next comment.. The remaining code can be merged with the existing test for pools deletion.

KKoukiou (changes requested):

test/check-machines-storage-pools
yunmingyang: I think there are some different between them, as this check will check the other files which is not changed by chattr  will be deleted, although there is an error shown on the dialog and the dialog is not closed. Or do you think this behavior is not our expected?

test/check-machines-storage-pools
yunmingyang: OK, I add a pixel test in the new push, but it seems that I doesn't have the permission of cockpit-project/pixel-test-reference.git.

test/check-machines-storage-pools
```diff
@@ -978,6 +978,56 @@ class TestMachinesStoragePools(VirtualMachinesCase):
...
+        b.click("#card-pf-storage-pools .pf-c-card__header button")
+        b.wait_visible("#storage-pools-listing")
+
+        m.execute(f'chattr +a {args["image"]}')
+        self.addCleanup(m.execute, f'chattr -a {args["image"]}')
+        b.click("#pool-images-system-action-kebab button")
+        b.click("#pool-images-system-action-kebab a:contains(Delete)")
+        b.set_checked("#storage-pool-delete-volumes", True)
+        b.click(".pf-c-backdrop footer button[aria-disabled=false]:contains(Delete)")
+        b.assert_pixels("div.pf-c-modal-box.pf-m-align-top.pf-m-sm", "storage-pool-deletion-error", skip_layouts=["rtl"])
```
KKoukiou, line 1028: Can you remove the skip layouts? Any specific reason you put it or just wrong copy?

KKoukiou (commented):
Just the skip_layouts commnet. I can update the pixel tests references.

test/check-machines-storage-pools
KKoukiou: But this is not a discussion for the test now. :)

test/check-machines-storage-pools
yunmingyang: Sure I can, the reason is that I am not sure where we will use the rtl layout. I will remove it.

test/check-machines-storage-pools
yunmingyang: Maybe we need a issue to track this behavior

KKoukiou (approved):

KKoukiou (changes requested):
@yunmingyang The new test lines fail on some distros. Can you please take a look? Only the fedora-37 pixel tests faluresa are irrelevant. This can be fixed by alphabetically sorting the list of the displayed 'to be deleted' volumes names.

yunmingyang:
> The new test lines fail on some distros. Can you please take a look? Only the fedora-37 pixel tests faluresa are irrelevant. This can be fixed by alphabetically sorting the list of the displayed 'to be deleted' volumes names.

Sure, will check it later

yunmingyang:
It seems that the test failed as the too quick check, let's see whether the wait function could resolve this.

yunmingyang:
@KKoukiou Ah, sorry, I forgot to pull the reference related in my force-pushed. Could you re-push them? Many thanks

KKoukiou (dismissed):

KKoukiou (approved):

KKoukiou:
This is failing because of:

Unused reference image TestMachinesStoragePools-testStoragePoolsDeletion-storage-pool-deletion-error-dark-pixels.png
Unused reference image TestMachinesStoragePools-testStoragePoolsDeletion-storage-pool-deletion-error-medium-pixels.png
Unused reference image TestMachinesStoragePools-testStoragePoolsDeletion-storage-pool-deletion-error-mobile-pixels.png
Unused reference image TestMachinesStoragePools-testStoragePoolsDeletion-storage-pool-deletion-error-rtl-pixels.png
Unused reference image TestMachinesStoragePools-testStoragePoolsDeletion-storage-pool-deletion-error-pixels.png

This is not your fault, it's just misdetecting unused reference images. Some wrong prefix string matching for test names in test/common/testlib.py

Will fix it there and retrigger.

Issue #536, opened 2022-01-21
non-privileged user is able to try to create a new virtual network


jelly:
The normal admin user can't create a new virtual network but is shown the option resulting in this issue:

![image](https://user-images.githubusercontent.com/67428/150564370-bad866ff-8338-4d97-9011-499e9d4617b9.png)

It would be better if cockpit-machines didn't show the create option if the user is not allowed to create it.

PR #992, opened 2023-03-27
Non-privileged user should not create a new virtual network

dCO2:
Attempt fix for https://github.com/cockpit-project/cockpit-machines/issues/536

### DIFF
--- a/src/components/networks/networkList.jsx
+++ b/src/components/networks/networkList.jsx
@@ -63,7 +63,7 @@ export class NetworkList extends React.Component {
                                     <Text component={TextVariants.h2}>{_("Networks")}</Text>
                                 </CardTitle>
                                 <CardActions>
-                                    { unlocked ? <CreateNetworkAction /> : <></> }
+                                    { unlocked && <CreateNetworkAction /> }
                                 </CardActions>
                             </CardHeader>
                             <CardBody className="contains-list">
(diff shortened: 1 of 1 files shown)


src/components/networks/networkList.jsx
```diff
@@ -61,7 +63,7 @@ export class NetworkList extends React.Component {
                                     <Text component={TextVariants.h2}>{_("Networks")}</Text>
                                 </CardTitle>
                                 <CardActions>
-                                    <CreateNetworkAction />
+                                    { unlocked ? <CreateNetworkAction /> : <></> }
```
skobyda, line 66: In javascript/react, instead of ternary operator and filling the other option with empty `<></>`, you can use also `&& condition `, which I think is what we should use here
```suggestion
                                    { unlocked && <CreateNetworkAction /> }
```
You can image it as `if` condition, where if unclocked is `true`, <CreateNetworkAction /> gets rendered

More here: https://react.dev/learn/conditional-rendering#logical-and-operator-

skobyda (changes requested):
Thanks for fix :)

src/components/networks/networkList.jsx
dCO2: oh yes that is true. thanks for the suggestion

skobyda (approved):
LGTM. I just noticed that this is split into 2 separate commits, even though it should be 1 commit, since it's 1 single functionality.
I'll squash them upon merge

PR #993, opened 2023-03-27
Move from webpack to esbuild bundler

KKoukiou:
Let's merge this first https://github.com/cockpit-project/cockpit-machines/pull/995 to clear out pixel tests.


martinpitt:
TF fails with

    make: *** No rule to make target 'webpack.config.js', needed by 'dist/manifest.json'.  Stop.

Obvious fix in Makefile, but @KKoukiou please let me know if you have trouble with this.

martinpitt:
Retrying [pybridge scenario](https://cockpit-logs.us-east-1.linodeobjects.com/pull-993-20230328-171452-a2315904-fedora-37-pybridge/log.html#43-2) for the `TestMachinesLifecycle.testDelete` failure.

TF is broken since yesterday, it completely misses artifact installation. I'll sort that out with the TF team today, and until then we can ignore it.

build.js
```diff
@@ -0,0 +1,105 @@
...
+                { from: ['./src/manifest.json'], to: ['./manifest.json'] },
+                { from: ['./src/index.html'], to: ['./index.html'] },
+            ]
+        }),
+        replace({
+            include: /ip.js/,
+            values: {
+                "var os": 'var os = {}; // HACK: os is not really used in our used functions from ip.js',
+            }
+        }),
```
martinpitt, line 65: We don't have such a hack in webpack.config.js. This could use a comment why it's necessary, possibly a link to a bug report?

build.js
```diff
@@ -0,0 +1,105 @@
...
+    },
+    minify: production,
+    nodePaths,
+    outdir,
+    target: ['es2020'],
+    plugins: [
+        cleanPlugin(),
+        ...lint ? [
+            stylelintPlugin({ filter: new RegExp(cwd + '\/src\/.*\.(css?|scss?)$') }),
+            eslintPlugin({ filter: new RegExp(cwd + '\/src\/.*\.(jsx?|js?)$') })
```
martinpitt, line 50: That filter value is the default.

build.js
```diff
@@ -0,0 +1,105 @@
...
+        ".py": "text",
+    },
+    minify: production,
+    nodePaths,
+    outdir,
+    target: ['es2020'],
+    plugins: [
+        cleanPlugin(),
+        ...lint ? [
+            stylelintPlugin({ filter: new RegExp(cwd + '\/src\/.*\.(css?|scss?)$') }),
```
martinpitt, line 49: Let's make that the default in cockpit's pkg/lib/ as well, to be consistent with eslintPlugin, and avoid repetition in all projects. (not a blocker for this PR, of course)

build.js
```diff
@@ -0,0 +1,105 @@
...
+try {
+    await context.rebuild();
+} catch (e) {
+    if (!watchMode)
+        process.exit(1);
+    // ignore errors in watch mode
+}
+
+if (watchMode) {
+    // Attention: this does not watch subdirectories -- if you ever introduce one, need to set up one watch per subdir
```
martinpitt, line 90: That was true for c-podman, but isn't true for machines -- we do have several subdirs in src/. I.e. we need the "full strength" recursive watching from cockpit's build.js. Do you want me to do that?

martinpitt (changes requested):
Thanks! I know that this is annoying (node and pixel updates). If you prefer, I would be okay with landing this now, and fixing watch (and the mystery ip.js hack comment) mode immediately aftewards as a follow-up.

build.js
KKoukiou: We actually had this hack in webpack config but indeed it had a comment, https://github.com/cockpit-project/cockpit-machines/blob/main/webpack.config.js#L59

Adding the comment back

build.js
KKoukiou: Not really. Without he cwd prefix it also matched files in node_modules that have a src/ directory.

build.js
KKoukiou: It's fine - will updat ethe PR.

build.js
martinpitt: Eww, yes -- I didn't realize that is a substring match and an absolute path. So let's fix https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/esbuild-eslint-plugin.js#L9 as well (in a follow-up).

KKoukiou:
Doh protection checks now fail - some new module available, repushing.

martinpitt (approved):
Purrfect! :joy_cat: Epharisto!

Are you in the mood for starter-kit, too, or should I do this? (I'll be glad to)

KKoukiou:
> Purrfect! joy_cat Epharisto!
> 
> Are you in the mood for starter-kit, too, or should I do this? (I'll be glad to)

I can do it as part of piloting.

martinpitt:
[This failure](https://artifacts.dev.testing-farm.io/a34e91a9-9c68-4102-80fa-f1c3edddba39/) is another piece of fallout from the TF update. I relayed it to the TF team.

PR #996, opened 2023-03-29
When failing to delete storage after vm deletion show all related messages

KKoukiou:
Previously by using Promise.all and not Promise.allSettled we were rejecting once the first rejection happened. That means we were not attempting to delete the rest of the storage files if one failed.

Also remove some unused properties when calling the domainDelete function.


skobyda (approved):
LGTM

skobyda:
Re-trigerred tests, once they pass let's merge it

skobyda:
All green, thanks for this PR

PR #998, opened 2023-03-30
Show label when VM needs shutdown

skobyda:
- [x] TODO after merge, make issue based on https://github.com/cockpit-project/cockpit-machines/pull/998#discussion_r1155850577

Split from https://github.com/cockpit-project/cockpit-machines/pull/986 , based on design discussion there

# Machines: Show "changes pending" when configuration changes require a shutdown to take effect

The VM list and detail pages now display a "changes pending" status label to indicate configuration changes that require a shutdown to take effect. Clicking the label reveals the specific changes causing the inconsistency.

image


skobyda:
@garrett This PR is based on our discussion at https://github.com/cockpit-project/cockpit-machines/pull/986

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,168 @@
...
+    // Watchdog
+    if (needsShutdownWatchdog(vm))
+        devices.push(_("Watchdog"));
+
+    return devices;
+}
+
+export const NeedsShutdownTooltip = ({ iconId, tooltipId }) => {
+    return (
+        <Tooltip id={tooltipId} content={_("Changes will take effect after shutting down the VM")}>
```
KKoukiou, line 117: String repeated. Maybe let's make this string a constant and move it up?

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,168 @@
...
+            <PendingIcon color="orange" id={iconId} />
+        </Tooltip>
+    );
+};
+
+NeedsShutdownTooltip.propTypes = {
```
KKoukiou, line 123: Honestly we stopped adding proptypes I think but fine by me. We need to sync about it though.

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,168 @@
...
+};
+
+export const VmNeedsShutdown = ({ vm }) => {
+    const devices = getDevicesRequiringShutdown(vm);
+
+    if (devices.length === 0)
+        return;
+
+    const body = (
+        <>
```
KKoukiou, line 142: Unneeded '<>' wrapper.

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,168 @@
...
+            <List isPlain isBordered>
+                { devices.map((device, idx) => {
+                    return (
+                        <ListItem key={idx}>
```
KKoukiou, line 146: Why don't you reuse the 'device' variable for key? Also it's suboptimal to use only indexes for keys.

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,168 @@
...
+                            {cockpit.format(_("$0 configuration was changed"), device)}
+                        </ListItem>
+                    );
+                }) }
+            </List>
+        </>
+    );
+
+    return (
+        <Popover aria-label="vm-needs-shutdown"
```
KKoukiou, line 156: Aria labels should be translated strings.

src/components/common/needsShutdown.jsx
skobyda: I just renamed this file and renamed the component, the proptypes were already there, but sure I can remove them

KKoukiou: Up to you - I don't mind

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,161 @@
...
+    if (devices.length === 0)
+        return;
+
+    const body = (
+        <>
+            <List isPlain isBordered>
+                { devices.map(device => {
+                    return (
+                        <ListItem key={device}>
+                            {cockpit.format(_("$0 configuration was changed"), device)}
```
KKoukiou, line 139: This is missing a period in the end, as the next sentence has one.

Also when more elements are there I find it very repetitive. Maybe let's remove the list, bring up the other string from the footer into the body and write something like:
```
<TextContent>
<Text>_("Changes to the domain's configuration only take effect after a complete shutdown and restart of the domain.")}
</Text>
<Text>cockpit.format(_("The are  pending changes for $0."), devices.join(", ")) </Text>
</TextContent>
```

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,161 @@
...
+        </>
+    );
+
+    const header = _("VM needs shutdown");
+    return (
+        <Popover aria-label={header}
+            alertSeverityVariant="warning"
+            headerContent={header}
+            headerIcon={<ExclamationTriangleIcon />}
+            headerComponent="h1"
```
KKoukiou, line 153: <h1> headings should be used for main heading of a page. This is definitely not a main heading. Why did you want to change whatever it uses by default?

src/components/common/needsShutdown.jsx
skobyda: Hey. That was initial idea, but the issue is that in https://github.com/cockpit-project/cockpit-machines/pull/986 I was to add an additional message, like "Watchdog configuration was changed, but the VM needs shutdown. Watchdog live edit was not possible: Not enough PCI slots available".

So @garrett suggested the list: https://github.com/cockpit-project/cockpit-machines/pull/986#pullrequestreview-1362435505 , and we can render additional information for each device (each list item).

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,161 @@
...
+                    );
+                }) }
+            </List>
+        </>
+    );
+
+    const header = _("VM needs shutdown");
+    return (
+        <Popover aria-label={header}
+            alertSeverityVariant="warning"
```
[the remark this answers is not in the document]
skobyda: Fixed.

src/components/common/needsShutdown.jsx
skobyda: Copied from patternfly example, didn't stop to think about it :D 
Fixed

src/components/common/needsShutdown.jsx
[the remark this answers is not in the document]
skobyda: Unfix.
I just realized we use orange icons (warning variant) everywhere for letting user know that VM needs reboo.
We need to be consistent. So do you think we should switch it everywhere else to info variant?

Screenshot from 2023-04-03 12-10-10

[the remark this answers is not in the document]
skobyda: Anyway, this is what it would look like as an info variant:
@garrett what do you think?

Screenshot from 2023-04-03 12-17-40
Screenshot from 2023-04-03 12-17-29

src/components/common/needsShutdown.jsx
```diff
@@ -0,0 +1,161 @@
...
+                        <ListItem key={device}>
+                            {cockpit.format(_("$0 configuration was changed"), device)}
+                        </ListItem>
+                    );
+                }) }
+            </List>
+        </>
+    );
+
+    const header = _("VM needs shutdown");
```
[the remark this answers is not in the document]
garrett: Yeah, agreed.

It should either be "Needs reboot" (as it's running already) or as @KKoukiou suggests "Updates pending" / "Pending updates" (although "updates" sounds like software updates, which might be problematic).

I guess we're trying to message that it's basically out of sync with the configuration.

src/components/common/needsShutdown.jsx
garrett: Agreed.

I wonder if we should make this a list of configuration items are out of sync?

"Some configuration changes only take effect after a fresh boot:"

And then a list.

src/components/common/needsShutdown.jsx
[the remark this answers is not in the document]
garrett: Agreed.

src/components/common/needsShutdown.jsx
garrett: Yep. You're not supposed to change the order of headings.

src/components/common/needsShutdown.jsx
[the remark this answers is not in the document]
skobyda: Ok, updated the PR to info variant

src/components/common/needsShutdown.jsx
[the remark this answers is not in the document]
garrett: "changes pending" / "pending changes"?

src/components/common/needsShutdown.jsx
skobyda: Patternfly's Popover uses by default --pf-c-popover__title--FontSize which is 1rem by default, and looks like this:
Screenshot from 2023-04-03 12-17-40

So I think we can just keep patterfly's default

src/components/common/needsShutdown.jsx
skobyda: @garrett Something like this:
Screenshot from 2023-04-03 13-08-55

skobyda: Hmm, there should probably be some padding between the list and the sentence above, updated it to this:
Screenshot from 2023-04-03 13-14-29

skobyda: Something like this:
Screenshot from 2023-04-03 13-17-55

garrett: We're already using blue for the action and "running" on the list. Perhaps it should be some other color.

Red: Error
Yellow/Orange: Warning
Green: Everything's good
Blue: Action
Grey: Actionless (stopped, etc.)

What's left?

https://www.patternfly.org/v4/guidelines/colors/#color-palette

Purple? Or reusing grey? (There's also cyan, but that's still blueish/greenish. And there's light green too. Purple might be too close to blue however.)

garrett: But, yeah, that's looking more correct.

We could even add a

"[Reboot now] Cancel" to the bottom and treat it like a specifically-placed dialog. (Cancel just closing it like the X). This would give someone more guidance as what to do.

Additionally, we need to change the rebooting action everywhere to do a shutdown and restart if the configuration state is dirty like this. It's basically the same thing from the user's point of view, but we'd handle the details to make sure the machine is put into the expected state.

skobyda:
Fixed mobile, this is how it looks like in mobile:

Screenshot from 2023-04-03 15-25-09

skobyda:
> > this is how it looks like in mobile
> 
> How does it look when the popup is open on mobile?

It moves the screen a bit so the popover fits into the screen. That's the basic behavior of all PF's Popovers we use:
Screenshot from 2023-04-03 16-55-28

garrett:
> It moves the focus of screen a bit so the popover fits into the screen. That's the basic behavior of all PF's Popovers we use:

That doesn't always work. Hence me asking. :wink: 

> So I fixed it with flex-wrap

But it isn't fixed. Where's the proper spacing? There's 0 at the left, which is wrong.

skobyda:
> But it isn't fixed. Where's the proper spacing? There's 0 at the left, which is wrong.

Screenshot from 2023-04-03 20-30-35

I just noticed that this is broken also on main (see screenshot above). Also the vm actions are inaccessible when vm name is too long. That's quite serious bug! So I made a fix in separate PR so it can land in next release in case this PR doesn't make it: https://github.com/cockpit-project/cockpit-machines/pull/1003

skobyda:
Ok, the mobile issues were split out into a [different PR](https://github.com/cockpit-project/cockpit-machines/pull/1003) since they are also present on main and not related to changes done here. Also the request for a Reboot button at the bottom of Popover was created to be fixed as followup: https://github.com/cockpit-project/cockpit-machines/issues/1004

I think that solves all of @garrett issues with this PR.

Also the tests are all green, @KKoukiou mind giving one more review?

src/components/common/needsShutdown.scss
```diff
@@ -0,0 +1,3 @@
+.configuration-changes-list {
```
KKoukiou, line 1: This should be better done by putting the two items in a Stack container with hasGutter. can be followup.

KKoukiou (dismissed):

skobyda:
Current UI:

Screenshot from 2023-04-03 14-58-43
Screenshot from 2023-04-03 14-58-30

garrett (changes requested):
Thanks again for splitting off the padding PR!

This PR is almost ready. There's one big oversight: The label doesn't look clickable at all. 

For example, the mouse cursor is an I beam instead of a pointer, and there's nothing else that happens when you hover:

Screenshot from 2023-04-04 14-06-28

It looks like PF labels look clickable when you add a link inside.

See https://www.patternfly.org/v4/components/label and mouse over the ones that are "{color} link"

The difference in React:

```jsx
    <Label color="blue">Blue</Label>
```

```jsx
    <Label color="blue" href="#filled">
      Blue link
    </Label>
```

The HTML and CSS difference are that it has `<a>` inside. It seems adding `href` on `<Label>` does that for you in React. As this isn't a link that goes anywhere, it should be `href="#"` so it doesn't try to load a page. If this were HTML, I might even suggest trying an `<a>` without `href` inside. (...As that's optional... but as it triggers the creation of the `<a>` element in React, I guess it's not in the PF widget?)

And there's too much space between labels in the label stack.

It's like this:

image

According to PF React examples and design docs, it should be spaced @ xs, like this:
image
(HTML examples seem to have extra spacing, but not as much as we have in this PR.)

---

Anyway, I think once these two small visual things are addressed, it should be good. Thanks! :+1:

garrett:
> Turns out we were doing bunch of custom padding, so I removed that and changed flex spacing to xs:

Awesome! Looks good in the screenshots.

I should switch branches and test out this PR again. :grin:

garrett (approved):
I thought I reviewed and approved this last night, but I don't see it on GitHub, so I just looked at it again today and am approving it now. :+1:

KKoukiou (changes requested):
Discussed on IRC> pixel test show regression on the label positioning in the table

skobyda:
Packit centos-stream-9 fails:
```
Error: GPG check FAILED

Copr build error: Mock build failed
```
But that's unrelated, so I guess let's merge it anyway?

src/components/common/stateIcon.scss
```diff
@@ -1,6 +1,4 @@
 .resource-state-text {
-  margin-right: 1rem;
```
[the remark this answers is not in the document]
KKoukiou: But anyway - leaving it up to you - since it's not regressing I am fine with that as well.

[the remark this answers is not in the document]
skobyda: Both of these are deliberate. Removing  `margin-right: 1rem;` otherwise items will be spaced far too apart from each other, which is opposite of the design mockup: https://github.com/cockpit-project/cockpit-machines/pull/998#pullrequestreview-1370865494
Screenshot from 2023-04-05 20-35-07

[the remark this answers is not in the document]
skobyda: Ahhh, now I noticed that without that margin-right, it will break state in vmDetailsPage.... Aaahh, css, I'll look at that more
Screenshot from 2023-04-05 20-44-55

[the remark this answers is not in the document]
skobyda: Ok, I changed the css a bit again. I cannot revert all the changes to the file, since we want to increase the spaces to xs as discussed in [mockup](https://github.com/cockpit-project/cockpit-machines/pull/998#pullrequestreview-1370865494)

KKoukiou (approved):

PR #1000, opened 2023-03-30
test: increase timeout when adding disk

KKoukiou:
Noticed it failed sometimes on CI with the spinner still on.


martinpitt (approved):
15 to 20 is enough? if that legitimately takes that long, then I'd bump it a little further even.

skobyda (dismissed):

martinpitt (approved):

Issue #778, opened 2022-07-26
request: adding categories


Larse99:
Hey!

I would like to have the ability to have categories within the web interface, where I can filter on each category. I run a couple of "labs" which i want to categorize. At the moment I use prefixes, which get messy.

Something like VMWare Workstation, would be great.

jelly:
I don't see anything in the libvirt API for adding categories to a machine. However it seems libvirt supports [custom metadata](https://libvirt.org/formatdomain.html#general-metadata) as another alternative would be using localstorage but that of course is not persistent.

1ras:
I was also surprised that libvirt has no concept of grouping, like e.g. [VirtualBox](https://www.virtualbox.org/manual/UserManual.html#gui-vmgroups). I'd like something like that too, but I also think it's a feature for libvirt.

KKoukiou:
We dont want to support something on top of what libvirt supports. This will introduce technical depth to this plugin out of it's initial goal and is something the team does not work to support long term.

PR #1001, opened 2023-03-31
adds VM (domain) title and description to UI

Britz:
The libvirt domain XML has the title and description element which is not represented in the UI. 
This commit adds, if available, the Title to the VM List in the form "<TITLE> (<NAME>)" and adds Title and description to the VM detail page including the option to modify it. 

* [ ] Blocked on new design mockups for the overview from @garrett

### DIFF
--- a/src/components/vm/overview/descriptionModal.jsx
+++ b/src/components/vm/overview/descriptionModal.jsx
@@ -1,0 +1,64 @@
+import React, { useState } from 'react';
+import cockpit from 'cockpit';
+import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
+import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form/index.js";
+import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput/index.js";
+import { Modal } from "@patternfly/react-core/dist/esm/components/Modal/index.js";
+import { useDialogs } from 'dialogs.jsx';
+
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { domainSetDescription } from '../../../libvirtApi/domain.js';
+
+const _ = cockpit.gettext;
+
+export const DescriptionModal = ({ vm }) => {
+    const Dialogs = useDialogs();
+    const [error, setError] = useState({});
+    const [newDescription, setDescription] = useState(vm.description);
+    const [isLoading, setIsLoading] = useState(false);
+
+    function save() {
+        setIsLoading(true);
+        domainSetDescription({
+            name: vm.name,
+            id: vm.id,
+            connectionName: vm.connectionName,
+            description: newDescription
+        }).then(Dialogs.close, exc => {
+            setIsLoading(false);
+            setError({ dialogError: _("Description could not be changed"), dialogErrorDetail: exc.message });
+        });
+    }
+
+    const defaultBody = (
+        <Form isHorizontal>
+            <FormGroup id="description-model-select-group" label={_("New description")}
+                fieldId="rename-dialog-new-description"
+            >
+                <TextInput id='rename-dialog-new-description'
+                    value={newDescription}
+                    onChange={setDescription} />
+            </FormGroup>
+        </Form>
+    );
+
+    return (
+        <Modal position="top" variant="small" isOpen onClose={Dialogs.close}
+            title={cockpit.format(_("Set description of $0"), vm.name)}
+            footer={
+                <>
+                    <Button variant='primary' isDisabled={isLoading} isLoading={isLoading} id="description-config-dialog-apply" onClick={save}>
+                        {_("Save")}
+                    </Button>
+                    <Button variant='link' onClick={Dialogs.close}>
+                        {_("Cancel")}
+                    </Button>
+                </>
+            }>
+            <>
+                {error && error.dialogError && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
+                {defaultBody}
+            </>
+        </Modal>
+    );
+};
--- a/src/components/vm/overview/descriptionModal.jsx
+++ b/src/components/vm/overview/descriptionModal.jsx
@@ -14,7 +14,7 @@ const _ = cockpit.gettext;
 export const DescriptionModal = ({ vm }) => {
     const Dialogs = useDialogs();
     const [error, setError] = useState({});
-    const [newDescription, setDescription] = useState(vm.description);
+    const [newDescription, setNewDescription] = useState(vm.description);
     const [isLoading, setIsLoading] = useState(false);
 
     function save() {
@@ -32,12 +32,12 @@ export const DescriptionModal = ({ vm }) => {
 
     const defaultBody = (
         <Form isHorizontal>
-            <FormGroup id="description-model-select-group" label={_("New description")}
-                fieldId="rename-dialog-new-description"
+            <FormGroup id="description-modal-select-group" label={_("New description")}
+                fieldId="set-description-dialog"
             >
-                <TextInput id='rename-dialog-new-description'
+                <TextInput id='set-description-dialog'
                     value={newDescription}
-                    onChange={setDescription} />
+                    onChange={setNewDescription} />
             </FormGroup>
         </Form>
     );
@@ -47,7 +47,7 @@ export const DescriptionModal = ({ vm }) => {
             title={cockpit.format(_("Set description of $0"), vm.name)}
             footer={
                 <>
-                    <Button variant='primary' isDisabled={isLoading} isLoading={isLoading} id="description-config-dialog-apply" onClick={save}>
+                    <Button variant='primary' isDisabled={isLoading} isLoading={isLoading} id="set-description-dialog-apply" onClick={save}>
                         {_("Save")}
                     </Button>
                     <Button variant='link' onClick={Dialogs.close}>
--- a/src/components/vm/overview/titleModal.jsx
+++ b/src/components/vm/overview/titleModal.jsx
@@ -14,7 +14,7 @@ const _ = cockpit.gettext;
 export const TitleModal = ({ vm }) => {
     const Dialogs = useDialogs();
     const [error, setError] = useState({});
-    const [newTitle, setTitle] = useState(vm.title);
+    const [newTitle, setNewTitle] = useState(vm.title);
     const [isLoading, setIsLoading] = useState(false);
 
     function save() {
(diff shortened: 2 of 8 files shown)


src/libvirtApi/domain.js
```diff
@@ -865,6 +865,36 @@ export function domainSetCpuMode({
     ], opts);
 }
 
+export function domainSetTitle({
```
KKoukiou, line 868: domainSetTitle and domainSetDescription have essentially the same function body. Can you maybe parameterize it?

src/components/vm/vmDetailsPage.jsx
```diff
@@ -87,15 +87,29 @@ export const VmDetailsPage = ({
...
     }, []);
 
+    const vmNameHeadline = (
+        <h2 className="vm-name">{vm.name}</h2>
+    );
+
+    const vmTitleHeadline = (
+        <h2 className="vm-name">
+            {vm.title}
+            <span style={{ fontSize: "var(--pf-global--FontSize--xl)", paddingLeft: "0.83em" }}>
```
KKoukiou, line 97: This 0.83 feels like a bit magic number.. Can you just maybe wrap this in a flex container with some spacing? 

```<Flex spaceItems={{ default: 'spaceItemsSm' }}>```

File: src/components/vm/vmDetailsPage.jsx [hunk 3]
```diff
@@ -87,15 +87,29 @@ export const VmDetailsPage = ({
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    const vmNameHeadline = (
```
KKoukiou, line 90: No need to keep this seperate. Just create a const like this:

```
    const vmHeadline = (
    <Flex spaceItems={{ default: 'spaceItemsSm' }} component='h2'>
            {vm.title && <FlexItem>{vm.title}</FlexItem>}
            <FlexItem style={{ fontSize: "var(--pf-global--FontSize--xl)">
                {vm.name}
            </span>
    </Flex>
```

I did not test it, please adjust if needed.

src/components/vm/overview/descriptionModal.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { domainSetDescription } from '../../../libvirtApi/domain.js';
+
+const _ = cockpit.gettext;
+
+export const DescriptionModal = ({ vm }) => {
+    const Dialogs = useDialogs();
+    const [error, setError] = useState({});
+    const [newDescription, setDescription] = useState(vm.description);
```
KKoukiou, line 17: Just a nitpick - but we like to use the same name for the variable as the setter. Just use const [description, setDescription] = ...

src/components/vm/overview/titleModal.jsx
```diff
@@ -0,0 +1,64 @@
...
+            title: newTitle
+        }).then(Dialogs.close, exc => {
+            setIsLoading(false);
+            setError({ dialogError: _("Title could not be changed"), dialogErrorDetail: exc.message });
+        });
+    }
+
+    const defaultBody = (
+        <Form isHorizontal>
+            <FormGroup id="title-model-select-group" label={_("New title")}
```
KKoukiou, line 35: s/model/modal/g

src/components/vm/overview/titleModal.jsx
```diff
@@ -0,0 +1,64 @@
...
+                fieldId="rename-dialog-new-title"
```
KKoukiou, line 36: It's not rename dialog;

s/rename/set-title/

KKoukiou (changes requested):
@Britz thanks a lot for the enhancement


@garrett can you please do a design review?

 @Britz As we are doing a test driven development, would you be interested to add tests? if yes I am happy to guide you, if not I can write them for you and push to your branch.

skobyda:
Hey @Britz . Thanks a lot for this PR.

Before we dive deeper into code review, do you mind explaining why did you decide to implement this specific feature (VM title and description)? I'm asking, since I have not seen an issue or request for it before, so I don't really know what use case does this solve, and what kind of additional information would use input into title/description which cannot be part of VM's name.
In cockpit-machines, we are quite careful about adding new features and expanding the UI, because we don't want the UI to become overcrowded with a lot of inputs, modals, and random information. So i'm just trying to categorize what kind of use case this functionality solve, and how many users would use it..., so for this so we can find the best place for this functionality in our UI :)

Did you decide to implement it based on some issue or some request from somebody?
Or did you implement it because you personally set title/description for your own VMs? Can you then perhaps explain and give is examples what kind of information do you input into it?

Lastly, virt-manager also supports setting title/description. But a lot of users use it for quasi-categories. Since virt-manager doesn't support VM categories, users input the category into the title/description, they can categorize it that way. That's not ideal from virt-manager, and if this is the aim here, I would like to avoid going exactly the same way (btw we have a [request for adding categories](https://github.com/cockpit-project/cockpit-machines/issues/778) )

skobyda:
Btw, here are the screenshots I took when I built this PR locally, so other people who just wanna talk about the design don't have to build it:

(seems like this PR breaks our grid in overview tab, that needs to be fixed)
Screenshot from 2023-04-03 01-16-01
Screenshot from 2023-04-03 01-16-19
ng)
Screenshot from 2023-04-03 01-16-06

src/libvirtApi/domain.js
Britz: Yes I will do it, no problem. Acually I thought the same, but tried to match the rest of the code and doing as little changes as possible.  All other field seems to have their own functions and there are a lot cockpit.spawn methods which call virt-xml with basically the same arguments and I didn't want to refactor a whole file on my first contribution ;-).

src/components/vm/vmDetailsPage.jsx
Britz: The intention was having the same space ob both sides and the space in the right side is , according to the chrome debugger, determined by `margin-block-end` of `h2`  wich is set to 0.83em. I could not find it as a css variable so I just hardcoded it. There are probably better ways but I'm not a React expert. The Flex spaceItems seems to put spaces on both sides.

src/components/vm/vmDetailsPage.jsx
Britz: Great, this is how spaceItems is used. This code does not the same, but I now understand the idea and will change mine accordingly.

src/components/vm/overview/descriptionModal.jsx
Britz: Makes sense. I will change it,  but i will use `[newDescription, setNewDescription]` to be consistent with the rename dialog code.

Britz:
@KKoukiou: According to the test driven development, It would be great if you could this, I will try to help then.  I've see that there are tests but no documentation on what and how to test and how to setup tests, so I just skipped it for now. I had not time to reverse engineer that by myself.

Britz:
@skobyda: Thanks for adding the screenshots. The rename dialog actually did not change, I will add additional screenshots to pinpoint the changes.

Your question about why is quite simple. I simply implemented it, be cause we need that feature to better describe the purpose of the VM. Of cause you can code a lot of information into naming patterns, but i don't like extensively long domain names, especially when you need to type them regularly on shell, e.g., using `virsh`. We started with naming patterns, but since the setup gained a lot of complexity, this was just not feasible anymore and we switch to a way cleaner naming using the hostnames and domain names.  
   
To our setup/project: We are currently developing a small-scale on premises trusted research environment to do research on medical data in Germany (so strict laws like GDPR apply) and we use plain libvirt on our servers with an increasing amount of machines. We also tried oVirt, proxmox and even OpenStack and they do not fit our current requirements, so we need to user pure libvirt for now.  Cockpit machine is a nice way to have a quick overview over the machines but it is far away from being a productive replacement of the underlaying console tools, thus we need both.

src/components/vm/vmDetailsPage.jsx
KKoukiou: Additionally - I just realized that here we put the vm.name in parenthesis. Since most users don't use title, we defiintely want to have the name outside of parenthesis, when the user did not specify any title for the VM.

Britz:
I committed my changes and tried to fix all already mentioned issues, except the missing tests. 
I also did not really understand how to edit the PO files, since they seem to be generated.
 
Since I am a german, I just added the German translations of the added text snippets. Should I add these the other languages without translation, such that other can easily spot and contribute to missing translations? How is the process there?

Britz:
Here are some screenshot of the changes including explanations.

### The VM List 
If the title is set, it is shown instead of the name and the vm-name is shown in parenthesis.  
Since VM names cannot contain whitespaces, this is unambiguous.  
Bildschirm­foto 2023-04-03 um 17 02 43

### VM details Page 
Title, name and description are set and shown both in the headline and in the overview.
The elements in the overview are for clarification and to be able to edit the values, since the VM  is running, the edit buttons are not active. An alternative would be to add "edit Title" and "edit Description" entries to the VM actions menu.
Bildschirm­foto 2023-04-03 um 17 02 29

If the VM  is shut down, the edit buttons are active and can be clicked.
Bildschirm­foto 2023-04-03 um 17 04 06

The change title dialog. For some reasons, the specified translations do not apply here. Empty values are allowed to delete the title.  Perhaps it is better to add a dedicated delete button  in addition.
Bildschirm­foto 2023-04-03 um 17 04 28

The dialog to change the description. Basically the same as the title dialog. Also here, empty values are allowed to delete the description.  
Bildschirm­foto 2023-04-03 um 17 04 40

If the title is not set, the name element is also not shown in the overwiew page, since it is indistinguishably shown in the header and it can be edited via the VM actions menu. Empty title and empty description entries are shown, since otherwise there would be no possibility to set them. 
Bildschirm­foto 2023-04-03 um 17 04 54

KKoukiou:
@Britz I had a discussion yesterday with our designer (@garrett) and he believes adding these few extra options in the overview makes the UI rather cramped. He is working on a new design which will fit this in nicely, and then after I can pick up your commits to incorporate in the new design, add tests etc.

Thanks for the contributions and sorry for the waiting time on this. I hope this is not super urgent.

Britz:
@KKoukiou That's great to here. It's in deed not the best design and really pragmatic. My first approach was to only put it in the header and add three field inside the rename dialogue, but that actually also  fails, if the name of the vm does not change and thus it would always create errors if only the title or the description would be changed, which is not nice. Since it is easy to install from Git, we can also use the version in my branch until the feature is there.

When you have final idea of how to incorporate this information, perhaps I can help to implement. I also has other required changes on my todo list, but I had not yet time to implement or to check if there are already issues. This includes, e.g., the name of the created vnet interface on the host in the network interface overview table, which is a very important information.

mvollmer:
> My first approach was to only put it in the header and add three field inside the rename dialogue,

I think that's a reasonable approach since it leads to fewer changes to the page. I think it's also discoverable: if people care about the names of their machines and want to improve them, they will see the additional options like title and description, and might change those instead of the id.

I wouldn't use "Title" in the UI, but rather something like "Display name" or "Long name".

> but that actually also fails, if the name of the vm does not change and thus it would always create errors if only the title or the description would be changed, which is not nice.

I am pretty sure we can make this work somehow.

mvollmer:
I have done my own version of this (mostly to get more familiar with the code in general): #1644. Let's continue there, ok?

PR #1015, opened 2023-04-11
Fix flaky testDetachDisk, testNICAdd, testCreateDownloadRhel

martinpitt:
This fixes our [current top flake](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?days=14&repo=cockpit-project%2Fcockpit-machines&test=%2Ftest%2Fcheck-machines-disks+TestMachinesDisks.testDetachDisk) in `TestMachinesDisks.testDetachDisk`:

image

Log example [one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1014-20230411-125144-732b5374-ubuntu-stable/log.html#26-2), [two](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1013-20230410-090748-0a28319b-fedora-38/log.html#26).

It also fixes [this timeout](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1014-20230411-130612-732b5374-arch/log.html#77), which I've seen twice today.

This also fixes our second top flake: testCreateDownloadRhel. [example 1](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1012-20230409-232913-0777ec7b-fedora-37-firefox/log.html#12), [example 2](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4620-20230407-230006-3c56b05e-fedora-37-cockpit-project-cockpit-machines/log.html#12). This is more serious, it causes a page crash even. So let's land this ASAP.

### DIFF
--- a/src/components/vm/vmDetailsPage.jsx
+++ b/src/components/vm/vmDetailsPage.jsx
@@ -31,6 +31,8 @@ import { Popover } from "@patternfly/react-core/dist/esm/components/Popover/inde
 import { ExpandIcon, HelpIcon } from '@patternfly/react-icons';
 import { WithDialogs } from 'dialogs.jsx';
 
+import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
+
 import { vmId } from "../../helpers.js";
 
 import { VmFilesystemsCard, VmFilesystemActions } from './filesystems/vmFilesystemsCard.jsx';
@@ -62,6 +64,10 @@ export const VmDetailsPage = ({
         };
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
+
+    // not initialized yet?
+    if (!vm.capabilities)
+        return <EmptyStatePanel title={ _("Loading...") } loading />;
 
     const vmActionsPageSection = (
         <PageSection className="actions-pagesection" variant={PageSectionVariants.light} isWidthLimited>
--- a/src/components/vm/disks/vmDiskColumns.jsx
+++ b/src/components/vm/disks/vmDiskColumns.jsx
@@ -16,7 +16,7 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
-import React, { useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import cockpit from 'cockpit';
 
@@ -110,9 +110,7 @@ DiskExtras.propTypes = {
     idPrefix: PropTypes.string.isRequired,
 };
 
-export const DiskActions = ({ vm, vms, disk, supportedDiskBusTypes, idPrefixRow }) => {
-    const [isActionOpen, setIsActionOpen] = useState(false);
-
+export const DiskActions = ({ vm, vms, disk, supportedDiskBusTypes, idPrefixRow, isActionOpen, setIsActionOpen }) => {
     const Dialogs = useDialogs();
 
     const onRemoveDisk = () => {
--- a/src/components/vm/disks/vmDisksCard.jsx
+++ b/src/components/vm/disks/vmDisksCard.jsx
@@ -16,7 +16,7 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import cockpit from 'cockpit';
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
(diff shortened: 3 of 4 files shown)


martinpitt:
I looked at our second top flake: testCreateDownloadRhel. [example 1](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1012-20230409-232913-0777ec7b-fedora-37-firefox/log.html#12), [example 2](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4620-20230407-230006-3c56b05e-fedora-37-cockpit-project-cockpit-machines/log.html#12). This is more serious, it causes a page crash even. So let's land this ASAP.

src/components/vm/vmDetailsPage.jsx
```diff
@@ -63,6 +65,10 @@ export const VmDetailsPage = ({
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    // not initialized yet?
+    if (!vm.capabilities)
+        return <EmptyStatePanel title={ _("Loading... ") } loading />;
```
KKoukiou, line 70: You have an extra space here, I guess it will not pick up existing translations?

martinpitt: Eek, that wasn't intended -- I surely want to use the existing translation, I explicitly grepped for that. Thanks for spotting! I'll let the tests finish and re-push.

martinpitt:
OMG, all green except for [this unrelated one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1015-20230412-071242-0e56ff99-rhel-8-9/log.html#78)! I'll debug that separately, and fixed the spacing issue above.

KKoukiou (approved):

src/components/vm/vmDetailsPage.jsx
```diff
@@ -63,6 +65,10 @@ export const VmDetailsPage = ({
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    // not initialized yet?
+    if (!vm.capabilities)
+        return <EmptyStatePanel title={ _("Loading...") } loading />;
+
```
skobyda, line 71: I don't think checking for vm.capabilities does what is indented here. 
If it's to handle only dummy VMs for the duration ofVM creation, then this should already be handled in [app.jsx](https://github.com/cockpit-project/cockpit-machines/blob/main/src/app.jsx#L297), where we show "Creating [VM]":
Screenshot from 2023-04-12 10-12-06.

If on the other hand, this should fix some situation where we have object of a regular existing VM fetched through libvirt API, then checking for `vm.capabilities.` It's true it's a last property of VM which gets fetched through libvirt API, but as we can see [here](https://github.com/cockpit-project/cockpit-machines/blob/main/src/libvirtApi/domain.js#L712), vm object in our redux store doesn't get updated until every property about the VM is fetched.

So there are only 2 states:
1. There is no "vm" object in redux store, since the [domainGet](https://github.com/cockpit-project/cockpit-machines/blob/main/src/libvirtApi/domain.js#L654) hasn't finished yet. This is already handled by in [app.jsx](https://github.com/cockpit-project/cockpit-machines/blob/main/src/app.jsx#L150) with "Loading resources" spinner
2. there is "vm" object in redux store with every single property, including capabilities.

The point is, I think there should never be a point in time where there is a "vm" object without "capabilities" property.

skobyda (changes requested):

src/components/vm/vmDetailsPage.jsx
skobyda: @martinpitt can you give me an example of a reproducer where you run into an app crash because vm.capabilities was not initialized yet?

martinpitt: @skobyda : The reproducers are in the description: testCreateDownloadRhel is our #2 top flake, and it reproduces very well locally. (I used `TEST_BROWSER=firefox` on fedora-37 as that's what fails most often in CI).

> I think there should never be a point in time where there is a "vm" object without "capabilities" property

That would be very nice, but is unfortunately and demonstrably (as indicated by the oopses in the test) not true. I tried to describe it in the commit message. There are several places which put a stub VM into the store, e.g. `finishVmCreateInProgress()`, `usageStartPolling()`, `doUsagePolling()`.

I initially started with a [patch like this](https://paste.centos.org/view/4bca7b2e), but it's horribly error prone (as we'll keep forgetting this in the future). This seems much cleaner and more robust to me.

It would certainly be possible to rework all the places which do `store.dispatch(updateVm(props))` and `updateOrAddVm` etc. to not do the store update if the detection isn't finished yet. But I don't know the data model well enough to do that. My suggestion would be to keep this fallback to fix the oops, and remove it if/once you want to rework the data store that it meets this assumption?

skobyda (approved):
So the issue with downloadRhelImages flake is here: https://github.com/cockpit-project/cockpit-machines/blob/main/src/libvirtApi/domain.js#L365 . finishVmCreateInProgress sets VM's "createInProgress: false". However for a short period of time, new VM object is still not fetched through libvirtAPI, but our UI VM object is already not "createInProgress", so this doesn't help: https://github.com/cockpit-project/cockpit-machines/blob/main/src/app.jsx#L293

So as Martin said, this needs some more work in our data model, so I'm fine with having that fallback which checks for vm.capabilities, but we still need to fix the above metioned issue.

src/components/vm/vmDetailsPage.jsx
skobyda: Resolved, see my comment at https://github.com/cockpit-project/cockpit-machines/pull/1015#pullrequestreview-1380908918

PR #1017, opened 2023-04-12
test: add some general checks for watchdog

yunmingyang:


KKoukiou (changes requested):
Please remove this pixel test (without-watchdog), as it offers nothing more compared to the existing image.

yunmingyang:
> Please remove this pixel test (without-watchdog), as it offers nothing more compared to the existing image.

But, by default, no selection in the watch dog dialog if a VM without watchdog device, do you think it is necessary for our test?

test/check-machines-settings
```diff
@@ -599,8 +599,31 @@ class TestMachinesSettings(VirtualMachinesCase):
...
+            b.click(selector)
+            b.wait_not_present("#vm-subVmTest1-watchdog-modal")
+
+        # Cancel
+        openWatchDogDialog()
+        closeWatchDogDialog("#watchdog-dialog-apply + button")
+
+        # "X" of the dialog
+        openWatchDogDialog()
+        closeWatchDogDialog("#vm-subVmTest1-watchdog-modal button[aria-label=Close]")
```
skobyda, line 617: I don't really understand point of these tests. Just opening and closing a modal is not related to watchdog, but more to the modal itself. We use patternfly's modals, so this feels like testing general functionality of pattenfly. Then we also have some custom wrapper at cockpit's [pkg/lib/dialogs.jsx](https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/dialogs.jsx), but watchdog tests still doesn't feel like a place to test it, no?

KKoukiou: @skobyda we indeed use the dialogs.jsx from lib but we can still forget to add the [onClose property to the Modal](https://github.com/cockpit-project/cockpit-machines/blob/main/src/components/vm/overview/watchdog.jsx#L81), or just wrongly code the Cancel button handlers. These are indeed trivial tests, but they can break on each Modal independently.

skobyda: Fair enough

test/check-machines-settings
```diff
@@ -599,8 +599,31 @@ class TestMachinesSettings(VirtualMachinesCase):
...
+
+        # Click "Add" without selecting action
+        openWatchDogDialog()
+        b.click("#watchdog-dialog-apply")
+        b.wait_in_text("#vm-subVmTest1-watchdog-modal .pf-m-danger", "unsupported configuration")
+        closeWatchDogDialog("#vm-subVmTest1-watchdog-modal button[aria-label=Close]")
+
```
skobyda, line 624: > But, by default, no selection in the watch dog dialog if a VM without watchdog device, do you think it is necessary for our test?

You mean that when no watchdog is specified, the dialog has no radio button selected?
Screenshot from 2023-04-12 20-02-59

Yeah, we can test that. But I think we should not test that using pixel tests. Pixel tests are usually meant to check that layout, spacing, colours... do not unexpectedly change. But that something is selected/not selected by default should be check in code.

You can test that no single radio button is selected with something like this. This gets a boolean of whether radio button for "Reset" is unchecked:
` b.assertFalse(b.get_checked("#vm-subVmTest1-watchdog-modal #reset"))`


```suggestion
        openWatchDogDialog()
        
        b.assertFalse(b.get_checked("#vm-subVmTest1-watchdog-modal #reset"))
        b.assertFalse(b.get_checked("#vm-subVmTest1-watchdog-modal #poweroff"))
        b.assertFalse(b.get_checked("#vm-subVmTest1-watchdog-modal #inject-nmi"))
        b.assertFalse(b.get_checked("#vm-subVmTest1-watchdog-modal #pause"))
        
        b.click("#watchdog-dialog-apply")
        b.wait_in_text("#vm-subVmTest1-watchdog-modal .pf-m-danger", "unsupported configuration")
        closeWatchDogDialog("#vm-subVmTest1-watchdog-modal button[aria-label=Close]")
        
```

And then we no longer need pixel_test_tag="without-watchdog" pixel test

skobyda (changes requested):

test/check-machines-settings
yunmingyang: OK, will modify it, thanks

skobyda (approved):
LGTM, thanks

PR #1018, opened 2023-04-12
test: Fix MAC overflow, better debugging of VM boot failures

martinpitt:
During my investigations in #1015 I saw `TestMachinesNICs.testNICDelete` [fail](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1015-20230412-073028-72a21a50-centos-8-stream/log.html#78-2) [several](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1015-20230412-071242-0e56ff99-rhel-8-9/log.html#78) [times](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1015-20230412-074921-72a21a50-centos-8-stream/log.html#78-1). It's impossible to see why, so I made that easier to debug. [This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1015-20230412-080736-72a21a50-centos-8-stream/log.html) is similar (although there the screenshot makes it easier to compare the log with the console output).

I also found a [TestMachinesNetworks.testNetworkSettings failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1015-20230412-073028-72a21a50-rhel-8-9/log.html#57) where it tried to increment a MAC 52:54:00:b6:51:FF to :100 (which is invalid). The second commit fixes this.

 - [x] Goes on top of #1015, as it's otherwise impossible to get tests to green

PR #1019, opened 2023-04-12
test: Run with CirrOS 0.5.2

martinpitt:
Check if that fixes any of the kernel oopses in the nested subVmTest.

---
 - [x] #1018 
 - [x] #1023

With #1018 we get more info about the boot failures:

 - VM is slow to boot: [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1018-20230412-083643-20237448-centos-8-stream/log.html#62) -- looks like too little entropy; but both our outer bots VM as well as the inner subVmTestN already has the `virtio-rng-pci` device
 - inner VM has kernel oops on "IO-APIC + timer doesn't work!": [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1018-20230412-083643-20237448-centos-8-stream/log.html#81), [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1018-20230412-083643-20237448-rhel-8-9/log.html#50), [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1018-20230412-083643-20237448-rhel-8-9/log.html#78)

The inner VMs already don't use KVM, as the outer VMs (our rhel-8-9 bots image, for example) don't have /dev/kvm. The only remaining thing to try that comes to my mind is to vary the CirrOS version, or perhaps try the x86_64 one (which will need more RAM, though)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ prepare-check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common test/reference
 # run the browser integration tests
 # this will run all tests/check-* and format them as TAP
 check: prepare-check
-	test/common/run-tests ${RUN_TESTS_OPTIONS}
+	test/common/run-tests --nondestructive-memory-mb=1400 ${RUN_TESTS_OPTIONS}
 
 bots: tools/make-bots
 	tools/make-bots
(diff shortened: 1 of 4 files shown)


martinpitt:
[This result](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1019-20230412-144402-aec713f4-rhel-8-9/log.html#76) ruins it from both sides -- allocating more RAM to the subVMs makes the host VM OOM in `testAddDiskSCSI`, and `testBasicLibvirtUserUnprivileged` still shows the `IO-APIC + timer doesn't work!` oops.

martinpitt:
Obsolete -- #1590 runs with Alpine now, let's rather investigate that direction.

PR #1020, opened 2023-04-12
Fix broken ESLint check

martinpitt:
If you look at a [recent test](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1018-20230412-105746-a06b072c-rhel-9-3/log.html), it starts with 

> 1 /static-code/test-eslint # SKIP no eslint

and we missed a few ESLint issues due to that.

### DIFF
--- a/build.js
+++ b/build.js
@@ -71,8 +71,8 @@ const context = await esbuild.context({
         cleanPlugin(),
         ...lint
             ? [
-                stylelintPlugin({ filter: new RegExp(cwd + '\/src\/.*\.(css?|scss?)$') }),
-                eslintPlugin({ filter: new RegExp(cwd + '\/src\/.*\.(jsx?|js?)$') })
+                stylelintPlugin({ filter: new RegExp(cwd + '/src/.*\\.(css?|scss?)$') }),
+                eslintPlugin({ filter: new RegExp(cwd + '/src/.*\\.(jsx?|js?)$') })
             ]
             : [],
         // Esbuild will only copy assets that are explicitly imported and used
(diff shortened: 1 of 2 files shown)


build.js
```diff
@@ -71,8 +71,8 @@ const context = await esbuild.context({
         cleanPlugin(),
         ...lint
             ? [
-                stylelintPlugin({ filter: new RegExp(cwd + '\/src\/.*\.(css?|scss?)$') }),
-                eslintPlugin({ filter: new RegExp(cwd + '\/src\/.*\.(jsx?|js?)$') })
+                stylelintPlugin({ filter: new RegExp(cwd + '/src/.*\\.(css?|scss?)$') }),
```
KKoukiou, line 74: Why did you add the new backlash? we need to escape the dot, and that's why we need the backlash before the dot, but I don't get why we are now escaping that.

martinpitt: Heh, I suppose this is a bit subtle. Note that these are *not* regular expressions, like `/a\.b/` -- they are *strings* which get *compiled into* a RE. In a standard string, `\.` does nothing different than `.`, both are a simple period character. That's why eslint declares it as "useless":

```
> console.log(/a\.b/)
/a\.b/
> console.log('a\.b')
a.b
```

But we actually want a *RE* with `\.`, and the way to represent `\` in a string is `\\`:

```
> console.log('a\\.b')
a\.b
```
I.e. before it was wrong -- the regexp said "any character followed by css or scss" (more or less, ignoring the `?`). But we don't want that, we want "a period followed by css or scss".

Does that make sense?

KKoukiou (approved):

PR #1021, opened 2023-04-12
Show error message if snapshot deletion fails during VM deletion

skobyda:
Before we delete a VM, we first delete snapshots. However, if snapshot deletion fails, no error message is shown in dialog and the dialog is just left open.

Screenshot from 2023-04-12 13-37-07

### DIFF
--- a/src/components/vm/deleteDialog.jsx
+++ b/src/components/vm/deleteDialog.jsx
@@ -143,9 +143,7 @@ export class DeleteDialog extends React.Component {
                             .then(() => {
                                 Dialogs.close();
                                 cockpit.location.go(["vms"]);
-                            })
-                            // Fail implicitly returns the exception, which is useful so that cleanup operation doesn't get invoked
-                            .fail(exc => this.dialogErrorSet(cockpit.format(_("Could not delete $0"), vm.name), exc.message));
+                            });
                 })
                 .then(() => { // Cleanup operations
                     return domainDeleteStorage({ connectionName: vm.connectionName, storage, storagePools })
@@ -154,7 +152,8 @@ export class DeleteDialog extends React.Component {
                                 detail: exc,
                                 type: "warning"
                             }));
-                });
+                })
+                .catch(exc => this.dialogErrorSet(cockpit.format(_("Could not delete $0"), vm.name), exc.message));
     }
 
     render() {
--- a/test/check-machines-lifecycle
+++ b/test/check-machines-lifecycle
@@ -430,6 +430,30 @@ class TestMachinesLifecycle(VirtualMachinesCase):
             qemu_img_info = m.execute(f"qemu-img info /var/lib/libvirt/images/{name}-2.img")
             self.assertNotIn("snapshotA", qemu_img_info)
 
+            # Snapshot deletion was failing for VMs with file-directory storage-pool volumes as disks.
+            # It was fixed since libvirt 7.1.0 at https://gitlab.com/libvirt/libvirt/-/commit/6423e3082835db932dd285f3ae75b5dad9d91c83
+            # Only debian-stable has libvirt lesser than 7.1.0
+            # Add this condition to notify us when the condition below is no longer needed
+            if m.image == "debian-stable":
+                self.assertLess(m.execute("virsh --version"), "7.1.0")
+            else:
+                # Check deleting VM fails if deleting snapshot is not possible and error message is shown
+                name = "vm-with-failed-snapshots"
+                self.createVm(name, running=False)
+
+                m.execute("virsh vol-create-as --pool images --name tmpdisk --capacity 1 --format qcow2")
+                m.execute(f"virt-xml {name} --add-device --disk source.pool=images,source.volume=tmpdisk,target.dev=sdb,driver.type=qcow2")
+                m.execute(f"virsh snapshot-create-as --domain {name} --name snapshotFails")
+                b.reload()
+                b.enter_page('/machines')
+                self.performAction(name, "delete")
+                b.wait_visible(f"#vm-{name}-delete-modal-dialog")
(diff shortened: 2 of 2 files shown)


src/components/vm/deleteDialog.jsx
```diff
@@ -131,8 +131,12 @@ export class DeleteDialog extends React.Component {
         const storagePools = getVmStoragePools(vm);
 
         Promise.all(
-            (Array.isArray(vm.snapshots) ? vm.snapshots : [])
-                    .map(snapshot => snapshotDelete({ connectionName: vm.connectionName, domainPath: vm.id, snapshotName: snapshot.name }))
+            (Array.isArray(vm.snapshots) ? vm.snapshots : []).map(snapshot => {
+                return snapshotDelete({ connectionName: vm.connectionName, domainPath: vm.id, snapshotName: snapshot.name })
+                        .catch(ex => {
+                            return Promise.reject(new Error(cockpit.format(_("Could not delete snapshot $0: "), snapshot.name) + ex.message));
```
KKoukiou, line 137: There is not need to change this here. If you leave it how it was before the error will be now propagated to the new catch you added in line 160.

After all you probably don't want a message like:

Coult not delete my-vm-name: Could not deleted snapshot vm-vm-snapshot0: error from libvirt 

It's fine to omit the middle string.

skobyda: If we omit that, then there is no mention of the snapshot in error message. And I don't if user would not be confused where exactly does the message come

KKoukiou: The snapshot APIs error seem to contain the name. In your case it fails on deleting the storage, and the error is again correct.

Screen Shot 2023-04-14 at 19 25 51

KKoukiou:
@skobyda sent a seperate commit to fix tests.

KKoukiou (dismissed):
Assuming tests pass.

martinpitt:
This has [failed on debian-stable](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1021-20230418-025732-93dea615-debian-stable/log.html#80) after three retries ("Requested operation is not valid: can't manipulate inactive snapshots of disk 'sdb'"), this failure is real I'm afraid.

skobyda:
The failure was caused by a bug in libvirt which was fixed since 7.1.0 with https://listman.redhat.com/archives/libvir-list/2021-February/216096.html
I added a condition which triggers this test case only if libvirt >= 7.1.0

test/check-machines-lifecycle
```diff
@@ -430,6 +430,26 @@ class TestMachinesLifecycle(VirtualMachinesCase):
             qemu_img_info = m.execute(f"qemu-img info /var/lib/libvirt/images/{name}-2.img")
             self.assertNotIn("snapshotA", qemu_img_info)
 
+            # Snapshot deletion was failing for VMs with file-directory storage-pool volumes as disks.
+            # It was fixed since libvirt 7.1.0 at https://gitlab.com/libvirt/libvirt/-/commit/6423e3082835db932dd285f3ae75b5dad9d91c83
+            if m.execute("virsh --version") >= "7.1.0":
```
KKoukiou, line 435: When we add conditions like this we always have this else which will notify us when the condition is not needed anymore. 

Since we know it's only ubuntu-stable that's reaches this else, what do you think if we would do something like:
else:
   assertEqual(m.image, "ubuntu-stable")

KKoukiou (dismissed):

test/check-machines-lifecycle
```diff
@@ -430,6 +430,30 @@ class TestMachinesLifecycle(VirtualMachinesCase):
             qemu_img_info = m.execute(f"qemu-img info /var/lib/libvirt/images/{name}-2.img")
             self.assertNotIn("snapshotA", qemu_img_info)
 
+            # Only debian-stable has libvirt lesser than 7.1.0
+            # Add this condition to notify us when the condition below is no longer needed
+            if m.image == "debian-stable":
+                self.assertTrue(m.execute("virsh --version") < "7.1.0")
```
martinpitt, line 436: assertLessThan() for more useful failures, in general (but it should be quite obvious)

test/check-machines-lifecycle
```diff
@@ -430,6 +430,30 @@ class TestMachinesLifecycle(VirtualMachinesCase):
...
+            # Snapshot deletion was failing for VMs with file-directory storage-pool volumes as disks.
+            # It was fixed since libvirt 7.1.0 at https://gitlab.com/libvirt/libvirt/-/commit/6423e3082835db932dd285f3ae75b5dad9d91c83
+            if m.execute("virsh --version") >= "7.1.0":
```
martinpitt, line 439: this would better be an `else:`

martinpitt (commented):
[This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1021-20230427-111045-aad09964-fedora-37-pybridge/log.html#80) seems relevant? Seems it's not reliable on f-37 either? (I retried once to compare)

test/check-machines-lifecycle
skobyda: Fixed

KKoukiou (changes requested):
AttributeError: 'TestMachinesLifecycle' object has no attribute 'assertLesserThan'

You need assertLessThan

martinpitt:
TestMachinesLifecycle.testDelete still fails in a lot of cases :cry:

KKoukiou (commented):
Oop still fails. AttributeError: 'TestMachinesLifecycle' object has no attribute 'assertLessThan'

assertLess :P ?

martinpitt:
Right, [assertLess](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLess).. :see_no_evil:

KKoukiou (approved):

martinpitt (approved):
Cheers!

PR #1022, opened 2023-04-12
Remove dummy UI VM only after we have fetched all the data about the VM

skobyda:
- Remove dummy UI VM only after we have fetched all the data about the VM
    
    Removing dummy UI VM immediately after the creation of VM is finished
    created a race condition, where the dummy UI VM no longer exists, but we
    have not yet fetched all the data about the real VM through libvirt
    APIs.

-  Don't update dummy UI VM before deleting it
    
    After we finish the creation of VM, we update its dummy UI VM's
    properties to 'createInProgress: false', and then we immediately call
    clearUiState which deletes this dummy UI VM.
    This makes no sense, and it creates space for race conditions where VM
    is neither in creation nor it is

### DIFF
--- a/src/components/vm/vmDetailsPage.jsx
+++ b/src/components/vm/vmDetailsPage.jsx
@@ -31,8 +31,6 @@ import { Popover } from "@patternfly/react-core/dist/esm/components/Popover/inde
 import { ExpandIcon, HelpIcon } from '@patternfly/react-icons';
 import { WithDialogs } from 'dialogs.jsx';
 
-import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
-
 import { vmId } from "../../helpers.js";
 
 import { VmFilesystemsCard, VmFilesystemActions } from './filesystems/vmFilesystemsCard.jsx';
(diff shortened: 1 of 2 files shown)


martinpitt:
Nice, this looks a lot better than my original fix! I'm just a bit concerned that there are [so](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1022-20230412-124634-9b7da9a2-rhel-9-3/log.html) [many](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1022-20230412-124634-9b7da9a2-rhel-8-9/log.html) [random](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1022-20230412-124634-9b7da9a2-debian-testing/log.html) [failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1022-20230412-124634-9b7da9a2-centos-8-stream/log.html) everywhere, but they don't look related to this change.

There are a few in e.g. #1018 or #1023 as well, but not nearly as many. I'll re-run these to make sure it's not something systematic.

martinpitt (approved):

martinpitt:
Better now!

PR #1023, opened 2023-04-12
test: Use standard "wait for boot" in TestMachinesLifecycle

martinpitt:
There is no particular reason why it waits for a different message than the login prompt. That line does not exist any more with newer CirrOS versions.

### DIFF
--- a/test/check-machines-lifecycle
+++ b/test/check-machines-lifecycle
@@ -176,7 +176,7 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         # force reboot
         self.machine.execute(f"echo '' > {args['logfile']}")
         self.performAction("subVmTest1", "forceReboot")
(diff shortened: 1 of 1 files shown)


KKoukiou (approved):

skobyda (approved):
Tried to find where it was introduced, and this seems like a leftover from a loooong time ago where we didn't have unified way how to check VM is booted.
LGTM

martinpitt:
Yes, I did some git blame archaeology, and it's all the way from the initial import from cockpit.

skobyda:
> Yes, I did some git blame archaeology, and it's all the way from the initial import from cockpit.

I checked in the original repository and it was introduced from old selenium tests, when there were no other tests which check that VM is booted, so again, should be fine to make this unified

PR #1024, opened 2023-04-13
Revert "test: Run static-code verbosely"

martinpitt:
Outputting TAP interferes with the integration TAP plan, and messes up the log.html results viewer. It's awkward to parse, merge, and present multiple TAP plans; also, it's not really interesting to show the static-code results in log.html, as they are empty anyway.

Skipped tests still show a "WARNING" even without --tap.

This reverts commit c68e548863151109a006350c5433b21e6b66ccd5.

----

This avoids the messy results count and interleaving of static-code and browser tests that we get [e.g. here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1019-20230413-072512-3cc72fd3-rhel-8-9/log.html):

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,7 +184,7 @@ print-vm:
 	@echo $(VM_IMAGE)
 
 codecheck: test/static-code $(NODE_MODULES_TEST)
-	test/static-code --tap
+	test/static-code
 
 # convenience target to setup all the bits needed for the integration tests
 # without actually running them


martinpitt:
[better](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1024-20230413-084952-3164fa72-debian-testing/log.html)!

KKoukiou (commented):
Which are the missing packages that made this --tap option wanted in the first place? Some linter?

KKoukiou (approved):

martinpitt:
Yes, something like `flake8` or so. But they've been present in our tasks container for a long time now. Of course we need to watch out if we add new tests.

PR #1028, opened 2023-04-13
build.js: esbuild-wasm, build duration, fix sourcemap

subhoghoshX:
Changes:
1. Add esbuild-wasm support for non x86 architectures.
2. Log the time it takes to finish building.
3. Fix sourcemap.


package.json
```diff
@@ -22,6 +22,7 @@
     "esbuild-plugin-copy": "^2.0.2",
     "esbuild-plugin-replace": "^1.3.0",
     "esbuild-sass-plugin": "2.6.0",
+    "esbuild-wasm": "^0.17.16",
```
martinpitt, line 25: This should be consistent with `"esbuild": "0.17.10"` above.

martinpitt: I fixed this while I was at it.

martinpitt:
Indeed updating package.json for cockpit{,-machines,-podman} is harder, as it involves pushing to https://github.com/cockpit-project/node-cache/  -- which you cannot do (yet). FYI, we try to avoid running `npm install` on our own and CI machines, as that's harmful and was actually already exploited in the past. So we do this in an isolated container with `tools/node-modules install` instead, and push it to that git repo.

I took the liberty to do that rebuild, rebase your PR to current main, and update the commit message. In the future: Please be more descriptive what a commit changes, and why. Also, please split unrelated changes into separate commits. It's not too bad in this case, so I'll let that slide here.

Thanks!

martinpitt (approved):
Thanks! TF is having a bad day, and the /pybridge flake is known/unrelated.

Issue #1027, opened 2023-04-13
Details page redesign


garrett:
I've gone through a few iterations and revisions, and he's what I have so far:

![cockpit-machines details redesign](https://user-images.githubusercontent.com/10246/231728235-7a6c0cf6-8fbf-4692-b9aa-958e6a772d5b.png)

Goals:

1. Group metadata (all metadata about the VM is now in the header at the top)
2. Group editable data together
3. Make editable data consistent (autostart was odd as a live effect toggle prior to this)
4. Group read-only values together
5. Provide more space for the Console (trimming down the space used for other things, making the VM go flush to the edges, etc.)
6. Group CPU related items together
7. Group Memory related items
8. Prioritize common machine-level editable
9. Add in a place for description (when it's added from the kebab menu)
10. Find a place for vsock

garrett:
So: What do you think of this so far?

garrett:
Parts of this was originally discussed in https://github.com/cockpit-project/cockpit-machines/issues/971 (good reference for some background info), but the redesign encompasses more than just changing one card (as mentioned above).

skobyda:
The one thing which struck me the most that vCPU and CPU is now combined into 1 row with only one editable options.
Does that mean that both CPU mode and vCPU dialogs:
![Screenshot from 2023-04-13 13-28-53](https://user-images.githubusercontent.com/42733240/231745354-78656ea9-fbc6-4bd2-ab03-b4bc1ea24bfd.png)
![Screenshot from 2023-04-13 13-28-46](https://user-images.githubusercontent.com/42733240/231745359-ab910f59-0206-43a6-865c-8d14e631ba0d.png)

Would be combined into one dialog, something like this?
![Screenshot from 2023-04-13 13-31-53](https://user-images.githubusercontent.com/42733240/231746015-e10d5678-4bc3-4e05-9e45-fab3fbb0980d.png)

garrett:
> Would be combined into one dialog, something like this?

Yes, correct. Something like that.

That dialog should be improved (the number entries shouldn't be quite that wide, for example, and this layout implies that vCPU and Sockets are related and that vCPU maximum and Cores per socket are also related), but all of that stuff would be in it, just reorganized a bit.

skobyda:
And secondly, I see that description has an **edit** button:
[...]

PR #1030, opened 2023-04-14
Redesign CPU in VM details

skobyda:
This is a preparation PR for a redesign of Details page: https://github.com/cockpit-project/cockpit-machines/issues/1027

This PR only redesign CPU-related things:
- It merges CPU model dialog and vCPU dialog into a single dialog
- Get's rid of column layout for vCPU, using a simple vertical form

It also does few small design improvements suggesteed by @garrett :
- Uses PF's NumberInput component for vCPU count and vCPU maximum
- Moves vCPU maximum before vCPU count
- Changes dialog size from medium to small
- Uses HelpIcon instead of InfoAltIcon
- Makes width of selectors smaller, since they were unecessary wide

There is also issue with HelpIcon being too low, but that's not issue of only CPU dialog, but everywhere in cockpit. I'm trying to fix it at https://github.com/cockpit-project/cockpit/pull/18640

### Before
Screenshot from 2023-04-14 13-53-54
### After
Screenshot from 2023-04-14 13-47-22
### Before
Screenshot from 2023-04-14 13-54-08
Screenshot from 2023-04-14 13-54-00
### After
Screenshot from 2023-04-14 13-47-00


skobyda:
Btw this design was discussed yesterday with @garrett where we quasi-mocked it, so I believe the design will not change anymore (or only marginally), so once I fix the test, I think we can immediatelly start with code-review.

src/components/vm/overview/cpuModal.jsx
```diff
@@ -0,0 +1,326 @@
+import React, { useState } from 'react';
```
KKoukiou, line 1: Missing licence

src/components/vm/overview/cpuModal.jsx
```diff
@@ -0,0 +1,326 @@
...
+            return;
+        }
+
+        const newValue = clamp(value, max, 1);
+        setCount(parseInt(newValue));
+    }
+
+    function onSocketChange (value) {
+        const state = { sockets, cores };
+        state.sockets = parseInt(value);
```
KKoukiou, line 112: This is an anti-practice - this is changing the component state without the set* function.

src/components/vm/overview/cpuModal.jsx
```diff
@@ -0,0 +1,326 @@
...
+        // Get divisors of Max VCPU number divided by number of sockets
+        const divs = dividers(max / state.sockets);
+
+        // If current cores value is not in divisors array, then change it to max divisor
+        if (divs.indexOf(cores) === -1) {
+            state.cores = divs[divs.length - 1];
+        }
+
+        // Likewise: Max value = Sockets * Cores * Threads. Sockets = Max value / ( Threads * Cores )
+        state.threads = (max / (state.sockets * state.cores));
```
KKoukiou, line 123: Same here.

KKoukiou (commented):
You have a few state updates without using the react's state modifier functions. Otherwsie looks good from first pass.

src/components/vm/overview/cpuModal.jsx
skobyda: But this is not the `state` itself, it's just a local variable, local object `const state = { sockets, cores };`, So changing `state.socket = something` doesn't change react state, only the local variable.
But I admit the name of variable is confusing... perhaps let's call it stateDelta, or newState?

martinpitt:
RPM build failure in COPR/TF is fixed, but as our own integration tests fail as well, this needs another push anyway.

skobyda:
@garrett mind giving it review?

garrett (changes requested):
Oh, I wasn't aware this was ready. The host "mode" is still at the top, which makes the order on label on the card and in the dialog inconsistent. They should be consistent, so the mode should probably go at the bottom of the modal, as we discussed in our call.

image

When I tried to change to passthrough, I got this error message and the spinner never completes. It's possibly unrelated to this PR, but it might also be a side effect of this PR, as we're changing both at the same time now:

image

skobyda:
> When I tried to change to passthrough, I got this error message and the spinner never completes. It's possibly unrelated to this PR, but it might also be a side effect of this PR, as we're changing both at the same time now:

This is unrelated to this redesign, you can achieve the same with the old design:
Screenshot from 2023-04-20 11-39-44

But I'm more interested, how did you manage to get into this situation that you have 2 vCPUs buy only 1 socket, 1 core per socket and 1 thread per core. You should not be able to set that from our UI. But you can set it like that through virsh or virt-manager, since they still support the "old way (before qemu 2.6"), where maxVCPU didn't have to match the cpu topology, maybe that's how you got that configuration?

skobyda:
> When it's host passthrough, are you able to specify all of the values of the topology? If not, we should hide what you can't change (and make sure the values work with it).


You are still able to specify all fields, no matter what mode you choose. 

> However, if the mode does influence the topology, then we probably want to keep mode at the top and change the label in the overview card to something like: "**CPU** Host mode, 2vCPUs" and "**CPU** Passthrough mode, 2vCPUs" right?

No, mode doesn't influence topology
 
Updated the design:
Screenshot from 2023-04-20 11-46-40

skobyda:
@KKoukiou @martinpitt mind giving this a code-review?

garrett (dismissed):
Thanks for flipping the dropdown to the bottom! This is good from a design viewpoint now.

We _could_ redesign it further later, but for a simple merge with a simple design, this now works for me! :+1: Thanks for taking the initiative at merging it and implementing this, @skobyda!

skobyda:
Fixed pixel tests, now it should be all green

PR #1031, opened 2023-04-14
Add option to create new RAW volume at Create VM dialog

skobyda:
This split "Create new volume" into two options:
- [Create new] qcow2 volume
- [Create new] raw volume

Screenshot from 2023-04-14 16-08-29

But maybe the options should not start with uppercase, e.g. "qcow2 volume" and "raw volume". Or maybe format should be all uppercase, e.g. "QCOW2 volume" or "RAW volume" ?

This PR also adds format into disk listing:
Screenshot from 2023-04-14 16-06-45

Fixes https://issues.redhat.com/browse/RHELPLAN-121361
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2083032


src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -859,7 +859,10 @@ const StorageRow = ({ connectionName, allowNoDisk, storageSize, storageSizeUnit,
                 <FormSelect id="storage-pool-select"
                             value={storagePoolName}
                             onChange={e => onValueChanged('storagePool', e)}>
-                    <FormSelectOption value="NewVolume" key="NewVolume" label={_("Create new volume")} />
+                    <FormSelectOptionGroup key="Create new volume" label={_("Create new volume")}>
+                        <FormSelectOption value="NewVolumeQCOW2" key="NewVolumeQCOW2" label={_("Qcow2 volume")} />
```
KKoukiou, line 863: It's weird to title case Qcow2.
Maybe use something like: 

format: qcow2
format: raw

@garrett ?

KKoukiou (commented):
Straightforward change - codewise approvied.  I think it's fine addition for the UI especially when you think that many people requested it.

garrett (changes requested):
Thanks for the first pass at this PR!

I was thinking something more like this when I was talking about it on the call:

image

qcow2 would be default, of course, and the sparse/preallocate toggle would only show up when using qcow2 (as raw is fully preallocated and cannot be changed). But we'd use dividers between the types of options.

In order to actually get dividers and better looking groups, we'd need to use the PF "select" widget instead of the "form select" widget.

As far as strings go, we could do something like "Create new volume (qcow2)" and "Create new volume (raw)" instead too. This might even work better for translations? (Unsure)

src/components/create-vm-dialog/createVmDialog.jsx
garrett: Answered in my review. (And I agree the case is wrong.)

skobyda:
@garrett Updated the design, this is how it looks now:
Screenshot from 2023-04-18 15-05-39

> qcow2 would be default, of course, and the sparse/preallocate toggle would only show up when using qcow2 (as raw is fully preallocated and cannot be changed). But we'd use dividers between the types of options.

I will leave that up to the next PR, as specifying qcow2/raw format is part of [more urgent jira](https://issues.redhat.com/browse/COCKPIT-925), while sparse/preallocate is not so urgent, so I would not want qcow2/raw format feature to be slowed down by that.
So let's merge this qcow2/raw as soon as possible, and leave sparse/preallocate for other PR.

skobyda:
Oops, didn't notice that in your mockup, No storage is first in the list, updated the PR again:
Screenshot from 2023-04-18 15-19-35

skobyda:
> As far as strings go, we could do something like "Create new volume (qcow2)" and "Create new volume (raw)" instead too. This might even work better for translations? (Unsure)

I always thought whole sentences are better for translators, since they give more freedom for translations, and translators know can better deal with languages where sentences are constructed differently?

garrett (approved):
Thanks! I think this works. :+1:

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -230,7 +231,7 @@ function validateParams(vmParams) {
         validationFailed.memory = _("Memory must not be 0");
     }
 
-    if (vmParams.storagePool == 'NewVolume' && vmParams.storageSize === 0) {
+    if (vmParams.storagePool == 'NewVolumeQCOW2' && vmParams.storagePool == 'NewVolumeRAW' && vmParams.storageSize === 0) {
```
KKoukiou, line 234: The check is wrong here. Will be always false.

KKoukiou (changes requested):

src/components/create-vm-dialog/createVmDialog.jsx
skobyda: Fixed

KKoukiou:
testing farm failure is unrelated.

shodanshok:
> qcow2 would be default, of course, and the sparse/preallocate toggle would only show up when using qcow2 (as raw is fully preallocated and cannot be changed). But we'd use dividers between the types of options.

Both qcow2 and raw files can be preallocated, with slightly different meanings:
- for qcow2, preallocation generally refers to *metadata preallocation inside the qcow2 file itself* (leaving the other space as sparse)
- for raw, preallocation means to `fallocate` all metadata and data blocks at the filesystem level. It is the same difference between `truncate --size=1G test.img` and `fallocate -l 1G test.img`: the first command will create a sparse file, the latter a fully-preallocated file.

I strongly suggest to leave the "preallocate" flag/checkbox for both qcow2 and raw files, as users can have a legitimate use for it.

skobyda:
> > qcow2 would be default, of course, and the sparse/preallocate toggle would only show up when using qcow2 (as raw is fully preallocated and cannot be changed). But we'd use dividers between the types of options.
> 
> Both qcow2 and raw files can be preallocated, with slightly different meanings:
> 
>     * for qcow2, preallocation generally refers to _metadata preallocation inside the qcow2 file itself_ (leaving the other space as sparse)
> 
>     * for raw, preallocation means to `fallocate` all metadata and data blocks at the filesystem level. It is the same difference between `truncate --size=1G test.img` and `fallocate -l 1G test.img`: the first command will create a sparse file, the latter a fully-preallocated file.
> 
> 
> I strongly suggest to leave the "preallocate" flag/checkbox for both qcow2 and raw files, as users can have a legitimate use for it.

There is a [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2183986) for preallocating qcow2 disks. I''ll copy your comment there so we don't forget about it once we will implement allocation of disks.

PR #1035, opened 2023-04-18
Remember VM state filter settings

### LINKED ISSUE #1032: Remember State filter settings in Machines interface
In the Virtual Machines interface, after a user filters the view to show only running VMs, the view reverts back to 'All' when navigating away and then back to the interface. 

Steps to reproduce:

1. Filter VM view by 'State: Running'
2. Switch to the Storage Pools or Networks tab.
3. Switch back to the main VM view, filter reverts back to 'State: All'

The ask is for the filter settings (regardless of which state is being selected) to be remembered during the active session.

subhoghoshX:
Fixes #1032


subhoghoshX:
Oops, tests failed because I used a semicolon in Python. fixed it :)

src/components/vms/hostvmslist.jsx
```diff
@@ -93,6 +105,29 @@ const HostVmsList = ({ vms, config, ui, storagePools, actions, networks, onAddEr
                     .map(state => { return { value: rephraseUI('resourceStates', state), apiState: state } })
                     .sort((a, b) => (prioritySorting[a.apiState] || 0) - (prioritySorting[b.apiState] || 0) || a.value.localeCompare(b.value)));
 
+    useState(() => {
```
KKoukiou, line 108: useState is not supposed to be used like this, in fact what you need here a useEffect hook and options as a dependency.

You can replace the whole useState body with this:
```
    useEffect(() => {
        setStatusSelected(getStatusFilterOption(options));
    }, [options]);
```

src/components/vm/vmDetailsPage.jsx
```diff
@@ -233,14 +233,16 @@ export const VmDetailsPage = ({
         );
     });
 
+    console.log(cockpit.location.encode('/', cockpit.location.options));
```
KKoukiou, line 236: Please remove

test/check-machines-lifecycle
```diff
@@ -272,6 +272,64 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         b.wait_in_text(".pf-c-popover", "VM subVmTest2 failed to start")
         b.click('#vm-subVmTest2-system-state button[aria-label=Close]')
 
+    def testVmFilterSettingsOnNavigate(self):
```
KKoukiou, line 275: This test does not really test what this commit introduces. Instead you want to check that the URL is affected when you apply some status filter, and vice versa, when you apply a status filter, and refresh and filter is persisting.

Use something like:   b.wait(lambda: "vm?name=test%25&connection=syste,&status=shut-off" in b.eval_js("window.location.href"))
for testing the URL.

subhoghoshX: Understood. Thanks! Last time my thought was the URL determines the state and the state in turn determines which vms are listed. So checking the vms listed also checks whether the URL works in an end to end way.

But I think I understand the subtle difference now.

URL -> State (The commit introduces this. I should check this)
State -> VMs listed (It was already implemented. No need to check this)

Fixed it!

I'm also checking for only one filter option to avoid repetition this time. Assuming that if it works for one (e.g. "running") it'd also work for others too (e.g. "shut off"). Happy to get any feedback on that :)

KKoukiou (changes requested):
I believe it's a bit weird to keep the 'status' filter in the URL when we are out of the VM list page. So VM details, Storage and Network pages should not have the 'status=running' in the URL.

This is because:
* The status variable does not have any effect in these subpages
* It's slightly confusing as Storage and Networks have also active/inactive state, which would be confused with the status property in the URL.

So please make this persist only for the main page, and adjust the test accordingly.

mvollmer:
I think we need to start over with this. I have made a "story" for this: https://issues.redhat.com/browse/COCKPIT-1311

PR #1036, opened 2023-04-18
FormHelper should not be inside InputGroup

skobyda:
This fixes broken position of helper text in the memory row

Before|After
:---:|:---:
Screenshot from 2023-04-19 01-25-32|Screenshot from 2023-04-19 01-25-38


martinpitt (approved):
Dakujem!

garrett:
(Thanks for the quick fix!)

PR #1043, opened 2023-04-21
Adjust Breadcrumb usage to match PF5 guidelines

KKoukiou:
This is still PF4 compatible.

See design suggestion: https://github.com/cockpit-project/cockpit/pull/18476#pullrequestreview-1343830115

This needs pkg/lib update from https://github.com/cockpit-project/cockpit/pull/18683
otherwise breaks dark-theme.

All pixel tests updates fall into two changes:

* Reducer spacer in consoles subpage - which was before too much
* Fixed alignment of help button in create dialog with the pkg/lib update

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
-COCKPIT_REPO_COMMIT = 5648586f0a43fcfea57982a792108af3c89866d8 # 289 + HelperText fixes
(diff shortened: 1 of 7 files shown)


martinpitt (approved):
Thanks! Looks reasonable. I was not sure about 
[this pixel change](https://github.com/cockpit-project/pixel-test-reference/compare/a8ee796a6775b88f74470d6c3c3968e1eaae5b08..4bd1819fff2c7053c3e7fdc4d216691dea9764f7#diff-b2500d6ee9af408fe33ea49fb754bac7f838f9e376b7a92f7438e16f4ba5870c) where the circled ? icon moves from "a little too south" to "a little too north", and not centered either way, but as it's not really making things worse, let's go ahead.

PR #1044, opened 2023-04-21
Fix FormHelper warning variants for Memory, Storage and Source rows

skobyda:
Because of missing tests, there is a regression for some form helper texts which should be warning variants, but are instead shown as plain text. 
Fix it and add the missing tests.

Before|After
:---:|:---:
Screenshot from 2023-04-21 13-51-28|Screenshot from 2023-04-21 13-46-37


KKoukiou (approved):

KKoukiou:
packit tests fail as it's not rebased.

PR #1049, opened 2023-04-26
tests: Remove special cases for old libvirt versions

skobyda:
We no longer test on distros with such old libvirt versions


martinpitt (approved):
Yay, thanks!

PR #1053, opened 2023-04-27
workflows: Move reposchutz back to Ubuntu 20.04

KKoukiou:
Moving to 22.04 brought back the tar errors.
To avoid these pin this workflow to Ubuntu 20.04 and add a comment to remind us.

See https://github.com/cockpit-project/cockpit/commit/269bf89276c679a03befc8a04244addd5d810048


martinpitt (approved):
Whoops, thanks!

PR #1055, opened 2023-04-28
Fix crash in VM details page

KKoukiou:
There is some raise condition when a bridge is added from the CLI /sys/class/net is not immediately in sync with the data ip command, when it comes to which bridges are present on the system.

Only update devices that /sys/class/net contains in order to fix the crash.

Fixes crash as seen here. https://cockpit-logs.us-east-1.linodeobjects.com/pull-1054-20230428-132647-f9d64efc-fedora-37/log.html#74-1


martinpitt (approved):
Thanks!

Issue #1057, opened 2023-05-03
testNICAdd is very unstable


martinpitt:
[testNICAdd](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1056-20230502-181230-073252a0-fedora-38/log.html#61) failed three times in a row. Retrying. The stats for this test look rather impressively bad:

![image](https://user-images.githubusercontent.com/200109/235842135-6abdc51e-ed0f-4c69-9cb5-8b2c19790ddf.png)

I can reproduce this locally on arch after some 10 attempts,  and on fedora-38 after around 5 attempts. The actual deletion works -- `virsh domiflist subVmTest1` does not show it any more, and neither does logging in with a parallel browser session. So the UI loses some event apparently.

I have debug logs for the [working](https://github.com/cockpit-project/cockpit-machines/files/11380193/works.txt) and [failing](https://github.com/cockpit-project/cockpit-machines/files/11380195/fail.txt) case, which confirms that 

> debug: signal on /org/libvirt/QEMU/domain/_ab8f1b68_c370_4b9c_904f_2af486a28826: org.libvirt.Domain.DeviceRemoved(["net1"])

is missing in the failing case. In fact, apart from the doUsagePolling() bits there are *no events at all* after removing the interface.

martinpitt:
I checked `virsh event --all --loop` and a series of

```
virsh attach-interface --domain subVmTest1 --type network --source test_network --model virtio --mac fe:54:00:a5:f8:99
virsh detach-interface --domain subVmTest1 --type network --mac fe:54:00:a5:f8:99
```
and that seems to be reliable.

I checked removing the interface on the CLI instead the UI:

```diff
--- test/check-machines-nics
+++ test/check-machines-nics
@@ -674,7 +676,8 @@ class TestMachinesNICs(VirtualMachinesCase):
                 self.browser.enter_page('/machines')
                 self.browser.wait_in_text("body", "Virtual machines")
             else:
-                self.deleteIface(self.nic_num)
+                # self.deleteIface(self.nic_num)
+                self.machine.execute(f"virsh detach-interface --mac {self.mac} --domain subVmTest1 --type {self.source_type}")
```

that fails in the same way:
```
+ virsh detach-interface --mac 52:54:00:a5:f8:02 --domain subVmTest1 --type network    
[... no events at all ... ]
timeout
```

However, I notice that the `virsh detach-interface` command took 5s, and

[...]

PR #1050, opened 2023-04-26
fmf: Drop qemu-virtiofsd from Fedora 39 on

martinpitt:
The package got merged back into QEMU [1]. Possibly this was overzealous, but it's the status quo. See [2] for details.

[1] https://src.fedoraproject.org/rpms/qemu/c/f2cb56a43cfeb660682837ccd90b58c276d2add2
[2] https://bugzilla.redhat.com/show_bug.cgi?id=2189368

---

This should fix the recent TF rawhide install failures: https://artifacts.dev.testing-farm.io/0f5ee502-36a5-4f23-99a0-c8becd340e77/

PR #1058, opened 2023-05-03
Fix testNICAdd and rawhide "CDP broken" failures

martinpitt:
This fixes [testNICAdd](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1056-20230502-181230-073252a0-fedora-38/log.html#61), which has an impressive failure rate right now, see #1057

It also fails the "CDP broken" parts of [TF rawhide](https://artifacts.dev.testing-farm.io/30382c0f-df95-47c6-9733-e8da1d80ce79/). This  is exactly the [same workaround as in Cockpit](https://github.com/cockpit-project/cockpit/commit/9038b849c97f1713f6b38e81dcd299991d26d9af).

There are more failed tests, at least `TestMachinesLifecycle.testBasic` and `TestMachinesFilesystems.testBasicSystemConnection`, but I'll look at them separately.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -29,6 +29,12 @@ if grep -q platform:el8 /etc/os-release; then
     dnf module switch-to -y nodejs:16
 fi
 
+# HACK: nodejs 20 is very segfaulty: https://bugzilla.redhat.com/show_bug.cgi?id=2190075
+if grep -q VERSION_ID=39 /etc/os-release; then
+    dnf install -y nodejs18
+    ln -sfn node-18 /usr/bin/node
+fi
+
 # RHEL/CentOS 8 and Fedora have this, but not RHEL 9; tests check this more precisely
 $DNF libvirt-daemon-driver-storage-iscsi-direct || true
 
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -676,8 +676,9 @@ class TestMachinesNICs(VirtualMachinesCase):
             else:
                 self.deleteIface(self.nic_num)
 
-                # Check NIC is no longer in list
-                self.browser.wait_not_present(f"#vm-subVmTest1-network-{self.nic_num}-mac")
+                # Check NIC is no longer in list; the device-removed signal can take *very* long to arrive
+                with self.browser.wait_timeout(90):
+                    self.browser.wait_not_present(f"#vm-subVmTest1-network-{self.nic_num}-mac")
                 self.browser.wait_not_present(".pf-c-modal-box")
 
 


martinpitt:
[TF rawhide run](https://artifacts.dev.testing-farm.io/1351e501-7a9e-4316-af6d-094efd02b802/) is better -- basic plan succeeded, and the "CDP broken" errors are gone. Much more work to be done, of course!

martinpitt:
The remaining big TF failures are `TestMachinesLifecycle.testBasic`, which I can't reproduce in a local `tmt` run on either Fedora 38 or rawhide. I suppose I'll send a debugging PR which enables debug logging.

But by far the worst offender in this round is `TestMachinesNetworks.testNetworksCreate`, which failed absolutely everywhere three times. At least that one reproduces effortlessly.

mvollmer (approved):

test/browser/browser.sh
```diff
@@ -29,6 +29,12 @@ if grep -q platform:el8 /etc/os-release; then
     dnf module switch-to -y nodejs:16
 fi
 
+# HACK: nodejs 20 is very segfaulty: https://bugzilla.redhat.com/show_bug.cgi?id=2190075
+if grep -q VERSION_ID=39 /etc/os-release; then
+    dnf install -y nodejs18
```
KKoukiou, line 34: dnf module switch-to does not work?

martinpitt: This is Fedora, so no modules. The nodejs:16 above is for RHEL 8.

martinpitt:
The last rawhide failure is `TestMachinesFilesystems.testBasicSystemConnection`, which I can reproduce locally with tmt. The [screenshot](https://artifacts.dev.testing-farm.io/5ac3549b-c0ac-4a1a-86a1-21801bd5f2e6/work-storagef1dxpm4q/plans/all/storage/execute/data/test/browser/storage/data/TestMachinesFilesystems-testBasicSystemConnection-fedora-39-localhost-22-FAIL.png) does not show the "Shared directories" section, but the [HTML dump](https://artifacts.dev.testing-farm.io/5ac3549b-c0ac-4a1a-86a1-21801bd5f2e6/work-storagef1dxpm4q/plans/all/storage/execute/data/test/browser/storage/data/TestMachinesFilesystems-testBasicSystemConnection-fedora-39-localhost-22-FAIL.html) shows that it does exist. But it isn't disabled. This is still the case when I log into a fresh Cockpit session in firefox (with port forwarding).

The problem is, I think, that the test runs the VM and expects it to start. But in the test's screenshot, the state is "Failed". This happens in my interactive browser as well, due to "operation failed: Unable to find a satisfying virtiofsd". Indeed we just dropped that in #1050 as the `qemu-virtiofsd` package does not exist any more. Curiously I actually *can* install qemu-virtiofsd again, it seems it's now an alias for the `virtiofsd` package. So it seems they fixed that up.

I'll land this PR, as it's already approved and helps a lot, and send another PR to fix this remaining test case.

PR #1059, opened 2023-05-03
fmf: Put back virtiofsd on Rawhide

martinpitt:
`TestMachinesFilesystems.testBasicSystemConnection` depends on a functioning virtiofsd, so this test got broken on Rawhide with commit 673147c52c41d.

Turns out that there is now a standalone Rust replacement "virtiofsd" which exists on all Fedoras. Use that for testing instead of the package from qemu, and extend our Suggests for either variant.

---

This addresses the last [rawhide failure](https://artifacts.dev.testing-farm.io/5ac3549b-c0ac-4a1a-86a1-21801bd5f2e6/). I'm testing with virtiofsd on all Fedoras first, and if that fails, go back to the version specific install.

### DIFF
--- a/packaging/cockpit-machines.spec.in
+++ b/packaging/cockpit-machines.spec.in
@@ -63,7 +63,7 @@ Requires: libvirt-dbus >= 1.2.0
 Recommends: virt-install >= 3.0.0
 Recommends: libosinfo
 Recommends: python3-gobject-base
-Suggests: qemu-virtiofsd
+Suggests: (qemu-virtiofsd or virtiofsd)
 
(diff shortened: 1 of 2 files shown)


martinpitt:
Still [one failure of TestMachinesLifecycle.testBasic](https://artifacts.dev.testing-farm.io/d4199701-6e60-4b77-b24b-9656d940dfd9/), but this is flaky enough to at least have a chance of succeeding -- so we don't have to block the release on it. I'll look at that next.

mvollmer (approved):

PR #1060, opened 2023-05-03
XXX debug TestMachinesLifecycle.testBasic on TF

martinpitt:
I cannot reproduce [this failure](https://artifacts.dev.testing-farm.io/d4199701-6e60-4b77-b24b-9656d940dfd9/) in a local tmt VM, I tried f-38 and rawhide. Let's debug this printf-style then.


martinpitt:
The [logs](https://artifacts.dev.testing-farm.io/57f2b84b-9d6d-432c-b543-b980e543be6f/) I get there seem fine:

> debug: signal on /org/libvirt/QEMU: org.libvirt.Connect.DomainEvent(["/org/libvirt/QEMU/domain/_eeb74ccf_9f3e_47e0_9708_a5557f45920f",3,0])
> debug: reducer vms: action={"type":"UPDATE_VM","vm":{"connectionName":"system","id":"/org/libvirt/QEMU/domain/_eeb74ccf_9f3e_47e0_9708_a5557f45920f","state":"paused"}

[DomainEvent](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainEventType) 3 means `VIR_DOMAIN_EVENT_SUSPENDED`, i.e. exactly what we expect, and the UPDATE_VM follows along with `"state":"paused"`. However, the next thing that happens is this, which resets it back to `running`:

> debug: reducer vms: action={"type":"UPDATE_ADD_VM","vm":{"inactiveXML":{...},"connectionName":"system","id":"/org/libvirt/QEMU/domain/_a5c12289
_84a9_4931_8fa8_9671d1c1a1de","state":"running","name":"subVmTest1","persistent":true,"autostart":false,"ui":{"initiallyExpanded":false,"initiall
yOpenedConsoleTab":false},"capabilities":{...},"osType":null,"osBoot":[{"order":1,"type":"disk"},{"order":2,"type":"network"}],"firmware":null,"a
rch":"x86_64","currentMemory":131072,"memory":131072,"memoryBacking":false,"vcpus":{"count":"1","placement":"static","max":"1"},"disks":{...},"em
ulatedMachine":"pc-i440fx-8.0","cpu":{"mode":"custom","model":"qemu64","topology":{}},"displays":[],"interfaces":[...],"redirectedDevices":[],"ho
stDevices":[],"filesystems":[],"watchdog":{},"metadata":{"hasInstallPhase":false,"installSourceType":null,"installSource":null,"osVariant":null,"
rootPassword":null,"userLogin":null,"userPassword":null}}}
> debug: reducer vms: action={"type":"DELETE_UI_VM","vm":{"name":"subVmTest1","connectionName":"system"}}    

So I don't think there's something wrong with libvirt, but with our state updating logic?

martinpitt:
@skobyda , @KKoukiou : Does the above ring a bell? Did you happen to look into something similar recently?

martinpitt:
It's suspicious that that domainGet() calls both updateOrAddVm() and clearVmUiState() right after another, and that happens *before* logging the GET_VM  event. So it feels like there is a domainGet() in progress, which somehow never finishes?

martinpitt:
I added the fix from #1061 to give this some intense testing.

martinpitt:
All green! :+1:

PR #1061, opened 2023-05-03
api: fetch domain state as last when re-fetching all domain propertie…

KKoukiou:
… from the api

There is a race where domainGetAPI is called, it reads the state of a VM as running and then execute all the other API calls for parsing the various domain properties. At this point we receive an event that the domain got suspended and we update the domain state in the store to 'paused', with UPDATE_VM reducer. At this point all API calls are now finished from the domainGetAPI and the method also cals UPDATE_ADD_VM which however sets in the store the old state, as it was prematurely fetched.

With this commit we make sure that fetching the domain state always happens as last when reading the properties of a domain.

 - [x] Ensure that this works with the torture testing in #1060

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -671,18 +671,10 @@ export function domainGet({
             .then(domInactiveXml => {
                 const dumpInactiveXmlParams = parseDomainDumpxml(connectionName, domInactiveXml[0], objPath);
                 props.inactiveXML = dumpInactiveXmlParams;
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! I have a gut feeling that there will still be race conditions with that, as no matter in which order you collect multiple properties from multiple API calls, there could always be a signal in between. That's difficult to avoid. However, tests look good, and this at least won't make matters worse.

I added this fix to #1060 to give it some torture testing, let's see how well it holds up there.

martinpitt:
Running out of patience on the c8s run in #1060, but the other contexts look very good!

PR #1063, opened 2023-05-04
Some UI fixes

KKoukiou:


src/components/aggregateStatusCards.jsx
```diff
@@ -44,11 +44,13 @@ export class AggregateStatusCards extends React.Component {
...
-                            {cockpit.format(cockpit.ngettext("$0 Storage pool", "$0 Storage pools", this.props.storagePools.length), this.props.storagePools.length)}
-                        </Button>
-                        <Flex>
+                        <Flex alignItems={{ default: 'alignItemsCenter' }}>
+                            <ServerIcon size="md" />
+                            <Button onClick={() => cockpit.location.go(['storages'])} variant="link">
+                                {cockpit.format(cockpit.ngettext("$0 Storage pool", "$0 Storage pools", this.props.storagePools.length), this.props.storagePools.length)}
+                            </Button>
+                        </Flex>
+                        <Flex alignItems={{ default: 'alignItemsCenter' }}>
```
skobyda: How does this not break any pixel tests? didn't we have any pixel test which has shown those cards? If so, then they should be added

KKoukiou: haah ok reposting with pixel test

PR #1065, opened 2023-05-04
Return a promise from networkGet and storagePoolGet functions

KKoukiou:
So that getApiData can correctly return when all nested promises are also resolved.


martinpitt (approved):
:tada: let's hope this fixes some more flakes as a by-product!

PR #1066, opened 2023-05-04
Hide console card if there are no consoles present for the VM

KKoukiou:
* [x]  needs the scrollbar hide hack from main repo

We don't provide the users a way to add/remove consoles from the UI yet.

It's valid configuration to define VMs without console. In this case hide the Consoles card.


skobyda (commented):
This looks good to me, but since it's a draft, it's still waiting for some changes?

Edit: Oh wait, just noticed the "blocked" label, nevermind

skobyda (approved):
LGTM

PR #1073, opened 2023-05-10
Use 'type' instead of 'which' which is standard unix

KKoukiou:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=2196701

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -177,7 +177,7 @@ class AppActive extends React.Component {
         if (this.props.error)
             this.onAddErrorNotification({ text: _("Failed to fetch some resources"), detail: this.props.error });
 
-        cockpit.spawn(['which', 'virt-install'], { err: 'ignore' })
+        cockpit.script('type virt-install', { err: 'ignore' })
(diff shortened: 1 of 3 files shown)


martinpitt (approved):
Right, `type` or `command`. Thanks for figuring this out! I'm glad it was something "easy" after all.

KKoukiou:
There is a problem with ubuntu, debian, arch, will check it tomorrow.

jelly:
> There is a problem with ubuntu, debian, arch, will check it tomorrow.

Well type does not really exists as binary right, it's a shell buildin. So you should probaly use `cockpit.shell` instead of spawn or `/bin/sh -c`?

It works on Fedora because that has `/usr/bin/type`

/usr/bin/type: a /usr/bin/sh script, ASCII text executable

martinpitt:
Ah yes -- cockpit.shell() sounds good then. `virt-install --version` would also work, but as it's Python, that's a lot more expensive.

martinpitt (approved):
Danke!

PR #1076, opened 2023-05-15
test: Debian 11 does not support target.hotplug either

martinpitt:
This was previously shadowed because the test got skipped on known issue https://github.com/cockpit-project/bots/issues/1543

---

Fixes [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4762-20230514-225923-f68610a8-debian-stable-cockpit-project-cockpit-machines/log.html) in https://github.com/cockpit-project/bots/pull/4762 . It failed the same way before, but got naughtied, like [in this run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1073-20230511-092016-77e25d29-debian-stable/log.html).

### DIFF
--- a/test/check-machines-settings
+++ b/test/check-machines-settings
@@ -644,10 +644,15 @@ class TestMachinesSettings(VirtualMachinesCase):
         # Check rebooting machine will not unexpectedly affect watchdog configuration
         setWatchdogActionLive(action="poweroff", reboot_machine=True)
 
-        # Older libvirt in rhel-8-9 and centos-8-stream doesn't support target.hotplug=off
-        if not m.image.startswith("rhel-8") and m.image != "centos-8-stream":
+        m.execute("virsh destroy subVmTest1")
+
+        # Some OSes don't support target.hotplug=off
+        if m.image.startswith("rhel-8") or m.image in ["centos-8-stream", "debian-stable"]:
+            self.assertIn("Unknown --controller options", m.execute(
+                "! virt-xml subVmTest1 --edit model=pci-root --controller target.hotplug=off 2>&1")
+            )
+        else:
             # Disable hotplugging for subVmTest1
-            m.execute("virsh destroy subVmTest1")
             m.execute("virt-xml subVmTest1 --edit model=pci-root --controller target.hotplug=off")
             # Cleanup watchdog configuration
             m.execute("virt-xml subVmTest1 --remove-device --watchdog 1")


skobyda (approved):
LGTM, thanks

test/check-machines-settings
```diff
@@ -644,8 +644,8 @@ class TestMachinesSettings(VirtualMachinesCase):
         # Check rebooting machine will not unexpectedly affect watchdog configuration
         setWatchdogActionLive(action="poweroff", reboot_machine=True)
 
-        # Older libvirt in rhel-8-9 and centos-8-stream doesn't support target.hotplug=off
-        if not m.image.startswith("rhel-8") and m.image != "centos-8-stream":
+        # Older libvirt doesn't support target.hotplug=off
+        if not m.image.startswith("rhel-8") and m.image not in ["centos-8-stream", "debian-stable"]:
             # Disable hotplugging for subVmTest1
             m.execute("virsh destroy subVmTest1")
             m.execute("virt-xml subVmTest1 --edit model=pci-root --controller target.hotplug=off")
```
skobyda, line 651: On the second thought, I just realized maybe we should have something that tells us that this condition is no longer needed. With rhel-8 and centos-8-stream that was simple, we would remove the condition once we will not support those versions. But since "debian-stable" does not contain version in its name, than here is a chance that the condition will linger forever.
So maybe we should also add condition to check that version of libvirt is indeed too log. Accoding to [libvirt notes](https://libvirt.org/news.html), the target.hotplug functionality was added at 7.9.0.:

```suggestion
        # Older libvirt doesn't support target.hotplug=off
        if m.image.startswith("rhel-8") or m.image in ["centos-8-stream", "debian-stable"]:
            self.assertLess(m.execute("virsh --version"), "7.9.0")
        else:
            # Disable hotplugging for subVmTest1
            m.execute("virsh destroy subVmTest1")
            m.execute("virt-xml subVmTest1 --edit model=pci-root --controller target.hotplug=off")
```

skobyda (changes requested):

skobyda (dismissed):
LGTM, once tests pass, let's merge it

martinpitt:
Interesting, now the test fails o [rhel-8-9](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1076-20230515-081850-c223215c-rhel-8-9/log.html) and [c8s](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1076-20230515-081850-c223215c-centos-8-stream/log.html) because libvirt is at 8.0.0.

skobyda:
> Interesting, now the test fails o [rhel-8-9](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1076-20230515-081850-c223215c-rhel-8-9/log.html) and [c8s](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1076-20230515-081850-c223215c-centos-8-stream/log.html) because libvirt is at 8.0.0.

Oh, so I guess they updated libvirt version there already. At least we know that the check works and will inform us for other distro too

martinpitt:
I locally ran the test on c8s with libvirt 8.0.0, and it still fails on

    ERROR    Unknown --controller options: ['target.hotplug']

So maybe that's less of a version thing, and more of a "feature flag at build" thing? I reworked the check. For some reason my machine is acting up and I couldn't locally run the updated check, let's see what CI says.

skobyda:
> I locally ran the test on c8s with libvirt 8.0.0, and it still fails on
> 
> ```
> ERROR    Unknown --controller options: ['target.hotplug']
> ```

This is not an libvirt error, but virt-xml error.

Ahh. So the hotplug option was added since libvirt 7.9.0. But virt-xml, which is used in the test `m.execute("virt-xml subVmTest1 --edit model=pci-root --controller target.hotplug=off")` has the support for target.hotplug option only since v4.0.0 . So now we are in the state where c8s has new libvirt, but not new enough virt-xml.

So I guess the better solution is to assert `virt-xml --version` < 4.0.0 . Or maybe assert both libvirt with `virsh --version` >= 7.9.0 and virt-xml since we can't guarantee which one will make it sooner to the debian-stable?

test/check-machines-settings
```diff
@@ -644,10 +644,15 @@ class TestMachinesSettings(VirtualMachinesCase):
...
-        # Older libvirt in rhel-8-9 and centos-8-stream doesn't support target.hotplug=off
-        if not m.image.startswith("rhel-8") and m.image != "centos-8-stream":
+        m.execute("virsh destroy subVmTest1")
+
+        # Some OSes don't support target.hotplug=off
+        if m.image.startswith("rhel-8") or m.image in ["centos-8-stream", "debian-stable"]:
+            self.assertIn("Unknown --controller options", m.execute(
+                "! virt-xml subVmTest1 --edit model=pci-root --controller target.hotplug=off 2>&1")
+            )
+        else:
```
skobyda, line 654: > So I guess the better solution is to assert virt-xml --version < 4.0.0 . Or maybe assert both libvirt with virsh --version < 7.9.0 and virt-xml since we can't guarantee which package update will make it sooner to the debian-stable?

Or this approach. That should also be fine

skobyda (approved):

martinpitt:
[This is unrelated](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1076-20230515-093758-490aa896-rhel-8-9/log.html#81-2), the usual "inner VM sometimes gets stuck on boot". So this seems to work fine.

PR #1081, opened 2023-05-17
Refresh volumes list upon mount of VM Create dialog and Volumes tab

skobyda:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2185235


src/components/storagePools/storagePoolVolumesTab.jsx
```diff
@@ -52,6 +53,17 @@ export class StoragePoolVolumesTab extends React.Component {
         return null;
     }
 
+    componentDidMount() {
+        console.log("componentDidMount");
```
KKoukiou, line 57: Please remove.

src/libvirtApi/storagePool.js
```diff
@@ -116,7 +116,8 @@ export function storagePoolGetAll({
                     return call(connectionName, path, 'org.freedesktop.DBus.Properties', 'Get', ['org.libvirt.StoragePool', 'Active'], { timeout, type: 'ss' })
                             .then(active => {
                                 if (active[0].v)
-                                    return storagePoolRefresh({ connectionName, objPath: path });
+                                    return storagePoolRefresh({ connectionName, objPath: path })
+                                            .then(() => storagePoolGet({ connectionName, id: path }));
```
KKoukiou, line 120: Why you need the Get after the Refresh? I think this is handled then by the refresh signal.

skobyda:
Closing this PR because this is not correct solution. Refreshing all the storage pools and volumes each time a "Disk add" dialog is opened takes unacceptable long time (especially if we have some storage on NFS).
On the other hand, adding "Refresh" button does to storage pool's page is against our design practices, where we try to maintain the illusion that all the data in cockpit is always up to date.
But looks like with [this issue](https://issues.redhat.com/browse/RHEL-4043) we run into our limitations where we cannot keep data about storage pools all the time up to date.

PR #1084, opened 2023-05-22
Show helper to set up SSH tunnel for Desktop Viewer 

### LINKED ISSUE #1078: desktop console show wrong address
Hello, the desktop console in cockpit-machine doesn't show the good address.... he show 127.0.0.1
image
So my cockpit is proxied over apache2 if this can help... anyone know how to have the good address ?

skobyda:
The first one of many upcoming changes, fixes to consoles.

What this fixes:
- Desktop viewer for situation when VMs are hosted on a remote server

What this doesn't fix, and  still need to be changed/fixed in the greater VNC redesign:
- [ ] Desktop viewer when VMs are hosted on secondary server which is channelled through primary server
- [ ] Embedded VNC viewer when VMs are hosted on secondary server which is channelled through primary server
- [ ] Situations where users want VNC server genuinely to listen on non-localhost addresses. In such cases we need to guide user to update firewall, open ports...
- [ ] Situation where user specifies multiple addresses where VNC is listening to (and only one of them is localhost)
- [ ] Situations where user already has some VM running on localhost which takes up th same port as the VM running on a remote server
- [ ] Solution for full screen VNC
- [ ] Optional: show user the virt-viewer command (e.g. `virt-viewer qemu+ssh://user@remoteaddress/system domainname`) which opens a VNC connection to a remote VM. The benefit is that virt-viewer command sets up SSH tunnel for you. Might be a simpler solution than downloading .vv file and setting up SSH tunnel manually (which we do now)

This PR adds a message to the details of Desktop Viewer if VM's <graphics> is listening on localhost, but VMs are located at  a remote server.
Screenshot from 2023-05-24 18-37-50

Fixes #1078

### DIFF
--- a/src/components/vm/consoles/desktopConsole.jsx
+++ b/src/components/vm/consoles/desktopConsole.jsx
@@ -17,9 +17,12 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 import React from "react";
+import { CodeBlock, CodeBlockCode } from "@patternfly/react-core/dist/esm/components/CodeBlock";
 import { DesktopViewer } from '@patternfly/react-console';
 
+import { getServerAddress, needsTunnel } from "./utils.js";
 import cockpit from "cockpit";
+import store from './../../../store.js';
 
 const _ = cockpit.gettext;
 
@@ -37,6 +40,11 @@ function fmt_to_fragments(fmt) {
 }
 
 const DesktopConsoleDownload = ({ vnc, spice, onDesktopConsole }) => {
+    // DesktopViewer prefers spice over vnc
+    const address = (spice && spice.address) || (vnc && vnc.address);
+    const serverAddress = getServerAddress();
+    const loggedUser = store.getState().systemInfo.loggedUser;
+
     return (
         <DesktopViewer spice={spice}
                        vnc={vnc}
@@ -57,6 +65,12 @@ const DesktopConsoleDownload = ({ vnc, spice, onDesktopConsole }) => {
                            <p>
                                {fmt_to_fragments(_("Clicking \"Launch remote viewer\" will download a .vv file and launch $0."), <i>Remote Viewer</i>)}
                            </p>
+                           {needsTunnel(address, serverAddress) && <p>
+                               {_("SSH tunnel needs to be set up on client:")}
+                               <CodeBlock>
+                                   <CodeBlockCode>{`ssh -L 5900:localhost:5900 -N ${loggedUser.name}@${serverAddress}`}</CodeBlockCode>
+                               </CodeBlock>
+                           </p>}
                            <p>
                                {fmt_to_fragments(_("$0 is available for most operating systems. To install it, search for it in GNOME Software or run the following:"), <i>Remote Viewer</i>)}
                            </p>
--- a/src/components/vm/consoles/utils.js
+++ b/src/components/vm/consoles/utils.js
@@ -1,0 +1,43 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2023 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+export function getServerAddress() {
+    return window.location.hostname;
+}
+
+export function isLocalhost(address) {
+    return address === "localhost" || address.startsWith("127");
+}
+
+export function isAmbigious(address) {
+    return isLocalhost(address) || ["0", "0.0.0.0"].includes(address);
+}
+
+// Get address where VNC or SPICE server is located
+export function getConsoleAddress(consoleDetails) {
+    let address = consoleDetails.address;
+
+    if (!address || isAmbigious(address))
+        address = getServerAddress();
+
+    return address;
+}
+
+export function needsTunnel(consoleAddress, serverAddress) {
+    return isLocalhost(consoleAddress) && !isLocalhost(serverAddress);
+}
--- a/test/check-machines-consoles
+++ b/test/check-machines-consoles
@@ -67,6 +67,8 @@ class TestMachinesConsoles(VirtualMachinesCase):
         b.wait_in_text('.pf-c-expandable-section__content',
                        'Clicking "Launch remote viewer" will download')
 
+        b.wait_not_in_text('.pf-c-expandable-section__content', 'SSH tunnel needs to be set up on client')
+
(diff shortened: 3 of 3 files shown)


src/components/vm/consoles/utils.js
```diff
@@ -0,0 +1,43 @@
...
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+export function getServerAddress() {
+    return window.location.hostname;
```
skobyda, line 20: @martinpitt Is there any way for me to mock window.location.hostname so I can test a situation where it's value is different from hostname/127.x.x.x ?

martinpitt: Not directly -- if you write to that attribute, you will actually change the page. But you can add a destructive test with `provision` to *actually* get a second machine, like you already did in test/check-machines-migrate. That's IMHO a better way to validate that this really works.

src/components/vm/consoles/utils.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function isLocalhost(address) {
+    return address === "localhost" || address.startsWith("127");
```
martinpitt, line 24: Do we need to be concerned about IPv6? `::1` and `localhost6` and such?

src/components/vm/consoles/utils.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function isAmbigious(address) {
```
martinpitt, line 27: Ambig*u*ous. But this is weirdly named, or at least needs to be explained in a comment. Neither "localhost" nor "0.0.0.0" are ambiguous, the latter just means "listen on all interfaces". But what is address "0"?

Perhaps "isRemoteIP()", and invert the logic?

src/components/vm/consoles/utils.js
```diff
@@ -0,0 +1,43 @@
...
+    return isLocalhost(address) || ["0", "0.0.0.0"].includes(address);
+}
+
+// Get address where VNC or SPICE server is located
+export function getConsoleAddress(consoleDetails) {
+    let address = consoleDetails.address;
```
martinpitt, line 33: This looks like a server *listening* address then, judging by the test that you run on it. Would this ever be an actually real address? Even if it's interface specific, it's usually a netmask then, such as 10.1.2.0/24

src/components/vm/consoles/utils.js
```diff
@@ -0,0 +1,43 @@
...
+
+    if (!address || isAmbigious(address))
+        address = getServerAddress();
```
martinpitt, line 36: I don't understand why the UI wouldn't just always show window.location.hostname -- in what situations would that be wrong?

It could be if Cockpit is running behind a reverse proxy -- but then you have no chance of figuring out the real server name/IP anyway, and the UI will be wrong. But I don't see much that you can do about it.

src/components/vm/consoles/desktopConsole.jsx
```diff
@@ -37,6 +40,11 @@ function fmt_to_fragments(fmt) {
 }
 
 const DesktopConsoleDownload = ({ vnc, spice, onDesktopConsole }) => {
+    // DesktopViewer prefers spice over vnc
+    const address = (spice && spice.address) || (vnc && vnc.address);
```
martinpitt, line 44: `spice?.address || vnc?.address`

src/components/vm/consoles/desktopConsole.jsx
```diff
@@ -57,6 +65,12 @@ const DesktopConsoleDownload = ({ vnc, spice, onDesktopConsole }) => {
                            <p>
                                {fmt_to_fragments(_("Clicking \"Launch remote viewer\" will download a .vv file and launch $0."), <i>Remote Viewer</i>)}
                            </p>
+                           {needsTunnel(address, serverAddress) && <p>
+                               {_("SSH tunnel needs to be set up on client:")}
+                               <CodeBlock>
+                                   <CodeBlockCode>{`ssh -L 5900:localhost:5900 -N ${loggedUser.name}@${serverAddress}`}</CodeBlockCode>
```
martinpitt, line 71: This is actually really nice! Even if the serverAddress is slightly incorrect (due to reverse proxy and such), it will at least give admins an idea what needs to happen. Same for the local port -- it might be busy, and this doesn't work if you run this for more than one VM (you need to pick a different local port then), but this is meant as a guideline.

However, is it always 5900 on the remote side? Are VNC and spice using the same port? Can we ask libvirt about the port?

test/check-machines-consoles
```diff
@@ -67,6 +67,8 @@ class TestMachinesConsoles(VirtualMachinesCase):
         b.wait_in_text('.pf-c-expandable-section__content',
                        'Clicking "Launch remote viewer" will download')
 
+        b.wait_not_in_text('.pf-c-expandable-section__content', 'SSH tunnel needs to be set up on client')
```
martinpitt, line 70: Needs a positive test case, too. You already asked about this, just opening a thread to track it.

martinpitt (changes requested):
Thanks! This is a helpful guide indeed, and better than trying to set this up automatically.

You also mentioned something like telling the VNC viewer to directly connect to something like a qemu+ssh://server/.. address. This could be shown as an alternative?

src/components/vm/consoles/utils.js
martinpitt: As you just said in the chat, this won't help, as the browser runs outside the VMs and the cockpit port *always* gets tunneled to 127.0.0.2. We can't run a browser inside the VM. So this is actually exactly the unfixable [cockpit behind a reverse proxy](https://github.com/cockpit-project/cockpit-machines/pull/1084#discussion_r1205002727) scenario below.

But there's a workaround: You can restrict the "localhost" check to 127.0.0.1, treat any other 127.* address as "remote", and then have one test set up a `socat` from 127.0.0.2:cockpit_port to 127.0.0.1:5555 (or any other port which is definitively not used); then that test can connect the browser to that forwarded port and it'll appear as truly "localhost". See e.g. [this cockpit test](https://github.com/cockpit-project/cockpit/blob/main/test/verify/check-connection#L940) where we do this. The caveat is that we run socat in the VM in these tests. But (1) our [tasks container has socat](https://github.com/cockpit-project/cockpituous/blob/main/tasks/Dockerfile#L52), and (2) you can also use `ssh -L` to tunnel the port (which is incidentally the very thing that this PR advertises, so perhaps that's better :grin: )

src/components/vm/consoles/utils.js
skobyda: Yeah that is an address at which a VNC server is listening to. 
By default, it's limited to localhost, so if the host is connected to the wider network ad has some IP assigned, nobody should be able to just connect to a VNC of some VM which is running on the server.
But of course, the user might want the VM's VNC to be exposed to connections coming from the outside. Then they might configure this address to the address of the host.

src/components/vm/consoles/utils.js
skobyda: Server can have multiple IP addresses, and VNC server can be configured to listen only to one specific IP address.
But I'm overthinking here, in the vast majority of cases console address is either going to be  a localhost (in which case user needs to make a tunnel) or it's going to be a host's IP address (in which case VNC should work fine without a tunnel)

So I think I'll drop this logic here, and see if somebody asks for it in the future

mvollmer:
This is obsolete now. We have https://issues.redhat.com/browse/COCKPIT-1228 now for the `virt-viewer qemu+ssh://` bit.

PR #1091, opened 2023-05-26
Don't show empty columns in vNIC table

skobyda:
Just a small fix which was always bothering me. Don't show the column in Network Interfaces table when the column is empty:

### Before
Screenshot from 2023-05-26 12-32-52

### After
Screenshot from 2023-05-26 12-33-05


skobyda:
All tests green

martinpitt (approved):
Thanks!

Issue #1090, opened 2023-05-25
The default SPICE listening address will block the autostart of VMs in environments without routable IPv4 address


fybmain:
I encountered this issue when I was setting up a PC as the soft router of my home lab. I created a VM running OpenWRT by cockpit-machines and let this VM manage all physical NICs via PCIe passthrough to maximize the network performance. However, after I checked the 'Autostart' option in the Cockpit panel to make it an autostart service, the VM failed to start on system bootup. I managed to login back into the machine again and found the failure log below.
```
internal error: Failed to autostart VM 'OpenWRT': internal error: process exited while connecting to monitor: 2023-05-25T19:28:34.603591Z qemu-system-x86_64: warning: Spice: ../server/reds.cpp:2551:reds_init_socket: getaddrinfo(127.0.0.1,5901): Address family for hostname not supported
2023-05-25T19:28:34.603607Z qemu-system-x86_64: warning: Spice: ../server/reds.cpp:3442:do_spice_init: Failed to open SPICE sockets
2023-05-25T19:28:34.603612Z qemu-system-x86_64: failed to initialize spice server
```
I searched for the solution to this problem but found that this was [a bug from the "stone age"](https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621). This was actually caused by `getaddrinfo` which could not work on IPv4 loopback addresses when there were no routable IPv4 addresses configured in the system. It was hard to see the code fix in the foreseeable future since there has already existed a workaround by configuration. I tried the [workaround](https://www.reddit.com/r/archlinux/comments/n1ra5y/vm_always_fails_to_start_but_starts_if_tried/) by configuring the `spice_listen` in `/etc/libvirt/qemu.conf` as `::1`, which should work because it was an IPv6 address. But the workaround didn't take any effect.
After checking here and there, I finally found that there existed a configuration below of the spice listening address in the auto-generated domain(VM) configuration file.
```
    <graphics type='spice' autoport='yes' listen='127.0.0.1'>
      <listen type='address' address='127.0.0.1'/>
      <image compression='off'/>
    </graphics>
```
I changed the listening address here to `::1` and it worked. Now my VM can autostart when the host system boots.
Domain configuration file after change:
```
    <graphics type='spice' autoport='yes' listen='::1'>
      <listen type='address' address='::1'/>
      <image compression='off'/>
    </graphics>
```
[...]

PR #1092, opened 2023-05-26
Fix the default SPICE listening address as a workaround of the VM autostart failure

fybmain:
Resolves the issue #1090

### DIFF
--- a/src/scripts/install_machine.py
+++ b/src/scripts/install_machine.py
@@ -46,7 +46,8 @@ def get_graphics_capabilies(connection):
 
 def prepare_graphics_params(connection):
     graphics_config = {
-        'spice': {'listen': '127.0.0.1'},
+        # Use ipv6 as workaround for https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621
+        'spice': {'listen': '::1'},
         'vnc': {'listen': '127.0.0.1'}
     }
     try:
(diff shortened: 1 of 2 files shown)


skobyda:
Thanks. So far, this looks good. I triggered test, let's see if they pass

src/scripts/install_machine.py
```diff
@@ -46,7 +46,7 @@ def get_graphics_capabilies(connection):
 
 def prepare_graphics_params(connection):
     graphics_config = {
-        'spice': {'listen': '127.0.0.1'},
+        'spice': {'listen': '::1'},
         'vnc': {'listen': '127.0.0.1'}
```
skobyda, line 50: Could you add a comment to the bug, so in the future we know why we use ipv6 for spice

```suggestion
        # https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621
        'vnc': {'listen': '127.0.0.1'}
```

skobyda (changes requested):

src/scripts/install_machine.py
```diff
@@ -46,7 +46,7 @@ def get_graphics_capabilies(connection):
 
 def prepare_graphics_params(connection):
     graphics_config = {
-        'spice': {'listen': '127.0.0.1'},
+        'spice': {'listen': '::1'},
```
skobyda, line 49: Could you add a comment to the bug, so in the future we know why we use ipv6 for spice

```suggestion
        # Use ipv6 as workaround for https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621
        'spice': {'listen': '::1'},
```

skobyda (changes requested):

src/scripts/install_machine.py
fybmain: Wow, thank you!

skobyda (approved):
LGTM, thanks

PR #1094, opened 2023-05-30
Install into /usr/local/ by default

martinpitt:
/usr is package manager territory, and not even writable on OSTree based systems such as Fedora CoreOS or RHEL Edge. This has been common and good practice in autotools for decades.

You can still override this with `make install PREFIX=...`. Use that in the RPM spec file.

Cherry-picked from https://github.com/cockpit-project/starter-kit/commit/57d7c73d6e32

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ export TEST_OS
 TARFILE=$(RPM_NAME)-$(VERSION).tar.xz
 NODE_CACHE=$(RPM_NAME)-node-$(VERSION).tar.xz
 SPEC=$(RPM_NAME).spec
+PREFIX ?= /usr/local
 APPSTREAMFILE=org.cockpit-project.$(PACKAGE_NAME).metainfo.xml
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 # stamp file to check for node_modules/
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ export TEST_OS
 TARFILE=$(RPM_NAME)-$(VERSION).tar.xz
 NODE_CACHE=$(RPM_NAME)-node-$(VERSION).tar.xz
 SPEC=$(RPM_NAME).spec
+PREFIX ?= /usr/local
```
allisonkarlitskaya, line 12: I think `=` is fine here, actually.  You can still override it on the commandline of `make`.

packaging/debian/rules
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/make -f
 
+export PREFIX=/usr
```
allisonkarlitskaya, line 3: This is weird.  It should be passed as an arg to `make`.  Is that why you used `?=` ?

martinpitt: I used `?=` as that's coming from starter-kit, and it intuitively feels more correct. But keeping this local to the install rule is nicer indeed, I changed it.

Makefile
martinpitt: Hmm, it's an "external" variable, unlike the others which are internal and not meant to be overridden. But I suppose that's more of a visual distinction. "not from command line" is `override` I suppose, which is rarely used.

I temporarily hanged it to `=`, and it works. I suppose we should clean this up more thoroughly in starter-kit, probably with rearranging and commenting, so I changed it back for this PR (which is supposed to be a cherry-pick).

packaging/debian/rules
```diff
@@ -8,3 +8,6 @@ override_dh_auto_clean:
 
 override_dh_auto_test:
 	# don't call `make check`, these are integration tests
+
+override_dh_auto_install:
+	make install DESTDIR=debian/cockpit-machines PREFIX=/usr
```
allisonkarlitskaya, line 13: Surprised to see that there's not an easier way to do this...

allisonkarlitskaya (approved):
Thanks!

PR #1109, opened 2023-06-20
test: Adjust special cases for debian-stable moving to Debian 12 "bookworm"

martinpitt:
https://github.com/cockpit-project/bots/pull/4886 moves our "debian-stable" image from Debian 11 to 12. Adjust the test special cases.

---

This needs to land in lockstep with the bots PR.


jelly (approved):
That's a healthy set of bug fixes \o/

PR #1114, opened 2023-06-22
Configure and fix ruff errors

martinpitt:
The new tasks container got ruff, which now [causes many failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1114-20230622-081049-8f5d78e2-fedora-38/log.html).


martinpitt:
Seems we don't need pixel updates after all, just the ruff parts.

jelly (approved):

PR #1116, opened 2023-06-23
workflows: Move npm-update back to Ubuntu 20.04

jelly:
Similarly to reposchutz in Cockpit (commit 269bf89276c6), the npm-update* workflows pipeline large tar archives, which is broken in Ubuntu 22.04's version.

npm-update works here, but npm-update-pf fails https://github.com/cockpit-project/cockpit-machines/actions/runs/5307247859/jobs/9605610658

### DIFF
--- a/.github/workflows/npm-update-pf.yml
+++ b/.github/workflows/npm-update-pf.yml
@@ -10,7 +10,7 @@ jobs:
     permissions:
       pull-requests: write
       contents: write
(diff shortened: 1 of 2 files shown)


martinpitt (approved):

jelly:
Re-triggered rawhide to see if the new naughty get's picked up and it goes green.

PR #1118, opened 2023-06-23
Don't allow watchdog/vsock editing on transient VM

skobyda:
Libvirt doesn't allow live-editing watchdog and vsock device. Transient is a VM which only has a live configuration. Editing such device on such VM therefore doesn't make sense and would result in an exception, see https://bugzilla.redhat.com/show_bug.cgi?id=2213740


martinpitt (approved):
Thanks!

PR #1119, opened 2023-06-23
Fix margin for disabled listing actions buttons

skobyda:
Disabled buttons get wrapped in a <div> element. Therefore applying "machines-listing-actions > button" will not apply the styling.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2215597


martinpitt:
Looks good, thanks! Please [update the affected pixels](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1119-20230623-120435-296e9280-fedora-38/log.html) and also add the new one.

garrett (dismissed):
Thanks for catching this and fixing it!

garrett (dismissed):
PatternFly is working on switching button wrappers from div to span (for other reasons), but it won't fix this problem, and this PR is still relevant and the correct approach to fix the problem we're seeing here.

So, still a thumbs up from me. :+1:

jelly:
This needs a rebase for the new pixel tests.

martinpitt (approved):
Thanks!

PR #1122, opened 2023-06-26
Cores should equal to number of vCPUs if no topology is specified

skobyda:
Sometimes VMs can have no topology specified in their XML (e.g. when a new VM is created). In such cases, libvirt automatically assigns a topology where a number of cores is equal to a number of vcpus. We should mimic that in our UI

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2213064


martinpitt (approved):
Dakujem!

PR #1123, opened 2023-06-27
Set mode to bridge when the type of the NIC is direct in some extra scenarios

yunmingyang:
Related with: https://bugzilla.redhat.com/show_bug.cgi?id=2217832

### DIFF
--- a/src/components/create-vm-dialog/pxe-helpers.js
+++ b/src/components/create-vm-dialog/pxe-helpers.js
@@ -135,8 +135,11 @@ export function getPXEInitialNetworkSource(nodeDevices, virtualNetworks) {
         virtualNetworks
     );
 
+    // Don't use localhost as default PXE source
+    const source = netNodeDevices.find(dev => dev.capability.interface !== "lo");
+
     if (netNodeDevices.length > 0)
-        return cockpit.format('type=direct,source=$0,source.mode=bridge', netNodeDevices[0].capability.interface);
+        return cockpit.format('type=direct,source=$0,source.mode=bridge', source.capability.interface);
 }
 
 /**
--- a/src/components/create-vm-dialog/pxe-helpers.js
+++ b/src/components/create-vm-dialog/pxe-helpers.js
@@ -136,7 +136,7 @@ export function getPXEInitialNetworkSource(nodeDevices, virtualNetworks) {
     );
 
     if (netNodeDevices.length > 0)
-        return cockpit.format('type=direct,source=$0', netNodeDevices[0].capability.interface);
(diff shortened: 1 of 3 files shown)


src/components/create-vm-dialog/pxe-helpers.js
```diff
@@ -136,7 +136,7 @@ export function getPXEInitialNetworkSource(nodeDevices, virtualNetworks) {
     );
 
     if (netNodeDevices.length > 0)
-        return cockpit.format('type=direct,source=$0', netNodeDevices[0].capability.interface);
+        return cockpit.format('type=direct,source=$0,source.mode=bridge', netNodeDevices[0].capability.interface);
```
skobyda, line 139: Hey. From what I understand, this specifies a source of the VM, meaning from where the VM will get installed (in this case, from network). This however doesn't specify virtual network interface which will persist as part of VM configuration once the VM gets installed right?
And from what I understand, main problem with using default VEPA mode, as according to https://bugzilla.redhat.com/show_bug.cgi?id=2095967, was that VMs with VEPA network interfaces could not talk to each other.
But since here we only specify source, and not network interface which may be used to communicate with other VMs, it's not a problem that we use VEPA here right?
It's just unless somebody complains that is is broken, I'm afraid to change something that is working. I'm afraid to change something which has been working without issues for years (i.e. if it works, don't touch it)
Or is there any other reason why using

cockpituous, line 139: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1123-20230627-082351-ccab151a-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

skobyda: Nevermind, I just looked how we call virt-install and the the code from where gets used as parameters for "--pxe --network", which are then used as vNIC of the newly created VM. So retracting my review

skobyda (changes requested):
This needs some test fixes

yunmingyang:
Hi @skobyda, fix failed tests, and I also add a case for covering the line mentioned by the bot. But there is a problem which is the installation source may be "lo" sometimes, then the new cases will fail. So maybe should we filter out "lo" from installation source?

skobyda:
> Hi @skobyda, fix failed tests, and I also add a case for covering the line mentioned by the bot. But there is a problem which is the installation source may be "lo" sometimes, then the new cases will fail. So maybe should we filter out "lo" from installation source?

Sure we can do that. But I'm wondering, why did this test failure never occurred when using macvtap?

skobyda (approved):

PR #1124, opened 2023-06-27
Run disk format detection also for CD/DVD discs

skobyda:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2215599

### DIFF
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -1186,7 +1186,7 @@ class TestMachinesDisks(VirtualMachinesCase):
         # https://bugzilla.redhat.com/show_bug.cgi?id=1977810
         self.VMAddDiskDialog(
             self,
-            expected_target='sda',
+            expected_target='sdb',
             file_path='/var/lib/libvirt/novell.iso',
             mode='custom-path'
         ).execute()
@@ -1221,7 +1221,7 @@ class TestMachinesDisks(VirtualMachinesCase):
         b.click(f"{prefix}-dialog-cancel")
         self.VMAddDiskDialog(
             self,
-            expected_target='sda',
+            expected_target='sdc',
             file_path='/var/lib/libvirt/novell.iso',
             mode='custom-path'
         ).execute()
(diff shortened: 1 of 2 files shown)


skobyda:
So there is a bug where it's appareant we just automatically set every CD/DVD disc to qcow2.
But a while ago, @jelly added format autodetection at https://github.com/cockpit-project/cockpit-machines/pull/646/ . But I see the autodetection is only limited to Disk image files:
Screenshot from 2023-06-27 12-13-08

But doesn't run for CD/DVD discs:
Screenshot from 2023-06-27 12-13-08

@jelly you wrote that code. Do you remember if there was any reason why you limited it only for Disk images, and didn't extend it to CD/DVD discs?

jelly:
@skobyda I can't recall why, but I think it was to as I never ever considered the cd/dvd image option. The option did exist back then.

test/check-machines-disks
```diff
@@ -1140,6 +1140,18 @@ class TestMachinesDisks(VirtualMachinesCase):
...
+        self.VMAddDiskDialog(
+            self,
+            device='cdrom',
+            bus_type='scsi',
+            expected_target='sdb',
+            expected_format='raw',
+            file_path='/var/lib/libvirt/images/foobar.raw',
+            mode='custom-path',
+        ).execute()
+
```
jelly, line 1154: Should we not also test that a valid cdrom is correctly detected? What happens if we attach an iso now?

```
[jelle@t14s][~/Downloads]%file webui_latest_install.iso
webui_latest_install.iso: ISO 9660 CD-ROM filesystem data (DOS/MBR boot sector) 'Fedora-38-BaseOS-x86_64' (bootable)
[jelle@t14s][~/Downloads]%head --bytes=16 webui_latest_install.iso
3퐐%
```

Or does QEMU expect it to be attached as raw?

skobyda: QEMU handles iso file as raw (which it is).
But adding a test case for iso file too and expecting it to be raw makes sense. 
I updated the PR with a new test

martinpitt (approved):
Ah, so an ISO file counts as "raw" format -- that's only relevant in the sense of "is it qcow or not"? Then this is straightforward indeed. Thanks!

PR #1125, opened 2023-06-27
Makefile: bump cockpit lib for pybridge allowed_messages for DBusChannel

jelly:
This currently occurs in the new Python bridge Arch quite often.


martinpitt:
That's stuck on the PF5 update, see PR #1103

PR #1141, opened 2023-07-07
onKeyPress is replaced with onKeyDown

themukeshmahiya:
trying to solve this good first issue

https://github.com/cockpit-project/cockpit/issues/19061


jelly (changes requested):
Please test the memory changes field, it does not seem to work and break all tests.

jelly:
https://github.com/cockpit-project/cockpit-machines/pull/1233

PR #1145, opened 2023-07-10
Drop noVerticalAlign from Icon's

jelly:
PF 5 dropped noVerticalAlign from all Icon's so this no longer does anything.


jelly:
See https://staging.patternfly.org/get-started/upgrade/release-notes

skobyda (approved):

PR #1153, opened 2023-07-18
package.json: Bump most npm modules

martinpitt:


martinpitt:
Meh, xterm 5.2.0 has a broken accessibility-tree, the contents does not appear anywhere. See https://github.com/xtermjs/xterm.js/issues/4571 . Going back to 5.1

PR #1165, opened 2023-07-26
build.js: support flags

jelly:
Adds flags to build.js to use rsync, disable linting and use watch mode.


martinpitt (changes requested):
Can you please also adopt the README bits from https://github.com/cockpit-project/starter-kit/pull/679/files ?

martinpitt (approved):
See https://github.com/cockpit-project/cockpit-podman/pull/1363#issuecomment-1652450599 ff.

PR #1168, opened 2023-07-31
test: Don't wait for VM to boot when only asserting uptime value

skobyda:
This fixes a flake where sometimes the value on the reference picture
    says that uptime is "1 minute" instead of expected "less than a minute".
    The reason why this race condition exists is that we after starting up
    the VM, we wait for booting up to finish `waitCirrOSBooted`. This however
    sometimes may take more than a minute.
    
    Fortunately, waiting for VM to boot here is unecessary, since we don't
    really do anything with the VM where we expect the OS to be responsive,
    so we can remote the wait which should eliminate the race condition.


skobyda:
Example of the flake: https://cockpit-logs.us-east-1.linodeobjects.com/pull-5058-20230731-102621-a09ef4d1-fedora-38-pybridge-cockpit-project-cockpit-machines/log.html#37

martinpitt (approved):
Makes sense, thanks!

PR #1169, opened 2023-07-31
Update pixel tests for fedora-38

skobyda:
It seems in new fedora-38, there is miniature difference in few pixels when redering numbers in modal title.


skobyda:
Example of the difference: https://cockpit-logs.us-east-1.linodeobjects.com/pull-5058-20230731-102625-a09ef4d1-fedora-38-cockpit-project-cockpit-machines/pixeldiff.html#TestMachinesNICs-testNICEdit-vm-edit-nic-modal

martinpitt:
Hmm, how is that specific to the tested image? That rather seems like a generic pixel flake to me, which depends on the browser, fonts, and possibly timing? as expected it [failed on the current fedora-38 image](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1169-20230731-121439-11142be5-fedora-38/log.html). You can try your luck with triggering a run against the refreshed image (it will then have to land in lockstep). If it succeeds, I'll look the other way and ack :grin:

skobyda:
> Hmm, how is that specific to the tested image? That rather seems like a generic pixel flake to me, which depends on the browser, fonts, and possibly timing? as expected it [failed on the current fedora-38 image](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1169-20230731-121439-11142be5-fedora-38/log.html). You can try your luck with triggering a run against the refreshed image (it will then have to land in lockstep). If it succeeds, I'll look the other way and ack grin

I always thought fonts are specific to operating system, since they are not distributed through npm packages, but are part of fedora packages (e.g. every "font" package in https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/r/ ). So maybe some change there could cause the diff?

Anyway, I don't exactly know what is change is causing the pixel diff, but I think it's small enough that we could not consider it regression and if it passes, just let it slide :smile: . Also I retrigerred the tests on image-refresh, the pixel diff persists, so I think it's not just a flake. Will trigger it here now

martinpitt:
> I always thought fonts are specific to operating system,

Ah, I know --  In general, the "operating system" is the cockpit/tasks container here. That's where the browser runs, where it pulls fronts from, etc. But the exception is the Red Hat font (which this one actually is), which is coming from cockpit-ws -- and *that* updated on the fedora-38 image. So this is actually correct for c-machines. Sorry, this wasn't entirely obvious!

skobyda:
`fedora-38@bots#5058` Passed, so let's merge it in uniform with https://github.com/cockpit-project/bots/pull/5058

martinpitt (approved):
Ack, thanks. Pre-approving, so that you can land these in lockstep once ready.

PR #1170, opened 2023-07-31
buildjs: fix rsync flag

tomasmatus:


jelly:
This change does not affect Packit tests so ignoring the F37 failure.

PR #1173, opened 2023-08-02
fmf: Don't disable SELinux message check globally

martinpitt:
It's useful to run cockpit-machines tests to gate changes to selinux-policy. But we can only do that if we actually pay attention to its messages.

The point of this is to not break cockpit-machines tests when they run as gating for other packages in Fedora/RHEL. But let's rather do that in the downstream dist-git test plan. We can inject an environment variable there [1], so just pass it on from the top-level entry point browser.sh to run-test.sh.

[1] https://tmt.readthedocs.io/en/latest/spec/tests.html#environment

---

I'll apply this to starter-kit and friends, too, but first wanted to see if it even passes.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -24,7 +24,7 @@ jobs:
     - fedora-development
     - centos-stream-8
     - centos-stream-9
-  
+
   - job: tests
     trigger: pull_request
     targets:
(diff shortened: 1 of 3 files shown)


jelly (approved):
Makes sense for the reverse dependency testing!

PR #1175, opened 2023-08-03
README: update virt-install's deprecated '--nographics' to '--graphic…

skobyda:
…s none'

Just a trivial change I run into when grepping for graphics options usage in our code


martinpitt (approved):
Dakujem!

PR #1178, opened 2023-08-14
Replace deprecated font redhat--monospace with basic PF's monospace

skobyda:
PF dropped pf-v5-global--FontFamily--redhat--monospace since patternfly/patternfly@4ec21e34fc3

Example (see "Disk identifier")
Screenshot from 2023-08-14 14-59-49

### DIFF
--- a/src/machines.scss
+++ b/src/machines.scss
@@ -81,7 +81,7 @@
 }
 
 .ct-monospace {
-    font-family: var(--pf-v5-global--FontFamily--redhat--monospace);
+    font-family: var(--pf-v5-global--FontFamily--monospace);
 }
 
 // Do not add a box-shadow to a "subsection", while it's technically correct it looks weird.
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -302,6 +302,7 @@ class TestMachinesDisks(VirtualMachinesCase):
             pool_name=None, volume_name=None,
             file_path=None,
             force=False,
+            pixel_tag=None,
         ):
             self.test_obj = test_obj
             self.mode = mode
@@ -311,6 +312,7 @@ class TestMachinesDisks(VirtualMachinesCase):
             self.volume_name = volume_name
             self.target = target
             self.force = force
+            self.pixel_tag = pixel_tag
 
         def _get_disks(self):
             m = self.test_obj.machine
@@ -402,6 +404,9 @@ class TestMachinesDisks(VirtualMachinesCase):
             b = self.test_obj.browser
             b.click(f"#vm-{self.vm_name}-disks-{self.target}-eject-button")  # button
             b.wait_in_text(".pf-v5-c-modal-box__title", "Eject disc from VM")
+
+            if self.pixel_tag:
+                b.assert_pixels(".pf-v5-c-modal-box", self.pixel_tag)
 
             if self.mode == "custom-path":
                 b.wait_in_text(f"#vm-{self.vm_name}-disks-{self.target}-modal-description-file", self.file_path)
@@ -1632,6 +1637,7 @@ class TestMachinesDisks(VirtualMachinesCase):
             mode='custom-path',
             target='sda',
             file_path='/var/lib/libvirt/novell.iso',
+            pixel_tag='vm-disc-eject-modal',
         ).execute()
 
         self.VMInsertMediaDialog(


skobyda:
Added pixel tests

garrett:
...is this screenshot the **before** or **after**? It's not clear. It has a bunch of problems (including font ones), which is why I'm asking.


---

The issues related to this PR:

- "Size" is a number; it should have a proprortional character. Does it? (It's hard to tell with just the "1" there, but I think it's supposed to have a bottom bar serif.)
- The font looks wrong. Is it using a bold version?

It should look like this (inside the box):

image
(From https://www.patternfly.org/design-foundations/typography/#our-font-family)

---

I see a few additional problems:

- What's up with the radio buttons?! The radios and text are all misaligned. Radios are too high and labels are too low.
   image

- The split is weird with labels on the side and the entries not constrained; it visually groups the wrong elements.
   image
  - There are a few ways to fix this: Stack the items vertically instead of a split, restricting the size of the "Size" and "Cache" fields so there's more of a gap between items, increase the gap size for the split, or shrink the label of "Format" and "Bus" (which would be inconsistent).

Restricting the width (we should probably do this with size anyway, just like we do elsewhere, to help indicate the relative size):

image

Shaving down the size of the labels looks like this, which is nice, but it's inconsistent and could be a problem with translations:

image

 
Bumping the gap to be another 1 PatternFly spacing units (equivalent to 16px) seems like the best approach here, perhaps:

image

---

We'll want to open up issues for the additional set of problems and focus the rest of this PR on the font issues.

jelly:
Needs a rebase for pixel test changes which landed after the Makefile update.

skobyda:
It looks like it's bold because patternfly uses RedHatMono for it, which seems to make it bold

skobyda:
> * The font looks wrong. Is it using a bold version?

It looks like it's bold because patternfly uses RedHatMono for it, which seems to make it bold

skobyda:
@garrett to avoid bold font, we can just use basic monospace:
Screenshot from 2023-08-21 14-05-00

skobyda:
Another update, so it seems for some reason, in cockpit the class `pf-v5-u-font-family-monospace` uses `Red Hat Mono Medium`, while on patternfly website, the same font is calculateed as `Red Hat Mono Light`, which as 2 degrees thinner/lighter than medium, as seen at https://github.com/RedHatOfficial/RedHatFont

skobyda:
I filed an issue with PF about not being able to select a lighter font: https://github.com/patternfly/patternfly/issues/5855

Until then, maybe we could work around this by defining our own font-weight, e.g. `300` (which was the value PF used for lighter font before it broke). So it could be something like
```
.ct-monospace-weight {
    font-weight: 300;
}
```

Then it would look like this:
Screenshot from 2023-08-22 12-15-22


@garrett WDYT?

garrett:
Why are you doing a 1-off patch to a PatternFly problem as a local and custom `ct-*` class? Shouldn't that be an override if it is a problem with PatternFly?

skobyda:
Ever since https://github.com/cockpit-project/cockpit/pull/19516 was landed, the weight of of monospace now looks correct:

Screenshot from 2023-10-31 02-11-55
Screenshot from 2023-10-31 02-11-41

martinpitt:
The change looks great, but tests [are](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1178-20231031-012029-ffb64b18-centos-8-stream/log.html#4-2) [a big](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1178-20231031-012029-ffb64b18-fedora-38/log.html#31-2) [train wreck](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1178-20231031-012029-ffb64b18-ubuntu-stable/log.html). Due to the many affected retries? I run them once more to compare.

martinpitt:
[pixel noise](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1178-20231031-044308-ffb64b18-fedora-38/log.html) in the new reference. Retrying to compare which one it will be.

martinpitt:
Yup, failed again in the same way, so I take it the original refs from this PR were outdated. I [pushed the new ones](https://github.com/cockpit-project/pixel-test-reference/compare/85f5d7aa3d629c80af2cdf8ac18fa1330c6b018f..21a39ee203617464c3c5a95e60b50d660b91bbb4)

martinpitt (approved):

martinpitt:
@jelly [f38](https://artifacts.dev.testing-farm.io/b9194c0c-db16-4342-bf7c-97183a526c78/#artifacts-/plans/all/storage) started to consistently fail `TestMachinesStoragePools.testStoragePoolsDeletion` on TF. Some regression in F38? Yesterday's landed PRs still seemed fine. Do you have time to investigate?

martinpitt:
The only recent plausible F38 update is [libvirt-dbus](https://koji.fedoraproject.org/koji/buildinfo?buildID=2313859)

PR #1190, opened 2023-08-25
packaging: add virtiofsd dependency for fedora-39

tomasmatus:


tomasmatus:
So this probably isn't the right way of doing it but I believe it has to be added here somehow 😄

martinpitt (changes requested):
That looks wrong -- c-machines is supposed to work without virtiofsd, that's why it's only a suggests, not a requires. The *tests* assume that it's installed, to test device passthrough.

tomasmatus:
ah I see so does that mean this can stay unchanged?

martinpitt:
Yes, I think so.

tomasmatus:
Closing then. As change isn't needed

PR #1156, opened 2023-07-20
test: Update our https server setup

skobyda:
Ever since python 3.12, ssl library got rid of already deprecated ssl.wrap_socket and instead replaced it with context.wrap_socket This required few changes to how we setup our http server. 

This should stop breaking fedora-rawhide test:

```
> error: spawn 'vm creation' returned error: "{"problem":null,"exit_status":1,"exit_signal":null,"message":"ERROR    internal error: process exited while connecting to monitor: 2023-07-18T08:24:43.721525Z qemu-system-x86_64: -blockdev {\"driver\":\"https\",\"url\":\"https://archive.fedoraproject.org:443/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso\",\"node-name\":\"libvirt-1-storage\",\"auto-read-only\":true,\"discard\":\"unmap\"}: CURL: Error opening file: Failed to connect to archive.fedoraproject.org port 443 after 0 ms: Couldn't connect to server\nDomain installation does not appear to have been successful.\nIf it was, you can restart your domain by running:\n  virsh --connect qemu:///system start subVmTestCreate7\notherwise, please restart your installation.\nTraceback (most recent call last):\n  File \"<stdin>\", line 352, in <module>\n  File \"<stdin>\", line 254, in create_vm\n  File \"/usr/lib64/python3.12/subprocess.py\", line 466, in check_output\n    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/lib64/python3.12/subprocess.py\", line 571, in run\n    raise CalledProcessError(retcode, process.args,\nsubprocess.CalledProcessError: Command '['virt-install', '--connect', 'qemu:///system', '--quiet', '--os-variant', 'fedora28', '--memory', '128', '--name', 'subVmTestCreate7', '--check', 'path_in_use=off', '--wait', '-1', '--noautoconsole', '--disk', 'none', '--graphics', 'vnc,listen=127.0.0.1', '--graphics', 'spice,listen=::1', '--cdrom', 'https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso']' returned non-zero exit status 1."}"
Traceback (most recent call last):
[...]

PR #1192, opened 2023-08-28
test: Update our RHSM server setup

skobyda:
Ever since python 3.12, ssl library got rid of already deprecated ssl.wrap_socket and instead replaced it with context.wrap_socket This required few changes to how we setup our http server. This should fix fedora-39's testCreateDownloadRhel


skobyda:
This is the same fix as at https://github.com/cockpit-project/cockpit-machines/pull/1156
I just forgot that we mock HTTP server also here

martinpitt (approved):
Cheers!

PR #1195, opened 2023-09-04
test: Use wait_text to check an empty string

skobyda:
Using wait_in_text(selector, "") always validates as true (since empty string is always subset of any string), rendering the test useless.

### DIFF
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -235,6 +235,19 @@ class VirtualMachinesCaseHelpers:
         else:
             return m.execute(cmd)
 
+    def deleteDisk(self, target, vm_name="subVmTest1"):
+        b = self.browser
+
+        b.wait_visible(f"#vm-{vm_name}-disks-{target}-device")
+        b.wait_visible(f"#vm-{vm_name}-disks-{target}-action-kebab")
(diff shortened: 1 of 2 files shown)


martinpitt:
> wait_js_cond(ph_text_is("#pool-myPoolTwo-system-volume-VolumeOne-usedby","")): Uncaught (in promise) Error: actual text: subVmTest1

... and it fails

skobyda:
> > wait_js_cond(ph_text_is("#pool-myPoolTwo-system-volume-VolumeOne-usedby","")): Uncaught (in promise) Error: actual text: subVmTest1
> 
> ... and it fails

So the test was really useless

test/machineslib.py
```diff
@@ -235,6 +235,19 @@ def run_admin(self, cmd, connectionName='system'):
         else:
             return m.execute(cmd)
 
+    def deleteDisk(self, target, vm_name="subVmTest1"):
```
martinpitt, line 238: You just defined this function without using this anywhere. Forgot to commit something?

martinpitt (changes requested):
Thanks for fixing the test properly!

martinpitt (approved):
Thanks!

PR #1197, opened 2023-09-05
test: drop pybridge allowed messages

jelly:
These landed in Cockpit a long time ago and our cockpit-lib should be up to date enough by now.

### DIFF
--- a/test/run
+++ b/test/run
@@ -11,12 +11,6 @@ export RUN_TESTS_OPTIONS=--track-naughties
 # linters are off by default for production builds, but we want to run them in CI
 export LINT=1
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
[rhel-9-3 failed](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1197-20230905-075345-8bf3fb5d-rhel-9-3/log.html#38) with an unexpected message "self.send_message(notify=notify)"; it's part of https://github.com/cockpit-project/cockpit/issues/18355 but curiously not one of the dropped messages here. So this needs fixing in Cockpit, and this is okay. Thanks!

PR #1200, opened 2023-09-06
Makefile: support the updates-testing scenario

jelly:
When we run with the `updates-testing` scenario install updates from Fedora's updates-testing repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,8 @@ DIST_TEST=dist/manifest.json
 COCKPIT_REPO_STAMP=pkg/lib/cockpit-po-plugin.js
 # common arguments for tar, mostly to make the generated tarballs reproducible
 TAR_ARGS = --sort=name --mtime "@$(shell git show --no-patch --format='%at')" --mode=go=rX,u+rw,a-s --numeric-owner --owner=0 --group=0
+# default customize_flags
+VM_CUSTOMIZE_FLAGS = --no-network
 
 all: $(DIST_TEST)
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Same issue with /pybridge: Preferably drop it, or fix it to add `--no-network`.

martinpitt (approved):

jelly:
@martinpitt should be good to merge no? centos-8-stream failure is unexpected but unrelated:

```
error: Failed to get libvirt version from the dbus API:Rejected send message, 1 matched rules; type="method_call", sender=":1.234" (uid=1000 pid=9865 comm="/usr/bin/python3 /usr/local/bin/cockpit-bridge " label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023") interface="org.freedesktop.DBus.Properties" member="Get" error name="(unset)" requested_reply="0" destination="org.libvirt" (uid=981 pid=2891 comm="/usr/sbin/libvirt-dbus --system " label="system_u:system_r:unconfined_service_t:s0-s0:c0.c1023")
warning: failed to poll tuned 
log: listInactiveInterfaces action for path /org/libvirt/QEMU/interface/52_3a54_3a00_3a12_3a34_3a56 XML error: Invalid value for attribute 'speed' in element 'link': '-1'. Expected integer value
log: listInactiveInterfaces action for path /org/libvirt/QEMU/interface/52_3a54_3a01_3a00_3a00_3a01 XML error: Invalid value for attribute 'speed' in element 'link': '-1'. Expected integer value
```

martinpitt:
Yes, let's! Can you please add a workflow to enable it?

PR #1206, opened 2023-09-12
WIP: Use '--tpm none' as workaround for broken virt-install+cloudinit

skobyda:


skobyda:
Just testing if this would break any tests on any distros. It's not a new function nor a fix, do not merge it

skobyda:
virt-manager managed to fix it upstream, so this workaround is unecessary: https://github.com/virt-manager/virt-manager/commit/ec434948a8384541c56bfa04e4985f4fc709bc76

PR #1212, opened 2023-09-18
test: add a case which are adding disks with all combianations of cache, bus and format

yunmingyang:


skobyda (changes requested):
arch and debian-testing are failing. Looks like the disk is not visible after attaching to the VM

yunmingyang:
Not reproduce locally, then add some debug log to see whether it helps

KKoukiou (commented):
What's the reason behind this? Was there some related bug? 

I would like to think this test suite as non-exhaustive, rather covering all possible code paths. By covering the grid map of all possible options, we just increase the time that tests run without necessarily increasing test code coverage. Does this make sense?

testAddDiskAdditionalOptions already tests all the options, just not the whole grid.
Testing that options don't break each other is something that could theoretically be part of the libvirt tests. Here we are testing that the UI works as expected, and not all possible libvirt VM configurations.

yunmingyang:
The reason is there is a manual case about this, and if executing this case manually, it will take about 1 ~ 2 hours, even more time. So automating it is the best practice. However, no directly related bug for this automated case, but, I think maybe we could use this case to prevent bugs like https://bugzilla.redhat.com/show_bug.cgi?id=1784297 or https://bugzilla.redhat.com/show_bug.cgi?id=1948366. Although, for now, this case just cover disks in the dir type storage pool, we could abstract it into a function just like _testBasic in the future, to cover more scenarios.

mvollmer:
> The reason is there is a manual case about this, and if executing this case manually, it will take about 1 ~ 2 hours, even more time.

What is the reason for this manual test?  It seems excessive, but if it is important, it should indeed be automated, and preferably in the libvirt testsuite.

PR #1219, opened 2023-09-27
test: Generalize OS special cases

martinpitt:
Replace the remaining specific rhel-8-* with pattern based matching.

---

This suffices for rhel-8-10.


jelly:
```
# test/check-machines-filesystems:21:37: F401 [*] `testlib.skipImage` imported but unused
```

jelly (approved):

PR #1224, opened 2023-09-29
Switch to dependabot for node_modules updates

jelly:
Pin our devDependencies, as they will now be updated via dependabot in a more controlled fashion.


martinpitt (approved):
Thanks!

PR #1233, opened 2023-10-02
replace onKeyPress with onKeyUp

jelly:
onKeyPress has been deprecated for a while, it is unlikely to be removed but it's nice to be already future proof.


skobyda (approved):
LGTM. As long as all tests pass

PR #1255, opened 2023-10-12
test: fix waiting for reboot in performAction()

yunmingyang:
That check for "the reboot does not happen when the dialog is open" was
broken:

 - It checked the log once right after opening the dialog without any
   delay, so neither the dialog nor the VM had any chance to actually do
   anything.

 - The caller clears the log file right before the call, so initialTime
   was always empty.

Wait 5s after opening the dialog to give the VM a good chance to boot (if
it would, which it isn't supposed to). Also add a positive log check that the
VM actually does start to boot after confirming the dialog.

Finally, move the cleaning of the log file from the callers into
performAction(), so that the whole logic is in one place.

Co-Authored-By: Martin Pitt <mpitt@redhat.com>

### DIFF
--- a/test/check-machines-lifecycle
+++ b/test/check-machines-lifecycle
@@ -160,14 +160,12 @@ class TestMachinesLifecycle(machineslib.VirtualMachinesCase):
         self.performAction("subVmTest1", "resume")
 
         # reboot
-        self.machine.execute(f"echo '' > {args['logfile']}")
-        self.performAction("subVmTest1", "reboot")
+        self.performAction("subVmTest1", "reboot", logPath=args['logfile'])
         self.waitLogFile(args['logfile'], "reboot: Power down")
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
 
         # force reboot
-        self.machine.execute(f"echo '' > {args['logfile']}")
-        self.performAction("subVmTest1", "forceReboot")
+        self.performAction("subVmTest1", "forceReboot", logPath=args['logfile'])
         self.waitCirrOSBooted(args['logfile'])
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
 
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -17,6 +17,7 @@
 
 import os
 import sys
+import time
 import traceback
 
 import netlib
@@ -42,21 +43,21 @@ class VirtualMachinesCaseHelpers:
         with self.browser.wait_timeout(30):
             self.browser.wait_in_text("body", "Virtual machines")
 
-    def performAction(self, vmName, action, checkExpectedState=True, connectionName="system"):
+    def performAction(self, vmName, action, checkExpectedState=True, connectionName="system", logPath=None):
         b = self.browser
         m = self.machine
 
-        def getStartTime():
-            return m.execute(f"grep 'starting up' /var/log/libvirt/qemu/{vmName}.log | tail -1")
+        if logPath:
+            m.write(logPath, '')
 
-        initialTime = getStartTime()
         b.click(f"#vm-{vmName}-{connectionName}-action-kebab button")
         b.wait_visible(f"#vm-{vmName}-{connectionName}-action-kebab > .pf-v5-c-dropdown__menu")
         b.click(f"#vm-{vmName}-{connectionName}-{action} a")
-        if (action == "reboot" or "forceReboot") and connectionName == "system":
-            # Check VM doesn't get rebooted after opening dialog
+        if action in ["reboot", "forceReboot"] and logPath:
             # https://bugzilla.redhat.com/show_bug.cgi?id=2221144
-            self.assertEqual(initialTime, getStartTime())
+            # The VM should not be rebooted when the confirmation dialog is shown
+            time.sleep(5)
+            self.assertNotIn("Linux version", m.execute(f"cat {logPath}"))
 
         # Some actions, which can cause expensive downtime when clicked accidentally, have confirmation dialog
         if action in ["off", "forceOff", "reboot", "forceReboot", "sendNMI"]:
@@ -69,8 +70,10 @@ class VirtualMachinesCaseHelpers:
 
         if action == "pause":
             b.wait_in_text(f"#vm-{vmName}-{connectionName}-state", "Paused")
-        if action == "resume" or action == "run":
+        if action in ["resume", "run", "reboot", "forceReboot"]:
             b.wait_in_text(f"#vm-{vmName}-{connectionName}-state", "Running")
+            if logPath:
+                testlib.wait(lambda: "Linux version" in m.execute(f"cat {logPath}"))
         if action == "forceOff" or action == "off":
             b.wait_in_text(f"#vm-{vmName}-{connectionName}-state", "Shut off")
 


KKoukiou (commented):
I don't understand what the changes improve to be honest. 

The bug related to the Reboot dialog pre-applying the change is actually already tested with             self.assertEqual(initialTime, getStartTime()) from what I see.

yunmingyang:
getStartTime checks the log which is "/var/log/liibviirt/qemu/{vm}.log", this log record the log about qemu start or shut off the VM. But it will not record events of the VM rebooting or force-rebooting. Thus, self.assertEqual(initialTime, getStartTime()) will be always true as getStartTime check the VM starting event actually. So we need to check the VM console log to know whether the VM is rebooted. And checking "Linux version" in the log is because maybe there is no the log like "reboot: Power down" in some other architectures.

test/machineslib.py
```diff
@@ -42,21 +42,18 @@ def waitPageInit(self):
...
         b.click(f"#vm-{vmName}-{connectionName}-action-kebab button")
         b.wait_visible(f"#vm-{vmName}-{connectionName}-action-kebab > .pf-v5-c-dropdown__menu")
         b.click(f"#vm-{vmName}-{connectionName}-{action} a")
-        if (action == "reboot" or "forceReboot") and connectionName == "system":
+        if (action == "reboot" or "forceReboot") and connectionName == "system" and logPath:
             # Check VM doesn't get rebooted after opening dialog
             # https://bugzilla.redhat.com/show_bug.cgi?id=2221144
-            self.assertEqual(initialTime, getStartTime())
+            # Need to clear the log before invoking
+            wait(lambda: "Linux version" not in m.execute(f"cat {logPath}"))
```
KKoukiou, line 56: ok I see you point now. My worry is that this is promt to a race condition. The "Linux version" will re-appear eventually in the log. wait retries very often, but it's really a matter of correct timing for this to work.

jelly:
As packit tests don't rebase and likely these where known failures. Do we want to rebase this one? And see if all tests pass. This PR last night went green so this should go green after a rebase https://github.com/cockpit-project/cockpit-machines/pull/1304

jelly:
@yunmingyang do you mind rebasing this PR then we can land it.

yunmingyang:
Ah sorry just see that. And of course, will do that

martinpitt:
> # test/machineslib.py:56:13: F821 Undefined name `wait`

It's `testlib.wait()` now.

test/machineslib.py
martinpitt: right, and this is especially true as "Linux version" is pretty much the first thing that the kernel writes at boot. It would be better with checking for the login prompt. Unfortunately this is hard to avoid in general -- waiting until the log file shrinks in size will have the same problem.

test/machineslib.py
```diff
@@ -42,21 +42,18 @@ def waitPageInit(self):
...
         b.click(f"#vm-{vmName}-{connectionName}-action-kebab button")
         b.wait_visible(f"#vm-{vmName}-{connectionName}-action-kebab > .pf-v5-c-dropdown__menu")
         b.click(f"#vm-{vmName}-{connectionName}-{action} a")
-        if (action == "reboot" or "forceReboot") and connectionName == "system":
+        if (action == "reboot" or "forceReboot") and connectionName == "system" and logPath:
             # Check VM doesn't get rebooted after opening dialog
             # https://bugzilla.redhat.com/show_bug.cgi?id=2221144
-            self.assertEqual(initialTime, getStartTime())
+            # Need to clear the log before invoking
+            testlib.wait(lambda: "Linux version" not in m.execute(f"cat {logPath}"))
```
martinpitt, line 56: The comment here is misleading -- it sounds like the code would clear the log, but it doesn't -- it expects that the log gets cleared through the reboot. If you don't mind, I'll change this, and also flesh out the commit message.

martinpitt: Also, the whole comment above is also bogus/outdated -- at this point the click already happened, so this isn't "after opening the dialog (what bz#2221144 was about), but after the action.

martinpitt:
@yunmingyang @KKoukiou I updated the commit and description. Can you please check that I understood what you meant/wanted to do here? Thanks!

test/check-machines-lifecycle
```diff
@@ -161,13 +161,13 @@ class TestMachinesLifecycle(machineslib.VirtualMachinesCase):
 
         # reboot
         self.machine.execute(f"echo '' > {args['logfile']}")
```
martinpitt, line 163: Let's please move that into `performAction`. It's easy to forget, and apparently it's an assumption.

test/machineslib.py
```diff
@@ -42,21 +42,16 @@ def waitPageInit(self):
...
         b.click(f"#vm-{vmName}-{connectionName}-action-kebab button")
         b.wait_visible(f"#vm-{vmName}-{connectionName}-action-kebab > .pf-v5-c-dropdown__menu")
         b.click(f"#vm-{vmName}-{connectionName}-{action} a")
-        if (action == "reboot" or "forceReboot") and connectionName == "system":
-            # Check VM doesn't get rebooted after opening dialog
-            # https://bugzilla.redhat.com/show_bug.cgi?id=2221144
-            self.assertEqual(initialTime, getStartTime())
+        if (action == "reboot" or "forceReboot") and connectionName == "system" and logPath:
+            # Rebooting clears the log, wait for that to happen
+            testlib.wait(lambda: "Linux version" not in m.execute(f"cat {logPath}"))
```
martinpitt, line 54: The `not` makes no sense to me here -- this is already satisfied, as the callers already do the `echo '' > logfile` thing. The only thing that would change here is that "Linux version" appears again, to make sure the boot happened.

martinpitt (changes requested):
Thanks for the Matrix chat -- I *did* misunderstand what's going on here. This needs another round of cleanup, and explanation. Do you want to work on that, or want me to?

test/machineslib.py
martinpitt: Ah, you wanted to ensure that boot did *not* happen. So:
 - This doesn't work, it's immediately true due to the synchronous truncation of the file. Please change to a sleep(5) and a single-shot test.
 - Please do that positive test after clicking the modal confirmation, to ensure that the reboot actually works.

test/machineslib.py
```diff
@@ -42,21 +42,18 @@ def waitPageInit(self):
...
         b.click(f"#vm-{vmName}-{connectionName}-action-kebab button")
         b.wait_visible(f"#vm-{vmName}-{connectionName}-action-kebab > .pf-v5-c-dropdown__menu")
         b.click(f"#vm-{vmName}-{connectionName}-{action} a")
-        if (action == "reboot" or "forceReboot") and connectionName == "system":
-            # Check VM doesn't get rebooted after opening dialog
+        if (action == "reboot" or "forceReboot") and connectionName == "system" and logPath:
             # https://bugzilla.redhat.com/show_bug.cgi?id=2221144
-            self.assertEqual(initialTime, getStartTime())
+            # The VM should not be rebooted when the confirmation dialog is shown
+            testlib.wait(lambda: True, tries=5)
```
martinpitt, line 55: That shuold just be `time.sleep(5)`

martinpitt (changes requested):
Thanks! The commit message is wrong, and there's still the outstanding comment about factoring the log clearing. I'll send some updates.

martinpitt (approved):
@yunmingyang I made the updates mentioned above, and corrected the commit message. Can you please cross-check again? Thanks!

Issue #1249, opened 2023-10-10
rawhide testWatchdog failure


jelly:
On rawhide

![image](https://github.com/cockpit-project/cockpit-machines/assets/67428/c98fe5d3-c8dd-4512-a291-6aa386210445)

Packages compared to F39

```
  libvirt-client (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-common (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-config-network (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-interface (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-network (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-nodedev (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-qemu (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-storage-core (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-storage-disk (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-storage-iscsi (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-storage-iscsi-direct (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-driver-storage-logical (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-daemon-log (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  libvirt-libs (9.7.0-1.fc39 -> 9.8.0-1.fc40)
  qemu-audio-spice (2:8.1.1-1.fc39 -> 2:8.1.1-1.fc40)
  qemu-block-curl (2:8.1.1-1.fc39 -> 2:8.1.1-1.fc40)
  qemu-char-spice (2:8.1.1-1.fc39 -> 2:8.1.1-1.fc40)
  qemu-common (2:8.1.1-1.fc39 -> 2:8.1.1-1.fc40)
  qemu-device-usb-host (2:8.1.1-1.fc39 -> 2:8.1.1-1.fc40)
[...]

PR #1256, opened 2023-10-12
Watchdog: Use 'reset' as default when no device is attached

skobyda:
Fixes https://github.com/cockpit-project/cockpit-machines/issues/1249

### Before

[Screencast from 2023-10-12 12-25-24.webm](https://github.com/cockpit-project/cockpit-machines/assets/42733240/7d0aabc2-90aa-4d2b-b365-6783b6834264)

### After

[Screencast from 2023-10-12 12-26-09.webm](https://github.com/cockpit-project/cockpit-machines/assets/42733240/65a8d668-9e87-4db8-af6a-2817c861599f)

### DIFF
--- a/src/components/vm/overview/watchdog.jsx
+++ b/src/components/vm/overview/watchdog.jsx
@@ -37,6 +37,8 @@ import { WATCHDOG_INFO_MESSAGE } from './helpers.jsx';
 
 const _ = cockpit.gettext;
 
+const SUPPORTEDACTIONS = ["reset", "poweroff", "inject-nmi", "pause"];
+
 const WatchdogModalAlert = ({ dialogError }) => {
     const [isExpanded, setIsExpanded] = useState(false);
 
@@ -68,7 +70,7 @@ const WatchdogModalAlert = ({ dialogError }) => {
 
 export const WatchdogModal = ({ vm, isWatchdogAttached, idPrefix }) => {
     const [dialogError, setDialogError] = useState();
-    const [watchdogAction, setWatchdogAction] = useState(vm.watchdog.action || "no-device");
+    const [watchdogAction, setWatchdogAction] = useState(vm.watchdog.action || SUPPORTEDACTIONS[0]); // use first option as default
     const [inProgress, setInProgress] = useState(false);
     const [offerColdplug, setOfferColdplug] = useState(false);
 
@@ -133,8 +135,6 @@ export const WatchdogModal = ({ vm, isWatchdogAttached, idPrefix }) => {
                 .then(Dialogs.close, exc => setDialogError({ text: _("Failed to detach watchdog"), detail: exc.message, variant: "danger" }))
                 .finally(() => setInProgress(false));
     };
-
-    const supportedActions = ["reset", "poweroff", "inject-nmi", "pause"];
 
     const needsShutdown = () => {
         if (vm.state === 'running' && isWatchdogAttached && vm.watchdog.action !== watchdogAction)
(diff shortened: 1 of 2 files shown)


src/components/vm/overview/watchdog.jsx
```diff
@@ -110,6 +112,11 @@ export const WatchdogModal = ({ vm, isWatchdogAttached, idPrefix }) => {
     };
 
     const save = (coldplug) => {
+        if (!watchdogAction) {
+            setValidationMessage(_("You need to select the action"));
+            return;
```
cockpituous, line 117: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1256-20231012-104029-22248714-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
The "you need to select the action" string is some bikeshed territory, but my gut feeling is that the "Add" button could simply be disabled while there is nothing selected?

skobyda:
> The "you need to select the action" string is some bikeshed territory, but my gut feeling is that the "Add" button could simply be disabled while there is nothing selected?

Do you mean that Add button would be disabled immediately when dialog is opened? That this is not good user experience. I.e. user only opens dialog, they didn't do anything yet, and we are basically telling them that something is already wrong by having a disabled button. That's confusing. I remember we had discussion about this a year ago with @garrett  when implementing watchdog.

Or do you mean that user would click on "Add" and when nothing is selected, it would get disabled, but "you need to select the action" message is not shown? I also find that a bit confusing, since we are disabling the button as a result of user action, but not providing any feedback.

jelly:
> > The "you need to select the action" string is some bikeshed territory, but my gut feeling is that the "Add" button could simply be disabled while there is nothing selected?
> 
> Do you mean that Add button would be disabled immediately when dialog is opened? That this is not good user experience. I.e. user only opens dialog, they didn't do anything yet, and we are basically telling them that something is already wrong by having a disabled button. That's confusing. I remember we had discussion about this a year ago with @garrett when implementing watchdog.
> 
> Or do you mean that user would click on "Add" and when nothing is selected, it would get disabled, but "you need to select the action" message is not shown? I also find that a bit confusing, since we are disabling the button as a result of user action, but not providing any feedback.

I think that is the correct approach yes, disabling until a valid choice has been made. And not showing the warning that you have to make a selection.

martinpitt:
> Do you mean that Add button would be disabled immediately when dialog is opened?

Yes.

> That this is not good user experience. I.e. user only opens dialog, they didn't do anything yet, and we are basically telling them that something is already wrong by having a disabled button

But with an enabled button we'd then show the error message, so it's the same result -- the only valid action without picking a value is "Cancel".

skobyda:
> But with an enabled button we'd then show the error message, so it's the same result -- the only valid action without picking a value is "Cancel".

Maybe we could at least have a Tooltip when "Add" button is hovered over? @garrett WDYT?

garrett:
Whooooooah... I'd suggest we step back a little. :grin: 

This is a list of radios. Radios, by default, should have one preselected — and that's almost always the first item. The fact that there isn't one preselected here is the bug. If the person using Cockpit Machines encounters this modal and doesn't want the first item — or any other — then they can always click cancel.

So: Please make the first default when adding a watchdog. If first isn't a good default, then we should reorder the list so it is.

(As it's a list of radios, as long as one is selected, then you cannot unselect, so someone would then never encounter this issue, and we'll never have to explain it.)

I think that'll solve all these issues. Right?

---

TL;DR: Preselect the first radio item, and that should solve this.

skobyda:
Updated the PR, so the "Reset" option is pre-selected as default when no watchdog device is attached yet:

[Screencast from 2023-10-13 12-23-18.webm](https://github.com/cockpit-project/cockpit-machines/assets/42733240/48f27091-7723-4658-a745-9ab7efda8523)

martinpitt (approved):
Works for me, thanks!

PR #1290, opened 2023-11-06
Test configure before install flake

jelly:
It turns out that we clicked on the memory modal button when the memory usage in our UI is still 150 MiB and the memory modal either does not update or updates too slow to get the new 130 MiB value.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1986,9 +1986,7 @@ vnc_password= "{vnc_passwd}"
         b.click("#vm-VmNotInstalled-memory-count button")
         b.wait_visible("#vm-memory-modal")
         b.set_input_text("#vm-VmNotInstalled-memory-modal-max-memory", "150")
-        b.blur("#vm-VmNotInstalled-memory-modal-max-memory")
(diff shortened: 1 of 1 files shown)


test/check-machines-create
```diff
@@ -2060,6 +2060,7 @@ vnc_password= "{vnc_passwd}"
 
         # Check configuration changes survived installation
         # Check memory settings have persisted
+        b.wait_in_text("#vm-subVmTest1-memory-count", "130.0 MiB")
```
martinpitt, line 2063: ITYM `#vm-VmNotInstalled-memory-count` ?

jelly: Yes..

jelly:
Not sure what is going on with fedora-39/firefox

martinpitt:
The [rawhide failures](https://artifacts.dev.testing-farm.io/0a63986c-a4fa-4872-9e67-d1f8abdd496d/) have been going on for a few days, that's the next yak to shave.

martinpitt (approved):
Thanks for tracking this down!

Issue #1294, opened 2023-11-07
Rawhide test failures


jelly:
### Snapshots revert

```
Traceback (most recent call last):
  File "/home/jelle/projects/cockpit-machines/./test/check-machines-snapshots", line 285, in testSnapshotRevert
    b.wait_visible("#vm-subVmTest1-snapshot-0-current")
  File "/home/jelle/projects/cockpit-machines/test/common/testlib.py", line 618, in wait_visible
    self._wait_present(selector)
  File "/home/jelle/projects/cockpit-machines/test/common/testlib.py", line 609, in _wait_present
    self.wait_js_func('ph_is_present', selector)
  File "/home/jelle/projects/cockpit-machines/test/common/testlib.py", line 603, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
  File "/home/jelle/projects/cockpit-machines/test/common/testlib.py", line 588, in wait_js_cond
    self.raise_cdp_exception("timeout\nwait_js_cond", cond, result["exceptionDetails"], trailer)
  File "/home/jelle/projects/cockpit-machines/test/common/testlib.py", line 292, in raise_cdp_exception
    raise Error("%s(%s): %s" % (func, arg, msg))
testlib.Error: timeout
wait_js_cond(ph_is_present("#vm-subVmTest1-snapshot-0-current")): Uncaught (in promise) Error: condition did not become true
```

```
> warning: GET_VM failed for /org/libvirt/QEMU/domain/_88a43112_22c0_43d9_9848_5ff0828b85e2, undefining: Cannot recv data: Connection reset by peer
```

### testHostDevAddSingleDevice

Locally
```
audit: type=1400 audit(1699378740.302:509): avc:  denied  { create } for  pid=2599 comm="qemu-system-i38" anonclass=[io_uring] scontext=system_u:system_r:virtqemud_t:s0 tcontext=system_u:object_r:io_uring_t:s0 tclass=anon_inode permissive=1
````

martinpitt:
Two recent example logs:
 - https://artifacts.dev.testing-farm.io/0aa53011-f549-412e-bb65-792d6ca6a3b7/
 - https://artifacts.dev.testing-farm.io/0a63986c-a4fa-4872-9e67-d1f8abdd496d/

martinpitt:
[...]

PR #1296, opened 2023-11-08
Random test cleanups

jelly:


martinpitt (approved):
Cheers! Rawhide failures are known, see #1294

PR #1305, opened 2023-11-10
Allow inputting a ssh key without a comment

jelly:


test/check-machines-create
```diff
@@ -291,11 +291,11 @@ class TestMachinesCreate(VirtualMachinesCase):
         self.machine.execute("ssh-keygen -t rsa -N '' -f /tmp/rsakey")
         rsakey = self.machine.execute("cat /tmp/rsakey.pub").strip()
         self.addCleanup(self.machine.execute, "rm -f /tmp/rsakey*")
-        self.machine.execute("ssh-keygen -t dsa -N '' -C '' -f /tmp/dsakey")  # public key with empty comment
+        self.machine.execute("ssh-keygen -t dsa -N '' -C '' -f /tmp/dsakey").strip()  # public key with empty comment
         dsakey = self.machine.execute("cat /tmp/dsakey.pub")
         self.addCleanup(self.machine.execute, "rm -f /tmp/dsakey*")
 
-        # Try to create VM with one SSH key
+        # # Try to create VM with one SSH key
```
martinpitt, line 298: There is lots of debugging leftover in the first commit, which the second commit then cleans up again. However, this double-commit is still left over, so merely squashing the two commits doesn't completely clean it up. ISTM that this should really just be one commit.

martinpitt (changes requested):
Thanks! Fix/test look good, just needs some cleanup.

martinpitt (approved):
Thanks! Having the test case separate breaks bisecting, so let's squash.

PR #1309, opened 2023-11-14
Allow adding multiple ssh keys at once

### LINKED ISSUE #1300: There should be an hint to clarify that only the first SSH key could take effect for each "Public key" input
Description of problem:
For each "Public key" input, only the first SSH key takes effect, it is better to set a hint for clarifying that.

cockpit-300.1-1.el8_9.x86_64
cockpit-machines-301.1.2.gb0140b7-1.fc38.noarch
libvirt-dbus-1.3.0-2.module+el8.9.0+18724+20190c23.x86_64

How reproducible:
100%

Steps to Reproduce:
1 Log in cockpit, then switch to the machines page
2 Click "Create VM", then select "Installation type" to "Cloud base image"
3 Go to the Automation tab, then click "Add SSH keys"
4 Input a SSH key without comment to the input appeared after clicking "Add SSH keys"

Actual results:
1 In step 3 and 4, there is no hint to clarify that only the first SSH key could take effect for each "Public key" input

Expected results:
1 There should be a hint to clarify that only the first SSH key could take effect for each "Public key" input

Additional info:

jelly:
The initial version of the ssh key functionality would only add the first key if multiple where provided. Allow adding multiple keys in via a new helper, the validation now needs to happen via SshKeysRow as they are added dynamically and the value provided via properties.

Closes #1300

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -752,8 +752,7 @@ const UsersConfigurationRow = ({
     );
 };
 
-// This method needs to be outside of component as re-render would create a new instance of debounce
-const parseKey = debounce(500, (key, setKeyObject, setKeyInvalid) => {
+const parseKey = (key, setKeyObject, setKeyInvalid) => {
     if (isEmpty(key))
         return;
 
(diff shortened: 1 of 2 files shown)


src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -776,22 +775,41 @@ const parseKey = debounce(500, (key, setKeyObject, setKeyInvalid) => {
...
-        // While handling that in the future might be a nice user experience, now we only parse one key out of input
-        value = value.split(/\r?\n/)[0];
-
-        onChange(idx, "value", value);
-        parseKey(value, setKeyObject, setKeyInvalid);
-    };
+    useEffect(() => {
+        let timeoutId = null;
+        if (keyValue !== "" && !keyObject) {
+            timeoutId = setTimeout(() => parseKey(keyValue, setKeyObject, setKeyInvalid), 500);
```
martinpitt, line 805: I think the [original debounce comment](https://github.com/cockpit-project/cockpit-machines/pull/1309/files#diff-c3b5a228e7299618756cc30089b5e5f0b1e3662e48459df036e31676a874bec2L755) still applies here -- this useEffect() re-runs on *every* input, so it creates multiple parallel timers and calls parseKey() for each one as well. Some debouncing still needs to happen.

martinpitt (changes requested):

src/components/create-vm-dialog/createVmDialog.jsx
jelly: The issue with a global debounce is that it is used for multiple components, so if you add two keys quickly only one will get parsed as the other call is debounced. So debouncing needs to happen per Component and we can't keep the debounce function in the component as every re-render it would be re-instantiated. So I believe the current approach with setTimeout debounces correctly, when a new as a new value would cause a re-render so clearTimeout() to be called?

If I'm wrong I'm happy to fix it if there is a better way.

jelly: We can not make onChange call setKeyValue simply because onChange will add a new row if multiple keys are added using `additem()`. That is the hard part, dynamically creating a new row component.

martinpitt: That sounds weird -- in JSX it's just a single TextArea. If this somehow multiplies the TextArea, this needs fixing -- either we shouldn't multiply it, or have a full-blown dynamic set of input lines where each one just has a *single* key, and :heavy_plus_sign: :wastebasket: buttons

KKoukiou (commented):
Does this allow to input multiple keys in one text area field? If so, this seems a bit off by design perspective.

mvollmer:
My version of this: https://github.com/cockpit-project/cockpit-machines/pull/1515

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -776,22 +775,41 @@ const parseKey = debounce(500, (key, setKeyObject, setKeyInvalid) => {
...
                 console.warn("Could not validate the public key");
             });
-});
+};
+
+const handleSSHKeysValue = (value, idx, onChange, additem) => {
+    let index = idx;
+    for (const part of value.split("\n")) {
+        if (part.trim() === "") {
+            continue;
```
mvollmer, line 784: This means that the last character in the textarea can never be deleted.

PR #1315, opened 2023-11-16
Enable creating VM without user password if using SSH keys

### LINKED ISSUE #1314: The VM could not be created with a SSH key and without "User password"
Description of problem:
Create a VM with a SSH key and without "User password", there will be an warning shown below "User password" and the creation is blocked.

cockpit-300.1-1.el8_9.x86_64
cockpit-machines-301.1.2.gb0140b7-1.fc38.noarch
libvirt-dbus-1.3.0-2.module+el8.9.0+18724+20190c23.x86_64

How reproducible:
100%

Steps to Reproduce:
1 Create a VM with a SSH key and without "User password"

Actual results:
1 There will be an warning shown below "User password" and the creation is blocked

Expected results:
1 The VM could be created with a SSH key and without "User password", and the SSH key could be put in the guest correctly

Additional info:

yunmingyang:
Fixes #1314

### DIFF
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -245,7 +245,7 @@ function validateParams(vmParams) {
         );
     }
 
-    if (!vmParams.userPassword && vmParams.userLogin) {
+    if (vmParams.userLogin && !((!!vmParams.userPassword) || vmParams.sshKeys.length > 0)) {
         validationFailed.userPassword = _("User password must not be empty when user login is set");
     }
     if (vmParams.userPassword && !vmParams.userLogin) {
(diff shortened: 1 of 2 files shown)


KKoukiou (commented):
Can you please add the test for this?

yunmingyang:
Yeah, sure, will add the test later

yunmingyang:
Add tests, and as some tests related with "Create and edit" need manually execution to make sure the key has been put in the VM, so no tests about "Create and edit"

martinpitt:
The patch LGTM and the previous [failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1315-20231123-075029-0027874e-fedora-39/log.html) was an unrelated flake which I believe got fixed since then. Retrying.

Just re-asking here my question from the issue: That user is the only way to log in. How does that work with sudo? Is it passwordless success then, or doesn't work at all? In the latter case we shouldn't allow that.

PR #1321, opened 2023-11-21
Change default behavior when adding disks to a VM

KKoukiou:
Now the 'Attach permanently' will be set by default.


jelly:
What issue is related to this change? I can't find any reasoning to change the default behavior?

KKoukiou:
> What issue is related to this change? I can't find any reasoning to change the default behavior?

The currently default behavior is not intuitive. @marusak had feedback from PM, there is not issue about it.

PR #1325, opened 2023-11-23
Explicitly set the host option for the VNC websocket channel

jelly:
Without setting the host option Cockpit connects to the main Cockpit's machine VNC port instead of the currently open host VNC port. This is a workaround for a bug in cockpit-ws which should set the host option automatically for us depending on what iframe makes the connection.

Resolves: RHEL-3959

---

@martinpitt you have opted that this should not happen in machines but in Cockpit-ws. I suppose I agree but then we need to find a way to detect who establishes the websocket connection and that requires more search.

So a merge-able approach might be:
* [x] Fill a Cockpit issue
* [x] Make a unit test / real test? Unit test would be hard I suppose
* [ ] ~Find a way to reliably detect the websocket origin.~ not trivial, let's discuss in the cockpit epic
* [x] #1329
* [x] #1334

### DIFF
--- a/src/components/vm/consoles/vnc.jsx
+++ b/src/components/vm/consoles/vnc.jsx
@@ -78,6 +78,9 @@ class Vnc extends React.Component {
                 binary: "raw",
                 address: consoleDetail.address,
                 port: parseInt(consoleDetail.tlsPort || consoleDetail.port, 10),
+                // https://issues.redhat.com/browse/COCKPIT-870
+                // https://issues.redhat.com/browse/RHEL-3959
+                host: cockpit.transport.host,
             });
             this.setState({
                 path: `${prefix.slice(1)}?${window.btoa(query)}`,
--- a/test/check-machines-multi-host-consoles
+++ b/test/check-machines-multi-host-consoles
@@ -1,0 +1,96 @@
+#!/usr/bin/python3 -cimport os, sys; os.execv(os.path.dirname(sys.argv[1]) + "/common/pywrap", sys.argv)
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2021 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import testlib
+from machineslib import VirtualMachinesCase
+
+
+class TestMultiMachineVNC(VirtualMachinesCase):
+    """
+    Test for showing the wrong VNC console when adding another host via the add
+    host feature the raw stream channel does not set the host option so would
+    connect to the console of machine you connect to instead of the newly added
+    host.
+
+    https://issues.redhat.com/browse/RHEL-3959
+    """
+
+    provision = {  # noqa: RUF012
+        "machine1": {"address": "10.111.113.1/20", "memory_mb": 660},
+        "machine2": {"address": "10.111.113.2/20", "memory_mb": 660},
+    }
+
+    def setUp(self):
+        super().setUp()
+        self.machine1 = self.machines['machine1']
+        self.machine2 = self.machines['machine2']
+
+        self.setup_ssh_auth()
+
+    def testBasic(self):
+        b = self.browser
+        m = self.machine
+        m2 = self.machine2
+        m2_host = "10.111.113.2"
+
+        m.execute(f"ssh-keyscan {m2_host} > /etc/ssh/ssh_known_hosts")
+        self.startLibvirt(m2)
+
+        # Start a VM with a VNC on machine1
+        name = "subVmTest1"
+        self.createVm(name, graphics="vnc", ptyconsole=True)
+
+        self.login_and_go("/machines")
+        self.waitPageInit()
+
+        self.waitVmRow(name)
+        self.goToVmPage(name)
+        b.wait_in_text(f"#vm-{name}-system-state", "Running")
+
+        self.add_machine(m2_host, known_host=True, password=None)
+
+        b.switch_to_top()
+        b.click("#hosts-sel button")
+
+        b.click(f"a[href='/@{m2_host}']")
+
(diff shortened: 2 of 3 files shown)


src/components/vm/consoles/vnc.jsx
```diff
@@ -73,12 +73,16 @@ class Vnc extends React.Component {
...
+            const data = {
                 payload: "stream",
                 binary: "raw",
                 address: consoleDetail.address,
                 port: parseInt(consoleDetail.tlsPort || consoleDetail.port, 10),
-            });
+            };
+            if (cockpit.transport.host !== "localhost") {
+                data.host = cockpit.transport.host;
+            }
```
martinpitt, line 84: No need to make this conditional, please always set it. Plus, please add a link to the jira or github issue.

test/browser/run-test.sh
```diff
@@ -73,6 +73,8 @@ EXCLUDES="$EXCLUDES
           TestMachinesNICs.NICAddDialog
 
           TestMachinesSettings.testMultipleSettings
+
+          TestMultiMachineVNC.testBasic
```
martinpitt, line 77: This isn't necessary. The test isn't nondestructive, so it won't even be considered.

test/check-machines-multi-host-consoles
```diff
@@ -0,0 +1,135 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import testlib
+from machineslib import VirtualMachinesCase
+
+
+def start_machine_troubleshoot(b, new=False, known_host=False, password=None):
```
martinpitt, line 24: This is copied from cockpit's check-shell-multi-machine . OK for now, but can you please create a cockpit PR that moves these to testlib.py? Especially as it changed recently, and may again at some point.

test/check-machines-multi-host-consoles
```diff
@@ -0,0 +1,135 @@
...
+            b.wait_in_text('#hosts_setup_server_dialog', "You are connecting to")
+            b.wait_in_text('#hosts_setup_server_dialog', "for the first time.")
+            b.click("#hosts_setup_server_dialog button:contains('Trust and add host')")
+    if password:
+        b.wait_in_text('#hosts_setup_server_dialog', "Unable to log in")
+        b.set_input_text('#login-custom-password', password)
+        b.click('#hosts_setup_server_dialog button:contains(Log in)')
+
+
+def add_machine(b, address, known_host=False, password="foobar"):
```
martinpitt, line 40: Dito

test/check-machines-multi-host-consoles
```diff
@@ -0,0 +1,135 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.machine1 = self.machines['machine1']
+        self.machine2 = self.machines['machine2']
+
+        # Disable preloading on all machines
+        # Preloading on machines debug build can overload the browser and cause slowness and browser crashes,
+        # and failing to load sofware updates breaks pixel tests in release builds
+        self.setup_provisioned_hosts(disable_preload=True)
```
martinpitt, line 82: Please check if it's really necessary. If so, please adjust the  comment, as that's a copy pasta. If not, please drop.

test/check-machines-multi-host-consoles
```diff
@@ -0,0 +1,135 @@
...
+        self.setup_ssh_auth()
+
+    def testBasic(self):
+        b = self.browser
+        m = self.machine
+        m2 = self.machine2
+        m2_host = "10.111.113.2"
+
+        m.execute(f"ssh-keyscan {m2_host} > /etc/ssh/ssh_known_hosts")
```
martinpitt, line 91: Why is that bit not in setup_ssh_auth()? It feels to me that it rather belongs there.

test/check-machines-multi-host-consoles
```diff
@@ -0,0 +1,135 @@
...
+        b.wait_js_cond(f'window.location.pathname == "/@{m2_host}/system"')
+        b.enter_page("/system", host=m2_host)
+        b.become_superuser()
+
+        b.go(f"/@{m2_host}/machines")
+        b.enter_page("/machines", host=m2_host)
+
+        self.createVm(name, graphics="vnc", ptyconsole=True, machine=m2)
+        self.waitVmRow("subVmTest1")
+        b.wait_in_text("#vm-subVmTest1-system-state", "Running")  # running or paused
```
martinpitt, line 121: What does that comment mean?

test/machineslib.py
```diff
@@ -144,8 +144,11 @@ def startLibvirt(self, m):
         m.execute("virsh net-start default || true")
         m.execute(r"until virsh net-info default | grep 'Active:\s*yes'; do sleep 1; done")
 
-    def createVm(self, name, graphics='none', ptyconsole=False, running=True, memory=128, connection='system'):
-        m = self.machine
+    def createVm(self, name, graphics='none', ptyconsole=False, running=True, memory=128, connection='system', machine=None):
+        if machine is None:
+            m = self.machine
+        else:
+            m = machine
```
martinpitt, line 151: `m = machine or self.machine` ?

martinpitt (changes requested):
Thanks for fixing this! Some minor cleanups, and please file a ticket to fix this more generally. I propose adding a task to https://issues.redhat.com/browse/COCKPIT-870 that links to this PR/discussion/commit.

test/check-machines-multi-host-consoles
martinpitt: https://github.com/cockpit-project/cockpit/pull/19659 . Need to retry some tests, but I expect to land this in 30 mins

martinpitt:
@jelly Thanks! I resolved the threads which your recent push fixes, there's still a few outstanding ones.

test/check-machines-multi-host-consoles
martinpitt: Landed. I triggered a lib update to get this out of the way: PR #1329 . Once that is in, you can rebase and remove the duplication.

martinpitt: Lib update landed.

test/check-machines-multi-host-consoles
martinpitt: Yes, I'd rather use the cockpit helpers and keep it here. It doesn't look ubuntu specific, I figure it's there to avoid the "unknown host key" prompt.

jelly: I see, with ubuntu I get 
image

So not sure how this works in Cockpit.

jelly:
> @jelly Thanks! I resolved the threads which your recent push fixes, there's still a few outstanding ones.

add_machine awaits https://github.com/cockpit-project/cockpit/pull/19663 which needs some fixing

martinpitt (approved):
Thanks for bearing with me! This looks really nice now, great that you figured out a test! :100:

PR #1326, opened 2023-11-23
test: add some checks about the source of NICs

yunmingyang:


martinpitt:
The [testVmNICs test failed](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1326-20231123-132559-f4b6bdda-fedora-39/log.html#83), but at first sight that shouldn't be due to your changes. Retrying.

martinpitt (approved):
Thanks!

PR #1327, opened 2023-11-24
Add disk dialog bug fixes

martinpitt:
This has turned into quite a can of worms... This has just renewed my aversion to abusing `useEffect()` -- honestly they should *all* die in diskAdd.jsx, I did not see any valid use case of it. They look like sloppy "let's clean up after our own  bad/wrong decisions", and make state handling immensely brittle and hard to follow. The logic from all of them should be moved into a single `useInit()` to set up the dialog, and the corresponding event handlers (which ought to set up the state correctly right away instead of relying on effects to second-guess them).

This PR at least gets rid of two of them.

Fixes #1272

Fixes #1273

PR #1331, opened 2023-11-28
Disk add dialog: Change format together with the storage pool

martinpitt:
Otherwise we create a race condition: The ModalBody already gets re-rendered with the new pool name, but with possibly with a format that is invalid for the new pool type. That triggers our new assertion.

This is very redundant with the effect which fine-tunes the format and disk type according to the other data. This effect is evil and should be dropped, but doing so requires more intrusive code cleanup. Introduce a helper function to avoid repeated code.

---

Follow-up from https://github.com/cockpit-project/cockpit-machines/pull/1327#issuecomment-1825682485

### DIFF
--- a/src/components/vm/disks/diskAdd.jsx
+++ b/src/components/vm/disks/diskAdd.jsx
@@ -395,6 +395,19 @@ export const AddDiskModalBody = ({ disk, idPrefix, isMediaInsertion, vm, vms, su
     }), [defaultPool, mode, vm]);
     const storagePoolName = diskParams.storagePool?.name;
 
+    const getPoolFormatAndDevice = (pool, volName) => {
+        const params = { format: getDefaultVolumeFormat(pool), device: "disk" };
+        if (['dir', 'fs', 'netfs', 'gluster', 'vstorage'].indexOf(pool.type) > -1) {
+            const volume = pool.volumes.find(vol => vol.name === volName);
+            if (volume?.format) {
+                params.format = volume.format;
+                if (volume.format === "iso")
+                    params.device = "cdrom";
+            }
+        }
+        return params;
+    };
+
     useEffect(() => {
         // Refresh storage volume list before displaying the dialog.
         // There are recently no Libvirt events for storage volumes and polling is ugly.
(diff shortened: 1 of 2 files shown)


martinpitt:
[Some flakes](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1331-20231128-090529-818a8cd3-fedora-39-firefox/log.html#76), not sure if relevant.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -395,6 +395,19 @@ export const AddDiskModalBody = ({ disk, idPrefix, isMediaInsertion, vm, vms, su
     }), [defaultPool, mode, vm]);
     const storagePoolName = diskParams.storagePool?.name;
 
+    const getPoolVolumeAndDevice = (pool, volName) => {
```
KKoukiou, line 398: It would be more self-explanatory to name this:
getPoolFormatAndDevice

KKoukiou (commented):
I like the approach.

KKoukiou:
> A better step would obviously be to drop the effect entirely. I'll see if I can find some time to experiment with that. In the meantime, @KKoukiou do you think this is an improvement, or does it make the thing even more complicated?

If you want to take it a step further, check the useEffect for the 'diskParams.file' state. It also has logic for changing the device and the format.

src/components/vm/disks/diskAdd.jsx
martinpitt: Err yes, that's what I actually meant :see_no_evil: Updated.

martinpitt:
> check the useEffect for the 'diskParams.file' state

Thanks for pointing out! It's not quite the same, as that's for a pool vs. a file, but it would make sense to factorize the logic. I'll think about that.

martinpitt:
@KKoukiou I didn't see any significant overlap in the end. But I did clean up the effect a bit to avoid an unnecessary and racy state update.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -471,20 +473,19 @@ export const AddDiskModalBody = ({ disk, idPrefix, isMediaInsertion, vm, vms, su
...
+                        const params = {};
+
+                        if (file.endsWith(".iso"))
+                            params.device = "cdrom";
 
                         // https://git.qemu.org/?p=qemu.git;a=blob;f=docs/interop/qcow2.txt
                         if (file_header[0] == 81 && file_header[1] == 70 &&
                             file_header[2] == 73 && file_header[3] == 251) {
-                            format = "qcow2";
+                            params.format = "qcow2";
```
cockpituous, line 488: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1331-20231128-160227-c60fcb6b-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
Meh, every time I take a stab at this code, it crumbles into pieces. I'll back out the last two commits, let's at least fix the first issue.

KKoukiou (approved):

PR #1332, opened 2023-11-28
test: Ignore noVNC TLS console error

martinpitt:
When opening a plain http connection to a non-localhost machine, noVNC writes a console error. This doesn't happen in our CI, as 127.0.0.2 is considered as `window.isSecureContext == true`, but it does happen for downstream testing against an actually remote machine.

https://issues.redhat.com/browse/RHEL-17309

[1] https://github.com/novnc/noVNC/blob/85a465288b314/core/rfb.js#L99

@yunmingyang FYI


jelly (approved):
Thanks!

PR #1342, opened 2023-12-07
Hide SPICE on RHEL 8

martinpitt:
SPICE got deprecated in RHEL 8.3 [1]. Stop configuring SPICE for new
VMs as requested by the libvirt team. libvirt still announces spice
as supported in the API, so we need to do a "check os-release" hack.

https://issues.redhat.com/browse/RHEL-18058

[1] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/8.3_release_notes/rhel-8-3-0-release#deprecated-functionality_virtualization

### DIFF
--- a/src/scripts/install_machine.py
+++ b/src/scripts/install_machine.py
@@ -39,6 +39,15 @@ def get_graphics_capabilies(connection):
         for value in graphics.find('enum').findall('value'):
             consoles.append(value.text)
 
+    # HACK: Ignore spice on RHEL 8; https://issues.redhat.com/browse/RHEL-18058
+    try:
+        with open("/etc/os-release") as f:
+            if "platform:el8" in f.read():
+                logging.debug("get_graphics_capabilies: ignoring spice on RHEL 8")
+                consoles.remove('spice')
+    except FileNotFoundError:
+        pass  # not RHEL then
+
     logging.debug('get_graphics_capabilies: %s', ', '.join(consoles))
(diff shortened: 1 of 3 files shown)


martinpitt:
The remaining failures are unrelated and on the [pilot task board](https://github.com/orgs/cockpit-project/projects/4). I'll look at them next.

src/scripts/install_machine.py
```diff
@@ -39,6 +39,15 @@ def get_graphics_capabilies(connection):
...
             consoles.append(value.text)
 
+    # HACK: Ignore spice on RHEL 8; https://issues.redhat.com/browse/RHEL-18058
+    try:
+        with open("/etc/os-release") as f:
+            if "platform:el8" in f.read():
+                logging.debug("get_graphics_capabilies: ignoring spice on RHEL 8")
+                consoles.remove('spice')
+    except FileNotFoundError:
+        pass  # not RHEL then
```
jelly, line 49: So hmm spice is more then just the console node. Who adds the other things like spice-serial? OR does this happen before virt-install is done?

I do spot a test below for any occurrence of spice in the XML but I don't know enough details if that is 100% correct.

Also this would be a "breaking" change in RHEL8, is that allowed/normal?

jelly (approved):

martinpitt:
I landed the two flake fixes, retrying.

Issue #1164, opened 2023-07-26
Missing spacing between input and form helper text


jelly:
![image](https://github.com/cockpit-project/cockpit-machines/assets/67428/186ee729-c794-4d3d-9db4-056b1bd2fffa)

satyam4p:
i'll be owrking on this one. @jelly  is that fine?

arup1221:
i want to work on it

Srihari2222:
@KKoukiou  Can you briefly give some points about the issue and what to be expected to solve. I can work on this issue if you assign me.

Vishalydv07:
@jelly I can work on this issue if you assign me.

jelly:
@Vishalydv07 we don't assign issues, feel free to look into it. See our docs for how to contribute https://cockpit-project.org/external/wiki/Contributing.html

mvollmer:
In our backlog now: https://issues.redhat.com/browse/COCKPIT-1312

mvollmer:
It's the `isStack` property of the FormGroup for "IP configuration". I make a quick PR.

PR #1354, opened 2023-12-16
fix: Missing spacing between input and form helper text solves #1164

ashutosh7i:
Hello @jelly & @KKoukiou , i have worked on this issue,
### Image with issue-
Screenshot from 2023-12-17 01-55-22

After my Solution to the issue-
image

### The fix i implemented is that i replaced  
```
<FormHelper ... /> component
```
with the official component-
```
<HelperText>
                    <HelperTextItem variant="error">{ ... } </HelperTextItem>
</HelperText>
```

image

### And tweaked css in Machines.scss
image

this workaround fixed the issue of missing spacing between input and helper text solving #1164

### DIFF
--- a/src/components/networks/createNetworkDialog.jsx
+++ b/src/components/networks/createNetworkDialog.jsx
@@ -22,6 +22,7 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { Checkbox } from "@patternfly/react-core/dist/esm/components/Checkbox";
 import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form";
 import { FormSelect, FormSelectOption, FormSelectOptionGroup } from "@patternfly/react-core/dist/esm/components/FormSelect";
+import { HelperText, HelperTextItem } from '@patternfly/react-core';
 import { Grid } from "@patternfly/react-core/dist/esm/layouts/Grid";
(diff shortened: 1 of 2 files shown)


garrett (changes requested):
This is an improvement, but why are the spaces different?

"Name should not be empty" has 13 pixels of space between the form and the string.

While the rest have 8 px.

PatternFly has 10px.

It seems your fonts are larger though (not the default), so the references above don't matter when comparing to PatternFly (which also doesn't use px to set these values, of course), so I don't know which is correct here, but I do know that they're inconsistent.

So I've checked locally on my machine and it seems the first one here is 10px and the rest are 6px. The first is correct; the rest still are not correct.

image

garrett:
For reference:

Here's a screenshot of JSX React on PatternFly's demo page @ https://www.patternfly.org/components/helper-text/react-demos

image

ashutosh7i:
sure @garrett i am onto it.

mvollmer:
This is still broken after the update to PF6:

image

mvollmer:
Patternfly says we should wrap `HelperText` with `FormHelperText` inside a `FormGroup` (which is what our `FormHelper` does), so we need to find a different solution.

I have put this into our backlog: https://issues.redhat.com/browse/COCKPIT-1312

PR #1359, opened 2023-12-21
Add Suspend To Disk Option to the VM control menu

DlpOne:
The stop function can only pause the VM, which means that the VM continues to allocate resources and the status of the VM is lost when the Host is rebooted. 

It would be useful to have a function that allows you to save the state of a VM to disk and resume it at a later point in time.
Libvirt offers the managedsave function for this purpose.

This patch adds this function to the control menu of the VM 

[
Bildschirmfoto vom 2023-12-21 01-57-54
](url)

### DIFF
--- a/src/libvirtApi/domain.ts
+++ b/src/libvirtApi/domain.ts
@@ -95,6 +95,8 @@ export const domainCanShutdown = (vmState: VMState) => domainCanReset(vmState);
 export const domainCanPause = (vmState: VMState) => vmState == 'running';
 export const domainCanRename = (vmState: VMState) => vmState == 'shut off';
 export const domainCanResume = (vmState: VMState) => vmState == 'paused';
+export const domainCanSuspendToDisk = (vmState: VMState) => vmState == 'running' || vmState == 'paused';
+export const domainCanSuspenImageRemove = (vmState: VMState, saveImage: Boolean) => vmState == "shut off" && saveImage == true;
 export const domainIsRunning = (vmState: VMState) => domainCanReset(vmState);
 export const domainSerialConsoleCommand = ({ vm, alias } : { vm: VM, alias: string }) => {
     if (vm.displays.find(display => display.type == 'pty'))
@@ -859,6 +861,9 @@ export async function domainGet({
             supportsTPM: getDomainCapSupportsTPM(domCaps),
         };
 
+        const [suspendImage] = await call<[number]>(connectionName, objPath, 'org.libvirt.Domain', 'HasManagedSaveImage', [0], { timeout, type: 'u' });
+        props.suspendImage = suspendImage?true:false;
+
         const [state] = await call<[number[]]>(connectionName, objPath, 'org.libvirt.Domain', 'GetState', [0], { timeout, type: 'u' });
         const stateStr = DOMAINSTATE[state[0]];
 
@@ -1122,6 +1127,36 @@ export function domainResume({
     id: string,
 }): Promise<void> {
     return call(connectionName, objPath, 'org.libvirt.Domain', 'Resume', [], { timeout, type: '' });
+}
+
+export function domainSuspendToDisk({
+    connectionName,
+    id: objPath,
+    flags
+} : {
+    connectionName: ConnectionName,
+    id: string,
+    flags: Number,
+}): Promise<void> {
+    return call(connectionName, objPath, 'org.libvirt.Domain', 'ManagedSave', [flags], { timeout, type: 'u' });
+}
+export function domainSuspendImageRemove({
+    connectionName,
+    id: objPath,
+} : {
+    connectionName: ConnectionName,
+    id: string,
+}): Promise<void> {
+    return call(connectionName, objPath, 'org.libvirt.Domain', 'ManagedSaveRemove', [0], { timeout, type: 'u' });
+}
+export function domainHasSuspendImage({
+    connectionName,
+    id: objPath,
+} : {
+    connectionName: ConnectionName,
+    id: string,
+}): Promise<void> {
+    return call(connectionName, objPath, 'org.libvirt.Domain', 'HasManagedSaveImage', [0], { timeout, type: 'u' });
 }
 
 export function domainSendKey({
--- a/src/components/vm/overview/vmOverviewCard.jsx
+++ b/src/components/vm/overview/vmOverviewCard.jsx
@@ -163,7 +163,7 @@ class VmOverviewCard extends React.Component {
                             <DescriptionListTerm>{_("State")}</DescriptionListTerm>
                             <DescriptionListDescription>
                                 <StateIcon error={vm.error}
-                                               state={vm.state}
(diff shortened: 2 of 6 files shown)


garrett:
1. Suspending to disk is similar in concept to pausing, so there should _not_ be a divider between pausing and suspending to disk.
2. Cockpit uses sentence capitalization, so it should be "Suspend to disk".

What should the menu look like when resuming? With both resuming, I'm assuming at least some of the other options aren't available (you can't unpause something that's suspended or send a non-maskable interrupt, for example). Right now, it just looks like a "shut off" VM.

Additionally:

- Is "Resume" the correct word to use for both paused and suspended VMs?
- Do we have a suspended state shown in the list and on details page? (There's a "Paused" state right now.) At the moment, it just shows as "Shut off" instead of "Suspended".
- Why does pausing a VM still show suspend to disk as an option? Does it unpause and then suspend? Or does it suspend a paused VM so you have to resume twice (once from suspend, another from pausing)?
   image

DlpOne:
Thanks for the Response.

The difference between pause and managedsave are:

pause
- Pausing stops the virtual machine (VM), but the VM domain continues to exist in libvirt, so the virtual memory resources of the VM are still allocated in Host Memory. 
- If the Host (hypervisor) is restarted, the VM's state is lost, and the VM cannot be resumed.

managedsave
- managedsave stops the VM, the contents of the virtual memory are saved to the hard disk, the VM domain in libvirt is destroyed and the resources used are released.  
- The state of the VM can also be continued after a restart of the host (hypervisor)

After managedsave (Suspend to disk) the menu looks like the vm was "shut off", because in libvirt the vm Domain was destroyed. So Technically the vm is shutoff . To Resume the vm use normal start. Its the same behaviour as on a real PC after Suspend to Disk the PC is off, to continue use the power on  button.

Save could be a better name for this functionality, instead of Suspend to disk.

Currently in the list and detail view the State is shut-off. Technically is the vm Domain destroyed in libvirt. I have to check if libvirt is able to represent the mangedsave state. 

It is possible to Suspend to disk a paused vm because, mangedsave allows to save the state of an paused vm to disk. On resume (start) the state of an paused vm is still paused.

DlpOne:
I made some changes based on the points mentioned. The name of the option has been changed from "Suspend to Disk" to "Save," as this should better reflect the functionality.

Save


To indicate that a saved image (managed save state) is present for the  vm, I have added the tag "(Saved)" to the Shut Off status in both the list view and detail view. I chose not to use a separate status since the libvirt status of a vm with a saved image is "shut off," and there is no separate status for managed save.

However, there might be a better way to implement this in the code.

Save ListView
VM DetailView


In addition, I have implemented the option to remove an existing saved image. However, without a confirmation dialog. Perhaps it would be advisable to add one.

Unfortunately, libvirt does not provide an event for a deleted saved image, so I had to choose an approach that deviates from the behavior of other actions in code.

Bildschirmfoto vom 2024-01-20 18-07-40

Currently, it is not possible to filter based on the saved image tag. I am not sure if it makes sense to implement this function at the moment. The VMs are listed under "shut off" because that is the libvirt status.

FilterState

garrett:
I do not like the word "Save" for this functionality. It sounds like it will save the VM as a file and download it, which is how "Save" is used in basically every other app (web-based or native).

"Suspend to disk" was clearer to me, at least.

The state should probably also be "Suspended". (It should probably still show up when filtering by the "shut off" state.)

How is "Remove save image" any different from "Delete"? If it's supposed to remove the suspended state, it's pretty much the same starting it back up and choosing "Force shut down", right?

DlpOne:
I had considered that 'Suspend to Disk' might not be suitable because 'managedsave' also allows saving the content of paused vm, and the libvirt command is named 'managedsave.' However, I can change it to 'Suspend to disk' and the state to Suspended if you prefer.

Internally, 'Remove save image' uses the Libvirt 'ManagedSaveRemove' command, which simply deletes the saved image without starting the vm. In principle, this image is also deleted when the vm is started, making the behaviour comparable to turning on and immediately turning off the vm. The reason I implemented this function is to provide a solution in case of an error where the vm fails to start due to an issue with the save image.

jelly:
New screenshots:

image

image

image

@garrett mind taking a look? I find "remove Suspended Image" a bit confusing and not quite clear.

DlpOne:
The purpose of the function is to remove the saved VM state using ManagedSaveRemove (libvirt), which makes the VM behave as if it were switched off.
 

However, “Clear Suspend State”, “Clear Suspend” or “Set Power Off” might be better names

garrett:
There still shouldn't be a divider between pause and suspend.

Should both be called "Resume"? It's a technical implementation whether it's a quick suspend or suspending to disk, but it's basically the same experience, and they're mutually exclusive. You wouldn't keep the stave state after resuming, would you? (The place where you save to, whether RAM or disk doesn't matter much when resuming in other words. You just want to undo the action and continue, basically.)

- Pause
- Suspend to disk

And replace both with:

- Resume

It could also be "Resume from disk" for the opposite "Save to disk" to be clear when a machine is in one state versus the other, _perhaps_... although there's a status indicator for that, but there should only be one "Resume" action in the menu, regardless of the method of how it's suspended.

garrett:
@DlpOne: Yes, that's all understood. My point is that you wouldn't suspend and then pause or pause then suspend. They're mutually exclusive... I was only talking about resuming it whether it is resumed from a paused state or resuming from disk (since you wouldn't do both pausing and suspending at the same time).

And, yes, I totally agree they should have different statuses and different menu entries. :+1:

src/libvirtApi/domain.ts
```diff
@@ -1124,6 +1129,36 @@ export function domainResume({
     return call(connectionName, objPath, 'org.libvirt.Domain', 'Resume', [], { timeout, type: '' });
 }
 
+export function domainSuspendToDisk({
```
mvollmer, line 1159: I think in the code we should call these things by their libvirt names, like `domainManagedSave`. Only the UI will say "Suspend to disk".

mvollmer:
@DlpOne, thanks for this!

I have put this into our backlog: https://issues.redhat.com/browse/COCKPIT-1313

I'll rebase this, resolve conflicts, and generally mess with this a bit.  We will need tests as well.

The [RHEL 9](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/configuring_and_managing_virtualization/saving-and-restoring-virtual-machines_configuring-and-managing-virtualization) documentation calls this "Saving & Restoring" and RHEL 10 doesn't seem to mention "managedsave" anymore at all. Contrarily, RHEL 10 documents "snapshot-create" while RHEL 9 doesn't mention snapshot at all. So, is "managedsave" maybe deprecated and has been replaced with the more general snapshots feature?

mvollmer:
Here are the UI changes I think we should make:

 - The state should be "Suspended" instead of "Shut off (suspended)".
 - The action should be "Resume from disk" instead of "Run" when there is a saved state.
 - There needs to be a confirmation dialog before removing the saved state.

PR #1369, opened 2024-01-04
test: Don't import {test,machines}lib into global namespace

martinpitt:
This makes c-machines' tests consistent with all our other projects, e.g. https://github.com/cockpit-project/cockpit/commit/a703c070bb4

It unblocks using pytest, and allows us to simply add `testlib.sit()` somewhere without having to mess around with the imports and getting yelled at by static code checks.

---

Yes, this was half an hour of busywork, but everytime I want to sit() in a c-machines tests, a little kitten died. We can't have that! :cat2:

### DIFF
--- a/test/check-machines-consoles
+++ b/test/check-machines-consoles
@@ -19,14 +19,14 @@
 
 import time
 
-from machineslib import VirtualMachinesCase
-from testlib import nondestructive, skipImage, test_main, wait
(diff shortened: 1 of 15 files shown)


mvollmer (approved):
I had the same `testlib.sit()` experience and was very happy when we made this change to cockpit.

PR #1370, opened 2024-01-04
Fix page crash on invalid domain startup times

martinpitt:
When the log file contains a "starting up" timestamp which `new Date()`
cannot parse, the page oopsed with a `RangeError` from
`distanceToNow()`.

Change `domainGetStartTime()` to return a Date object instead of a raw
string, to keep the knowledge of the date format/parsing entirely within
the function. Check for invalid dates and return `null` in the error
cases, which the confirmation dialog already handles.

Also, fix the `grep` to look for the same pattern that the `.split()`
uses, to avoid mismatches.

In the confirmation dialog, rename `uptime` to `startTime`, as that's
what it is -- uptime is a duration, which is only computed inside of the
render function, but that state is an absolute time stamp.

https://issues.redhat.com/browse/RHEL-19878

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -790,16 +790,20 @@ export function domainGetStartTime({
 
         // Use libvirt APIs for getting VM start up time when it's implemented:
         // https://gitlab.com/libvirt/libvirt/-/issues/481
-        return cockpit.script(`grep 'starting up' '${logFile}' | tail -1`, options);
+        return cockpit.script(`grep ': starting up' '${logFile}' | tail -1`, options);
     })
             .then(line => {
                 // Line from a log with a start up time is expected to look like this:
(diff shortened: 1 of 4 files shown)


src/libvirtApi/domain.js
```diff
@@ -790,16 +790,20 @@ export function domainGetStartTime({
...
-                const timeStr = line.split(': starting')[0];
-
-                return Promise.resolve(timeStr);
+                const timeStr = line.split(': starting up')[0];
+                const date = new Date(timeStr);
+                return isNaN(date) ? null : date;
+            })
+            .catch(ex => {
+                console.log("Unable to detect domain start time:", ex);
+                return null;
```
cockpituous, line 806: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1370-20240104-150421-edbe2c28-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt: Yeah, belt-and-suspenders. This isn't actually reachable.

KKoukiou (approved):

PR #1373, opened 2024-01-05
Fix crash if `<redirdev>` has no `<address>`

### LINKED ISSUE #1307: Virtual Machine page is stuck "Loading resources"
When I select "Virtual Machines" from the menu it just spins with "Loading Resources". I am probably doing something dumb or maybe have some older misconfigured VMs, lol. Some details:
cockpit-304-1.fc39
cockpit-machines-301.1-1.fc39

I notice the 404's have a link syntax like:
https://192.168.1.50:9090/cockpit/@localhost/*/po.manifest.js

Do I have something configured wrong? I am not viewing this from the server itself but from my desktop on my LAN using Chrome  Version 118.0.5993.120 (Official Build) (64-bit).

### Qemu XML's:
[qemu.tar.gz](https://github.com/cockpit-project/cockpit-machines/files/13342511/qemu.tar.gz)

### Console Log:

machines:13 
        
        
       GET https://192.168.1.50:9090/cockpit/@localhost/*/po.manifest.js net::ERR_ABORTED 404 (ERROR)
machines:14 
        
        
       GET https://192.168.1.50:9090/cockpit/@localhost/*/po.js net::ERR_ABORTED 404 (ERROR)
machines:1 Refused to execute script from 'https://192.168.1.50:9090/cockpit/@localhost/*/po.manifest.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
machines:1 Refused to execute script from 'https://192.168.1.50:9090/cockpit/@localhost/*/po.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
VM803:57 Refused to apply inline style because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-mjCXvCiI7QM+qc3wUjLfE81NQ63GSsJcdb/iy9A0DgM='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.

injectGlobalStyles @ VM803:57
injectStyleTag @ VM803:7531
extractTimeoutRules @ VM803:7506
applyRules @ VM803:7657
(anonymous) @ VM803:7630
index.js:582 Scrollbar test exception: TypeError: Cannot read properties of null (reading 'appendChild')
(anonymous) @ index.js:582
(anonymous) @ index.js:689
index.js:390 Uncaught TypeError: Cannot read properties of undefined (reading 'getAttribute')
    at G8r (index.js:390:13911)
    at Zee (index.js:390:9635)
    at Function.<anonymous> (index.js:521:5968)
    at s (index.js:30:5693)
    at index.js:30:5888
    at a (index.js:30:4740)

martinpitt:
Old existing domain definitions may have `<redirdev>` elements without any `<address>` child. It's impossible to actually create these with current libvirt, as `virsh import`ing such a definition will automatically add an address -- so we can't create a test for this.

Fixes #1307

### DIFF
--- a/src/libvirt-xml-parse.js
+++ b/src/libvirt-xml-parse.js
@@ -529,9 +529,9 @@ export function parseDumpxmlForRedirectedDevices(devicesElem) {
                 type: redirdevElem.getAttribute('type'),
                 bootOrder: bootElem?.getAttribute('order'),
                 address: {
-                    type: addressElem.getAttribute('type'),
-                    bus: addressElem.getAttribute('bus'),
-                    port: addressElem.getAttribute('port'),
+                    type: addressElem?.getAttribute('type'),
+                    bus: addressElem?.getAttribute('bus'),
+                    port: addressElem?.getAttribute('port'),
                 },
                 source: {
                     mode: sourceElem?.getAttribute('mode'),


carrowood:
> @carrowood : Please feel free to test this on your machine using the [COPR build](https://dashboard.packit.dev/results/copr-builds/1241764). Thanks!

That test build solved the issue. Thanks! Now I will go clean up the root problem (that VM definition).

jelly (approved):

PR #1374, opened 2024-01-05
Avoid creating duplicate MACs

martinpitt:
In the NIC add and edit dialogs, check if there is an already existing MAC with the same value. libvirt itself does not check that, and while it is technically possible to do that, it will wreak havoc in the guest system (which we cannot influence) and also in Cockpit (which we could fix, but it's hard to do so, as we currently use the MAC address as identifier everywhere).

https://issues.redhat.com/browse/RHEL-4064

### DIFF
--- a/src/components/vm/nics/nicAdd.jsx
+++ b/src/components/vm/nics/nicAdd.jsx
@@ -36,7 +36,7 @@ import './nic.css';
 const _ = cockpit.gettext;
 
 function getRandomMac(vms) {
-    // prevent getting cycled in inforseen case where all MACs will conflict with existing ones
+    // prevent getting cycled in the unforseen case where all MACs will conflict with existing ones
     for (let i = 0; i < 42; i++) {
         const parts = ["52", "54", "00"];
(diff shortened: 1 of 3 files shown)


jelly (approved):
LGTM, the generation code already checks if it doesn't generate a duplicate MAC so users can not really run into this issue often but it's good that we check.

Also note that `getRandomMac` has a leftover console.log(mac);` from an older PR.

martinpitt:
> the generation code already checks if it doesn't generate a duplicate MAC

What do you mean? On current main you can trivially create a duplicate MAC with the NIC add and edit dialogs.

jelly:
> > the generation code already checks if it doesn't generate a duplicate MAC
> 
> What do you mean? On current main you can trivially create a duplicate MAC with the NIC add and edit dialogs.

Oh I mean, I read the generateMac code and that already makes sure that it doesn't generate duplicates so that does not require additional fixing.

martinpitt:
Ah right! Yes, that already avoids conflicts. Thanks for reviewing!

PR #1381, opened 2024-01-08
Import DynamicListForm fix from latest cockpit

martinpitt:
Adjust to changed import file name.

Fixes state confusion when deleting SSH keys in the "Create VM" dialog.

Add an integration test which exercises the "delete SSH key" button (which wasn't covered at all before), which reproduces the bug. Also test validation of an ED25519 key while we are at it.

https://issues.redhat.com/browse/RHEL-19598

-----

 - [x] Requires https://github.com/cockpit-project/cockpit/pull/19825 - update cockpit SHA after landing.


jelly (approved):
Thanks!

PR #1387, opened 2024-01-12
Add action to replace SPICE support

martinpitt:
There are some hosts which don't support SPICE, e.g. RHEL 9 or any other
OS where the SPICE support packages are not installed. This action
prepares a VM so that upgrading/changing the host or migrating to a host
without SPICE support will work without failing to restart the VM again.
In particular, it converts "spice" graphics to "vnc", "qxl" video to
"vga", and drops all other spice devices which don't have a replacement
(audio and USB passthrough).

Only show the action if the domain actually has any SPICE devices.

https://issues.redhat.com/browse/RHEL-17434

-----

TODO:
 - [x] Don't migrate a spice graphics console to vnc if there is already a vnc one (covered by test)
 - [x] Handle running VMs, change only gets effective after restart → works, but needs verbiage/sandclock/alert
 - [x] Proper explanation text
 - [x] Integration test: remove spice from VM which doesn't already have VNC
 - [x] Integration test: remove spice from stopped VM
 - [x] Integration test: remove/disable spice packages and ensure that the VM still starts up
 - [x] Integration test: failed migration; it's not possible to add something "funny" to the original XML, as both `virsh edit` and `virsh define` reject invalid XML. So let's hack `virt-xml-validate` instead.
 - [x] Fix [flake with faking virt-xml-validate](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-135703-d3dbdd79-fedora-39/log.html#85-1); the `cockpit.spawn()` promise finishes too early

## Machines: Action to Replace SPICE devices

There are some VM host machines which don't support [SPICE features](https://www.spice-space.org/spice-user-manual.html). In particular, Red Hat Enterprise Linux 9 does not support SPICE at all, and on other operating systems the SPICE support packages may not be installed.

The VM action menu got a "Replace SPICE support" action which prepares a VM so that upgrading/changing the host or migrating to a host without SPICE support will work without failing to start the VM afterwards. In particular, it converts SPICE graphics to VNC, "qxl" video to "vga", and drops all other spice devices which don't have a replacement (audio and USB passthrough).

Screen Shot 2024-01-24 at 10 03 30

**TODO**: Add spice notification label from PR #1398 if it lands in time

### DIFF
--- a/src/components/common/needsShutdown.jsx
+++ b/src/components/common/needsShutdown.jsx
@@ -77,6 +77,10 @@ export function needsShutdownWatchdog(vm) {
     return vm.persistent && vm.state === "running" && vm.inactiveXML.watchdog.action !== vm.watchdog.action;
 }
 
+export function needsShutdownSpice(vm) {
+    return vm.hasSpice !== vm.inactiveXML.hasSpice;
+}
+
 export function getDevicesRequiringShutdown(vm) {
     if (!vm.persistent)
         return [];
@@ -112,6 +116,10 @@ export function getDevicesRequiringShutdown(vm) {
     // Watchdog
     if (needsShutdownWatchdog(vm))
         devices.push(_("Watchdog"));
+
+    // SPICE
+    if (needsShutdownSpice(vm))
+        devices.push(_("SPICE"));
 
     return devices;
 }
--- a/src/components/vm/vmActions.jsx
+++ b/src/components/vm/vmActions.jsx
@@ -36,6 +36,7 @@ import { ConfirmDialog } from './confirmDialog.jsx';
 import { DeleteDialog } from "./deleteDialog.jsx";
 import { MigrateDialog } from './vmMigrateDialog.jsx';
 import { RenameDialog } from './vmRenameDialog.jsx';
+import { ReplaceSpiceDialog } from './vmReplaceSpiceDialog.jsx';
 import {
     domainCanDelete,
     domainCanInstall,
@@ -426,6 +427,20 @@ const VmActions = ({ vm, onAddErrorNotification, isDetailsPage }) => {
             </DropdownItem>
         );
         dropdownItems.push(<DropdownSeparator key="separator-migrate" />);
+    }
+
+    if (vm.inactiveXML?.hasSpice) {
+        dropdownItems.push(
+            <DropdownItem key={`${id}-replace-spice`}
+                          id={`${id}-replace-spice`}
+                          onClick={() => Dialogs.show(<ReplaceSpiceDialog vmName={vm.name}
+                                                                          vmId={vm.id}
+                                                                          connectionName={vm.connectionName}
+                                                                          vmRunning={vm.state == 'running'} />)}>
+                {_("Replace SPICE devices")}
+            </DropdownItem>
+        );
+        dropdownItems.push(<DropdownSeparator key="separator-spice" />);
     }
 
     if (domainCanRename(state)) {
--- a/src/components/vm/vmReplaceSpiceDialog.jsx
+++ b/src/components/vm/vmReplaceSpiceDialog.jsx
@@ -1,0 +1,89 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import cockpit from 'cockpit';
+import React, { useState } from 'react';
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
+import { Modal } from "@patternfly/react-core/dist/esm/components/Modal";
+import { Text, TextContent, TextList, TextListItem } from "@patternfly/react-core/dist/esm/components/Text";
+
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { useDialogs } from 'dialogs.jsx';
+import { fmt_to_fragments } from 'utils.jsx';
+
+import { NeedsShutdownAlert } from '../common/needsShutdown.jsx';
+import { domainReplaceSpice } from '../../libvirtApi/domain.js';
+
+const _ = cockpit.gettext;
+
+export const ReplaceSpiceDialog = ({ vmName, vmId, connectionName, vmRunning }) => {
+    const Dialogs = useDialogs();
+    const [error, dialogErrorSet] = useState(null);
+    const [inProgress, setInProgress] = useState(false);
+
+    function onReplace() {
+        setInProgress(true);
+
+        return domainReplaceSpice({ connectionName, id: vmId })
+                .then(() => Dialogs.close())
+                .catch(exc => {
+                    // we don't know of any case where this would fail, so this is all belt-and-suspenders
+                    // still, give the user some hint how to go on
+                    console.warn("Failed to replace SPICE devices:", exc);
+                    dialogErrorSet({
+                        dialogError: _("Failed to replace SPICE devices"),
+                        dialogErrorDetail: fmt_to_fragments(
+                            _("Please see $0 how to reconfigure your VM manually."),
+                            <a href="https://access.redhat.com/solutions/6955095" target="_blank" rel="noopener noreferrer">
+                                https://access.redhat.com/solutions/6955095
+                            </a>),
+                    });
+                })
+                .finally(() => setInProgress(false));
+    }
+
+    return (
+        <Modal position="top" variant="small" isOpen onClose={Dialogs.close}
+           title={cockpit.format(_("Replace SPICE devices in VM $0"), vmName)}
+           footer={
+               <>
+                   <Button variant='primary'
+                           id="replace-spice-dialog-confirm"
+                           isDisabled={inProgress}
+                           onClick={onReplace}>
+                       {_("Replace")}
+                   </Button>
+                   <Button variant='link' onClick={Dialogs.close}>
+                       {_("Cancel")}
+                   </Button>
+               </>
+           }>
+            { vmRunning && !error && <NeedsShutdownAlert idPrefix="spice-modal" /> }
+            {error && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
+            <TextContent>
+                <Text>{_("Reconfigure the virtual machine for a host which does not support SPICE, for host upgrades or live migration:")}</Text>
+                <TextList>
+                    <TextListItem>{_("Convert SPICE graphics console to VNC.")}</TextListItem>
+                    <TextListItem>{_("Convert QXL video card to VGA.")}</TextListItem>
+                    <TextListItem>{_("Remove SPICE audio and host devices.")}</TextListItem>
+                </TextList>
+            </TextContent>
+        </Modal>
+    );
+};
--- a/src/libvirt-xml-parse.js
+++ b/src/libvirt-xml-parse.js
@@ -246,6 +246,7 @@ export function parseDomainDumpxml(connectionName, domXml, objPath) {
     const filesystems = parseDumpxmlForFilesystems(devicesElem);
     const watchdog = parseDumpxmlForWatchdog(devicesElem);
     const vsock = parseDumpxmlForVsock(devicesElem);
+    const hasSpice = parseDumpxmlForSpice(devicesElem);
 
     const hasInstallPhase = parseDumpxmlMachinesMetadataElement(metadataElem, 'has_install_phase') === 'true';
     const installSourceType = parseDumpxmlMachinesMetadataElement(metadataElem, 'install_source_type');
@@ -289,6 +290,7 @@ export function parseDomainDumpxml(connectionName, domXml, objPath) {
         watchdog,
         vsock,
         metadata,
+        hasSpice,
     };
 }
 
@@ -479,6 +481,21 @@ export function parseDumpxmlForVsock(devicesElem) {
     }
 
     return { cid };
+}
+
+function parseDumpxmlForSpice(devicesElem) {
+    for (let i = 0; i < devicesElem.children.length; ++i) {
+        const device = devicesElem.children.item(i);
+        // also catch spicevmc
+        if (device.getAttribute("type")?.startsWith("spice"))
+            return true;
+        // qxl video is also related to SPICE
+        if (device.tagName === "video" && getSingleOptionalElem(device, "model")?.getAttribute("type") === "qxl")
+            return true;
+    }
+
+    logDebug("parseDumpxmlForSpice: no SPICE elements found in", devicesElem.children);
+    return false;
 }
 
 export function parseDumpxmlForFilesystems(devicesElem) {
--- a/src/libvirt-xml-update.js
+++ b/src/libvirt-xml-update.js
@@ -1,5 +1,5 @@
 import { getDoc, getSingleOptionalElem } from './libvirt-xml-parse.js';
-import { getNextAvailableTarget } from './helpers.js';
+import { getNextAvailableTarget, logDebug } from './helpers.js';
 
 export function changeMedia({ domXml, target, eject, file, pool, volume }) {
     const s = new XMLSerializer();
@@ -295,4 +295,57 @@ export function updateMaxMemory(domXml, maxMemory) {
     memElem.textContent = `${maxMemory}`;
 
     return s.serializeToString(doc);
+}
+
+function xmlGetXPath(doc, xpath) {
+    const elements = [];
+    let node = null;
+    const iterator = doc.evaluate(xpath, doc);
+    while ((node = iterator.iterateNext()))
+        elements.push(node);
+    return elements;
+}
+
+export function replaceSpice(domXml) {
+    // this implements https://access.redhat.com/solutions/6955095
+    logDebug("replaceSpice original XML:", domXml);
+
+    const doc = getDoc(domXml);
+
+    xmlGetXPath(doc, "/domain/devices/audio[@type='spice']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/redirdev[@type='spicevmc']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/channel[@type='spicevmc']").forEach(n => n.remove());
+
+    xmlGetXPath(doc, "//domain/devices/video").forEach(video => {
+        const model = getSingleOptionalElem(video, "model");
+        if (model?.getAttribute("type") === "qxl") {
+            // remove all attributes except for primary; don't change list while iterating
+            const removeAttrs = [];
+            for (let i = 0; i < model.attributes.length; i++) {
+                const attr = model.attributes.item(i);
+                if (attr.name !== "primary")
+                    removeAttrs.push(attr.name);
+            }
+            removeAttrs.forEach(attr => model.removeAttribute(attr));
+            model.setAttribute("type", "vga");
+        }
+    });
+
+    const spice_graphics = xmlGetXPath(doc, "//domain/devices/graphics[@type='spice']");
+    // do we already have a VNC graphics device?
+    if (xmlGetXPath(doc, "//domain/devices/graphics[@type='vnc']").length > 0) {
+        // then simply remove the spice graphics device
+        spice_graphics.forEach(graphics => graphics.remove());
+    } else {
+        // otherwise replace spice graphics with VNC
+        spice_graphics.forEach(graphics => {
+            graphics.setAttribute("type", "vnc");
+            graphics.setAttribute("port", "-1");
+            getSingleOptionalElem(graphics, "image")?.remove();
+        });
+    }
+
+    const result = (new XMLSerializer()).serializeToString(doc);
+    logDebug("replaceSpice updated XML:", result);
+    return result;
 }
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -65,6 +65,7 @@ import {
     updateBootOrder,
     updateDisk,
     updateMaxMemory,
+    replaceSpice,
 } from '../libvirt-xml-update.js';
 import { storagePoolRefresh } from './storagePool.js';
 import { snapshotGetAll } from './snapshot.js';
@@ -1041,4 +1042,22 @@ export async function domainUpdateDiskAttributes({ connectionName, objPath, targ
     const [domXml] = await call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' });
     const updatedXML = updateDisk({ diskTarget: target, domXml, readonly, shareable, busType, existingTargets, cache });
     await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
+}
+
+export async function domainReplaceSpice({ connectionName, id: objPath }) {
+    /* Ideally this would be done by virt-xml, but it doesn't offer that functionality yet
+     * see https://issues.redhat.com/browse/RHEL-17436 */
+    const [domXML] = await call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' });
+    const updatedXML = replaceSpice(domXML);
+
+    // check that updatedXML is valid; if not, throw; it needs to be updated manually
+    await (cockpit.spawn(["virt-xml-validate", "-"], { err: "message" }).input(updatedXML));
+
+    try {
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
+    } catch (ex) { // not-covered: unreachable, belt-and-suspenders
+        // if this fails, libvirt should keep the original XML
+        console.warn("domainReplaceSpice defining updated XML failed:", updatedXML); // not-covered: see above
+        throw ex; // not-covered: see above
+    }
 }
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -2321,6 +2321,177 @@ vnc_password= "{vnc_passwd}"
                                               "expired",
                                               False)
 
+    def testReplaceSpice(self):
+        m = self.machine
+        b = self.browser
+        TestMachinesCreate.CreateVmRunner(self)
+        config = TestMachinesCreate.TestCreateConfig
+
+        self.login_and_go("/machines")
+        self.waitPageInit()
+
+        def checkConnnectInfo(expect_vnc=True, expect_spice=False):
+            # if VNC is available, the inline viewer is the default
+            if expect_vnc:
+                b.wait_visible(".pf-v5-c-console__vnc canvas")
+
+            b.click("#pf-v5-c-console__type-selector")
+            b.wait_visible("#pf-v5-c-console__type-selector + .pf-v5-c-select__menu")
+            b.click("#DesktopViewer button")
+            b.wait_not_present("#pf-v5-c-console__type-selector + .pf-v5-c-select__menu")
+            if expect_vnc:
+                b.wait_in_text(".pf-v5-c-console__manual-connection", "VNC")
+            else:
+                self.assertNotIn("VNC", b.text(".pf-v5-c-console__manual-connection"))
+            if expect_spice:
+                b.wait_in_text(".pf-v5-c-console__manual-connection", "SPICE")
+            else:
+                self.assertNotIn("SPICE", b.text(".pf-v5-c-console__manual-connection"))
+
+        def doReplaceSpice(vmName, expect_running=False):
+            self.performAction(vmName, "replace-spice")
+            if expect_running:
+                b.wait_visible("#spice-modal-idle-message")
+            else:
+                b.wait_in_text(".pf-v5-c-modal-box__body", "Reconfigure")
+                self.assertFalse(b.is_present("#spice-modal-idle-message"))
+
+            b.click("#replace-spice-dialog-confirm")
+            b.wait_not_present(".pf-v5-c-modal-box")
+
+            if expect_running:
+                b.click(f"#vm-{vmName}-needs-shutdown")
+                b.wait_in_text(".pf-v5-c-popover__body", "SPICE")
+            else:
+                b.wait_not_present(f"#vm-{vmName}-needs-shutdown")
+
+            # should not have any spice or qxl video any more
+            dump = m.execute(f"virsh dumpxml --inactive {vmName}")
+            self.assertNotIn("spice", dump)
+            self.assertNotIn("qxl", dump)
+
+        #
+        # c-machines standard VM (vnc + spice on non-RHEL), running
+        #
+
+        dialog = TestMachinesCreate.VmDialog(self, sourceType='cloud',
+                                             location=config.VALID_DISK_IMAGE_PATH,
+                                             create_and_run=True,
+                                             delete=False)
+        vmCockpit = dialog.name
+        dialog.open() \
+            .fill()
+        b.click(".pf-v5-c-modal-box__footer button:contains(Create and run)")
+        self.browser.wait_not_present("#create-vm-dialog")
+        self.waitVmRow(vmCockpit, "system")
+        self.goToVmPage(vmCockpit)
+        b.wait_text(f"#vm-{vmCockpit}-system-state", "Running")
+        domainXML = m.execute(f"virsh dumpxml --inactive {vmCockpit}")
+
+        # on RHEL 9, SPICE is unsupported; on RHEL 8, we disable SPICE in VM creation; so we don't expect any device
+        if m.image.startswith("rhel") or m.image.startswith("centos"):
+            self.assertNotIn("spice", domainXML)
+            self.assertNotIn("qxl", domainXML)
+            # and thus no "Replace SPICE" menu entry
+            b.click(f"#vm-{vmCockpit}-system-action-kebab button")
+            b.wait_in_text(f"#vm-{vmCockpit}-system-action-kebab > .pf-v5-c-dropdown__menu", "Delete")
+            self.assertNotIn("Replace SPICE",
+                             b.text(f"#vm-{vmCockpit}-system-action-kebab > .pf-v5-c-dropdown__menu"))
+        else:
+            # all other OSes configure SPICE by default
+            self.assertIn("graphics type='spice'", domainXML)
+            self.assertIn("channel type='spicevmc'", domainXML)
+            checkConnnectInfo(expect_vnc=True, expect_spice=True)
+
+            doReplaceSpice(vmCockpit, expect_running=True)
+
+            # "Replace SPICE" menu entry goes away (async)
+            for _retry in range(15):
+                time.sleep(1)
+                b.click(f"#vm-{vmCockpit}-system-action-kebab button")
+                b.wait_in_text(f"#vm-{vmCockpit}-system-action-kebab > .pf-v5-c-dropdown__menu", "Delete")
+                menu = b.text(f"#vm-{vmCockpit}-system-action-kebab > .pf-v5-c-dropdown__menu")
+                b.click(f"#vm-{vmCockpit}-system-action-kebab button")
+                b.wait_not_present(".pf-v5-c-dropdown__menu")
+                if "Replace SPICE" not in menu:
+                    break
+            else:
+                self.fail("timed out waiting for Replace SPICE menu item removal")
+
+            # ensure that the VM can start without any SPICE plugins
+            # avoid uninstalling packages to keep this nondestructive
+            plugins = m.execute("find /usr/lib* -path '*qemu*spice*' | xargs").strip()
+            try:
+                m.execute(f"mkdir {self.vm_tmpdir}/qemu; for f in {plugins}; do mv $f {self.vm_tmpdir}/qemu; done")
+                caps = m.execute("virsh domcapabilities")
+                if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+                    # on Debian, spice graphics is built in, spicevmc channels are a plugin
+                    self.assertNotIn("spicevmc", caps)
+                else:
+                    # in Fedora/RHEL we don't expect any remaining capability
+                    self.assertNotIn("spice", caps)
+
+                # restart the VM to make the change effective
+                self.performAction(vmCockpit, "forceOff")
+                b.wait_not_present(f"#vm-{vmCockpit}-needs-shutdown")
+                b.click(f"#vm-{vmCockpit}-system-run")
+                b.wait_text(f"#vm-{vmCockpit}-system-state", "Running")
+                checkConnnectInfo(expect_vnc=True, expect_spice=False)
+            finally:
+                m.execute(f"for f in {plugins}; do mv {self.vm_tmpdir}/qemu/$(basename $f) $f; done")
+
+            b.wait_not_present(f"#vm-{vmCockpit}-needs-shutdown")
+
+        self.goToMainPage()
+        self.machine.execute(f"virsh destroy {vmCockpit}; virsh undefine {vmCockpit}")
+
+        # RHEL 9 doesn't support SPICE, so skip the rest of the test
+        if m.image.startswith("rhel-9") or m.image.startswith("centos-9"):
+            return
+
+        #
+        # custom spice-only VM, not running
+        #
+
+        vmSpiceOnly = "SpiceOnly"
+        self.createVm(vmSpiceOnly, graphics="spice", running=False)
+        self.waitVmRow(vmSpiceOnly, "system")
+        self.goToVmPage(vmSpiceOnly)
+        b.wait_text(f"#vm-{vmSpiceOnly}-system-state", "Shut off")
+        doReplaceSpice(vmSpiceOnly, expect_running=False)
+
+        b.click(f"#vm-{vmSpiceOnly}-system-run")
+        b.wait_text(f"#vm-{vmSpiceOnly}-system-state", "Running")
+        checkConnnectInfo(expect_vnc=True, expect_spice=False)
+
+        self.goToMainPage()
+        self.machine.execute(f"virsh destroy {vmSpiceOnly}; virsh undefine {vmSpiceOnly}")
+
+        # failed migration
+        # we can't test this properly, as it's not possible to poke invalid XML into libvirt
+        # simulate it with hacking virt-xml-validate
+
+        self.write_file("/usr/local/bin/virt-xml-validate",
+                        # ensure to drain stdin (read the XML), to avoid exiting too early
+                        "#!/bin/sh\ncat >/dev/null\necho kaputt >&2\nexit 1",
+                        perm="0755")
+
+        vmKaputt = "Kaputt"
+        self.createVm(vmKaputt, graphics="spice", running=False)
+        self.waitVmRow(vmKaputt, "system")
+        self.goToVmPage(vmKaputt)
+        b.wait_text(f"#vm-{vmKaputt}-system-state", "Shut off")
+        domainXML = m.execute(f"virsh dumpxml {vmKaputt}")
+
+        self.performAction(vmKaputt, "replace-spice")
+        b.click("#replace-spice-dialog-confirm")
+        b.wait_in_text(".pf-v5-c-modal-box .pf-v5-c-alert", "Failed to replace SPICE devices")
+        b.click(".pf-v5-c-modal-box__footer button:contains(Cancel)")
+        b.wait_not_present(".pf-v5-c-modal-box")
+
+        # no change to the XML
+        self.assertEqual(m.execute(f"virsh dumpxml {vmKaputt}"), domainXML)
+
 
 if __name__ == '__main__':
     testlib.test_main()


src/components/vm/vmRemoveSpiceDialog.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+    function onRemove() {
+        setInProgress(true);
+
+        return domainRemoveSpice({ connectionName, id: vmId })
+                .then(() => Dialogs.close())
+                .catch(exc => {
+                    console.warn("Failed to remove spice support:", exc);
+                    dialogErrorSet({
+                        dialogError: _("Failed to remove spice support"),
```
cockpituous, line 46: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmRemoveSpiceDialog.jsx
```diff
@@ -0,0 +1,78 @@
...
+                        // FIXME: link to https://access.redhat.com/solutions/6955095
+                        dialogErrorDetail: _("Do this manually FIXME"),
```
cockpituous, line 48: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmRemoveSpiceDialog.jsx
```diff
@@ -0,0 +1,78 @@
...
+                           isDisabled={inProgress}
+                           onClick={onRemove}>
+                       {_("Remove")}
+                   </Button>
+                   <Button variant='link' onClick={Dialogs.close}>
+                       {_("Cancel")}
+                   </Button>
+               </>
+           }>
+            {error && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
```
cockpituous, line 72: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
```diff
@@ -481,6 +483,21 @@ export function parseDumpxmlForVsock(devicesElem) {
...
 
+function parseDumpxmlForSpice(devicesElem) {
+    for (let i = 0; i < devicesElem.children.length; ++i) {
+        const device = devicesElem.children.item(i);
+        // also catch spicevmc
+        if (device.getAttribute("type")?.startsWith("spice"))
+            return true;
+        // qxl video is also related to SPICE
+        if (device.tagName === "video" && getSingleOptionalElem(device, "model")?.getAttribute("type") === "qxl")
+            return true;
```
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+
+    const doc = getDoc(domXml);
+
+    xmlGetXPath(doc, "/domain/devices/audio[@type='spice']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/redirdev[@type='spicevmc']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/channel[@type='spicevmc']").forEach(n => n.remove());
+
+    xmlGetXPath(doc, "//domain/devices/video").forEach(video => {
+        const model = getSingleOptionalElem(video, "model");
+        if (model?.getAttribute("type") === "qxl") {
```
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+            // remove all attributes except for primary; don't change list while iterating
+            const removeAttrs = [];
+            for (let i = 0; i < model.attributes.length; i++) {
+                const attr = model.attributes.item(i);
+                if (attr.name !== "primary")
+                    removeAttrs.push(attr.name);
```
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+            }
+            removeAttrs.forEach(attr => model.removeAttribute(attr));
+            model.setAttribute("type", "vga");
```
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+    // do we already have a VNC graphics device?
+    if (xmlGetXPath(doc, "//domain/devices/graphics[@type='vnc']").length > 0) {
+        // then simply remove the spice graphics device
+        spice_graphics.forEach(graphics => graphics.remove());
+    } else {
+        // otherwise replace spice graphics with VNC
+        spice_graphics.forEach(graphics => {
+            graphics.setAttribute("type", "vnc");
+            graphics.setAttribute("port", "-1");
+            getSingleOptionalElem(graphics, "image")?.remove();
```
cockpituous, line 344: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,23 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+    const [domXML] = await call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' });
+    const updatedXML = removeSpice(domXML);
+
+    // check that updatedXML is valid; if not, throw; it needs to be updated manually
+    await (cockpit.spawn(["virt-xml-validate", "-"], { err: "message" }).input(updatedXML));
+
+    await call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [0], { timeout, type: 'u' });
+    try {
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
+    } catch (ex) {
```
cockpituous, line 1129: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,23 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+        // if this fails, we are in trouble; but at least try to restore the original XML
+        console.warn("domainRemoveSpice defining updated XML", updatedXML, "failed, restoring original:", ex);
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [domXML], { timeout, type: 's' });
+        throw ex;
```
cockpituous, line 1133: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240112-155815-cbcdc423-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmRemoveSpiceDialog.jsx
cockpituous, line 46: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240115-083924-04c4f941-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240115-083924-04c4f941-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
[race condition](https://artifacts.dev.testing-farm.io/6546d860-f4f6-4875-89cb-decd64411bfa/), trying a fix.

src/components/vm/vmRemoveSpiceDialog.jsx
cockpituous, line 46: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmRemoveSpiceDialog.jsx
cockpituous, line 48: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmRemoveSpiceDialog.jsx
cockpituous, line 72: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
cockpituous, line 1129: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
cockpituous, line 1133: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240117-024845-4e8cc37f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

garrett:
Here's a first pass at mockups for what we talked about with regard to the SPICE removal.

image

I'm still working on how it would be when SPICE removal isn't mandatory. Strings will change, and it's debatable if it should show up in the list or not for non-RHEL distros. But for RHEL where it is supported but won't be in a future update, we probably still want to show it in the list... but perhaps without the ⚠️ warning icon.

On the details page, I'm still thinking about having a alert warning at the top with a button that shows the same modal.

garrett:
I have a version with select all/none underneith, in case the inline might be a problem to implement.

select under

I also tried using a [PF widget for bulk selection](https://www.patternfly.org/patterns/bulk-selection/#bulk-selection), but it's just weird, looks out of place, and not obvious:

weird widget

martinpitt:
Thanks @garrett for the design! In this PR I'll implement the first step of it: migrating a single VM, from the kebab menu. This is to get the backend (XML fiddling) and all the test scenarios right. Then I'll follow up with the alert, multi-VM migration and so on.

> I'm still working on how it would be when SPICE removal isn't mandatory. Strings will change, and it's debatable if it should show up in the list or not for non-RHEL distros.

IMHO it should appear in the kebab menu everywhere. The "live migration" scenario makes this useful/applicable for all distros, i.e. if you want to migrate a VM from a SPICE-supporting host to a Debian/Fedora/etc. machine which doesn't have the spice packages installed, or even to a RHEL 9 machine. But of course in that case it should *not* annoy you with pro-active alerts. That should be restricted to RHEL 8/9. WDYT?

garrett:
> if you want to migrate a VM from a SPICE-supporting host to a Debian/Fedora/etc. machine which doesn't have the spice packages installed, or even to a RHEL 9 machine

Could we handle that in the migration process? Is there some way to tell what display support the remote host has?

martinpitt:
@garrett Potentially, yes; I'm not sure right now up to which point we control the SSH connection ourselves, or if libvirt does it. In the former case, we could run `virsh domcapabilities` on the remote and check `<graphics>`.

src/libvirt-xml-parse.js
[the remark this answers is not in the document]
martinpitt: Can't cover this on Fedora, as our Fedora images don't have the QXL driver. The tests cover it on RHEL 8 and other distros, though.

src/libvirt-xml-update.js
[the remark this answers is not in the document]
martinpitt: dito

src/libvirtApi/domain.js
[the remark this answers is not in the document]
martinpitt: belt-and-suspenders, I'll add an annotation.

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-104854-54dc3b42-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+
+    const doc = getDoc(domXml);
+
+    xmlGetXPath(doc, "/domain/devices/audio[@type='spice']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/redirdev[@type='spicevmc']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/channel[@type='spicevmc']").forEach(n => n.remove());
+
+    xmlGetXPath(doc, "//domain/devices/video").forEach(video => {
+        const model = getSingleOptionalElem(video, "model");
+        if (model?.getAttribute("type") === "qxl") {
```
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-104854-54dc3b42-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+            // remove all attributes except for primary; don't change list while iterating
+            const removeAttrs = [];
+            for (let i = 0; i < model.attributes.length; i++) {
+                const attr = model.attributes.item(i);
+                if (attr.name !== "primary")
+                    removeAttrs.push(attr.name);
```
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-104854-54dc3b42-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+            }
+            removeAttrs.forEach(attr => model.removeAttribute(attr));
+            model.setAttribute("type", "vga");
```
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-104854-54dc3b42-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,23 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+    await (cockpit.spawn(["virt-xml-validate", "-"], { err: "message" }).input(updatedXML));
+
+    await call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [0], { timeout, type: 'u' });
+    try {
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
+    } catch (ex) { // not-covered: unreachable, belt-and-suspenders
+        // if this fails, we are in trouble; but at least try to restore the original XML
+        console.warn("domainRplaceSpice defining updated XML", updatedXML, "failed, restoring original:", ex); // not-covered
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [domXML], { timeout, type: 's' }); // not-covered
+        throw ex; // not-covered
```
cockpituous, line 1133: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-104854-54dc3b42-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
@garrett : First iteration of the dialog, still without the multi-VM migration feature, and with what we discussed in chat:

image

garrett:
"Remove" instead of "Drop", especially since "drop" can now also mean "add"* in addition to "remove". (I still think this is absurd, but it is what it is."

\* It's originally "[drop a new album](https://idioms.thefreedictionary.com/to+drop+a+new+album)", but it's been extended to things like "drop a new trailer", and even as absurd as "Your devices keep getting better with [feature drops across Pixel phones](https://blog.google/products/pixel/feature-drop-december-2022/)" — where they are talking about _adding_ features, not _removing_ them. :exploding_head: It's really confusing as some articles will say "drop XYZ feature" and mean _add_ and while others — sometimes on the very same site — will be about _removing_ the feature.

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-135703-d3dbdd79-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-135703-d3dbdd79-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-135703-d3dbdd79-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-135703-d3dbdd79-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,23 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+    await (cockpit.spawn(["virt-xml-validate", "-"], { err: "message" }).input(updatedXML));
+
+    await call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [0], { timeout, type: 'u' });
+    try {
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
+    } catch (ex) { // not-covered: unreachable, belt-and-suspenders
+        // if this fails, we are in trouble; but at least try to restore the original XML
+        console.warn("domainRplaceSpice defining updated XML", updatedXML, "failed, restoring original:", ex); // not-covered:
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [domXML], { timeout, type: 's' }); // not-covered:
+        throw ex; // not-covered:
```
cockpituous, line 1133: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240118-135703-d3dbdd79-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
@garrett for your review:

I reached a point here where this is a single complete feature. The only interface right now is the menu entry:

image

It only appears for a VM which actually has any SPICE device. This brings up the dialog:

image

The "Changes will take effect" alert of course only appears on running VMs. It's a shared component, so it looks consistent with all the other dialogs. We talked about changing it yesterday, I'll look into that next (as a separate PR). Once you click "Replace", it does the XML migration (with restoring the original on any failure), and then goes back to the VM detail, where the main observable effect is that the consoles change. If the VM was running, there is also the pending indicator (again, shared component):

image

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-062100-f9c8ea53-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-062100-f9c8ea53-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-062100-f9c8ea53-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-062100-f9c8ea53-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
cockpituous, line 1133: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-062100-f9c8ea53-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-065449-19f5edda-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-065449-19f5edda-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-065449-19f5edda-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-065449-19f5edda-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
cockpituous, line 1133: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-065449-19f5edda-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
martinpitt: Can't cover this on Fedora, as our Fedora images don't have the QXL driver. The tests cover it on RHEL 8 and other distros, though.

src/libvirt-xml-update.js
martinpitt: dito

src/libvirt-xml-update.js
martinpitt: Dito. I tested this manually after installing the QXL package.

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-080518-037ed95d-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-080518-037ed95d-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-080518-037ed95d-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240119-080518-037ed95d-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
martinpitt: Can't cover this on Fedora, as our Fedora images don't have the QXL driver. The tests cover it on RHEL 8 and other distros, though. I tested it manually on F39 as well, and qxl → vga migration works fine.

src/libvirt-xml-update.js
martinpitt: Dito

src/libvirt-xml-update.js
martinpitt: Dito

src/libvirt-xml-update.js
martinpitt: Dito

marusak:
Nice, thank you @martinpitt ! What are we replacing the SPICE with? I though we just remove it. What followup do you plan?

martinpitt:
@marusak The [dialog](https://private-user-images.githubusercontent.com/200109/297972161-44ade067-ad22-4efd-a92e-fd237b2c6a76.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDU2NTg5MjIsIm5iZiI6MTcwNTY1ODYyMiwicGF0aCI6Ii8yMDAxMDkvMjk3OTcyMTYxLTQ0YWRlMDY3LWFkMjItNGVmZC1hOTJlLWZkMjM3YjJjNmE3Ni5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMTE5JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDExOVQxMDAzNDJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1hNTNjNmRhY2E4YmU1MjVmMjIxNDhiYTFlOWI5MTM0MDZjODc1ZTIzNjdiZDIxODIyZDE5NTI2NTk2ZDE3ZjcxJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.NuJizBUat7pMZ--SmQOK0DjxYKYZBjVeTc4PzPfegMg) hopefully explains the precise replacements? The commit message says the same.

Follow-ups: The [design](https://github.com/cockpit-project/cockpit-machines/pull/1387#issuecomment-1895348199) as well as our meeting GDoc has a few outstanding issues: (1) Show the popup/alert if the host doesn't support SPICE (and thus starting will fail), and (2) add mass conversion of multiple VMs to the dialog.

marusak:
> The [dialog](https://private-user-images.githubusercontent.com/200109/297972161-44ade067-ad22-4efd-a92e-fd237b2c6a76.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDU2NTg5MjIsIm5iZiI6MTcwNTY1ODYyMiwicGF0aCI6Ii8yMDAxMDkvMjk3OTcyMTYxLTQ0YWRlMDY3LWFkMjItNGVmZC1hOTJlLWZkMjM3YjJjNmE3Ni5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMTE5JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDExOVQxMDAzNDJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1hNTNjNmRhY2E4YmU1MjVmMjIxNDhiYTFlOWI5MTM0MDZjODc1ZTIzNjdiZDIxODIyZDE5NTI2NTk2ZDE3ZjcxJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.NuJizBUat7pMZ--SmQOK0DjxYKYZBjVeTc4PzPfegMg) hopefully explains the precise replacements? The commit message says the same.

Ah yes, it did not load the first time. Makes perfect sense and I like how you managed to "fix" the issue rather than just remove all the stuff.

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+        }
+    });
+
+    const spice_graphics = xmlGetXPath(doc, "//domain/devices/graphics[@type='spice']");
+    // do we already have a VNC graphics device?
+    if (xmlGetXPath(doc, "//domain/devices/graphics[@type='vnc']").length > 0) {
+        // then simply remove the spice graphics device
+        spice_graphics.forEach(graphics => graphics.remove());
+    } else {
+        // otherwise replace spice graphics with VNC
```
mvollmer, line 340: Say there are multiple spice entries. Then, if there is a single vnc entry, all spice entries are removed, leaving the single vnc entry. But if there is no vnc entry, each spice entry is converted into a vnc entry, leaving multiple vnc entries. Also, spice specific elements might still be in those VNC entries. That seems hairy.

What about always removing all spice entries, and then adding a generic vnc entry if there isn't one already?

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,23 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+    // check that updatedXML is valid; if not, throw; it needs to be updated manually
+    await (cockpit.spawn(["virt-xml-validate", "-"], { err: "message" }).input(updatedXML));
+
+    await call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [0], { timeout, type: 'u' });
+    try {
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
+    } catch (ex) { // not-covered: unreachable, belt-and-suspenders
+        // if this fails, we are in trouble; but at least try to restore the original XML
+        console.warn("domainRplaceSpice defining updated XML", updatedXML, "failed, restoring original:", ex); // not-covered: see above
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [domXML], { timeout, type: 's' }); // not-covered: see above
```
mvollmer, line 1132: We might want to ignore errors here.

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,23 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+export async function domainReplaceSpice({ connectionName, id: objPath }) {
+    /* Ideally this would be done by virt-xml, but it doesn't offer that functionality yet
+     * see https://issues.redhat.com/browse/RHEL-17436 */
+    const [domXML] = await call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' });
+    const updatedXML = replaceSpice(domXML);
+
+    // check that updatedXML is valid; if not, throw; it needs to be updated manually
+    await (cockpit.spawn(["virt-xml-validate", "-"], { err: "message" }).input(updatedXML));
+
+    await call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [0], { timeout, type: 'u' });
```
mvollmer, line 1126: Is there no way to update the XML of a existing domain?

mvollmer:
I am disappointed that Cockpit has to have all this explicit code for mangling the XML... There should at least be some script provided by Red Hat that Cockpit can run, and that people can also run without Cockpit.

src/libvirt-xml-update.js
martinpitt: There can only be one `<graphics>` of each type, otherwise libvirt will complain. So in *practice* this should be fine, but conceptually you are right. I'm just a bit wary of completely deleting the spice device, as they often have things like virtual PCI slot, or the IP to listen on, and I'm e.g. afraid of replacing a localhost or "my internal network" IP with a hardcoded one. Essentially I'm implementing the [official documentation](https://access.redhat.com/solutions/6955095). This would really be something where I'd appreciate the libvirt team's input. I asked in the Jira ticket a few days ago.

src/libvirtApi/domain.js
martinpitt: If this fails, we are in big big trouble. We just killed the user's VM and restoring the backup also failed. Then I think it's ok for the UI to show the "migration failed" error, instead of silently pretending that it worked. But of course your VM will not exist any more either way.

src/libvirtApi/domain.js
martinpitt: Nope. `virsh edit` may do this in a more robust way (doing undefine/redefine much closer together), but that's not accessible through the API.

martinpitt:
> I propose to block on that.

We can wait for two weeks or so, but remember that we need this in RHEL 8. TBH I'm not super hopeful that `virt-xml` will learn about this in two months, let alone in the one week that we'd need to test it, find the bugs, get them fixed, landed in RHEL 8, and have everything ready by mid-February. So if we decide to block on it, then the result is pretty much "we won't have this feature in RHEL 8, ever", and that feels strategically bad. I'm afraid we'll need to go with this.

src/libvirt-xml-update.js
martinpitt: > Also, spice specific elements might still be in those VNC entries

If there are entries which we don't recognize, I specifically *want* the auto-migration to fail, rather than stomping over the admin's customizations. I'm ok with converting known cases, but if we encounter something which we don't understand (i.e. when `virt-xml-validate` says :-1: ), we revert and say "sorry, go do it yourself".

mvollmer:
> I'm afraid we'll need to go with this.

Yes, convinced.

src/libvirtApi/domain.js
crobinso: @martinpitt `Define` works for an existing VM, virt-xml doesn't do anything with `undefine` for example. Were you hitting a specific error?

src/libvirt-xml-update.js
crobinso: @martinpitt yes I think `s/type='spice'/type='vnc'/` is the sensible way to handle it for `<graphics>`. For common use cases it should do the right thing.

Libvirt doesn't heavily validate `<graphics>` XML like it does for some other devices; most of the config bits that work for spice and not VNC will be thrown away at XML parse time. `<listen>` config works mostly the same for both types though

src/libvirtApi/domain.js
germanovm: In this apocalyptic scenario, would it be possible to dump the old and new xml to a file on the host and tell the full paths to the user in the console warning? So the user can go and troubleshoot from there, either making the old one define or fixing whatever is wrong with the new one.

src/libvirt-xml-update.js
martinpitt: @crobinso Thanks. Just to make sure: is it ever a thing to have more than one spice graphics device? My gut feeling is that doing an 1:1 conversion of spice to vnc is conceptually cleaner than removing them all and adding back a standard one, as that silently removes customizations.

src/libvirtApi/domain.js
martinpitt: @crobinso Indeed, nice! I don't remember any more where/how I saw this, but indeed both `virsh define` as well as the API seem to do this right -- if it fails due to invalid XML, the previous definition stays intact. That makes this much more robust. I dropped this, and simplified the cleanup below.

src/libvirtApi/domain.js
martinpitt: This fortunately went away entirely, see the [discussion above](https://github.com/cockpit-project/cockpit-machines/pull/1387/files#r1461069400). I think this should be ok now. Thanks!

src/components/common/needsShutdown.jsx
```diff
@@ -113,6 +117,10 @@ export function getDevicesRequiringShutdown(vm) {
     if (needsShutdownWatchdog(vm))
         devices.push(_("Watchdog"));
 
+    // SPICE
+    if (needsShutdownSpice(vm))
+        devices.push(_("SPICE"));
```
cockpituous, line 122: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmActions.jsx
```diff
@@ -428,6 +429,20 @@ const VmActions = ({ vm, onAddErrorNotification, isDetailsPage }) => {
...
     }
 
+    if (vm.inactiveXML?.hasSpice) {
+        dropdownItems.push(
+            <DropdownItem key={`${id}-replace-spice`}
+                          id={`${id}-replace-spice`}
+                          onClick={() => Dialogs.show(<ReplaceSpiceDialog vmName={vm.name}
+                                                                          vmId={vm.id}
+                                                                          connectionName={vm.connectionName}
+                                                                          vmRunning={vm.state == 'running'} />)}>
```
cockpituous, line 439: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+import { NeedsShutdownAlert } from '../common/needsShutdown.jsx';
+import { domainReplaceSpice } from '../../libvirtApi/domain.js';
+
+const _ = cockpit.gettext;
+
+export const ReplaceSpiceDialog = ({ vmName, vmId, connectionName, vmRunning }) => {
+    const Dialogs = useDialogs();
+    const [error, dialogErrorSet] = useState(null);
+    const [inProgress, setInProgress] = useState(false);
```
cockpituous, line 38: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+    function onReplace() {
+        setInProgress(true);
```
cockpituous, line 41: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+        return domainReplaceSpice({ connectionName, id: vmId })
+                .then(() => Dialogs.close())
+                .catch(exc => {
```
cockpituous, line 45: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+                    // we don't know of any case where this would fail, so this is all belt-and-suspenders
+                    // still, give the user some hint how to go on
+                    console.warn("Failed to replace SPICE devices:", exc);
+                    dialogErrorSet({
+                        dialogError: _("Failed to replace SPICE devices"),
+                        dialogErrorDetail: fmt_to_fragments(
+                            _("Please see $0 how to reconfigure your VM manually."),
+                            <a href="https://access.redhat.com/solutions/6955095" target="_blank" rel="noopener noreferrer">
```
cockpituous, line 53: These 6 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+                                https://access.redhat.com/solutions/6955095
+                            </a>),
```
cockpituous, line 55: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+                    });
+                })
+                .finally(() => setInProgress(false));
```
cockpituous, line 58: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    return (
+        <Modal position="top" variant="small" isOpen onClose={Dialogs.close}
+           title={cockpit.format(_("Replace SPICE devices in VM $0"), vmName)}
+           footer={
```
cockpituous, line 64: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+               <>
+                   <Button variant='primary'
+                           id="replace-spice-dialog-confirm"
+                           isDisabled={inProgress}
+                           onClick={onReplace}>
+                       {_("Replace")}
+                   </Button>
+                   <Button variant='link' onClick={Dialogs.close}>
+                       {_("Cancel")}
+                   </Button>
```
cockpituous, line 74: These 9 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/vmReplaceSpiceDialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+            {error && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
+            <TextContent>
+                <Text>{_("Reconfigure the virtual machine for a host which does not support SPICE, for host upgrades or live migration:")}</Text>
+                <TextList>
+                    <TextListItem>{_("Convert SPICE graphics console to VNC.")}</TextListItem>
+                    <TextListItem>{_("Convert QXL video card to VGA.")}</TextListItem>
+                    <TextListItem>{_("Remove SPICE audio and host devices.")}</TextListItem>
+                </TextList>
+            </TextContent>
+        </Modal>
```
cockpituous, line 87: These 11 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
     return s.serializeToString(doc);
 }
+
+function xmlGetXPath(doc, xpath) {
+    const elements = [];
+    let node = null;
+    const iterator = doc.evaluate(xpath, doc);
+    while ((node = iterator.iterateNext()))
+        elements.push(node);
+    return elements;
```
cockpituous, line 306: These 7 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+}
+
+export function replaceSpice(domXml) {
```
cockpituous, line 309: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+    // this implements https://access.redhat.com/solutions/6955095
+    logDebug("replaceSpice original XML:", domXml);
```
cockpituous, line 311: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+
+    const doc = getDoc(domXml);
```
cockpituous, line 313: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+
+    xmlGetXPath(doc, "/domain/devices/audio[@type='spice']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/redirdev[@type='spicevmc']").forEach(n => n.remove());
+    xmlGetXPath(doc, "//domain/devices/channel[@type='spicevmc']").forEach(n => n.remove());
```
cockpituous, line 317: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+                const attr = model.attributes.item(i);
+                if (attr.name !== "primary")
+                    removeAttrs.push(attr.name);
+            }
+            removeAttrs.forEach(attr => model.removeAttribute(attr));
+            model.setAttribute("type", "vga");
+        }
+    });
+
+    const spice_graphics = xmlGetXPath(doc, "//domain/devices/graphics[@type='spice']");
```
cockpituous, line 334: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+    // do we already have a VNC graphics device?
+    if (xmlGetXPath(doc, "//domain/devices/graphics[@type='vnc']").length > 0) {
```
cockpituous, line 336: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+        // then simply remove the spice graphics device
+        spice_graphics.forEach(graphics => graphics.remove());
```
cockpituous, line 338: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+    } else {
+        // otherwise replace spice graphics with VNC
+        spice_graphics.forEach(graphics => {
+            graphics.setAttribute("type", "vnc");
+            graphics.setAttribute("port", "-1");
+            getSingleOptionalElem(graphics, "image")?.remove();
```
cockpituous, line 344: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
```diff
@@ -296,3 +296,56 @@ export function updateMaxMemory(domXml, maxMemory) {
...
+        });
+    }
+
+    const result = (new XMLSerializer()).serializeToString(doc);
+    logDebug("replaceSpice updated XML:", result);
+    return result;
```
cockpituous, line 350: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,21 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
                 return call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
             });
 }
+
+export async function domainReplaceSpice({ connectionName, id: objPath }) {
```
cockpituous, line 1117: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,21 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+    /* Ideally this would be done by virt-xml, but it doesn't offer that functionality yet
+     * see https://issues.redhat.com/browse/RHEL-17436 */
+    const [domXML] = await call(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' });
+    const updatedXML = replaceSpice(domXML);
```
cockpituous, line 1121: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,21 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+
+    // check that updatedXML is valid; if not, throw; it needs to be updated manually
+    await (cockpit.spawn(["virt-xml-validate", "-"], { err: "message" }).input(updatedXML));
```
cockpituous, line 1124: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.js
```diff
@@ -1112,3 +1113,21 @@ export function domainUpdateDiskAttributes({ connectionName, objPath, target, re
...
+
+    try {
+        await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'DomainDefineXML', [updatedXML], { timeout, type: 's' });
```
cockpituous, line 1127: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-073835-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-082647-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-082647-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-082647-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-082647-c9864bc3-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-085740-ae88084a-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-085740-ae88084a-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-085740-ae88084a-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240122-085740-ae88084a-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
garrett: > I'm ok with converting known cases, but if we encounter something which we don't understand (i.e. when virt-xml-validate says 👎 ), we revert and say "sorry, go do it yourself".

How are we handling error cases? Will the VM still boot? (It would be a warning then, which should probably be dismissable if it's not a problem. If it is a problem, then we need to show it until it's not a problem, and provide some guidance on how to fix it, if we cannot already easily fix it from within Cockpit.)

martinpitt: If the conversion fails, the VM remains unchanged -- i.e. it will still boot. You'd just get an error message in the dialog:

image

At this point there isn't an automatic solution any more. But for the record, we don't currently know of a current case for this -- to test this we need to explicitly fake/break `virt-xml-validate`.

garrett: After someone fixes the problem, does the error message automatically go away?

martinpitt: No, it doesn't -- it's a dialog, at this point you can only cancel it. I.e. at this point the user only has the options to convert the XML manually (following https://access.redhat.com/solutions/6955095 but it *will* be incomplete as we implemented all the steps there), or reconsider and maybe move the VM to a host which does support spice. At this point we can really only throw our hands in the air, I'm afraid. (But again: This should be an exceptionally rare scenario)

garrett: Oh, sorry; for whatever reason, I thought it was at the top of the list after the modal was finished. OK. That works! Thanks!

I guess ordinarily, it'll be near-instant unless it hits an error like this?

martinpitt: Yes, both failure and success are near-instant. There is just some XML mangling going on, no variable-length or expensive operation.

crobinso: > @crobinso Thanks. Just to make sure: is it ever a thing to have more than one spice graphics device? My gut feeling is that doing an 1:1 conversion of spice to vnc is conceptually cleaner than removing them all and adding back a standard one, as that silently removes customizations.

Libvirt rejects multiple `<graphics>` with the same `type` value so I don't think you have to worry about that case. In theory someone could have a spice and vnc device already configured, in which case converting spice to another vnc device will fail validation since 2x vnc isn't allowed, but IMO that's obscure enough to leave it for the 'we tried, fix it yourself pls' case

martinpitt: We regularly encounter that case, and this PR even covers it with an integration test. That's why we have to do the if/else here, and `virt-xml` will have to do the same.

Thanks for reviewing and confirming! So I suppose this is good enough.

mvollmer: Thanks, I get it now, too. Just one thing: Do we need to explicitly look for "vnc" as the alternative? Or should we be ok to remove all spice entries as long as there is _any_ other graphics entry? Maybe vnc is going to cause problems? *shrug* I guess that's very unlikely.

src/libvirtApi/domain.js
mvollmer: Yay!

src/libvirtApi/domain.js
mvollmer, line 1124: Is there any advantage to doing a explicit validation step before calling "DomainDefineXML"? Wouldn't "DomainDefineXML" do the validation again and return the same error?

src/libvirt-xml-update.js
martinpitt: That's a question for @crobinso or @germanovm -- I understand the [request](https://issues.redhat.com/browse/RHEL-17434) was rather specific about converting spice to VNC consoles. I also agree that it makes sense -- e.g. having an additional local `sdl` or `gtk` console won't help you much as the point is that we want to keep a *remote* console.

src/libvirtApi/domain.js
martinpitt: Good point -- @crobinso @germanovm WDYT? Is this redundant/equivalent, or is virt-xml-validate stronger than what DomainDefineXML does? Thanks!

martinpitt: For history: I originally did that to get a safety check before undefining the VM. But since then I took that out and just redefine (which is more atomic). So indeed I figure this is redundant now, but I'll rather wait for confirmation from the libvirt devs.

garrett (dismissed):
This looks great for a first pass!

It gets the official seal of approval:

:seal: :+1:

src/libvirtApi/domain.js
martinpitt: I.e. I tried `virsh destroy test1`, edit ./.config/libvirt/qemu/test1.xml to add a `<foo bar='on'/>` into `<graphics>`. But `virsh start test1` works fine, and `virsh dumpxml --inactive test1` doesn't show the extra `<foo>` tag. I tried the same with adding an unknown attribute to `<redirdev bus='usb' type='spicevmc'>`, same effect -- unknown stuff gets ignored.

And when I replace e.g. `bus='usb'` to `bus='usc'` and try to start test1, then the VM disappears from libvirt completely.

crobinso: editing XML behind libvirt's back is tough. there's errors which still will load the XML and only fail at `virsh start` time, but a bad enum value like `usc` will cause it to completely fail to parse and disappear like you say.

maybe start with some spice specific very obscure config like '<channel name='cursor' mode='secure'/> in the spice `<graphics>` block. That bit isn't valid for VNC and the XML validator will reject it after s/spice/vnc/

different tactic: add `<gl enable='yes' rendernode='/fake/rendernode/path/make/cockpit/fail'/>` and then change cockpit code to match on that fake path and trigger error in some way. that XML will parse but the VM won't run obviously

martinpitt: Ah, @crobinso I didn't see your most recent reply yet when I wrote back. That `<channel name='cursor' mode='secure'/>` sounds like an excellent way -- it wouldn't even be a hack, but something that could *actually* occur on a real system. However, when I do that, defining the updated XML just works, and libvirt automatically cleans up the XML to

```xml
   <graphics type='vnc' port='-1' autoport='yes' listen='127.0.0.1'>
      <listen type='address' address='127.0.0.1'/>
   </graphics>
```
At the same time, virt-xml-validate seems to be happy with the vnc + cursor/secure thing as well. So it seems this is actually quite unbreakable :grin: 

So I think I'll actually stand by my comment above -- I'm happy to tweak this further in the future, but it feels "good enough"?

mvollmer: > I'm happy to tweak this further in the future, but it feels "good enough"?

Agreed. Covering the error case in the test is important and running virt-xml-validate is not wrong, and since it does a different kind of validation than DefineXML, it might even add value.

test/check-machines-create
```diff
@@ -2321,6 +2322,182 @@ vnc_password= "{vnc_passwd}"
...
+            m.execute(f"mkdir {self.vm_tmpdir}/qemu; for f in {plugins}; do mv $f {self.vm_tmpdir}/qemu; done")
+            caps = m.execute("virsh domcapabilities")
+            if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+                # on Debian, spice graphics is built in, spicevmc channels are a plugin
+                self.assertNotIn("spicevmc", caps)
+            else:
+                # in Fedora/RHEL we don't expect any remaining capability
+                self.assertNotIn("spice", caps)
+
+            yield
```
mvollmer, line 2340: What does this `yield` do?

mvollmer: Oh, I get it, context manager! Fancy!

test/check-machines-create
```diff
@@ -2321,6 +2322,182 @@ vnc_password= "{vnc_passwd}"
...
+            b.wait_text(f"#vm-{vmCockpit}-system-state", "Running")
+            checkConnnectInfo(expect_vnc=True, expect_spice=False)
+
+        b.wait_not_present(f"#vm-{vmCockpit}-needs-shutdown")
+
+        self.goToMainPage()
+        self.machine.execute(f"virsh destroy {vmCockpit}; virsh undefine {vmCockpit}")
+
+        #
+        # custom spice-only VM, not running
```
mvollmer, line 2457: Can this part of the test run on RHEL?

I am a bit worried that we don't cover the actual use case: Replacing SPICE on RHEL in a VM that was created on an older version of that RHEL...

mvollmer (approved):

mvollmer:
@martinpitt, I am sure you have considered this but: The tests don't run the new code on RHEL, only on OSes that still have SPICE. Is there any way to run that code on RHEL?

src/libvirtApi/domain.js
martinpitt: Ack, closing this thread then. @crobinso thanks a lot for your help! :hugs:

mvollmer:
> @martinpitt, I am sure you have considered this but: The tests don't run the new code on RHEL, only on OSes that still have SPICE. Is there any way to run that code on RHEL?

Ok, I learned that we can still create SPICE VMs on rhel-8 using the command line tools. Let's do that in the test so that we actually run the code where it is meant to be run (on rhel-8). As a followup, if you want.

test/check-machines-create
martinpitt: Right, you can mentally replace it with the code block inside the `with:`

test/check-machines-create
martinpitt: Yes, let's do that! We only need to skip the "create with UI" part.

martinpitt:
@mvollmer You are right -- let's do that right away here. I changed the test to run the non-UI creation bits on RHEL 8 as well.

src/libvirt-xml-parse.js
cockpituous, line 494: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240124-084953-cd9e826f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 321: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240124-084953-cd9e826f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 327: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240124-084953-cd9e826f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/libvirt-xml-update.js
cockpituous, line 330: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1387-20240124-084953-cd9e826f-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

mvollmer (approved):
Thanks!

PR #1388, opened 2024-01-15
test: add some checks about creating new qcow2 or raw disk in the VM creation dialog

yunmingyang:


martinpitt:
I don't understand the purpose of this -- these are already covered by testCreateFileSource(). Can you please describe what this does? Can it perhaps be added to that existing test?

yunmingyang:
Sure, it checks the details of "Storage" selection, to make sure there are two options which are "Create new qcow2 volume" and "Create new raw volume", and when select these two options, there will be no "Volume" option shown on the VM creation dialog. The reason why adding this case is that this case looks more intuitive. If adding this case to the existing test, checking "Volume" option is the only thing needed. I will try to do that

yunmingyang:
@martinpitt  Ah, after I re-check cases, I think the public method, `VmDialog.fill`,  had already done the checks in this PR, near `if self.storage_pool in [NEW_VOLUME_QCOW2, NEW_VOLUME_RAW]`. Also, as you said before, these lines are executed in testCreateFileSource. So please close this PR, thanks a lot.

martinpitt:
OK, thanks for checking nevertheless! You should be able to close your own PRs as well, but I'll do it now.

PR #1394, opened 2024-01-19
Restyle "needs shutdown" dialog alert and page label

martinpitt:
Tone down the dialog alert from "warning" to "custom" -- the user did nothing wrong, it's
an expected part of the workflow. Also use the "pending" icon.

Change the on-page label and popover to cyan and custom alert level to
match the dialog alert.

----

This was discussed with @garrett yesterday as a side issue in #1387. As this is a shared component, it should have a separate PR. This will apply to all dialogs, the vCPU one is easiest to test.

Current main:

needs-shutdown-alert-original

image

This PR:

needs-shutdown-alert-pr

image

### DIFF
--- a/src/components/common/needsShutdown.jsx
+++ b/src/components/common/needsShutdown.jsx
@@ -119,7 +119,7 @@ export function getDevicesRequiringShutdown(vm) {
 export const NeedsShutdownTooltip = ({ iconId, tooltipId }) => {
     return (
         <Tooltip id={tooltipId} content={NEEDS_SHUTDOWN_MESSAGE}>
-            <Icon status="info">
+            <Icon status="custom">
                 <PendingIcon id={iconId} />
             </Icon>
         </Tooltip>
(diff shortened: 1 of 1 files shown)


martinpitt:
@garrett Good idea! I did that and updated the screenshots in the description.

garrett (changes requested):
Looks good!

However, I also noticed that the hourglass icons are still blue in the details area:

image

(CPU and Watchdog in this screenshot)

martinpitt:
@garrett Well spotted, done:

image

The "edit" link buttons remain blue, though -- they are throughout cockpit and should stay IMHO.

garrett:
> The "edit" link buttons remain blue, though -- they are throughout cockpit and should stay IMHO.

Yeah, those are links and are distinct. :+1:

garrett (approved):
Thanks for making this a separate PR and adjusting it a few times too! Approved!

PR #1397, opened 2024-01-22
Port domain.js to async functions

martinpitt:
This makes the functions easier to read, in particular the rather deeply nested domainGet() and domainCreate().

Keep the few functions which run several queries in parallel with `Promise.all()`, as that can't be done nicely with `await`.

Also use structural decomposition for the `call()` return values instead of always having to use their `[0]` element.

---

This is purely syntactic change, no logic changes.

PR #1399, opened 2024-01-24
Use python.spawn() for calling install_machine.py

martinpitt:
This avoids duplicating the logic for detecting the Python interpreter.

---

@jelly That was a follow-up from https://github.com/cockpit-project/cockpit-machines/pull/1397#discussion_r1463114858


jelly (approved):
Very nice!

PR #1409, opened 2024-01-29
Add "Pre-formatted Block Device" to createStoragePoolDialog

ohruka:
Partition-based storage pool (use Pre-formatted Block Device) created by
virt-manager and virsh could not be created by cockpit-machines before.
With this change, when cockpit-machines creates a storage pool, it will be possible
to select a partition-based pool type.
fs-create1
fs-create2

## Machines: Add support for Pre-formatted Block Device storage pool type

This pool type is similar to the existing "Physical disk device", but more robust: It avoids that the [guest sees and possibly reformats the raw disk device on its own](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/configuring_and_managing_virtualization/managing-storage-for-virtual-machines_configuring-and-managing-virtualization#proc_creating-disk-based-storage-pools-using-the-web-console_assembly_managing-virtual-machine-storage-pools-using-the-web-console).

fs-create1

_Thanks to Haruka Ohata for this contribution!_

### DIFF
--- a/src/components/storagePools/createStoragePoolDialog.jsx
+++ b/src/components/storagePools/createStoragePoolDialog.jsx
@@ -152,6 +152,10 @@ const StoragePoolSourceRow = ({ onValueChanged, dialogValues }) => {
     let placeholder;
     const diskPoolSourceFormatTypes = ['dos', 'dvh', 'gpt', 'mac'];
     const diskPoolSourceFsTypes = ['auto', 'ext2', 'ext3', 'ext4', 'xfs'];
+    /*
+      FsTypes lists the types listed below:
+      https://libvirt.org/storage.html#valid-filesystem-pool-format-types
+    */
 
     if (dialogValues.type == 'netfs') {
         validationState = dialogValues.source.dir.length == 0 && dialogValues.validationFailed.source ? 'error' : 'default';
--- a/src/components/storagePools/createStoragePoolDialog.jsx
+++ b/src/components/storagePools/createStoragePoolDialog.jsx
@@ -151,11 +151,8 @@ const StoragePoolSourceRow = ({ onValueChanged, dialogValues }) => {
     let validationState;
     let placeholder;
     const diskPoolSourceFormatTypes = ['dos', 'dvh', 'gpt', 'mac'];
+    // https://libvirt.org/storage.html#valid-filesystem-pool-format-types
     const diskPoolSourceFsTypes = ['auto', 'ext2', 'ext3', 'ext4', 'xfs'];
-    /*
-      FsTypes lists the types listed below:
-      https://libvirt.org/storage.html#valid-filesystem-pool-format-types
-    */
 
     if (dialogValues.type == 'netfs') {
         validationState = dialogValues.source.dir.length == 0 && dialogValues.validationFailed.source ? 'error' : 'default';
--- a/src/components/storagePools/createStoragePoolDialog.jsx
+++ b/src/components/storagePools/createStoragePoolDialog.jsx
@@ -60,6 +60,7 @@ const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion, pool
         { type: 'iscsi', detail: _("iSCSI target") },
         { type: 'disk', detail: _("Physical disk device") },
         { type: 'logical', detail: _("LVM volume group") },
+        { type: 'fs', detail: _("Pre-formatted Block Device") },
     ];
     // iscsi-direct exists since 4.7.0
     if (libvirtVersion && libvirtVersion >= 4007000)
@@ -68,7 +69,6 @@ const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion, pool
     const supportedPoolTypes = poolTypes.filter(pool => poolCapabilities[pool.type] ? poolCapabilities[pool.type].supported === "yes" : true);
 
     /* TODO
-        { type: 'fs', detail _("Pre-formatted Block Device") },
         { type: 'gluster', detail _("Gluster Filesystem") },
         { type: 'mpath', detail _("Multipath Device Enumerator") },
         { type: 'rbd', detail _("RADOS Block Device/Ceph") },
@@ -98,7 +98,7 @@ const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion, pool
 const StoragePoolTargetRow = ({ onValueChanged, dialogValues }) => {
     const validationState = dialogValues.target.length == 0 && dialogValues.validationFailed.target ? 'error' : 'default';
 
-    if (['dir', 'netfs', 'iscsi', 'disk'].includes(dialogValues.type)) {
+    if (['dir', 'netfs', 'iscsi', 'disk', 'fs'].includes(dialogValues.type)) {
         return (
             <FormGroup fieldId='storage-pool-dialog-target' label={_("Target path")}
                        id="storage-pool-dialog-target-group">
@@ -151,6 +151,7 @@ const StoragePoolSourceRow = ({ onValueChanged, dialogValues }) => {
     let validationState;
     let placeholder;
     const diskPoolSourceFormatTypes = ['dos', 'dvh', 'gpt', 'mac'];
+    const diskPoolSourceFsTypes = ['auto', 'ext2', 'ext3', 'ext4', 'xfs'];
 
     if (dialogValues.type == 'netfs') {
         validationState = dialogValues.source.dir.length == 0 && dialogValues.validationFailed.source ? 'error' : 'default';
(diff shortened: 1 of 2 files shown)


src/components/storagePools/createStoragePoolDialog.jsx
```diff
@@ -60,6 +60,7 @@ const StoragePoolTypeRow = ({ onValueChanged, dialogValues, libvirtVersion, pool
         { type: 'iscsi', detail: _("iSCSI target") },
         { type: 'disk', detail: _("Physical disk device") },
         { type: 'logical', detail: _("LVM volume group") },
+        { type: 'fs', detail: _("Pre-formatted Block Device") },
```
martinpitt, line 63: This is a rather curious name -- could this be "Directory"? Does libvirt actually care/check if this is a full block device, or could it just be any directory? But then again, we already have "Filesystem directory" -- how does this  differ from `dir`?

Is there some reference in the libvirt documentation how it describes these types?

src/components/storagePools/createStoragePoolDialog.jsx
```diff
@@ -211,6 +212,35 @@ const StoragePoolSourceRow = ({ onValueChanged, dialogValues }) => {
...
+                                      placeholder={placeholder}
+                                      onChange={value => onValueChanged('source', { device: value })} />
+                    <FormHelper fieldId="storage-pool-dialog-source" helperTextInvalid={validationState == "error" && _("Source path should not be empty")} />
+                </FormGroup>
+                <FormGroup fieldId='storage-pool-dialog-source-format' label={_("Format")}
+                           className="pf-m-4-col">
+                    <FormSelect id='storage-pool-dialog-source-format'
+                                value={dialogValues.source.format}
+                                onChange={(_event, value) => onValueChanged('source', { format: value })}>
+                        { diskPoolSourceFsTypes
```
martinpitt, line 232: This whole new block almost perfectly duplicates the existing `disk` case, except for this list. Please merge them check for "disk" || "fs", and do this here:
```
dialogValues.type === 'fs' ? diskPoolSourceFsTypes : diskPoolSourceFormatTypes)
```

martinpitt (changes requested):
Thanks for working on this feature! I have some questions, and the test needs a little fix.

src/components/storagePools/createStoragePoolDialog.jsx
martinpitt: Also, there is "disk" type already. Between "dir" and "disk" there isn't a lot of room, so I'm interested in when users would pick that over the other two.

src/components/storagePools/createStoragePoolDialog.jsx
ohruka: Thank you for your advice!
Merged two blocks.
After the change, it seems to work fine.

src/components/storagePools/createStoragePoolDialog.jsx
ohruka: > This is a rather curious name -- could this be "Directory"?

'fs' functions differently from 'dir'.

'dir' registers the storage pool under the specified directory.
'disk' registers the entire specified disk as a storage pool.
'fs' registers the specified formatted partition as a storage pool.

The libvirt documentation describes fs as follows:
- https://libvirt.org/storage.html#filesystem-pool

martinpitt: Right, I know the syntactics difference -- I mean, what's the use case for this? when is "fs" better than "dir" or "disk"?

martinpitt:
@ohruka BTW, if you are planning to do more work on c-machines, I'm happy to help you with setting up a [development environment](https://github.com/cockpit-project/cockpit/blob/main/HACKING.md#setting-up-development-container) and [running tests](https://github.com/cockpit-project/cockpit-machines/blob/main/HACKING.md), and letting you get this PR into perfect shape. You are also welcome to [chat with me/us](https://matrix.to/#/#cockpit:fedoraproject.org).

If this was just an one-off contribution, I'm equally happy to clean up the remaining issues myself.

Thanks!

src/components/storagePools/createStoragePoolDialog.jsx
```diff
@@ -151,6 +151,7 @@ const StoragePoolSourceRow = ({ onValueChanged, dialogValues }) => {
     let validationState;
     let placeholder;
     const diskPoolSourceFormatTypes = ['dos', 'dvh', 'gpt', 'mac'];
+    const diskPoolSourceFsTypes = ['auto', 'ext2', 'ext3', 'ext4', 'xfs'];
```
martinpitt, line 154: Could you please add a comment with a reference URL which defines this list? This is prone to getting out of date, so our future selves in a few years will appreciate :grin:

test/check-machines-storage-pools
```diff
@@ -488,6 +488,19 @@ class TestMachinesStoragePools(machineslib.VirtualMachinesCase):
...
+        m.execute(f"parted {dev} mklabel gpt")
+        m.execute(f"parted -s {dev} mkpart primary ext4 0% 100%")
+        part_dev = f"{dev}p1"
+        m.execute(f"mkfs.ext4 {part_dev}")
+        StoragePoolCreateDialog(
+            self,
+            name="my_fs_pool",
+            pool_type="fs",
+            target="/media/",
+            source={"source_path": dev, "format":"ext4"},
```
martinpitt, line 501: Shouldn't this be `part_dev`? The test goes through the trouble of creating a partition, but then doesn't actually use it, and instead gives the raw device here?

ohruka: Thank you, you're right, it's `part_dev` correctly.
Now that the test set works on hand, I'm fixing it along with other minor fixes.

src/components/storagePools/createStoragePoolDialog.jsx
ohruka: I added it according to the advice.

src/components/storagePools/createStoragePoolDialog.jsx
ohruka: If you pass 'disk' to a guest, an error may occur when the host detects the guest's disk.
By using 'fs', you can avoid problems with the host grabbing the disk for the guest to use.

The RHEL documentation says to assign guests at the partition level instead of the entire disk to avoid the above problems:
```
When whole disks or block devices are passed to the VM, the VM will likely partition it or create its own LVM groups on it. This can cause the host machine to detect these partitions or LVM groups and cause errors.
These errors can also occur when you manually create partitions or LVM groups and pass them to the VM.
To avoid theses errors, use file-based storage pools instead.
```
- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/configuring_and_managing_virtualization/managing-storage-for-virtual-machines_configuring-and-managing-virtualization#proc_creating-disk-based-storage-pools-using-the-web-console_assembly_managing-virtual-machine-storage-pools-using-the-web-console

martinpitt: I see, thanks! That almost sounds like the "disk" should be deprecated in favor of "fs".. Thanks for the pointers! (Note, I am very much not a libvirt expert, so this is learning for me)

src/components/storagePools/createStoragePoolDialog.jsx
martinpitt: Thanks! The comment should go above the line though, and the URL is enough.

martinpitt:
Thanks @ohruka ! This  looks very nice now, except for a tiny nitpick. I triggered the full integration tests.

martinpitt:
Ah, haha:
```
# test/check-machines-storage-pools:501:55: E231 missing whitespace after ':'
```

You can validate that locally with `make codecheck`

src/components/storagePools/createStoragePoolDialog.jsx
ohruka: Thank you, I changed the location.

martinpitt (approved):
Thanks! Waiting for the last test, then I'll merge.

I added a release note to the bottom of the description. If you have suggestions for changes/improvements, please do!

martinpitt:
Congrats @ohruka for your first landed contribution! :100: Thanks a lot!

ohruka:
Thank you for writing the release notes! I think the writing is good. No additional suggestions.
Thank you very much for your help for Marge!

martinpitt:
@ohruka I just learned that the string capitalization is wrong -- it should be "Pre-formatted block device", not "Block Device", to conform with PatternFly and Cockpit standards. Do you want to send a follow-up PR to fix that, or want me to?

ohruka:
Thanks for pointing that out.
I posted a PR to correct hastily displayed names to lowercase. (#1418 )
I would appreciate it if you could confirm.

PR #1414, opened 2024-01-31
Show "Uses SPICE" warning on the overview

martinpitt:
It turns out that the overview already reads all the VM's XML, so we can show the alert label from commit 5258c32d57f285 also on the overview.

image


src/components/vms/hostvmslist.jsx
```diff
@@ -59,7 +60,7 @@ const VmState = ({ vm, dismissError }) => {
             error={vm.error}
             state={state}
             valueId={`${vmId(vm.name)}-${vm.connectionName}-state`}
-            additionalState={<VmNeedsShutdown vm={vm} />} />
+            additionalState={<><VmNeedsShutdown vm={vm} /><VmUsesSpice vm={vm} /></>} />
```
martinpitt, line 63: Note: These two can *never* appear at the same time. The "uses spice" warning can only happen on non-running VMs, while "needs shutdown" only on running VMs.

PR #1398, opened 2024-01-23
Show alert label for SPICE VM if the host does not support it

martinpitt:
This makes the operation more discoverable in situations where VM
startup is already broken.

----

Related to https://issues.redhat.com/browse/RHEL-17434

This implements the [next step in the design](https://github.com/cockpit-project/cockpit-machines/pull/1387#issuecomment-1895348199):

image

This only appears on the details page, not on the overview, as it's rather expensive to determine this for all VMs at the same time on the overview -- I don't feel like penalizing all users all the time for this fringe case.

 - [x] builds on top of PR #1387

PR #1439, opened 2024-02-12
Don't hide long dialog titles

martinpitt:
A lot of our dialogs include the VM name, which can be long. Also, the title templates themselves may get longer in some translations.

Revert PF's `no-wrap` whitespace option (which partially hides the title in conjunction with `overflow: hidden`) to `normal`.

----

This was spotted by @garrett in https://github.com/cockpit-project/cockpit-machines/pull/1398#pullrequestreview-1841420419

Before:
cpu-edit
force-shutdown
spice-replace

Now:

image
image
image

Wrapping the dialog title is not exactly pretty, but much more useful than hiding it.


garrett:
Hey! This is great. I was pretty certain we had the "don't truncate long titles" override in our PatternFly overrides (at least at some point in time), but I don't see it there.

I think we probably should override this for all of Cockpit, not just Machines. When you're making a decision about what to do (which is the point of the modals), it's important to have critical information about what it is you're doing, such as what you're modifying... and that's often partially in the title.

martinpitt:
@garrett OK! I sent https://github.com/cockpit-project/cockpit/pull/20001 . If that lands by tomorrow, I can update cockpit lib in machines instead.

garrett:
@martinpitt I approved that other PR! Thanks for opening it up as a workaround over there.

martinpitt:
I'll rework this PR to drop the CSS workaround and update cockpit lib instead. The pixel change should stay.

garrett (approved):
Thanks!

PR #1442, opened 2024-02-12
Update RHEL 8 docs to RHEL 9

jelly:
https://issues.redhat.com/browse/RHEL-3954

 - [x] land this after the final release to RHEL 8

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -14,7 +14,7 @@
         "docs": [
             {
                 "label": "Managing virtual machines",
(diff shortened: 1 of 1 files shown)


martinpitt:
Thanks! But let's stall this until we do the final RHEL 8 upload. I don't want RHEL 9 doc links there.

martinpitt (approved):
I relelased 308 (aimed for RHEL 8), so landing this now for RHEL 9.

PR #1467, opened 2024-02-28
Fix disk "Used" value

### LINKED ISSUE #1423: Wrong USED disk space reported
As reported in a old issue for the `cockpit` project, USED disk space is wrongly reported. Original issue link: https://github.com/cockpit-project/cockpit/issues/9052

NOTE: I am opening an issue here because I think it is the more appropriate place, but feel free to close this issue if you want.

As explained in https://github.com/cockpit-project/cockpit/issues/9052#issuecomment-1918033769 the issue is about reading the wrong field from `libvirt` stats: `cockpit-machines` reads `allocation` rather than `physical`. From libvirt man page:

> block.\<num\>.allocation - offset of highest written sector in bytes
> block.\<num\>.physical - physical size of source file in bytes

So `block.<num>.physical` should be read to show USED disk space (as seen by the host).

From my understanding the fix is very simple: replace `!isNaN(vm.disksStats[target].allocation)` with `!isNaN(vm.disksStats[target].physical)` in [vmDisksCard.jsx](https://github.com/cockpit-project/cockpit-machines/blob/fe8f018c72e310a3c3c7d2752596fe78b9f488b8/src/components/vm/disks/vmDisksCard.jsx)

martinpitt:
Showing "allocation" in the UI doesn't make much sense. According to the manpage [1] it describes "offset of highest written sector in bytes". Showing the "physical size of source file in bytes" is much more useful.

Thanks to @shodanshok for figuring this out!

Fixes #1423

[1] https://www.libvirt.org/manpages/virsh.html

---

This was one of these "looks like a quick fix" issues, but it's not -- this is definitively not working yet, but I want to see if we get a more meaningful test failure than just the unexpected "error: input is not a number" -- i.e. where we cover this.

### DIFF
--- a/src/components/storagePools/storagePool.jsx
+++ b/src/components/storagePools/storagePool.jsx
@@ -48,11 +48,11 @@ export const getStoragePoolRow = ({ storagePool, vms, onAddErrorNotification }) 
             { storagePool.name }
         </span>
     );
-    const allocation = parseFloat(convertToUnit(storagePool.allocation, units.B, units.GiB).toFixed(2));
+    const physical = parseFloat(convertToUnit(storagePool.physical, units.B, units.GiB).toFixed(2));
     const capacity = parseFloat(convertToUnit(storagePool.capacity, units.B, units.GiB).toFixed(2));
-    const sizeLabel = String(cockpit.format("$0 / $1 GiB", allocation, capacity));
+    const sizeLabel = String(cockpit.format("$0 / $1 GiB", physical, capacity));
     const size = (
-        <Progress value={Number(storagePool.allocation)}
+        <Progress value={Number(storagePool.physical)}
                   min={0}
                   max={Number(storagePool.capacity)}
                   label={sizeLabel}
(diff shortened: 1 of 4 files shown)


martinpitt:
Yeah, [nothing](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1467-20240228-114834-9d20f71f-fedora-39/log.html#82) -- so we definitively need a proper test case for "Used" for both running and stopped machines first. That value isn't very predictable, but at least a plausible range should be tested that guards us against `undefined`/`NaN`.

mvollmer:
In our backog now: https://issues.redhat.com/browse/COCKPIT-1315. Closing here.

mvollmer:
This also changes Storage Pools. Those are not in scope of the bug report, and I don't know what would be wrong with them.

I'll start over based on the investigation [here](https://github.com/cockpit-project/cockpit-machines/issues/1423#issuecomment-3279073775).

mvollmer:
> This also changes Storage Pools. Those are not in scope of the bug report, and I don't know what would be wrong with them.

Ah, sorry, they are involved in the reporting of the statistics.

mvollmer:
> > This also changes Storage Pools. Those are not in scope of the bug report, and I don't know what would be wrong with them.
> 
> Ah, sorry, they are involved in the reporting of the statistics.

But they are correct. The "allocation" number in a volume XML seems to always correspond to the actual size.

PR #1468, opened 2024-02-28
test: drop allowing COMM failure message

jelly:
It's been a long time since Fedora 32 and I can no longer reproduce this test failure.


martinpitt (approved):
Looks good, thanks!

PR #1469, opened 2024-02-28
PF fixes discovered by codemods

jelly:
Discovered by running:

```
npx @patternfly/pf-codemods@latest --fix src
```


martinpitt (approved):
Thanks!

PR #1474, opened 2024-03-01
fmf: Drop explicit start of virtinterfaced.socket

martinpitt:
We don't do anything with that service.


martinpitt:
Hmm, quo vadis packit?

martinpitt:
I expect this to fail, see https://bugzilla.redhat.com/show_bug.cgi?id=2266014 . tmt is running now, let's see if it confirms.

martinpitt:
Yep!

PR #1475, opened 2024-03-01
TEST: Run CI without virtinterfaced

martinpitt:
Don't land! This is conceptually questionable, and will also fail on distros with the monolithic daemon, just want to answer https://bugzilla.redhat.com/show_bug.cgi?id=2266014#c2


martinpitt:
OK, we very much need it. Followed up to the bug.

PR #1485, opened 2024-03-08
Use Cockpit's KebabDropdown component

jelly:


martinpitt (approved):
Thanks a lot! I'm really happy that we shed the deprecated components, they'll haunt us at some point :cold_sweat:

PR #1494, opened 2024-03-13
test: don't wait_visible when we click

jelly:
Our click function already calls `wait_visible` so there is no need to wait twice.


jelly:
TestMachinesDisks.testDetachDisk seems to be a bit flaky now with these changes.

martinpitt (approved):
Thanks!

PR #1501, opened 2024-03-15
workflows: add cockpit-container-update workflow

jelly:
This workflow tries to update our cockpit CI container on a weekly interval.

### DIFF
--- a/.github/workflows/tasks-container-update.yml
+++ b/.github/workflows/tasks-container-update.yml
@@ -1,0 +1,34 @@
+name: tasks-container-update
+on:
+  schedule:
+    - cron: '0 2 * * 4'
+  # can be run manually on https://github.com/cockpit-project/cockpit-machines/actions
+  workflow_dispatch:
+jobs:
+  tasks-container-update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      statuses: write
+    container:
+      image: quay.io/cockpit/tasks
+      options: --user root
+    steps:
+      - name: Set up configuration and secrets
+        run: |
+          printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
+          mkdir -p ~/.config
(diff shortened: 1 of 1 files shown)


.github/workflows/cockpit-container-update.yml
```diff
@@ -0,0 +1,31 @@
+name: cockpit-container-update
+on:
+  schedule:
+    - cron: '0 2 * * 4'
```
martinpitt, line 4: I'd actually like to do this daily. But, does the  bot check if it already opened a PR? If not, then let's keep weekly for now.

martinpitt (approved):
Thanks! Question below, but not a blocker.

.github/workflows/cockpit-container-update.yml
```diff
@@ -0,0 +1,31 @@
...
+          printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
+          echo '${{ secrets.GITHUB_TOKEN }}' > ~/.config/github-token
+
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      - name: Run cockpit-container-update
+        run: |
+          make bots
+          bots/cockpit-container-update
```
martinpitt, line 31: It was renamed in https://github.com/cockpit-project/bots/pull/6095 , can  you please follow suit and also rename the workflow and job name?

martinpitt (changes requested):

.github/workflows/cockpit-container-update.yml
jelly: Did anyone test it on a fork?

jelly: And it doesn't work at all.... https://github.com/jelly/cockpit-machines/actions/runs/8329606250/job/22792320551

jelly: And now it does https://github.com/jelly/cockpit-machines/actions/runs/8330370703/job/22794851154

allisonkarlitskaya (approved):

PR #1502, opened 2024-03-15
test/browser: run tests in the tasks container

allisonkarlitskaya:


test/browser/browser.sh
```diff
@@ -1,41 +1,24 @@
 #!/bin/sh
-set -eux
-
-# like "basic", passed on to run-test.sh
-PLAN="$1"
```
martinpitt, line 5: This needs to be put back. We need to tell the run-tests.sh in the container about which test plan to run.

martinpitt (changes requested):

test/vm.install
```diff
@@ -10,8 +10,9 @@ fi
 systemctl enable cockpit.socket
 
 # don't force https:// (self-signed cert)
+mkdir -p /etc/cockpit
 printf "[WebService]\\nAllowUnencrypted=true\\n" > /etc/cockpit/cockpit.conf
 
 if type firewall-cmd >/dev/null 2>&1; then
-    firewall-cmd --add-service=cockpit --permanent
+    firewall-cmd --add-service=cockpit --permanent || true
```
martinpitt, line 17: Reminder, please do The Treatment™

martinpitt:
This looks quite good already, the storage plan even passed. `testCreatePXE` failed on

> ERROR    error creating macvtap interface macvtap0@veth0 (52:54:00:9f:04:7f): Device or resource busy

which doesn't ring a bell -- the tests don't hardcode "veth0", but they do hardcode the "macvtap" device type.  The tasks container probably already claimed "veth0"?

`testVmNICs` looks similar: It fails on

```
_wait_present("#vm-subVmTest1-add-iface-source option:nth-of-type(3):contains(virbr0)")
```

but the [screenshot](https://artifacts.dev.testing-farm.io/6aac2781-6298-408b-8286-59371548a499/work-network3kvsg8gv/plans/all/network/execute/data/guest/default-0/test/browser/network-1/data/TestMachinesNICs-testVmNICs-fedora-39-10.88.0.1-22-FAIL-1.png) doesn't show that -- again, it's plausible that the bridge is already claimed by the test container.

So we could try `--network host`, or run the container as non-root (and thus use slirp/pasta), or change the test to determine the name more dynamically.

test/vm.install
martinpitt: I force-pushed it to apply the change you did to certificates and friends.

martinpitt:
@allisonkarlitskaya I force-pushed for the firewalld config, and added a separate temp commit to try with `--network host`. I don't really want to land this (at least it's not my prefered choice), but I'd like to get a comparison test run. Happy to do some real investigation on Monday.

martinpitt:
Interesting, so with `--network host`, [f39 passes](https://artifacts.dev.testing-farm.io/af1838d0-7363-4561-9fb3-2d690a84940b/) while the other platforms fail. But this is much more clear: The test does some   TEST_OS decision whether external snapshots are supported. This smells  a lot like the test wouldn't get a correct $TEST_OS.

And indeed, the fmf test detects `$TEST_OS` in run-test.sh, which **already runs in the tasks container**. This needs to move to the outside. We need to fix that in the other projects, too.

martinpitt:
Also, eh? This PR has a "no-test" label, and yet our CI tests get run. This already happened in the previous push.

martinpitt:
That is much better, all green. So let's discuss/investigate on Monday.

allisonkarlitskaya:
> Also, eh? This PR has a "no-test" label, and yet our CI tests get run. This already happened in the previous push.

https://github.com/cockpit-project/cockpituous/pull/601?

allisonkarlitskaya:
> And indeed, the fmf test detects `$TEST_OS` in run-test.sh, which **already runs in the tasks container**. This needs to move to the outside. We need to fix that in the other projects, too.

Oops :)

allisonkarlitskaya:
How about `--env-file=/usr/lib/os-release` on the host?  Or too "pollute-y"?

ps: feel free to force-push.

martinpitt:
env-file feels a bit pollute-y, but not too much, and since you proposed it too, I agree.

martinpitt:
> > Also, eh? This PR has a "no-test" label, and yet our CI tests get run. This already happened in the previous push.
> 
> [cockpit-project/cockpituous#601](https://github.com/cockpit-project/cockpituous/pull/601)?

Don't think so -- this is from a "synchronize" event on a PR, not an "opened". 

So more likely a tests-scan regression, will have a look tomorrow.

martinpitt:
Mh, what? That `--env-file` change has mightily upset qemu/libvirt on [c8s](https://artifacts.dev.testing-farm.io/d2f5f109-9140-4861-8301-77940d07c244/) and [c9s](https://artifacts.dev.testing-farm.io/e00a2c74-2347-44a4-8ac3-1cea51e29908/). So I suppose back to the previous version.

test/browser/browser.sh
```diff
@@ -83,10 +51,16 @@ if [ "$ID" = fedora ]; then
...
-RC=$(cat $LOGS/exitcode)
-exit ${RC:-1}
+# Run tests in the cockpit tasks container, as unprivileged user
+CONTAINER="$(cat .cockpit-ci/container)"
+exec podman \
+    run \
+        --rm \
+        --shm-size=1024m \
+        --security-opt=label=disable \
+        --network=host \
```
allisonkarlitskaya, line 61: Do you want to make `--network=host` the "standard way" of doing this, or will we keep this as an adaption for cockpit-machines only?

In that case, I think a comment might be helpful here to explain why we deviate from starter-kit.

allisonkarlitskaya (commented):
I can't review, but if you +1 this, I'm going to squash and merge it.  I don't think we need to waste so much time on this.

test/browser/browser.sh
martinpitt: I like the net ns separation by default. I'll add a comment. I just wanted to get an agreement with you before I push again.

allisonkarlitskaya: Can you see my comment in the starter-kit PR as well about `/run/host` and do that?

martinpitt: Done.

martinpitt (approved):
I'm fine with this (in fact I'm really happy :grin: )

test/browser/browser.sh
martinpitt: Changed the bind mount path to /run/host

martinpitt (approved):

PR #1515, opened 2024-03-25
create: Allow pasting of multiple SSH keys at once

mvollmer:


src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -776,21 +780,31 @@ const parseKey = debounce(500, (key, setKeyObject, setKeyInvalid) => {
                 setKeyInvalid(true);
                 console.warn("Could not validate the public key");
             });
-});
+    };
+
+    const debouncedParseKey = useInit(() => debounce(500, parseKey));
```
jelly, line 785: I need to wrap my brain around this, this makes it so that `debounceParseKey` lives "forever"? And thus keeps debouncing?

jelly: Seems an alternative is to use `useMemo` or `useCallback`

https://dmitripavlutin.com/react-throttle-debounce/

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -806,6 +820,7 @@ const SshKeysRow = ({
                         testdata={keyInvalid ? "key-invalid" : undefined}
                         fieldId='public-key'>
                         <TextArea value={item.value || ""}
+                            autoFocus
```
jelly, line 823: Is this wanted? I think there are multiple form fields in this modal no?

test/check-machines-create
```diff
@@ -408,8 +408,8 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                                     user_password="catsaremybestfr13nds",
                                                                     user_login="foo",
                                                                     root_password="dogsaremybestfr13nds",
-                                                                    ssh_keys=[f"{rsakey} \n {dsakey}"],
-                                                                    expected_ssh_keys=[rsakey],
+                                                                    ssh_keys=[f"{rsakey} \r {dsakey}"],
```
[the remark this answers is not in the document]
mvollmer: Well, this string is used with `set_input_text` which turns characters into keypresses, and one needs to use `\r` to get it to simulate the enter key...

This is not at all the same as pasting in the whole text at once. I have found no way to simulate that, unfortunately.

src/components/create-vm-dialog/createVmDialog.jsx
mvollmer: But do we need an alternative? :-) `useInit` avoids both the performance problem of `useCallback` and the threat of `useMemo` to sometimes throw away the memoized value.

src/components/create-vm-dialog/createVmDialog.jsx
mvollmer: The `autoFocus` makes it so that newly added text areas get the focus, and typing will continue in them. The test relies on that, since it doesn't simulate a paste operation, but rather a fast typist.

mvollmer: But eslint doesn't want us to use it: "The autoFocus prop should not be used, as it can reduce usability and accessibility for users". Hmm.

test/check-machines-create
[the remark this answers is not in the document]
jelly: Ah right, so in podman we have an identical feature and we use some trickery there https://github.com/cockpit-project/cockpit-podman/blob/main/test/check-application#L1771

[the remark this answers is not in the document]
jelly: Either way it should test the same-ish thing.

src/components/create-vm-dialog/createVmDialog.jsx
jelly: No, was just curious :)

src/components/create-vm-dialog/createVmDialog.jsx
jelly: I recall we exclude this check somewhere else:


[jelle@t14s][~/projects/cockpit/cockpit-defer-lib-spawn]%git grep autoFocus
pkg/shell/superuser.jsx:                        autoFocus // eslint-disable-line jsx-a11y/no-autofocus

test/check-machines-create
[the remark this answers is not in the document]
mvollmer: > Ah right, so in podman we have an identical feature and we use some trickery there https://github.com/cockpit-project/cockpit-podman/blob/main/test/check-application#L1771

Ah, interesting! By accident, I figured out that appending a `\r` to the value used with `set_val` also triggers an event. But appending a space with set_input_text is much cleaner, I guess.

The other way I found is to use ReactTestUtils.Simulate.change after set_val. But I really like the set_input_text space trick!

jelly (approved):

PR #1528, opened 2024-04-03
Fix Fedora 40 fmf test

martinpitt:
See the [recent](https://artifacts.dev.testing-farm.io/4f4438ff-ae10-4fd2-84b8-556bd608f75a/) [failures](https://artifacts.dev.testing-farm.io/b6cf6bbe-c859-4149-a96f-0254c7505d4c/) in Fedora 40.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -12,7 +12,7 @@ $DNF libvirt-daemon-driver-storage-iscsi-direct || true
 
 # Show critical packages versions
 rpm -q selinux-policy cockpit-bridge cockpit-machines
-rpm -qa | grep -E 'libvirt|qemu' | sort
+rpm -qa | grep -E 'virt|qemu' | sort
 
 # allow test to set up things on the machine
 mkdir -p /root/.ssh
--- a/test/browser/main.fmf
+++ b/test/browser/main.fmf
@@ -4,6 +4,8 @@
     - cockpit-machines
     - cockpit-system
     - cockpit-ws
+    # it's a Recommends:, but sometimes dnf ignores it
+    - virt-install
     # required by tests
     - podman
     - dbus-tools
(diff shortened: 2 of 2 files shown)


martinpitt:
Oh, that actually helped -- I totally expected a failure on c8s, but apparently the package exists. But [now it fails differently](https://artifacts.dev.testing-farm.io/5d5ed7a0-288e-49cd-b293-aef0d0769044/):
```
error: Failed to open file '/etc/libvirt/qemu/networks/default.xml': No such file or directory
```

But wait, this makes no sense -- our spec already has
```
Requires: (libvirt-daemon-driver-interface if virt-install)
Requires: (libvirt-daemon-config-network if virt-install)
```
and indeed we don't test-depend on `virt-install`, we just `Recommend` it. That *should* be installed by default, though!?

martinpitt:
I tried to reproduce that in `podman run -it --rm fedora:40`:
```
dnf install --allowerasing https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-machines-1528/fedora-40-x86_64/07252647-cockpit-machines/cockpit-machines-310-1.20240403060431800072.pr1528.5.gffb7e54.fc40.noarch.rpm
```

This *does* install `virt-install`. Of course the TF run does a lot of other things, but they don't seem particularly relevant/influencing. Well, I'll let the current run finish, triple-check that `virt-install` is missing, and then add it as an explicit test dependency.

martinpitt:
[confirmed](https://artifacts.dev.testing-farm.io/4c1007ed-c379-4f59-9816-8823f692a83c/), no virt-install, and for some weird reason [dnf does not want to install the Recommends:](https://artifacts.dev.testing-farm.io/4c1007ed-c379-4f59-9816-8823f692a83c/guest-setup-4ccbb53f-89a2-49ae-8783-340822640284/artifact-installation-4ccbb53f-89a2-49ae-8783-340822640284/4-Install-packages.txt).

martinpitt:
Works! OK, let's clean up and push the real thing.

jelly (approved):

PR #1530, opened 2024-04-03
tmt: Pass host environment into test container

mvollmer:
So that things like TEST_AUDIT_NO_SELINUX can be set from the outside.


martinpitt (approved):
:100: thanks!

PR #1533, opened 2024-04-03
src: drop Vagrant instructions

jelly:
We don't have a Vagrantfile in our repository so these instructions are outdated and not useful.


martinpitt (approved):
Dankjewel!

PR #1536, opened 2024-04-04
workflows: Run cockpit-lib-update.yml in "self" environment

martinpitt:
The updated branch needs to be pushed with SSH instead of https and the GitHub token, so that the reposchutz workflow can run. This is the same approach as podman's and cockpit's workflows already do.

---

See #1535. I landed #1520 and previous updates with my Spethial Project Admin Hammer :hammer_and_pick: , but let's fix this properly.

### DIFF
--- a/.github/workflows/cockpit-lib-update.yml
+++ b/.github/workflows/cockpit-lib-update.yml
@@ -6,9 +6,9 @@ on:
   workflow_dispatch:
 jobs:
   cockpit-lib-update:
+    environment: self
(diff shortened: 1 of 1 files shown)


martinpitt:
I [ran the workflow against this branch](https://github.com/cockpit-project/cockpit-machines/actions/runs/8551517543/job/23430687187), it succeeded, produced #1537 and reposchutz ran correctly there.

jelly:
Thanks for fixing!

PR #1539, opened 2024-04-05
Go back to external "open" state management for KebabDropdown

martinpitt:
Commit 5ec3a27f1fb417dc3 introduced a race condition: kebabps rendered as part of a list are sometimes not persistent. This leads to immediately closing the kebab again when the user/test opens it if the table changes at the same time (most prominently in `TestMachinesDisks.testDetachDisk`).

This can only be fixed with managing the state outside of the list, as we did before 5ec3a27f1fb417dc3. All that state was still present in `DiskActions`, so use the new API from https://github.com/cockpit-project/cockpit/pull/20265 to just pass in the existing external state management.

For `VmNetworkTab`, bring back the `dropdownOpenActions` (which still existed as state) translation into a per-interface property from 5ec3a27f1fb417dc3.

`VmActions` and `NetworkActions` already tracked the state inside the tables, so either they were already broken before 5ec3a27f1fb417dc3 or are fine.

-----

This is our current [top machines flake](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?days=7&repo=cockpit-project%2Fcockpit-machines)

 - [x] Land https://github.com/cockpit-project/cockpit/pull/20265 and update the checkout SHA
 - [x] Audit other usages of Kebab which have the same problem

PR #1545, opened 2024-04-08
testStoragePoolsBasic flake

martinpitt:
Remaining top flake on the weather report, fails [like this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1539-1695d83f-20240405-095751-ubuntu-2204/log.html#82). Initial analysis: https://github.com/cockpit-project/cockpit-machines/pull/1539#issuecomment-2039647747 There and with commit 0a23d3f0d8ca8bd38880de6f420fdf03e24e1a8b we can ascertain that the click failure is *not* due to the kebab entry being disabled.

Next theory: the `isActionOpen` bounces between true/false/true? It seems to survive is_present and even is_visible, but not the immediately following ph_mouse(). An updated debug log confirms that:

```
> log: XXX VmDisksCard open actions [] for vda is open {"type":"boolean","value":false}
> log: XXX VmDisksCard open actions [] for vdc is open {"type":"boolean","value":false}
> log: XXX DiskActions for subVmTest1 vm state shut off disk vda disabled {"type":"boolean","value":false} is open {"type":"boolean","value":false}
> log: XXX DiskActions for subVmTest1 vm state shut off disk vdc disabled {"type":"boolean","value":false} is open {"type":"boolean","value":false}
<- {'type': 'undefined'}
-> wait: ph_is_present("#vm-subVmTest1-disks-vdc-action-kebab:not([disabled]):not([aria-disabled=true])")
<- {'type': 'undefined'}
-> wait: ph_is_visible("#vm-subVmTest1-disks-vdc-action-kebab:not([disabled]):not([aria-disabled=true])")
<- {'type': 'undefined'}
-> ph_mouse("#vm-subVmTest1-disks-vdc-action-kebab:not([disabled]):not([aria-disabled=true])","click",0,0,0,false,false,false,false)
> log: XXX VmDisksCard open actions ["vdc"] for vda is open {"type":"boolean","value":false}
> log: XXX VmDisksCard open actions ["vdc"] for vdc is open {"type":"boolean","value":true}
> log: XXX DiskActions for subVmTest1 vm state shut off disk vda disabled {"type":"boolean","value":false} is open {"type":"boolean","value":false}
> log: XXX DiskActions for subVmTest1 vm state shut off disk vdc disabled {"type":"boolean","value":false} is open {"type":"boolean","value":true}
<- {'type': 'undefined'}
-> wait: ph_is_present("#delete-vm-subVmTest1-disks-vdc:not([disabled]):not([aria-disabled=true])")
<- {'type': 'undefined'}
-> wait: ph_is_visible("#delete-vm-subVmTest1-disks-vdc:not([disabled]):not([aria-disabled=true])")
<- {'type': 'undefined'}
-> ph_mouse("#delete-vm-subVmTest1-disks-vdc:not([disabled]):not([aria-disabled=true])","click",0,0,0,false,false,false,false)
> log: XXX VmDisksCard open actions ["vdc"] for vda is open {"type":"boolean","value":false}
```
:arrow_up: this is wrong, and bouncing.
```
> log: XXX VmDisksCard open actions ["vdc"] for vdc is open {"type":"boolean","value":true}
> log: XXX DiskActions for subVmTest1 vm state shut off disk vda disabled {"type":"boolean","value":false} is open {"type":"boolean","value":false}
> log: XXX DiskActions for subVmTest1 vm state shut off disk vdc disabled {"type":"boolean","value":false} is open {"type":"boolean","value":true}
<- {'exceptionDetails': {'text': '#delete-vm-subVmTest1-disks-vdc:not([disabled]):not([aria-disabled=true]) not found'}}
```

The bouncing always happens, it's just that most of the time the `ph_mouse()` doesn't happen at a time where it matters. The bouncing also doesn't happen interactively on a steady state -- but there are several renderings when opening the details page. These actions happen right after one another:
```
        self.goToVmPage("subVmTest1")
        self.deleteDisk("vdc")
```
so apparently the renders from the goToVmPage() aren't done yet?

This is some emergent behaviour, and really hard to pinpoint the bug to any specific location :cry:

### DIFF
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -41,6 +41,14 @@ class VirtualMachinesCaseHelpers:
         with self.browser.wait_timeout(30):
             self.browser.wait_in_text("body", "Virtual machines")
 
+    # TODO: generic, move to testlib.py?
+    def dropdownAction(self, kebab_selector, menu_selector):
+        b = self.browser
+        b.click(kebab_selector)
+        b.wait_visible(f"{kebab_selector}[aria-expanded=true]")
+        b.click(menu_selector)
+        b.wait_not_present(f"{kebab_selector}[aria-expanded=true]")
+
     def performAction(self, vmName, action, checkExpectedState=True, connectionName="system", logPath=None):
         b = self.browser
         m = self.machine
@@ -48,9 +56,7 @@ class VirtualMachinesCaseHelpers:
         if logPath:
             m.write(logPath, '')
 
-        b.click(f"#vm-{vmName}-{connectionName}-action-kebab")
-        b.wait_visible(".pf-v5-c-menu")
-        b.click(f"#vm-{vmName}-{connectionName}-{action}")
+        self.dropdownAction(f"#vm-{vmName}-{connectionName}-action-kebab", f"#vm-{vmName}-{connectionName}-{action}")
         if action in ["reboot", "forceReboot"] and logPath:
             # https://bugzilla.redhat.com/show_bug.cgi?id=2221144
             # The VM should not be rebooted when the confirmation dialog is shown
@@ -246,8 +252,7 @@ class VirtualMachinesCaseHelpers:
         b = self.browser
 
         b.wait_visible(f"#vm-{vm_name}-disks-{target}-device")
-        b.click(f"#vm-{vm_name}-disks-{target}-action-kebab")
-        b.click(f"#delete-vm-{vm_name}-disks-{target}")
+        self.dropdownAction(f"#vm-{vm_name}-disks-{target}-action-kebab", f"#delete-vm-{vm_name}-disks-{target}")
         b.wait_visible(".pf-v5-c-modal-box")
         b.wait_in_text("#delete-resource-modal-target", target)
         if expect_path:
@@ -259,8 +264,7 @@ class VirtualMachinesCaseHelpers:
     def deleteIface(self, iface, mac=None, vm_name=None):
         b = self.browser
 
-        b.click(f"#vm-subVmTest1-iface-{iface}-action-kebab")
-        b.click(f"#delete-vm-subVmTest1-iface-{iface}")
+        self.dropdownAction(f"#vm-subVmTest1-iface-{iface}-action-kebab", f"#delete-vm-subVmTest1-iface-{iface}")
         b.wait_in_text(".pf-v5-c-modal-box .pf-v5-c-modal-box__header .pf-v5-c-modal-box__title", "Remove network interface?")
         if mac and vm_name:
             b.wait_in_text(".pf-v5-c-modal-box__body .pf-v5-c-description-list", f"{mac} will be removed from {vm_name}")
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -242,13 +242,16 @@ class VirtualMachinesCaseHelpers:
         else:
             return m.execute(cmd)
 
-    def deleteDisk(self, target, vm_name="subVmTest1"):
+    def deleteDisk(self, target, vm_name="subVmTest1", expect_path=None):
         b = self.browser
 
         b.wait_visible(f"#vm-{vm_name}-disks-{target}-device")
         b.click(f"#vm-{vm_name}-disks-{target}-action-kebab")
         b.click(f"#delete-vm-{vm_name}-disks-{target}")
         b.wait_visible(".pf-v5-c-modal-box")
+        b.wait_in_text("#delete-resource-modal-target", target)
+        if expect_path:
+            b.wait_in_text("#delete-resource-modal-file", expect_path)
         b.click("#delete-resource-modal-primary")
         b.wait_not_present(".pf-v5-c-modal-box")
         b.wait_not_present(f"#vm-{vm_name}-disks-{target}-device")
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -46,6 +46,9 @@ class VirtualMachinesCaseHelpers:
         b = self.browser
         b.click(kebab_selector)
         b.wait_visible(f"{kebab_selector}[aria-expanded=true]")
+        # HACK: PF bug: Dropdown open state is bouncing
+        time.sleep(0.2)
+        b.wait_visible(f"{kebab_selector}[aria-expanded=true]")
         b.click(menu_selector)
         b.wait_not_present(f"{kebab_selector}[aria-expanded=true]")
 
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -1498,15 +1498,7 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
         self.goToVmPage("subVmTest1")
 
-        b.wait_visible("#vm-subVmTest1-disks-vdc-device")
-        b.click("#vm-subVmTest1-disks-vdc-action-kebab")
-        b.click("#delete-vm-subVmTest1-disks-vdc")
-        b.wait_visible(".pf-v5-c-modal-box")
-        b.wait_in_text(".pf-v5-c-modal-box__body .pf-v5-c-description-list", "subVmTest1")
-        b.wait_in_text("#delete-resource-modal-target", "vdc")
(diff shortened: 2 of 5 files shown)


martinpitt:
All green! Doing another round just to be sure.

@mvollmer @jelly can I please have your opinion here? I hope that there's some React approach that I missed. Do you have another idea?

mvollmer:
> so apparently the renders from the goToVmPage() aren't done yet?

If this is true, then I would put any hacks to avoid this into goToVmPage, and not into deleteDisks, I think.

mvollmer:
There some React keys missing in the `colums` array in `VmDisksCard`. I don't think adding them would necessarily help, but who knows.

mvollmer:
> There some React keys missing in the `columns` array in `VmDisksCard`. I don't think adding them would necessarily help, but who knows.

Hmm, I guess the missing ones are added automatically by `ListingTable`.

martinpitt:
> I would put any hacks to avoid this into goToVmPage, and not into deleteDisks, I think.

What would you add there, other than a static `sleep(5)`? The root problem is that there is no obvservable property (or even definition of) "done" for the list. libvirt events (via dbus) happen all the time, and there's constant re-rendering jitter.

mvollmer:
>      log: XXX VmDisksCard open actions ["vdc"] for vdc is open {"type":"boolean","value":true}
>      log: XXX DiskActions for subVmTest1 vm state shut off disk vda disabled {"type":"boolean","value":false} is open {"type":"boolean","value":false}
>      log: XXX DiskActions for subVmTest1 vm state shut off disk vdc disabled {"type":"boolean","value":false} is open {"type":"boolean","value":true}

What is bouncing here? vdc stays open, and vda is closed. Or am I misreading things?

martinpitt:
@mvollmer look at the log in the description, I explicitly marked the bounce for vdc.

martinpitt:
Ah, seems I misread things, sorry.

martinpitt:
@mvollmer right, seems I misread this. Then this is even more subtle, and I have no idea at all what happens..

mvollmer:
In this screenshot https://cockpit-logs.us-east-1.linodeobjects.com/pull-1539-1695d83f-20240405-095751-ubuntu-2204/TestMachinesStoragePools-testStoragePoolsBasic-ubuntu-2204-127.0.0.2-2301-FAIL.png the dropdown is open but in a funny place. Is PF maybe moving it around?

Do we know what the exact error is from ph_mouse? I can't see it anywhere...

test/machineslib.py
```diff
@@ -247,6 +247,10 @@ def deleteDisk(self, target, vm_name="subVmTest1", expect_path=None):
 
         b.wait_visible(f"#vm-{vm_name}-disks-{target}-device")
         b.click(f"#vm-{vm_name}-disks-{target}-action-kebab")
+        b.wait_visible(f"#vm-{vm_name}-disks-{target}-action-kebab[aria-expanded=true]")
+        # HACK: kebab open state rendered in lists is bouncing
+        time.sleep(0.2)
+        b.wait_visible(f"#vm-{vm_name}-disks-{target}-action-kebab[aria-expanded=true]")
```
mvollmer, line 253: So the theory is that the dropdown is closed and then reopened spontaneously, right?  I am not sure about opening/closing, but the `Dropdown` rabbit hole is very deep, with tens if not hundreds of `useEffect` and timers and at least two `useIsometricLayoutEffect`... It boggles the mind how complex it seems to be to get a dropdown menu on the screen. So I can believe that _something_ is going on behind the scenes that disturbs our test code somehow.

So this delay seems reasonable to me, to let the menu settle down after opening it. But we shouldn't give the impression that we know exactly what's going on, and maybe just say "HACK - PatternFly DropDown components seem to need some time to settle down after opening, sometimes, maybe."

mvollmer: [This failure screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1539-1695d83f-20240405-095751-ubuntu-2204/TestMachinesStoragePools-testStoragePoolsBasic-ubuntu-2204-127.0.0.2-2301-FAIL.png) shows a open dropdown, but in a funny place. Something is going on there. Maybe the Kebab gets moved around, and maybe the Dropdown has to catch up? We could also try to insert a delay after `goToVmPage` to let the page itself settle down before opening the dropdown.

martinpitt: > So the theory is that the dropdown is closed and then reopened spontaneously, right?

That's the observable and provable behaviour, yes. I originally thought it was the bouncing of the open-ness state due to delayed functional event handlers, but as you spotted I mis-read the output and this is actually fine. I also really believe that it is *not* related to missing `key=`s. First of all, React would complain about it, and second of all that's precisely why we keep a *global* open-ness state for all table entries, instead of tracking the state individually inside of the kebab.

So indeed I don't know at all what happens now. If the PF Dropdown is really so massively complicated (why oh why?), then maybe we should not use it at all, and implement our own simpler thing in cockpit-components-dropdown? Or go back to the deprecated one (but that's not a long-term option obviously). *throws hands into the air*.

With this kind of "we don't know what's going on", we should then introduce a testlib.py helper function for selecting something from a dropdown, and we can centrally add the debouncing there. I'm sure this affects other places, just perhaps less obviously due to sheer timing luck. WDYT?

mvollmer:
> > I would put any hacks to avoid this into goToVmPage, and not into deleteDisks, I think.
> 
> What would you add there, other than a static `sleep(5)`?

Yes, just a delay.

The page might get rerendered constantly, but my theory is that there some initial layout changes that mess with a dropdown that is already open.

I don't think we need to figure out the details, but if inserting a delay after openeing the dropdown helps, doesn't that mean we would have to insert the same kind of delay into every dropdown interaction from tests?  INserting the dealy into vmGoToPage seems nicer.(if it helps!)

test/machineslib.py
martinpitt: For the lolz, PF Dropdown also messes up in a completely different way: https://github.com/cockpit-project/starter-kit/pull/835

martinpitt:
@mvollmer With your observation about how complex the PF Dropdown is, and my refuted theory about delayed (but correct) state updates I now to doubt whether just waiting *before* opening the dropdown is even adequate. It may work for a little bit out of sheer luck, but all the data that we have shows that the menu closes and opens *after* opening it, and there is no observable change of the table any more -- i.e. it is already fully rendered when the click happens. I.e. I think it would be more robust to do that debouncing at *every* kebab opening, IOW we need a testlib helper for that.

mvollmer:
> @mvollmer With your observation about how complex the PF Dropdown is, and my refuted theory about delayed (but correct) state updates I now to doubt whether just waiting _before_ opening the dropdown is even adequate. It may work for a little bit out of sheer luck, but all the data that we have shows that the menu closes and opens _after_ opening it, and there is no observable change of the table any more -- i.e. it is already fully rendered when the click happens. I.e. I think it would be more robust to do that debouncing at _every_ kebab opening, IOW we need a testlib helper for that.

Right, hmm. But do we have evidence that other kebabs also suffer from this issue? Isn't it maybe just "open the kebab right after switching pages"?

But, in general, yeah, whatever works is good enough here. Let's just make it clear that we don't fully understand what's actually going on (so as to not mislead our future selves),

martinpitt:
@mvollmer: There was #1539 (an even worse flake), which in retrospect may have been due to the same root cause (although it looked completely different upon examination). So.. I don't know. My idea was to add a "select something from kebab dropdown" helper to machineslib, use it in the ~ 20 or so places which repeat the kebab selectors (factorizing this is a good idea anyway), add the debouncing to that, and if it works, move it to testlib.py.

test/machineslib.py
mvollmer: > For the lolz, PF Dropdown also messes up in a completely different way: [cockpit-project/starter-kit#835](https://github.com/cockpit-project/starter-kit/pull/835)

Autsch...

martinpitt:
I did the factorization. Let's see how it works, I may have screwed up something.

mvollmer (dismissed):
Nice, thanks!

martinpitt:
@mvollmer Thanks for investigating and cross-checking! I dropped the amplification commit.

mvollmer (approved):
Thanks!

PR #1543, opened 2024-04-08
Fix crash on checking for hw virtualization

martinpitt:
`virt-host-validate` does not actually include "Checking for hardware virtualization" on ARM, so the page crashes due to the grep exiting with non-zero.

While at it, restrict the checks to QEMU. We aren't interested in LXC or other hypervisors.

Thanks to @yunmingyang for debugging this!

Fixes #1523

PR #1546, opened 2024-04-08
Fix domainDetachHostDevice() for running VMs

yunmingyang:
https://issues.redhat.com/browse/RHEL-31082

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -515,7 +515,7 @@ export async function domainDetachHostDevice({ connectionName, vmId, live, dev }
         await call(connectionName, vmId, 'org.libvirt.Domain', 'DetachDevice', [hostdevInactiveXML, Enum.VIR_DOMAIN_AFFECT_CONFIG], { timeout, type: 'su' });
 
     if (live) {
-        const [domXml] = call(connectionName, vmId, 'org.libvirt.Domain', 'GetXMLDesc', [0], { timeout, type: 'u' });
+        const [domXml] = await call(connectionName, vmId, 'org.libvirt.Domain', 'GetXMLDesc', [0], { timeout, type: 'u' });
         const hostdevXML = getHostDevElemBySource(domXml, source);
 
         await call(connectionName, vmId, 'org.libvirt.Domain', 'DetachDevice', [hostdevXML, Enum.VIR_DOMAIN_AFFECT_LIVE], { timeout, type: 'su' });
--- a/test/check-machines-hostdevs
+++ b/test/check-machines-hostdevs
@@ -25,10 +25,6 @@ import machineslib
 import testlib
 from machinesxmls import PCI_HOSTDEV, SCSI_HOST_HOSTDEV, USB_HOSTDEV, USB_HOSTDEV_NONEXISTENT
 
-virt_xml_mock = """#!/usr/bin/python
-
-raise Exception("Mock error message")"""
-
 
 class HostDevAddDialog(object):
     def __init__(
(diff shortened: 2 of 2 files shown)


src/libvirtApi/domain.js
```diff
@@ -515,7 +515,7 @@ export async function domainDetachHostDevice({ connectionName, vmId, live, dev }
         await call(connectionName, vmId, 'org.libvirt.Domain', 'DetachDevice', [hostdevInactiveXML, Enum.VIR_DOMAIN_AFFECT_CONFIG], { timeout, type: 'su' });
 
     if (live) {
-        const [domXml] = call(connectionName, vmId, 'org.libvirt.Domain', 'GetXMLDesc', [0], { timeout, type: 'u' });
+        const [domXml] = await call(connectionName, vmId, 'org.libvirt.Domain', 'GetXMLDesc', [0], { timeout, type: 'u' });
```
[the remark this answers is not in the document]
jelly: Can we add a test for this somehow?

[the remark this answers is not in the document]
yunmingyang: @martinpitt Ah yeah, I will do that, thanks a lot.

test/check-machines-hostdevs
```diff
@@ -229,6 +229,15 @@ class TestMachinesHostDevs(machineslib.VirtualMachinesCase):
                 b.wait_not_present("#delete-resource-modal")
                 b.wait_not_present(f"#vm-subVmTest1-hostdev-{self.vm_dev_id}-product")
 
+        # Add USB devices when the VM is running
+        b.wait_in_text(f"#vm-subVmTest1-{connectionName}-state", "Running")
+        output = self.run_admin(f"virsh -c qemu:///{connectionName} nodedev-list --cap usb_device", connectionName)
```
martinpitt, line 234: Thanks! This new test indeed reproduces the error on the code from main.

However, I wonder if we really need the `if` here.. Don't all our test VMs have some USB device? That's a function of the hypervisor, i.e. we need to test this on both our CI and Testing Farm. It's brittle in the sense that it could silently fail here and then the test would never run.

So I'd like to test this without the condition. I'll push an extra test commit to see what happens. Somehow the previous test run got completely stuck anyway.

The fix here [breaks testHostDevAddSessionConnection](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1546-c966aabd-20240410-085437-fedora-39-firefox/log.html#85) -- I think the test is wrong, and the bug fixed here previously just hid it. I'll take a look.

martinpitt: Ah, I see this check already existed. I'll clean it up a bit to avoid duplicating it.

yunmingyang: I met a scenario that some VMs from ppc64le and s390x will not have USB or PCI devices at all, so that I add the if there. Also, I think assertNotEqual is also a good way, thanks.
Also, I think some virt-host-validate related fixes was merged in main by https://github.com/cockpit-project/cockpit-machines/pull/1543. Are there some duplications?
And about failures of testHostDevAddSessionConnection, I saw the screenshot, and it said permission denied. Is there a reason that the non-root user should not be allowed to attach host devices? Not sure about this.

martinpitt: Alright, I tried to make it unconditional, and it indeed [does fail on the TF](https://artifacts.dev.testing-farm.io/fbda1632-72af-4006-8ae7-baba5b9af107/). Alright, I've tried :grin: I'll revert the second patch.

> Is there a reason that the non-root user should not be allowed to attach host devices?

It would be dangerous if it *did* work -- if you can't access them as user on the host, it surely can't work if you forrward it to a VM of yours. So that test case is a bit too optimistic. As user we really can only check for the "not allowed" case. I'll rearrange this a bit.

martinpitt: I cleaned up the tests, they were quite messy. Both work now.

martinpitt:
This works now, and I'm happy with it. @jelly as I dabbled with this quite a lot, can you please do a final review?

martinpitt:
... or @mvollmer , as you expressed some interest in c-machines.

PR #1548, opened 2024-04-09
workflows: Run tasks-container-update in "self" environment

martinpitt:
Just like cockpit-lib-update, this workflow produces PRs which need to run reposchutz, so they need to be pushed via SSH, not the GitHub token.

----

I landed yesterday's #1541 with my admin hammer, but let's fix this properly. Exactly the same as https://github.com/cockpit-project/cockpit-podman/pull/1660

### DIFF
--- a/.github/workflows/tasks-container-update.yml
+++ b/.github/workflows/tasks-container-update.yml
@@ -6,6 +6,7 @@ on:
   workflow_dispatch:
 jobs:
   tasks-container-update:
+    environment: self
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


martinpitt:
I [ran the workflow against this branch](https://github.com/cockpit-project/cockpit-machines/actions/runs/8611777724), it succeeded, produced #1549 and [reposchutz succeeded there](https://github.com/cockpit-project/cockpit-machines/actions/runs/8611789182/job/23599643490?pr=1549).

jelly (approved):

PR #1553, opened 2024-04-11
packit: Land a new upstream release automatically to centos-stream c9s

KKoukiou:
This packit job only makes sure the changes happen in dist-git - no builds.

Relevant documentation: https://packit.dev/docs/configuration/upstream/propose_downstream

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -7,6 +7,13 @@ downstream_package_name: cockpit-machines
 # use the nicely formatted release description from our upstream release, instead of git shortlog
 copy_upstream_release_description: true
 
+packages:
+  cockpit-machines-fedora:
+    specfile_path: cockpit-machines.spec
+  cockpit-machines-centos:
+    specfile_path: cockpit-machines.spec
+    pkg_tool: centpkg
+
 srpm_build_deps:
 - npm
 - make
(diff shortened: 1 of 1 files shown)


packit.yaml
```diff
@@ -1,12 +1,24 @@
 upstream_project_url: https://github.com/cockpit-project/cockpit-machines
 # enable notification of failed downstream jobs as issues
 issue_repository: https://github.com/cockpit-project/cockpit-machines
-specfile_path: cockpit-machines.spec
```
KKoukiou, line 4: the specfile_path was duplicated because apparently there is a bug in packit, when multiple `packages` are defined, they don't inherit the common fields. 

https://github.com/packit/packit/issues/2272

martinpitt (approved):
Very cool! :clap:

martinpitt (changes requested):
Wait, the [copr build failed](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-machines-1553/srpm-builds/07296387/builder-live.log.gz):
```
ERROR  [Errno 2] No such file or directory: 'wget'
```

Please drop that -- we *really* want to treat the upstream .spec file as the single source of truth, and the Fedora/RHEL dist-gits should sync from it. So in theory you should just be able to drop the two post-upstream-clone commands.

martinpitt (approved):

PR #1557, opened 2024-04-16
packit: two fixes

martinpitt:
- [x] Rebase after #1559

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -55,6 +55,15 @@ fi
 # TODO: Run in "host" network ns, as some tests fail on unexpected veth/bridge claimed by the container
 # fix these and then use the isolation in starter-kit and friends
 CONTAINER="$(cat .cockpit-ci/container)"
+
+# this often fails in TF
+for i in $(seq 5); do
+    if podman pull "$CONTAINER"; then
+        break
+    fi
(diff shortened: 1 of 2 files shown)


martinpitt:
Argh, since commit db7e195812a7d125e6dd5f7072c83deb344cb4cb it now duplicates all *tests* to -centos and -fedora -- we only want this for propose_downstream. Let's fix that, see PR #1559

martinpitt:
So enabling c10 is blocked by `virsh net-start default` [falling over](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1557-586f4efc-20240416-103316-centos-10/log.html). I'll still wait for the [TF run](https://artifacts.dev.testing-farm.io/0af9b4e0-7aa3-4dbb-aee8-0d36c620c65b/), perhaps that VM is magically configured to have different/better results.

Reported to https://issues.redhat.com/browse/RHEL-32906 . So in the meantime I'll just land the two other fmf test adjustments.

martinpitt:
OK, [same thing on the TF](https://artifacts.dev.testing-farm.io/0af9b4e0-7aa3-4dbb-aee8-0d36c620c65b/). So removing the enablement commit.

jelly (approved):

PR #1558, opened 2024-04-16
Validate TEST_AUDIT_NO_SELINUX

martinpitt:
Debugging why the bodhi test failure https://artifacts.dev.testing-farm.io/8ce2302a-ee98-4b69-bc0f-4974597020fc/ happened. #1530 happened before the release.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -14,6 +14,9 @@ $DNF libvirt-daemon-driver-storage-iscsi-direct || true
 rpm -q selinux-policy cockpit-bridge cockpit-machines
 rpm -qa | grep -E 'virt|qemu' | sort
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Hmm, the [test log](https://artifacts.dev.testing-farm.io/90d2ac48-5fd4-4fa9-985e-5b25a1271c4a/) shows that `TEST_AUDIT_NO_SELINUX=1` makes it all the way through to run-test.sh. I think the bug is that the flag isn't set in https://src.fedoraproject.org/rpms/cockpit-machines/blob/rawhide/f/plans/all.fmf -- would be nice to sync this file with packit, but it's slightly different (`dist-git-source`). I'll commit these manually.

PR #1559, opened 2024-04-16
packit: Split copr_build and tests for fedora/centos

martinpitt:
Commit db7e195812a7d125e6d introduced the two -fedora and -centos package variants. Since then, *all* copr_build and test jobs got duplicated, which is a waste. Split them up properly.

---

See e.g. #1558 or any other recent PR. They get duplicated builds and tests:

image
image

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -25,19 +25,31 @@ actions:
 jobs:
   - job: copr_build
     trigger: pull_request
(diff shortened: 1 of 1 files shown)


KKoukiou (approved):
I guess you just need it for the specfile_path configuration? 

Update: oh now I saw in the commit message, the jobs got duplicated.

PR #1562, opened 2024-04-17
config: Load configuration values from manifest

mvollmer:
And use that mechanism to control storage during migration.


src/config.js
```diff
@@ -35,6 +37,34 @@ const VMS_CONFIG = {
...
         }
     },
+
+    StorageMigrationSupported: true,
 };
 
+function try_fields(dict, fields, def) {
+    for (let i = 0; i < fields.length; i++) {
+        if (fields[i] && dict[fields[i]] !== undefined)
+            return dict[fields[i]];
```
cockpituous, line 47: This added line is not executed by any test.

src/config.js
```diff
@@ -35,6 +37,34 @@ const VMS_CONFIG = {
...
+    const config = cockpit.manifests.machines?.config;
+    const os_release = await read_os_release();
+
+    function get_config(name) {
+        if (config) {
+            let val = config[name];
+            if (typeof val === 'object' && val !== null)
+                val = try_fields(val, [os_release.PLATFORM_ID, os_release.ID]);
+            if (val !== undefined)
+                VMS_CONFIG[name] = val;
```
cockpituous, line 62: This added line is not executed by any test.

martinpitt (approved):
Thanks! The best part is replacing that custom parser with `read_os_release()`, but I suppose it makes sense to make this configurable -- if someone in RHEL really wants to, they have a way to enable this.

mvollmer:
> if someone in RHEL really wants to, they have a way to enable this.

Yes, and there might be more RHEL-specific behavior to come with snapshots...

PR #1564, opened 2024-04-18
snapshots: Skip empty media drives for external snapshots

mvollmer:
Instead of falling back to internal ones. This makes external snapshots work in more cases.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -899,8 +899,8 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
         return false;
     }
 
-    // Currently external snapshots work only for disks of type "file" with a source.
-    if (!disks.every(disk => (disk.type === "file" && disk.source?.file))) {
+    // Currently external snapshots work only for disks of type "file"
+    if (!disks.every(disk => disk.type === "file")) {
         logDebug(`vmSupportsExternalSnapshots: vm ${vm.name} has unsupported disk type`);
         return false;
     }
--- a/src/libvirt-xml-create.js
+++ b/src/libvirt-xml-create.js
@@ -263,15 +263,24 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
 
         const disksElem = doc.createElement('disks');
         disks.forEach(disk => {
-            // Disk can have attribute "snapshot" set to "no", which means no snapshot should be created of the said disk
-            // This cannot be configured through cockpit, but we should uphold it nevertheless
-            // see "snapshot" attribute of <disk> element at https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms
-            if (disk.snapshot !== "no") {
-                const diskElem = doc.createElement('disk');
-                diskElem.setAttribute('name', disk.target);
-                diskElem.setAttribute('snapshot', 'external');
-                disksElem.appendChild(diskElem);
-            }
+            // Disk can have attribute "snapshot" set to "no", which
+            // means no snapshot should be created of the said disk
+            // This cannot be configured through cockpit, but we
+            // should uphold it nevertheless.
+            //
+            // See "snapshot" attribute of <disk> element at
+            // https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms
+            if (disk.snapshot == "no")
+                return;
+
+            // Skip disks without source, such as empty media drives.
+            if (isExternal && disk.type == "file" && !disk.source.file)
+                return;
+
+            const diskElem = doc.createElement('disk');
+            diskElem.setAttribute('name', disk.target);
+            diskElem.setAttribute('snapshot', 'external');
+            disksElem.appendChild(diskElem);
         });
         snapElem.appendChild(disksElem);
     }
--- a/test/check-machines-snapshots
+++ b/test/check-machines-snapshots
@@ -146,6 +146,7 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
                 if self.xfail is None:
                     self.verify_frontend()
                     self.verify_backend()
+                    self.revert()
                     if self.remove:
                         self.cleanup()
 
@@ -237,6 +238,11 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
                     print(m.execute(snap_xml))
                     print("------ end snapshot XML -------")
                     raise
+
+            def revert(self):
+                b.click(f"#vm-subVmTest1-snapshot-{self.snap_num}-revert")
+                b.click('.pf-v5-c-modal-box__footer button:contains("Revert")')
+                b.wait_not_present('.pf-v5-c-modal-box__footer button:contains("Revert")')
 
             def cleanup(self):
                 b.click(f"#delete-vm-subVmTest1-snapshot-{self.snap_num}")
--- a/test/check-machines-snapshots
+++ b/test/check-machines-snapshots
@@ -241,8 +241,9 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
 
             def revert(self):
                 b.click(f"#vm-subVmTest1-snapshot-{self.snap_num}-revert")
+                b.wait_visible(".pf-v5-c-modal-box")
                 b.click('.pf-v5-c-modal-box__footer button:contains("Revert")')
-                b.wait_not_present('.pf-v5-c-modal-box__footer button:contains("Revert")')
+                b.wait_not_present(".pf-v5-c-modal-box")
 
             def cleanup(self):
                 b.click(f"#delete-vm-subVmTest1-snapshot-{self.snap_num}")
@@ -355,7 +356,6 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
         ).execute()
 
         # With CD-ROM drive without media
-        # doesn't support external snapshots yet
         b.click("#vm-subVmTest1-disks-sdd-eject-button")
         b.click(".pf-v5-c-modal-box__footer button:contains(Eject)")
         b.wait_not_present(".pf-v5-c-modal-box")
@@ -364,7 +364,7 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
             self,
             name="nomedia",
             state="shutoff",
-            expect_external=False,
+            expect_external=supports_external,
         ).execute()
 
         # delete CD-DROM again so that the next test can use external snapshot again


mvollmer:
I think I'll expand the test to also check reverting each created snapshot.

src/helpers.js
```diff
@@ -904,7 +904,7 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
     // or media-less disks
     const supportedPools = storagePools.filter(pool => pool.type === "dir" && pool.connectionName === vm.connectionName);
     if (!disks.every(disk =>
-        (disk.type === "file" && disk.source?.file) ||
+        (disk.type === "file") ||
```
cockpituous, line 907: This added line is not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,29 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-
-                if (disk.type === "volume") {
-                    const poolPath = getStoragePoolPath(storagePools, disk.source.pool, connectionName);
-                    if (poolPath) {
-                        const sourceElem = doc.createElement('source');
-                        sourceElem.setAttribute('file', `${poolPath}/${disk.source.volume}.snap`);
-                        diskElem.appendChild(sourceElem);
-                    }
+            if (disk.snapshot == "no")
+                return;
```
cockpituous, line 253: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,29 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            // Disks of type "file" but without a source (such as
+            // empty media drives) are not automatically skipped for
+            // external snapshots. Let's skip them here.
+            if (isExternal && disk.type == "file" && !disk.source.file)
+                return;
```
cockpituous, line 259: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,29 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            const diskElem = doc.createElement('disk');
+            diskElem.setAttribute('name', disk.target);
+            diskElem.setAttribute('snapshot', 'external');
```
cockpituous, line 263: These 3 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,29 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            if (disk.type === "volume") {
+                const poolPath = getStoragePoolPath(storagePools, disk.source.pool, connectionName);
+                if (poolPath) {
+                    const sourceElem = doc.createElement('source');
+                    sourceElem.setAttribute('file', `${poolPath}/${disk.source.volume}.snap`);
+                    diskElem.appendChild(sourceElem);
```
cockpituous, line 270: These 6 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,29 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
                 }
-
-                disksElem.appendChild(diskElem);
             }
+
+            disksElem.appendChild(diskElem);
```
cockpituous, line 274: This added line is not executed by any test.

martinpitt:
Oh, this is nice! Makes much more sense.

src/helpers.js
cockpituous, line 907: This added line is not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 253: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 259: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 263: These 3 added lines are not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 270: These 6 added lines are not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 274: This added line is not executed by any test.

src/helpers.js
```diff
@@ -904,7 +904,7 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
     // or media-less disks
```
mvollmer, line 904: This comment should be removed as well.

mvollmer:
Hmm, we might need to extend this to read-only devices in general.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,29 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-                    const poolPath = getStoragePoolPath(storagePools, disk.source.pool, connectionName);
-                    if (poolPath) {
-                        const sourceElem = doc.createElement('source');
-                        sourceElem.setAttribute('file', `${poolPath}/${disk.source.volume}.snap`);
-                        diskElem.appendChild(sourceElem);
-                    }
+            if (disk.snapshot == "no")
+                return;
+
+            // Disks of type "file" but without a source (such as
```
mvollmer, line 255: This explanation is wrong. For external snapshots, all disks that should be part of the snapshots must be explicitly listed, even on the command line. There is nothing "automatic". We just need to skip it here.

mvollmer:
> Hmm, we might need to extend this to read-only devices in general.

No, snapshots of read-only devices work just fine. (I was confused by the general brokenness of snapshotting volumes in pools, which needs to be addressed as well.)

test/check-machines-snapshots
```diff
@@ -237,6 +238,11 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
                     print("------ end snapshot XML -------")
                     raise
 
+            def revert(self):
+                b.click(f"#vm-subVmTest1-snapshot-{self.snap_num}-revert")
+                b.click('.pf-v5-c-modal-box__footer button:contains("Revert")')
+                b.wait_not_present('.pf-v5-c-modal-box__footer button:contains("Revert")')
```
martinpitt, line 244: Hmm, this feels a little thin -- it doesn't do any checks if it actually worked (in libvirt). `testSnapshotRevert` covers this. It should also check that the snapshot goes away from the list after reverting it.

Would it be reasonable to move some of the `virsh snapshot-info` checks here, and move these out of `testSnapshotRevert`?

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,27 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-                    if (poolPath) {
-                        const sourceElem = doc.createElement('source');
-                        sourceElem.setAttribute('file', `${poolPath}/${disk.source.volume}.snap`);
-                        diskElem.appendChild(sourceElem);
-                    }
+            if (disk.snapshot == "no")
+                return;
+
+            // Skip disks without source.
+            if (isExternal && disk.type == "file" && !disk.source.file)
```
martinpitt, line 256: This is interesting -- so if you don't include a disk in a snapshot, and revert it, the disk will still be present? Apparently so, as our integration test checks that `sdd` is still present after reverting.

 When I saw that code the first time a few months ago, I was a bit afraid that the entire disk would get dropped from the VM on reverting. Do you know, would this also work for a disk *with* a source file? Like, you could revert your main/OS partition to an earlier state, but keep your separate /home or /opt disk?

src/helpers.js
```diff
@@ -901,10 +901,9 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
 
     // Currently external snapshots works only for disks where we can specify a local path of a newly created disk snapshot file
     // This rules out all disks which are not file-based, or pool-based where pool is of type "dir"
-    // or media-less disks
     const supportedPools = storagePools.filter(pool => pool.type === "dir" && pool.connectionName === vm.connectionName);
     if (!disks.every(disk =>
-        (disk.type === "file" && disk.source?.file) ||
+        (disk.type === "file") ||
```
cockpituous, line 906: This added line is not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,27 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-
-                if (disk.type === "volume") {
-                    const poolPath = getStoragePoolPath(storagePools, disk.source.pool, connectionName);
-                    if (poolPath) {
-                        const sourceElem = doc.createElement('source');
-                        sourceElem.setAttribute('file', `${poolPath}/${disk.source.volume}.snap`);
-                        diskElem.appendChild(sourceElem);
-                    }
+            if (disk.snapshot == "no")
+                return;
```
cockpituous, line 253: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,27 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            // Skip disks without source.
+            if (isExternal && disk.type == "file" && !disk.source.file)
+                return;
```
cockpituous, line 257: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,27 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            const diskElem = doc.createElement('disk');
+            diskElem.setAttribute('name', disk.target);
+            diskElem.setAttribute('snapshot', 'external');
```
cockpituous, line 261: These 3 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,27 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            if (disk.type === "volume") {
+                const poolPath = getStoragePoolPath(storagePools, disk.source.pool, connectionName);
+                if (poolPath) {
+                    const sourceElem = doc.createElement('source');
+                    sourceElem.setAttribute('file', `${poolPath}/${disk.source.volume}.snap`);
+                    diskElem.appendChild(sourceElem);
```
cockpituous, line 268: These 6 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -249,22 +249,27 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
                 }
-
-                disksElem.appendChild(diskElem);
             }
+
+            disksElem.appendChild(diskElem);
```
cockpituous, line 272: This added line is not executed by any test.

martinpitt (approved):
Thanks! Some thoughts, but this is already an improvement, so I'm happy for these to be follow-ups.

test/check-machines-snapshots
mvollmer: > It should also check that the snapshot goes away from the list after reverting it.
No, it will stay in the list.

> Would it be reasonable to move some of the virsh snapshot-info checks here, and move these out of testSnapshotRevert?

Yes, I think so too. Also stuff from testSnapshot.

src/helpers.js
```diff
@@ -899,14 +899,8 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
...
-    // Currently external snapshots works only for disks where we can specify a local path of a newly created disk snapshot file
-    // This rules out all disks which are not file-based, or pool-based where pool is of type "dir"
-    // or media-less disks
-    const supportedPools = storagePools.filter(pool => pool.type === "dir" && pool.connectionName === vm.connectionName);
-    if (!disks.every(disk =>
-        (disk.type === "file" && disk.source?.file) ||
-        (disk.type === "volume" && supportedPools.some(pool => pool.name === disk.source.pool))
-    )) {
+    // Currently external snapshots work only for disks of type "file"
+    if (!disks.every(disk => disk.type === "file")) {
```
cockpituous, line 903: This added line is not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -246,25 +263,23 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-                disksElem.appendChild(diskElem);
-            }
+            // Disk can have attribute "snapshot" set to "no", which
+            // means no snapshot should be created of the said disk
+            // This cannot be configured through cockpit, but we
+            // should uphold it nevertheless see "snapshot" attribute
+            // of <disk> element at
+            // https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms
+            if (disk.snapshot == "no")
+                return;
```
cockpituous, line 273: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -246,25 +263,23 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            // Skip disks without source, such as empty media drives.
+            if (isExternal && disk.type == "file" && !disk.source.file)
+                return;
```
cockpituous, line 277: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -246,25 +263,23 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            const diskElem = doc.createElement('disk');
+            diskElem.setAttribute('name', disk.target);
+            diskElem.setAttribute('snapshot', 'external');
+            disksElem.appendChild(diskElem);
```
cockpituous, line 282: These 4 added lines are not executed by any test.

test/check-machines-snapshots
martinpitt: > No, it will stay in the list.

Ah ok -- but does *anything* change in the UI? There's a "current" indicator or so? Can this be detected via the CLI somehow, or purely inside the guest VM? (which is hard to introspect)

src/helpers.js
```diff
@@ -899,8 +899,8 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
         return false;
     }
 
-    // Currently external snapshots work only for disks of type "file" with a source.
-    if (!disks.every(disk => (disk.type === "file" && disk.source?.file))) {
+    // Currently external snapshots work only for disks of type "file"
+    if (!disks.every(disk => disk.type === "file")) {
```
cockpituous, line 903: This added line is not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -263,15 +263,23 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-                disksElem.appendChild(diskElem);
-            }
+            // Disk can have attribute "snapshot" set to "no", which
+            // means no snapshot should be created of the said disk
+            // This cannot be configured through cockpit, but we
+            // should uphold it nevertheless see "snapshot" attribute
+            // of <disk> element at
+            // https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms
+            if (disk.snapshot == "no")
+                return;
```
cockpituous, line 273: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -263,15 +263,23 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            // Skip disks without source, such as empty media drives.
+            if (isExternal && disk.type == "file" && !disk.source.file)
+                return;
```
cockpituous, line 277: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -263,15 +263,23 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            const diskElem = doc.createElement('disk');
+            diskElem.setAttribute('name', disk.target);
+            diskElem.setAttribute('snapshot', 'external');
+            disksElem.appendChild(diskElem);
```
cockpituous, line 282: These 4 added lines are not executed by any test.

test/check-machines-snapshots
mvollmer: > Ah ok -- but does _anything_ change in the UI?

Actually, no. The snapshot was current before reverting it, and it is still current afterwards. I now think the test is fine as it is, providing a extra check that the created snapshot can actually be reverted to. testSnapshotRevert tests multiple snapshots and their relations.

src/helpers.js
cockpituous, line 903: This added line is not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 273: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 277: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
cockpituous, line 282: These 4 added lines are not executed by any test.

test/check-machines-snapshots
martinpitt: OK, let's keep that simple approach then.

test/check-machines-snapshots
```diff
@@ -238,6 +239,11 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
...
```
martinpitt, line 245: This is weird. Shouldn't this ensure that the whole dialog goes away? `b.wait_not_present(".pf-v5-c-modal-box")`

src/libvirt-xml-create.js
```diff
@@ -263,15 +263,23 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-            if (disk.snapshot !== "no") {
-                const diskElem = doc.createElement('disk');
-                diskElem.setAttribute('name', disk.target);
-                diskElem.setAttribute('snapshot', 'external');
-                disksElem.appendChild(diskElem);
-            }
+            // Disk can have attribute "snapshot" set to "no", which
+            // means no snapshot should be created of the said disk
+            // This cannot be configured through cockpit, but we
+            // should uphold it nevertheless see "snapshot" attribute
```
martinpitt, line 269: You dropped the line break between "nevertheless" and "see", this is now a bit hard to read. Add a `. ` or move it back to its own line?

martinpitt (commented):
Only trivial bits left. This is a good improvement, thanks!

test/check-machines-snapshots
mvollmer: Yeah, I think I cargo-culted it. I have a look.

src/libvirt-xml-create.js
mvollmer: Fixed.

test/check-machines-snapshots
mvollmer: Done.

src/helpers.js
cockpituous, line 903: This added line is not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -263,15 +263,24 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-            }
+            // Disk can have attribute "snapshot" set to "no", which
+            // means no snapshot should be created of the said disk
+            // This cannot be configured through cockpit, but we
+            // should uphold it nevertheless.
+            //
+            // See "snapshot" attribute of <disk> element at
+            // https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms
+            if (disk.snapshot == "no")
+                return;
```
cockpituous, line 274: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -263,15 +263,24 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            // Skip disks without source, such as empty media drives.
+            if (isExternal && disk.type == "file" && !disk.source.file)
+                return;
```
cockpituous, line 278: These 2 added lines are not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -263,15 +263,24 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
+
+            const diskElem = doc.createElement('disk');
+            diskElem.setAttribute('name', disk.target);
+            diskElem.setAttribute('snapshot', 'external');
+            disksElem.appendChild(diskElem);
```
cockpituous, line 283: These 4 added lines are not executed by any test.

martinpitt (approved):
Thanks! Please see my coverage replies above, but this is a good improvement.

mvollmer:
> Are you planning to add a test for a non-file disk (perhaps a partition or dir pool?) in the future?

Such a test was added in #1566.

mvollmer:
> This is the kind of thing that libvirt should set by default for empty drives..

Are you saying that libvirt already sets this attribute, or that it would be nice if libvirt would set it?

> Can this be controlled with the XML (I suppose yes?) then this might be possible to actually test, at least in a follow-up PR.

Yes, also for `snapshot="internal".`

mvollmer:
> Now, that is weird... Not covering this would mean we don't test external snapshots at all? 😕

This will only run for external snapshots, and we collect coverage on f39, which doesn't implement them yet. I am thinking to switch coverage to f40, actually.

martinpitt:
> Are you saying that libvirt already sets this attribute, or that it would be nice if libvirt would set it?

The latter. This feels like an "unbreak my API" flag that we have to set right now.

> This will only run for external snapshots, and we collect coverage on f39, which doesn't implement them yet. 

Ah, of course -- I keep forgetting how new they are.

mvollmer:
> > Are you saying that libvirt already sets this attribute, or that it would be nice if libvirt would set it?
> 
> The latter. This feels like an "unbreak my API" flag that we have to set right now.

I think the way it is supposed to work is that this attribute is for overriding the default, and wouldn't be set by libvirt itself. The default for a media-less drive should of course be "no", and it indeed is AFAICS. However, a external snapshot is done by overriding this attribute as well, and without detecting media-less drives ourselves, we would explicitly be telling libvirt "make a external snapshot of this" instead of letting it use the sensible default. It's a API design issue, imo.

PR #1566, opened 2024-04-19
snapshots: Only use external snapshots for disks of type "file"

mvollmer:
The previous method of changing the source of a "volume" disk to be a "file" disk leads to broken snapshots that can't be reverted or deleted.

And leaving the source alone leads to an error during snapshot creation.

Fixes: https://issues.redhat.com/browse/RHEL-33468

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -899,14 +899,8 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
         return false;
     }
 
-    // Currently external snapshots works only for disks where we can specify a local path of a newly created disk snapshot file
-    // This rules out all disks which are not file-based, or pool-based where pool is of type "dir"
-    // or media-less disks
-    const supportedPools = storagePools.filter(pool => pool.type === "dir" && pool.connectionName === vm.connectionName);
-    if (!disks.every(disk =>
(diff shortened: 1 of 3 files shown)


src/helpers.js
```diff
@@ -899,14 +899,8 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
...
-    // Currently external snapshots works only for disks where we can specify a local path of a newly created disk snapshot file
-    // This rules out all disks which are not file-based, or pool-based where pool is of type "dir"
-    // or media-less disks
-    const supportedPools = storagePools.filter(pool => pool.type === "dir" && pool.connectionName === vm.connectionName);
-    if (!disks.every(disk =>
-        (disk.type === "file" && disk.source?.file) ||
-        (disk.type === "volume" && supportedPools.some(pool => pool.name === disk.source.pool))
-    )) {
+    // Currently external snapshots work only for disks of type "file" with a source.
+    if (!disks.every(disk => (disk.type === "file" && disk.source?.file))) {
```
cockpituous, line 903: This added line is not executed by any test.

martinpitt (approved):
Thanks!

PR #1569, opened 2024-04-22
snapshots: Refresh VM after creating/reverting/deleting

mvollmer:
Managing external snapshots will change the source of the disks of the VM.

Demo: https://youtu.be/poUAVaBy3oo

### DIFF
--- a/src/components/vm/snapshots/vmSnapshotsCard.jsx
+++ b/src/components/vm/snapshots/vmSnapshotsCard.jsx
@@ -30,6 +30,7 @@ import { CheckIcon, InfoAltIcon } from '@patternfly/react-icons';
 import { DeleteResourceButton } from '../../common/deleteResource.jsx';
 import { RevertSnapshotModal } from './vmSnapshotsRevertModal.jsx';
 import { snapshotDelete, snapshotGetAll } from '../../../libvirtApi/snapshot.js';
+import { domainGet } from '../../../libvirtApi/domain.js';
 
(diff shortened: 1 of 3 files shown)


mvollmer:
I don't feel an urge to write a test for this, tbh.

martinpitt:
Hmm, that feels like something that should already happen via an event in the libvirt API. This is non-trivial, and easy to break in a future code restructuring, and shouldn't be hard to test. The original name is predictable, and for the new name you could just assert that it startswith('/var../foo') and isn't the same as the original name? That proves that the UI did react properly.

mvollmer:
> Hmm, that feels like something that should already happen via an event in the libvirt API.

Are you saying there is an event, but we don't listen to it?

mvollmer:
> This is non-trivial, and easy to break in a future code restructuring, and shouldn't be hard to test.

It's not hard to test, but it is testing for behavior that is arguably a bug. (The source of a disk shouldn't really change, that's just a detail of a half-assed implementation leaking out to the surface. It doesn't change with internal snapshots.) That's what made me hesitant.

martinpitt:
> Are you saying there is an event, but we don't listen to it?

No, just that there should be an event if the definition changes, so that the UI would auto-refresh.

> that's just a detail of a half-assed implementation leaking out to the surface

OK, that may explain why there isn't.

OK, *shrug*, I stop pretending that I know what's going on here :grin:

martinpitt (approved):

PR #1571, opened 2024-04-22
eslint: enable no-useless-fragment rule by default

jelly:
This rule is not part of the standard ESLint recommends set but is useful and can be auto-fixed.


martinpitt (approved):
Nice!

jelly:
> Nice!

I'll apply this to starter-kit, podman, certificates and ostree as well.

PR #1575, opened 2024-04-23
snapshots: Don't list disks in XML

mvollmer:
We can control whether or not an external snapshot is made with the "disk only" flag and the memory path, and libvirt will provide defaults for all the disks.

### DIFF
--- a/src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
+++ b/src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
@@ -147,22 +147,18 @@ export class CreateSnapshotModal extends React.Component {
 
     onCreate() {
         const Dialogs = this.context;
-        const { vm, isExternal, storagePools } = this.props;
+        const { vm, isExternal } = this.props;
         const { name, description, memoryPath } = this.state;
-        const disks = Object.values(vm.disks);
         const validationError = this.onValidate();
 
         if (!Object.keys(validationError).length) {
             this.setState({ inProgress: true });
             snapshotCreate({
-                connectionName: vm.connectionName,
-                vmId: vm.id,
+                vm,
                 name,
                 description,
-                memoryPath: vm.state === "running" && memoryPath,
-                disks,
                 isExternal,
-                storagePools
+                memoryPath: isExternal && vm.state === "running" && memoryPath,
             })
                     .then(() => {
                         // VM Snapshots do not trigger any events so we have to refresh them manually
--- a/src/libvirt-xml-create.js
+++ b/src/libvirt-xml-create.js
@@ -236,7 +236,7 @@ export function getPoolXML({ name, type, source, target }) {
 }
 
 // see https://libvirt.org/formatsnapshot.html
-export function getSnapshotXML(name, description, disks, memoryPath, isExternal, storagePools, connectionName) {
+export function getSnapshotXML(name, description, memoryPath) {
     const doc = document.implementation.createDocument('', '', null);
 
     const snapElem = doc.createElement('domainsnapshot');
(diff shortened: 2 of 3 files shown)


mvollmer:
I kinda like this, and I kinda don't. It's a bizarre API...

src/libvirtApi/snapshot.js
```diff
@@ -30,9 +30,27 @@ import { call, Enum, timeout } from './helpers.js';
...
+
+    // The behavior is as follows:
+    //
+    // VM state  |  disk only  |  memory path   =>  resulting snapshot type
+    // --------------------------------------------------------------------
+    // shutoff   |  false      |  null          =>  internal
+    // shutoff   |  true       |  null          =>  external
+    // running   |  false      |  null          =>  internal full system
+    // running   |  false      |  non-null      =>  external full system
+    // running   |  true       |  null          =>  external disk only
```
martinpitt, line 48: That's the case which I find crazy. Can we add some assertion that we never ever attempt this?

src/libvirtApi/snapshot.js
```diff
@@ -30,9 +30,27 @@ import { call, Enum, timeout } from './helpers.js';
 import { logDebug } from '../helpers.js';
 
 export function snapshotCreate({ connectionName, vmId, name, description, memoryPath, disks, isExternal, storagePools }) {
```
martinpitt, line 32: Please also drop the now unused properties from this function.

mvollmer: Done.

martinpitt:
More `make codecheck` fallout:
```
#   150:33  error  'storagePools' is assigned a value but never used  no-unused-vars
#   152:15  error  'disks' is assigned a value but never used         no-unused-vars
```

martinpitt (approved):
Thanks! There's still the "Can we add some assertion" question from above, but not blocking this PR. I'm surprised that the tests just continue to work without modification.

martinpitt (changes requested):
Seems the tests [do need adjusting](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1575-fcb022d5-20240425-132615-fedora-39/log.html#79-2) after all.

mvollmer:
> There's still the "Can we add some assertion" question from above,

I know you don't like that case, but I don't think it counts as a programming error to request it... :-)

mvollmer:
> > There's still the "Can we add some assertion" question from above,
> 
> I know you don't like that case, but I don't think it counts as a programming error to request it... :-)

Then again, the state handling here is a bit involved, and I just made a mistake which lead to the creation of external snapshots on fedora-39... (which apparently works just fine?) so... I am sympathetic to that now. Let's do it.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -147,22 +147,18 @@ export class CreateSnapshotModal extends React.Component {
...
-                connectionName: vm.connectionName,
-                vmId: vm.id,
+                vm,
                 name,
                 description,
-                memoryPath: vm.state === "running" && memoryPath,
-                disks,
                 isExternal,
-                storagePools
+                memoryPath: isExternal && vm.state === "running" && memoryPath,
```
cockpituous, line 161: This added line is not executed by any test.

src/libvirt-xml-create.js
```diff
@@ -253,36 +253,11 @@ export function getSnapshotXML(name, description, disks, memoryPath, isExternal,
...
-            diskElem.setAttribute('name', disk.target);
-            diskElem.setAttribute('snapshot', 'external');
-            disksElem.appendChild(diskElem);
-        });
-        snapElem.appendChild(disksElem);
+    if (memoryPath) {
+        const memoryElem = doc.createElement('memory');
+        memoryElem.setAttribute('snapshot', 'external');
+        memoryElem.setAttribute('file', memoryPath);
+        snapElem.appendChild(memoryElem);
```
cockpituous, line 260: These 5 added lines are not executed by any test.

src/libvirtApi/snapshot.js
```diff
@@ -29,11 +29,37 @@ import { parseDomainSnapshotDumpxml } from '../libvirt-xml-parse.js';
...
     const flags = (!isExternal || memoryPath) ? 0 : Enum.VIR_DOMAIN_SNAPSHOT_CREATE_DISK_ONLY;
-    const xmlDesc = getSnapshotXML(name, description, disks, memoryPath, isExternal, storagePools, connectionName);
-    return call(connectionName, vmId, 'org.libvirt.Domain', 'SnapshotCreateXML', [xmlDesc, flags], { timeout, type: 'su' });
+    const xmlDesc = getSnapshotXML(name, description, memoryPath);
+
+    // We really don't want to make disk only snapshots of running
+    // machines by accident.
+
+    if (vm.state === "running" && (flags & Enum.VIR_DOMAIN_SNAPSHOT_CREATE_DISK_ONLY))
+        throw new Error("Cowardly refusing to make a disk-only snapshot of a running machine");
```
cockpituous, line 59: This added line is not executed by any test.

src/libvirtApi/snapshot.js
martinpitt: Thanks!

src/libvirtApi/snapshot.js
martinpitt: Yeah, `// not-covered: assertion` or so.

martinpitt (approved):
Cheers! I'm glad that the tests found a bug then.

PR #1560, opened 2024-04-16
snapshots: Allow disk-only snapshots of running machines

mvollmer:
- [x] Use "Full system" instead of "Complete". That's the libvirt terminology.
- [x] Detect guest agent and add "quiesce" flag is found.
- [x] Test quiesce functionality.... Maybe switch from cirros to alpine, which has qemu-guest-agent

image

image

image

PR #1576, opened 2024-04-23
snapshots: Use --quiesce and --live during creation as appropriate

mvollmer:
This is the recommended way to create snapshots of running machines.


mvollmer:
I guess we should not use --live unless we also give people the option to use disk-only snapshots for running machines.  --live seems to be risky in that it might require lots of unexpected storage amounts.

src/libvirtApi/snapshot.js
```diff
@@ -31,7 +31,11 @@ import { logDebug } from '../helpers.js';
 
 export function snapshotCreate({ connectionName, vmId, name, description, memoryPath, disks, isExternal, storagePools }) {
     // that flag ought to be implicit for non-running VMs, see https://issues.redhat.com/browse/RHEL-22797
-    const flags = (!isExternal || memoryPath) ? 0 : Enum.VIR_DOMAIN_SNAPSHOT_CREATE_DISK_ONLY;
+    let flags = 0;
+    if (isExternal && !memoryPath)
+        flags |= Enum.VIR_DOMAIN_SNAPSHOT_CREATE_DISK_ONLY | Enum.VIR_DOMAIN_SNAPSHOT_CREATE_QUIESCE;
```
cockpituous, line 36: These 2 added lines are not executed by any test.

martinpitt: IMHO this belongs into PR #1560 -- this is the same functionality, with just some tweaks.

src/libvirtApi/snapshot.js
```diff
@@ -31,7 +31,11 @@ import { logDebug } from '../helpers.js';
...
+    if (memoryPath)
+        flags |= Enum.VIR_DOMAIN_SNAPSHOT_CREATE_LIVE;
```
martinpitt, line 38: I know this is a draft, and maybe it's just a reminder. But this can't go in like this. We should always support the non-live way where a running VM is paused first before taking the memory dump. That is slower (and some connections to the VM may time out during the pause), but much safer and predictable. The live option should be behind some big :warning:  "danger! don't do this unless you are know what you are doing!" kind of warning sign. That's the one which has indefinite runtime and disk requirements and thus is hard to predict, no?

mvollmer:
For my own record:

 --quiesce requires --disk-only
 --memspec conflicts with --disk-only
 --quiesce causes the snapshot creation to fail when there is no agent in the gues
 --quiesce with a stopped machine does nothing

Thus: quiesce can only be used for disk only snapshots of running machines, and it might cause the snapshot creation to fail in that case. Do we need to find a way to figure out whether the guest has an agent?

mvollmer:
> For my own record:
> 
> --quiesce requires --disk-only --memspec conflicts with --disk-only --quiesce causes the snapshot creation to fail when there is no agent in the gues --quiesce with a stopped machine does nothing
> 
> Thus: quiesce can only be used for disk only snapshots of running machines, and it might cause the snapshot creation to fail in that case. Do we need to find a way to figure out whether the guest has an agent?

"Currently, there is no reliable method/API to tell if GA is running", https://wiki.libvirt.org/Qemu_guest_agent.html

PR #1578, opened 2024-04-24
networks: Guide people to a valid IP address

### LINKED ISSUE #1567: Creating a Virtual Network, IPv4 address or IPv4 address space?
When trying to define a virtual network I'm met with an error that I can't understand. When defining a network I assume the 'IPv4 address' field means the 'Network Address' but an error occurs that says "IPv4 address cannot be same as the network identifier".
 
image

Looking at [virt-manager](https://virt-manager.org) documentation/screenshots, here I see a 'Network' field and not an 'IPv4 address'  field (see [here](https://wiki.libvirt.org/TaskRoutedNetworkSetupVirtManager.html))

image
image

What should I enter in cockpit under 'IPv4 address' if not the network address?

image

mvollmer:
... when they enter a network identifier such as 192.168.100.0.

Fixes #1567

### DIFF
--- a/src/components/networks/createNetworkDialog.jsx
+++ b/src/components/networks/createNetworkDialog.jsx
@@ -53,10 +53,22 @@ function validateParams(dialogValues) {
         validationFailed.name = _("Name should not be empty");
 
     if (dialogValues.ip === "IPv4 only" || dialogValues.ip === "IPv4 and IPv6") {
+        let ipv4_prefix = null;
+
         if (isEmpty(dialogValues.netmask.trim()))
             validationFailed.netmask = _("Mask or prefix length should not be empty");
-        else if (!utils.validateNetmask(dialogValues.netmask))
-            validationFailed.netmask = _("Invalid IPv4 mask or prefix length");
+        else {
+            ipv4_prefix = utils.parseNetmask(dialogValues.netmask);
+            if (ipv4_prefix === null)
+                validationFailed.netmask = _("Invalid IPv4 mask or prefix length");
+            else if ((ipv4_prefix % 8) != 0) {
+                // because we use localPtr="yes".
+                validationFailed.netmask = _("IPv4 prefix length must be a multiple of 8");
+            } else if (ipv4_prefix > 24) {
+                // because we use localPtr="yes".
+                validationFailed.netmask = _("IPv4 prefix length must be 24 or less");
+            }
+        }
 
         if (isEmpty(dialogValues.ipv4.trim()))
             validationFailed.ipv4 = _("IPv4 network should not be empty");
@@ -64,11 +76,13 @@ function validateParams(dialogValues) {
             validationFailed.ipv4 = _("Invalid IPv4 address");
         // During virtual network creation, address is assigned to bridge. However no interface can have the
         // address same as the network identifier, as it would disable the connectivity of the virtual network.
-        else if (utils.ipv4IsNetworkIdentifier(dialogValues.ipv4, dialogValues.netmask))
-            validationFailed.ipv4 = _("IPv4 address cannot be same as the network identifier");
+        else if (!validationFailed.netmask && utils.ipv4IsNetworkIdentifier(dialogValues.ipv4, ipv4_prefix)) {
+            validationFailed.ipv4 =
+                cockpit.format(_("Must be an address instead of the network identifier, such as $0"),
+                               utils.ipv4ExampleBridgeAddressForNetworkIdentifier(dialogValues.ipv4));
         // Using broacast address of network space as the address of virtual network's bridge
         // is forbidden and would disable the connectivity of the virtual network.
-        else if (utils.ipv4IsBroadcast(dialogValues.ipv4, dialogValues.netmask))
+        } else if (!validationFailed.netmask && utils.ipv4IsBroadcast(dialogValues.ipv4, ipv4_prefix))
             validationFailed.ipv4 = _("IPv4 address cannot be same as the network's broadcast address");
 
         if (dialogValues.ipv4DhcpEnabled) {
@@ -76,14 +90,14 @@ function validateParams(dialogValues) {
                 validationFailed.ipv4DhcpRangeStart = _("Start should not be empty");
             else if (!utils.validateIpv4(dialogValues.ipv4DhcpRangeStart))
                 validationFailed.ipv4DhcpRangeStart = _("Invalid IPv4 address");
-            else if (!utils.isIpv4InNetwork(dialogValues.ipv4, dialogValues.netmask, dialogValues.ipv4DhcpRangeStart))
+            else if (!validationFailed.netmask && !utils.isIpv4InNetwork(dialogValues.ipv4, ipv4_prefix, dialogValues.ipv4DhcpRangeStart))
                 validationFailed.ipv4DhcpRangeStart = _("Address not within subnet");
 
             if (isEmpty(dialogValues.ipv4DhcpRangeEnd.trim()))
                 validationFailed.ipv4DhcpRangeEnd = _("End should not be empty");
             else if (!utils.validateIpv4(dialogValues.ipv4DhcpRangeEnd))
                 validationFailed.ipv4DhcpRangeEnd = _("Invalid IPv4 address");
-            else if (!utils.isIpv4InNetwork(dialogValues.ipv4, dialogValues.netmask, dialogValues.ipv4DhcpRangeEnd))
+            else if (!validationFailed.netmask && !utils.isIpv4InNetwork(dialogValues.ipv4, ipv4_prefix, dialogValues.ipv4DhcpRangeEnd))
                 validationFailed.ipv4DhcpRangeEnd = _("Address not within subnet");
         }
     }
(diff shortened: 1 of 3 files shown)


src/components/networks/utils.js
```diff
@@ -53,6 +53,12 @@ export function ipv4IsNetworkIdentifier(address, netmask) {
     return address === ipaddr.IPv4.networkAddressFromCIDR(`${address}/${prefix}`).toString();
 }
 
+export function ipv4ExampleBridgeAddressForNetworkIdentifier(network) {
+    const net = ipaddr.parse(network);
+    net.octets[3] += 1;
```
martinpitt, line 58: Is that always right? AFAICS, `ipv4IsNetworkIdentifier()` would also return true if you specified e.g. a netmask of /32? And if the last octet is 255, then this would give an invalid address.

mvollmer: Yeah, that would break, that's why I called this ExampleBridgeAddress. If we use this to produce an address that gets stored in configuration, it would need to be a bit more careful, I guess. 

A prefix length of 32 is not a network, it's a single address, and we could reject it.

mvollmer: Ok, if we allow all addresses for a /32 "subnet", the next check will fail "IPv4 address cannot be same as the network's broadcast address". The same happens for /31: this is a subnet with only two "addresses": one is the network id, the other is the broadcast address. So that is out as well. Let's just reject /32 and /31 subnets.

mvollmer: Also, we include `localPtr="yes"` in our network XML definition, and that requires a prefix <= 30. So that's another reason to reject prefix >= 31.

src/components/networks/createNetworkDialog.jsx
```diff
@@ -55,8 +55,10 @@ function validateParams(dialogValues) {
     if (dialogValues.ip === "IPv4 only" || dialogValues.ip === "IPv4 and IPv6") {
         if (isEmpty(dialogValues.netmask.trim()))
             validationFailed.netmask = _("Mask or prefix length should not be empty");
-        else if (!utils.validateNetmask(dialogValues.netmask))
+        else if (!utils.parseNetmask(dialogValues.netmask))
             validationFailed.netmask = _("Invalid IPv4 mask or prefix length");
+        else if (!utils.validateNetmask(dialogValues.netmask))
+            validationFailed.netmask = _("IPv4 prefix length must be 30 bits or less");
```
cockpituous, line 61: This added line is not executed by any test.

src/components/networks/utils.js
```diff
@@ -79,6 +85,24 @@ export function validateNetmask(prefixOrNetmask) {
...
+/**
+ * validates correctness and suitability of ipv4 prefix length or mask
+ *
+ * @param {string} prefixOrNetmask
+ * @returns int prefix length, or null if invalid
+ */
+export function validateNetmask(prefixOrNetmask) {
+    const prefix = parseNetmask(prefixOrNetmask);
+    if (prefix === null)
+        return null;
```
cockpituous, line 97: This added line is not executed by any test.

src/components/networks/utils.js
```diff
@@ -79,6 +85,24 @@ export function validateNetmask(prefixOrNetmask) {
...
+
+    // Networks with zero or one bits of address are too small.
+    if (prefix >= 31)
+        return null;
```
cockpituous, line 101: This added line is not executed by any test.

src/components/networks/utils.js
mvollmer: Alright localPtr="yes" actually has more restrictions: IPv4 prefix must be divisible by 8.  It it's not, trying to activate the network will fail and actually disconnect Cockpit.

mvollmer:
I will refactor the prefix validation a bit more.

src/components/networks/utils.js
cockpituous, line 97: This added line is not executed by any test.

mvollmer:
For those curious: `localPtr="yes"` was added in https://github.com/cockpit-project/cockpit/pull/14564.

mvollmer:
About IPv6: It has similar restrictions when used with `localPtr="yes"` (prefix must be multipl of 4, maybe others). However, we don't set `localPtr` in the IPv6 portion of the XML definition.

src/components/networks/createNetworkDialog.jsx
```diff
@@ -53,37 +53,49 @@ function validateParams(dialogValues) {
...
+        else {
+            ipv4_prefix = utils.parseNetmask(dialogValues.netmask);
+            if (ipv4_prefix === null)
+                validationFailed.netmask = _("Invalid IPv4 mask or prefix length");
+            else if ((ipv4_prefix % 8) != 0) {
+                // because we use localPtr="yes".
+                validationFailed.netmask = _("IPv4 prefix length must be a multiple of 8");
+            } else if (ipv4_prefix == 32) {
+                // because we use localPtr="yes".
+                validationFailed.netmask = _("IPv4 prefix length must be 24 or less");
```
martinpitt, line 69: TIL! So far I thought that e.g. /28 networks would be a thing.

However, this looks inconsistent: You check for `== 32` but say "24 or less" -- should that be `> 24` then?

src/components/networks/createNetworkDialog.jsx
```diff
@@ -53,37 +53,49 @@ function validateParams(dialogValues) {
...
         if (isEmpty(dialogValues.ipv4.trim()))
             validationFailed.ipv4 = _("IPv4 network should not be empty");
         else if (!utils.validateIpv4(dialogValues.ipv4))
             validationFailed.ipv4 = _("Invalid IPv4 address");
         // During virtual network creation, address is assigned to bridge. However no interface can have the
         // address same as the network identifier, as it would disable the connectivity of the virtual network.
-        else if (utils.ipv4IsNetworkIdentifier(dialogValues.ipv4, dialogValues.netmask))
-            validationFailed.ipv4 = _("IPv4 address cannot be same as the network identifier");
+        else if (!validationFailed.netmask && utils.ipv4IsNetworkIdentifier(dialogValues.ipv4, ipv4_prefix))
+            validationFailed.ipv4 = cockpit.format(_("Not a valid IP address. Please specify an address instead of the network identifier, such as $0"), utils.ipv4ExampleBridgeAddressForNetworkIdentifier(dialogValues.ipv4));
```
martinpitt, line 80: Can you please break this long line?

test/check-machines-networks
```diff
@@ -425,6 +425,36 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
...
+        # Check "prefix must be multiple of 8"
+        NetworkCreateDialog(
+            self,
+            name="test_network",
+            forward_mode="open",
+            ip_conf="IPv4 only",
+            ipv4_address="192.168.100.1",
+            ipv4_netmask="26",
+            ipv4_dhcp_start="300.2.1",
+            ipv4_dhcp_end="168..1.1.1",
```
martinpitt, line 437: Can you  please make these valid? It's a bit confusing to have more than one wrong value per test case. These are already covered by the previous test.

src/components/networks/createNetworkDialog.jsx
mvollmer: Yes, it's complicated.

 /28 subnets are totally a thing, but libvirt is not able to handle them with localPtr="yes", which we set in libvirt-xml-create.js in the network XML. I don't know if that is a fundamental problem, or just a missing feature in libvirt. I don't even know what localPtr="yes" does.

In fact, the only valid values are 8, 16, and 24. We can as well have a dropdown for them.

I'll change "== 32" check as you suggest.

test/check-machines-networks
mvollmer: This does in fact test something useful: As long as the prefix is invalid, we don't want any fields to be checked that depend on a valid prefix. Thus, this checks that the DHCP fields do _not_ have errors.

This could be improved, I guess: We can call out syntax errors even without a valid prefix. (The prefix is only used to check whether the values are out of bound.)

Not sure how far we want to go...

I make these valid.

martinpitt (changes requested):
Thanks! Looks good, but `make codecheck` complains:

> #   83:9  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style

martinpitt:
@mvollmer This [failure](https://artifacts.dev.testing-farm.io/4c9d5996-8808-43da-b713-f456c179e1bd/) is relevant:
```
  File "/source/test/check-machines-networks", line 487, in testNetworksCreate
    ).execute()
      ^^^^^^^^^
  File "/source/test/check-machines-networks", line 161, in execute
    self.create()
  File "/source/test/check-machines-networks", line 220, in create
    self.test_obj.assertIn(self.xfail_error, b.eval_js(f"ph_input_form_helper_text(\"create-network-{xfo}\")"))
AssertionError: 'Not a valid IP address. Please specify an address instead of the network identifier, such as 192.168.100.1' not found in 'Must be an address instead of the network identifier, such as 192.168.100.1'

# Result testNetworksCreate (__main__.TestMachinesNetworks.testNetworksCreate) failed
# 1 TEST FAILED [26s on ip-172-31-19-191]
```

Straightforward string change.

martinpitt (changes requested):
Clearing my "review requested" list, this still needs a test adjustment.

martinpitt (approved):
Kiitos!

PR #1590, opened 2024-04-26
test: Use alpine image instead of cirros for guests

mvollmer:
Alpine has qemu-guest-agent, which we want to use to test snapshots.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -107,8 +107,8 @@ if [ "$ID" = "fedora" ]; then
     export TEST_TIMEOUT_FACTOR=3
 fi
 
-# pre-download cirros image for Machines tests
-bots/image-download cirros
+# pre-download alpine image for Machines tests
+bots/image-download alpine
 
 exclude_options=""
 for t in $EXCLUDES; do
--- a/test/check-machines-consoles
+++ b/test/check-machines-consoles
@@ -81,12 +81,12 @@ class TestMachinesConsoles(machineslib.VirtualMachinesCase):
         b.wait_visible(".pf-v5-c-console__vnc canvas")
 
         # make sure the log file is full - then empty it and reboot the VM - the log file should fill up again
-        self.waitCirrOSBooted(args['logfile'])
+        self.waitGuestBooted(args['logfile'])
 
         self.machine.execute(f"echo '' > {args['logfile']}")
         b.click("#subVmTest1-system-vnc-sendkey")
         b.click("#ctrl-alt-Delete")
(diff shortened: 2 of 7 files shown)


mvollmer:
Hmm, I am not really sure what's the deal is with Q35 vs i4400fx, and whether it is actually worth switching the tests over... All machines that I have created recently seem to have ended with Q35, but our own CI machines (created by vm-run etc) are i4400fx... I dig a bit more...

I guess qemu is about choice...

mvollmer:
Yeah, even CentOS 7 will give you a Q35 machine by default.

mvollmer:
There are disk space issues related to cloning and migration. Maybe we use "cirros" in those tests.

mvollmer:
The pixel test failures need to be investigated, they are not just cosmetic.

mvollmer:
> Or doesn't q35 work with cirros?

I'd say it's the other way around: Alpine works fine with i440fx, so we can indeed split that out.

mvollmer:
> Hmm, I am not really sure what's the deal is with Q35 vs i440fx, and whether it is actually worth switching the tests over...

There are rumors that i440fx will be deprecated on some platforms at some time... So let's indeed use it by default, and test on both where it matters.

Now, replacing Cirros with Alpine seems to be orthogonal to q35 vs i440fx. I think Alpine is a more "realistic" choice, as its a real OS that can do real work. It has a package manager etc, and there are interactions between the hypervisor and the guest that depend on the guest OS, such as running the qemu-guest-agent, and maybe others. So, I would be happy to use Alpine instead of Cirros everywhere as well, which seems very possible.

martinpitt:
> seems to be orthogonal to q35 vs i440fx

I thought virt-install would determine this through `--os-name`/`--os-version`? We could specify the `--machine` explicitly, but TBH I really don't want to.

> I would be happy to use Alpine instead of Cirros everywhere as well

FTR, me 2!

mvollmer:
> I thought virt-install would determine this through `--os-name`/`--os-version`? We could specify the `--machine` explicitly, but TBH I really don't want to.

Yes, but we can of course use the "cirros" image and tell virt-install "--osinfo linux2022". Or run the "alpine" image with "--osinfo=linux2016".

mvollmer:
I have split out the "move to q35" change into #1601.

mvollmer:
@martinpitt, I think this is ready for serious consideration.

martinpitt (approved):
Thanks! I understand this is the path towards q35 and better guest testing, even though this PR by itself doesn't have any noticeable benefit.

PR #1599, opened 2024-04-29
Initial backports to rhel-8 branch to set up CI

martinpitt:
This is a new branch, and will receive z-stream updates for 8.10/9.4.

### DIFF
--- a/.cockpit-ci/container
+++ b/.cockpit-ci/container
@@ -1,0 +1,1 @@
+ghcr.io/cockpit-project/tasks:2024-04-25
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,7 @@ jobs:
   source:
     runs-on: ubuntu-latest
     container:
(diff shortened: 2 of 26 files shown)


martinpitt:
[This flake](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1599-c3459150-20240429-130516-rhel-9-4/log.html#25-2) is unfortunately quite persistent. The retry worked on firefox, and it doesn't fail locally, and it's not on the weather report. I also don't see anything immediately obvious in `git shortlog 311.. -- pkg/lib/`, i.e. some test fix which we could import. So one more try..

martinpitt:
Wow, so it actually *is* the dropdowns.. I backported the commits, and [tests pass](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1599-16a9355f-20240430-104701-rhel-9-4/log.html) with amplification. :man_shrugging:

jelly (approved):

PR #1600, opened 2024-04-29
watchdog: Detect permanent ones and adapt accordingly

mvollmer:


src/components/vm/overview/watchdog.jsx
```diff
@@ -37,7 +37,7 @@ import { WATCHDOG_INFO_MESSAGE } from './helpers.jsx';
 
 const _ = cockpit.gettext;
 
-const SUPPORTEDACTIONS = ["reset", "poweroff", "inject-nmi", "pause"];
+const SUPPORTEDACTIONS = ["reset", "poweroff", "inject-nmi", "pause", "none"];
```
martinpitt, line 40: I see what this does in the test. So it is related to permanent watchdogs. However, this would/could still make sense as a separate commit (including the pixel change), if the "normal" watchdogs can have a "none" action as well? This also needs to gain a translation in helper.js's `watchdogAction` "human string" table.

src/components/vm/overview/watchdog.jsx
```diff
@@ -68,7 +68,7 @@ const WatchdogModalAlert = ({ dialogError }) => {
     return <ModalError dialogError={dialogError.text} dialogErrorDetail={dialogError.detail} />;
 };
 
-export const WatchdogModal = ({ vm, isWatchdogAttached, idPrefix }) => {
+export const WatchdogModal = ({ vm, isWatchdogAttached, isPermanent, idPrefix }) => {
```
martinpitt, line 71: Is "permanent" some libvirt lingo? Or could this be called `isRemovable`?

src/components/vm/overview/watchdog.jsx
```diff
@@ -227,9 +227,14 @@ export const WatchdogLink = ({ vm, idPrefix, onAddErrorNotification }) => {
     const Dialogs = useDialogs();
 
     const isWatchdogAttached = Object.keys(vm.watchdog).length > 0;
+    const isPermanent = (vm.watchdog.model === "itco");
```
martinpitt, line 230: This is a magic value. Is there some doc link to that? That'd be great as a comment for our future selves.

test/check-machines-settings
```diff
@@ -643,13 +648,13 @@ class TestMachinesSettings(machineslib.VirtualMachinesCase):
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
 
         # Test configuring watchdog for running VM
-        setWatchdogActionLive(action="reset")
+        setWatchdogActionLive(action="reset", removable=False)
```
martinpitt, line 651: Is this because it's q35, or because it's running, and watchdogs can't be removed from a running VM?

test/check-machines-settings
```diff
@@ -705,6 +710,28 @@ class TestMachinesSettings(machineslib.VirtualMachinesCase):
...
+
+        m.execute("virsh destroy subVmTest2")
+
+        # Create a Q35 machine, which has a permanent watchdog that
+        # can't be removed.  (Some older versions of libvirt don't add
+        # a watchdog.)
+
+        m.execute("virsh destroy subVmTest1; virsh undefine subVmTest1")
+        args = self.createVm("subVmTest1", os="linux2022")
+        has_watchdog = (m.execute("virsh dumpxml subVmTest1 | grep '<watchdog' || echo no") != "no\n")
```
martinpitt, line 727: This is a strange way to write that, and it also hides errors in the pipeline. How about `"<watchdog" in m.execute("virsh..")`?

But moreover, this is too dynamic for a test for my taste -- if the above comment is true, then this should always be `True`, no? So this could instead be an assertion? Or at least a more static `if m.image in [old, OSes]`?

martinpitt (changes requested):
Thanks! This could do with some verbiage in the commit message to describe the "why" and some overview of what this does. Is there a bug report about this somewhere?

src/components/vm/overview/watchdog.jsx
mvollmer: > This also needs to gain a translation in helper.js's watchdogAction "human string" table.

That's already there, afaics.

src/components/vm/overview/watchdog.jsx
mvollmer: No, I can change that to isRemovabel.

src/components/vm/overview/watchdog.jsx
mvollmer: I didn't find libvirt documentation that outright says that it is not removable. This seems to be part of emulating the q35 chipset, which has a real watchdog in its southbridge. So I added a comment explaining that.

(There is probably a way to dig out the removability of devices by checking to which kind of bus they are connected... I will have a look at that as well.)

test/check-machines-settings
mvollmer: This is because there isn't any watchdog yet. The dialog just shows "Add" and no "Remove", obvsly.

test/check-machines-settings
mvollmer: Agreed!

src/components/vm/overview/watchdog.jsx
mvollmer: I have split out the "none" action.

src/components/vm/overview/watchdog.jsx
mvollmer: > (There is probably a way to dig out the removability of devices by checking to which kind of bus they are connected... I will have a look at that as well.)

No, the "itco" watchdog hasn't any information about how it is connected:
```
 <watchdog model='itco' action='reset'>
      <alias name='watchdog0'/>
    </watchdog>
```

Trying to remove it gives this error: "Error attempting device hotunplug: Operation not supported: hot unplug of watchdog of model itco is not supported"

So I guess checking the model string is our best option.

src/components/vm/overview/watchdog.jsx
martinpitt: I was mentioning the translation because the test does

>  b.wait_in_text("#vm-subVmTest1-watchdog-state", "none")

which looks wrong -- all other states are capitalized. Indeed I see the entry now:

    none: _("Do nothing"),

but then something is wrong.

src/components/vm/overview/watchdog.jsx
martinpitt: Thanks, that's fine - I was just wondering where it came from. The comment is enough.

test/check-machines-settings
martinpitt: Is [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1600-e72e108e-20240503-103655-rhel-8-10/log.html#87) related? Sounds plausible to put RHEL 8 into the "old" bucket.

test/check-machines-settings
```diff
@@ -705,6 +710,28 @@ class TestMachinesSettings(machineslib.VirtualMachinesCase):
...
+        m.execute("virsh destroy subVmTest1; virsh undefine subVmTest1")
+        args = self.createVm("subVmTest1", os="linux2022")
+        has_watchdog = m.image not in ["debian-stable", "ubuntu-2204", "rhel-8-10"]
+
+        self.goToVmPage("subVmTest1")
+        if has_watchdog:
+            b.wait_in_text("#vm-subVmTest1-watchdog-state", "Reset")
+            setWatchdogActionLive(action="pause", previous_action="reset", reboot_machine=True, removable=False)
+        else:
+            b.wait_in_text("#vm-subVmTest1-watchdog-state", "none")
```
martinpitt, line 734: This is still wrong, like in the previous thread. The [pixel tests](https://github.com/cockpit-project/pixel-test-reference/compare/2e5564298318492cda67adbcffc3fe1e06db2fa0..31e5a6717ab7d31db535e983efc725352fb361b9) clearly show that the "none" gets humanized to "Do nothing", in the dialog. This check looks at the VM details page (not the dialog), and wants "none". So something is still wrong with "translating" the API value to the string in this case. It clearly works for "Reset".

test/check-machines-settings
martinpitt: Thanks -- please add this as a comment.

martinpitt (changes requested):

test/check-machines-settings
mvollmer: In the UI "none" means "No watchdog".  "Do nothing" means "watchdog with action 'none'".

test/check-machines-settings
mvollmer: I have added another flag, "add", that will do the right thing. I think that is even better.

test/check-machines-settings
martinpitt: Ah, sorry!

martinpitt (approved):
Cheers!

martinpitt:
@mvollmer This [rhel-8-10 test run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1600-3c7d0037-20240503-125614-rhel-8-10/log.html) is a disaster. Most of it looks like "not patient enough" (i.e. needs longer timeouts), presumably because the inner guest is all just emulated qemu. I'll do a retry to compare, as that's not directly related to *this* PR. But it's fallout from moving to Alpine.

PR #1601, opened 2024-04-30
test: Use a Q35 machine by default in tests

mvollmer:
This is a newer chipset and creating a virtual machine in the UI or
with virt-install will very likely use it. Thus, let's also use it in
our tests.

### DIFF
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -192,7 +192,7 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
 
         # Virtio bus type should not be shown for CDROM devices
         m.execute("touch /var/lib/libvirt/novell.iso")
-        m.execute("virsh attach-disk --domain subVmTest1 --source /var/lib/libvirt/novell.iso --target hda --type cdrom --mode readonly --persistent")
+        m.execute("virsh attach-disk --domain subVmTest1 --source /var/lib/libvirt/novell.iso --target hda --type cdrom --targetbus sata --mode readonly --persistent")
         # virsh attach-disk want send an event for offline VM changes
         b.reload()
         b.enter_page('/machines')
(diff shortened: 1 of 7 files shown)


martinpitt:
#1590 landed, so this now needs a rebase.

mvollmer:
Hmm, q35 in rhel-8-10 seems to not allow enough hotplugging for our tests.

mvollmer:
Pixel test flake is being fixed in #1614

mvollmer:
> Pixel test flake is being fixed in #1614

Turns out this is something new with linux2022, actually.

test/check-machines-disks
```diff
@@ -82,7 +82,11 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
             else:
                 b.wait_not_present(f"#vm-{name}-disks-{target}-edit-dialog")
 
-        self.createVm("subVmTest1")
+        if "rhel-8" in m.image:
+            # q35 in rhel-8 can't hotplug
+            self.createVm("subVmTest1", os="linux2016")
```
martinpitt, line 87: This is done several times -- perhaps factor this out into setUp(), create some `create_options` dict (empty by default, that option on rhel8), and call `self.createVm("subVmTest1", **create_options)`?

That should also apply to centos-8-stream, no?

martinpitt (commented):
Ooh, really nice cleanup of the xml parsing mess in the tests! :hugs: 

You may need a force-push anyway to get around that hanging TF test. One suggestion for code cleanup.

test/check-machines-disks
mvollmer: Yep. I might try sticking with linux2016 on rhel-8 in general, and then bumping it up to linux2022 in specific tests explicitly.

mvollmer: Oh, this just works.

mvollmer: > That should also apply to centos-8-stream, no?

We don't test on centos-8-stream here, do we?

martinpitt: Right, that overlapped with yesterday's c8s cleanup.

martinpitt (approved):
Yay!

PR #1611, opened 2024-05-03
test: Avoid racing the serial console

mvollmer:
The "Connected" message might already be scrolled out of the terminal when we check for it. We are already checking it as well after the boot is complete, and we can make that check a bit more precise.

### DIFF
--- a/test/check-machines-consoles
+++ b/test/check-machines-consoles
@@ -110,8 +110,6 @@ class TestMachinesConsoles(machineslib.VirtualMachinesCase):
         b.click("#SerialConsole button")
         b.wait_not_present("#pf-v5-c-console__type-selector + .pf-v5-c-select__menu")
 
-        b.wait_in_text(f"#{name}-terminal .xterm-accessibility-tree", f"Connected to domain '{name}'")
-
         # In case the OS already finished booting, press Enter into the console to re-trigger the login prompt
         # Sometimes, pressing Enter one time doesn't take effect, so loop to press Enter to make sure the console has accepted it
         for _ in range(0, 60):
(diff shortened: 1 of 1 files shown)


test/check-machines-consoles
```diff
@@ -110,8 +110,6 @@ class TestMachinesConsoles(machineslib.VirtualMachinesCase):
         b.click("#SerialConsole button")
         b.wait_not_present("#pf-v5-c-console__type-selector + .pf-v5-c-select__menu")
 
-        b.wait_in_text(f"#{name}-terminal .xterm-accessibility-tree", f"Connected to domain '{name}'")
```
martinpitt, line 113: That was strange indeed - this should rather check for something at the end of the boot sequence which doesn't scroll off, like the usual "login:" prompt?

test/check-machines-consoles
```diff
@@ -127,7 +125,8 @@ class TestMachinesConsoles(machineslib.VirtualMachinesCase):
         b.wait_text(f"#{name}-terminal", "Disconnected from serial console. Click the connect button.")
 
         b.click(f"#{name}-serialconsole-connect")
-        b.wait_in_text(f"#{name}-terminal .xterm-accessibility-tree > div:nth-child(1)", "Connected to domain")
+        b.wait_in_text(f"#{name}-terminal .xterm-accessibility-tree > div:nth-child(1)",
+                       f"Connected to domain '{name}'")
```
martinpitt, line 129: So, did I get that right? That'll scroll off as well, but we have a higher chance of catching it because we start polling for this immediately after clicking the connect button. As opposed to the place above where the VM was already running for some indeterminate time?

martinpitt (approved):
Assuming green, this is a good improvement, thanks! I have an idea/question whether we can make this predictable/robust, but I didn't look into it very far.

PR #1614, opened 2024-05-06
test: Wait for IPv6 address before testing pixels

mvollmer:
So that the picture is stable.

### DIFF
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -56,6 +56,7 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         b.wait_in_text('#vm-subVmTest1-networks td[data-label="Source"]', "TAP device")
         b.wait_in_text("#vm-subVmTest1-network-1-tapdevice", "vnet0")
         b.wait_in_text("#vm-subVmTest1-network-1-ipv4-address", "192.168.122.")
+        b.wait_visible("#vm-subVmTest1-network-1-ipv6-address")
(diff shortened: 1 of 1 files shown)


martinpitt:
Thanks! Do you have a link to an example failure?

mvollmer:
> Thanks! Do you have a link to an example failure?

Here: https://cockpit-logs.us-east-1.linodeobjects.com/pull-1601-922588b7-20240506-093822-fedora-40/log.html

mvollmer:
(The reference image doesn't have ipv6 in it, so we need to update it. That's why this is Draft.)

mvollmer:
Huh?  Does the IPv6 address actually only show up with `os="linux2022"`? That doesn't make sense to me right now, but...

mvollmer:
> Huh? Does the IPv6 address actually only show up with `os="linux2022"`? That doesn't make sense to me right now, but...

Ah, of course, missing change notifications. We have to wait until there is IPv6 address before switching to the page.

mvollmer:
> > Huh? Does the IPv6 address actually only show up with `os="linux2022"`? That doesn't make sense to me right now, but...
> 
> Ah, of course, missing change notifications. We have to wait until there is a IPv6 address before switching to the page.

Ok, the IPv6 link-local address gets reported by the guest agent, and the agent indeed only gets connected by default with linux2022. So, we only need to do this in #1601.

PR #1616, opened 2024-05-07
disks: Display source for disks of type "block"

mvollmer:
They have an entry like `<source dev="..." />`.

### DIFF
--- a/src/components/vm/disks/vmDiskColumns.jsx
+++ b/src/components/vm/disks/vmDiskColumns.jsx
@@ -39,6 +39,7 @@ const _ = cockpit.gettext;
 
 export const DISK_SOURCE_LIST = [
     { name: "file", label: _("File") },
+    { name: "dev", label: _("Block device") },
     { name: "device", label: _("Device") },
     { name: "protocol", label: _("Protocol") },
     { name: "pool", label: _("Pool") },
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -1667,6 +1667,28 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
                 force=True,
             ).execute()
 
+    def testBlockDeviceDisk(self):
+        b = self.browser
+        m = self.machine
+
+        loop_dev = self.add_loopback_disk()
+
+        self.createVm("subVmTest1", running=False)
+        m.execute(f"virt-xml subVmTest1 --add-device --disk source.type=block,source.dev={loop_dev}")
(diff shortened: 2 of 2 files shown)


martinpitt (dismissed):
Test s'il te plaît ? This doesn't seem to be caught by any pixel test either.

mvollmer:
> Test s'il te plaît ? This doesn't seem to be caught by any pixel test either.

Added!

martinpitt:
Thanks! This new test [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1616-79ef972d-20240508-114144-debian-stable/log.html#84) on all Debian/Ubuntu due to AppArmor not allowing access to /dev/loop0. But as it doesn't seem to break the VM startup on the outside (Just that supposedly the guest can't actually access it), I suggest to simply allow that message. There's plenty of precedent like
```
self.allow_journal_messages(f'.* type=1400 .* apparmor="DENIED" operation="open" profile="libvirt.* name="{self.vm_tmpdir}.*')
```

src/components/vm/disks/vmDiskColumns.jsx
```diff
@@ -39,6 +39,7 @@ const _ = cockpit.gettext;
 
 export const DISK_SOURCE_LIST = [
     { name: "file", label: _("File") },
+    { name: "dev", label: _("Block device") },
```
allisonkarlitskaya, line 42: The commit message says "type block", but this says "dev"...  I guess that's what the XML fragment is trying to explain?

allisonkarlitskaya (approved):

src/components/vm/disks/vmDiskColumns.jsx
mvollmer: Yes, exactly. This is how the code works in general.  A disk XML might look like this:

```
<disk type='file' device='disk'>
      <driver name='qemu' type='qcow2' discard='unmap'/>
      <source file='/var/lib/libvirt/images/foo.qcow2'/>
      <target dev='vda' bus='virtio'/>
      <address type='pci' domain='0x0000' bus='0x04' slot='0x00' function='0x0'/>
</disk>
```

Cockpit ignores the `type` attribute of `<disk>` and only looks at the `<source>` element. If that has a `file` attribute, for example, the "Source" column in the table gets a entry with "File  /var/lib/libvirt/images/foo.qcow2".

This PR extends this to also recognize `<source>` elements with `dev` attributes. They only happen with `<disk>` elements with `type="block"`.  I hope. :)

PR #1617, opened 2024-05-07
disks: Don't take defaults from existing volumes when creating a new one

mvollmer:
Specifically, don't declare the new device for a new volume to be a CDROM when the first volume in the selected pool happens to be a ISO.

### DIFF
--- a/src/components/vm/disks/diskAdd.jsx
+++ b/src/components/vm/disks/diskAdd.jsx
@@ -397,7 +397,7 @@ export const AddDiskModalBody = ({ disk, idPrefix, isMediaInsertion, vm, vms, su
 
     const getPoolFormatAndDevice = (pool, volName) => {
         const params = { format: getDefaultVolumeFormat(pool), device: "disk" };
-        if (['dir', 'fs', 'netfs', 'gluster', 'vstorage'].indexOf(pool.type) > -1) {
+        if (volName && ['dir', 'fs', 'netfs', 'gluster', 'vstorage'].indexOf(pool.type) > -1) {
             const volume = pool.volumes.find(vol => vol.name === volName);
             if (volume?.format) {
                 params.format = volume.format;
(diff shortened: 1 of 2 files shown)


mvollmer:
This was hella confusing. I have a Alpine ISO in my default pool, which meant that each newly created volume in that pool would be added as a CDROM to the machine. I didn't get to use a virtio bus, and the image format was always raw. Very magic.  I blame `useEffect`.

mvollmer:
Yes, yes, a test would be nice... :-)

martinpitt (dismissed):
Thanks!

mvollmer:
> Yes, yes, a test would be nice... :-)

Added, and it found a bug. :-)

martinpitt (dismissed):
Hah, nice!

martinpitt:
Rebased, to get a new run after re-dropping cloud-init from alpine. The previous run was way too red.

martinpitt:
This [breaks testAddDiskDirPool everywhere](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1617-1921153b-20240508-100425-arch/log.html#80), I'm afraid

mvollmer:
> This [breaks testAddDiskDirPool everywhere](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1617-1921153b-20240508-100425-arch/log.html#80), I'm afraid

That was the switch to q35 by default... now cdroms can't be hotplugged anymore apparently (except on rhel-9???). I rearrange the tests a bit.

src/components/vm/disks/diskAdd.jsx
```diff
@@ -397,7 +397,7 @@ export const AddDiskModalBody = ({ disk, idPrefix, isMediaInsertion, vm, vms, su
 
     const getPoolFormatAndDevice = (pool, volName) => {
         const params = { format: getDefaultVolumeFormat(pool), device: "disk" };
-        if (['dir', 'fs', 'netfs', 'gluster', 'vstorage'].indexOf(pool.type) > -1) {
+        if (volName && ['dir', 'fs', 'netfs', 'gluster', 'vstorage'].indexOf(pool.type) > -1) {
```
martinpitt, line 400: To make sure I understand: Is this functionally required to fix this bug, or just an unrelated optimization? I.e. are there pool volumes with undefined names which would unintentionally match here?

martinpitt (approved):
Danke!

PR #1618, opened 2024-05-07
test: Also attach console output of test VMs on failure

mvollmer:
I hope to get some insights into the frequent failures to shut off a vm, like here: https://cockpit-logs.us-east-1.linodeobjects.com/pull-1616-01ca2ff1-20240507-110227-fedora-40-devel/log.html#40

Alpine will ignore shut-off requests while it is still booting, but the tests are (mostly) careful to wait for that to be over before shutting it down again. But let's see.


martinpitt (approved):
Oh, that's nice!

PR #1619, opened 2024-05-07
test: Check creating and booting an EFI guest

martinpitt:
Use our new "alpine-efi" bots image as guest.

This isn't very comprehensive yet, mostly to validate that the alpine-efi image works, and that c-machines/libvirt can *actually* boot an EFI guest. We may want to extend this to more tests in the future, so that we'll get better ARM coverage. But that at least sets a baseline.

---

Requires https://github.com/cockpit-project/bots/pull/6351 . Also, the first image I tried this on was fedora-40, and it [failed right away](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1619-039a1ed8-20240507-145041-fedora-40-bots%236351/log.html#74). Turns out it's *drumroll* [SELinux](https://github.com/cockpit-project/bots/issues/5944), so this will need a naughty. That's added to the bots PR now.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -73,7 +73,7 @@ Rules configuration can be found in the `.stylelintrc.json` file.
 # Running tests locally
 
 Run `make vm` to build an RPM and install it into a standard Cockpit test VM.
-This will be `fedora-39` by default. You can set `$TEST_OS` to use a different
+This will be `fedora-40` by default. You can set `$TEST_OS` to use a different
 image, for example
 
     TEST_OS=centos-9-stream make vm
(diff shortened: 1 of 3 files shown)


martinpitt:
Why am I not surprised.. It doesn't boot:

image

mvollmer:
The alpine-efi image is not a ISO and can't be put into a cdrom, I think.

mvollmer:
Alright, using a i440fx machine and using "alpine-efi" as a "Cloud base image" works. 

I couldn't get to boot the official Alpine ISO with UEFI on neither q35 nor i440fx.. If that's what progress looks like, I don't want to be part of it...

mvollmer:
> Alright, using a i440fx machine and using "alpine-efi" as a "Cloud base image" works.

I have pushed a commit that implements this in the test. The hardest part was to get the dialog to give us a i440fx machine...

mvollmer:
> it gets past EFI and grub, and now just hangs on an empty screen

That is probably because the console is not on VNC but on ttyS0.

martinpitt:
@mvollmer Thanks for your investigations! TBH I'm not really that much interested in continuing to work on this -- it was a half-hearted attempt towards https://issues.redhat.com/browse/COCKPIT-970 , but this needs a *lot* more libvirt/qemu/images understanding and work than I currently have..

martinpitt:
No time for this, sorry.

PR #1626, opened 2024-05-09
Fix few french translations

arthurlm:
Few french translations were wrong.
I fix them to make application more consistent with common french software.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -96,7 +96,7 @@ msgid ""
 "shut off. Meanwhile, the origin host keeps its copy of the VM configuration."
 msgstr ""
 "Une copie de la machine virtuelle s’exécutera sur la destination et "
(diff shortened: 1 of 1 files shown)


martinpitt:
Merci beaucoup ! Je parle un petit peut français, et ça me semble bien.

Translations are not done via PRs, but on weblate: https://translate.fedoraproject.org/projects/cockpit-machines/main/fr/

I applied your changes there.

arthurlm:
Thanks a lot 😄!

I check the applied translation and there is only one small issue I see:
You have set "Source volume group" to "Renommer le groupe" (which will be closer now as "Rename group").

To be completely honest with you, I start reworking this file because of this specific translation.
When I first used cockpit-machine with LVM pool, I have to check the source code since I was not able to understand what this label was about 😅 ...

Can you also change this label please 🙏 ?

martinpitt:
Whoops, thanks -- I either forgot that one or didn't save correctly. Done!

PR #1629, opened 2024-05-10
test: Wait longer for Alpine to shut down

mvollmer:
It seems to be slower than Cirros.

Example failure: https://cockpit-logs.us-east-1.linodeobjects.com/pull-1616-e0c69eeb-20240510-071702-fedora-40-devel/log.html#30-2

The console log shows that shutdown has in fact started: https://cockpit-logs.us-east-1.linodeobjects.com/pull-1616-e0c69eeb-20240510-071702-fedora-40-devel/TestMachinesLifecycle-testBasic-fedora-40-127.0.0.2-2201-subVmTest1-console.log

Thus, waiting longer should indeed work.


martinpitt (approved):
Yeah, this is all emulated QEMU. I'm having fun with achingly slow guests in #1619 as well.. Kiitos!

PR #1644, opened 2024-05-20
vm: Support for VM descriptions

mvollmer:
Based on work by @Britz, thanks a lot!

This is also a proving ground for robust and convenient virt-xml invocations. I plan to renovate the rest of domain.js and other files accordingly.

--------------------------------

### Machines: Support for VM descriptions

Cockpit can now show and change the descriptions of virtual machines.

--------------------------------

Details page with description:

image

The description is a `<p>` so that it doesn't get too shouty when people put a lot of text there.

"Edit description" action:

image

"Edit description" dialog:

image

### DIFF
--- a/src/components/vm/vmDetailsPage.jsx
+++ b/src/components/vm/vmDetailsPage.jsx
@@ -75,6 +75,12 @@ export const VmDetailsPage = ({
                 <VmNeedsShutdown vm={vm} />
                 <VmUsesSpice vm={vm} />
             </div>
+            {
+                vm.inactiveXML.description &&
+                    <div className="vm-description">
+                        {vm.inactiveXML.description.split("\n").map((p, i) => <p key={i}>{p}</p>)}
+                    </div>
+            }
         </PageSection>
     );
 
--- a/src/components/vm/vmEditDescriptionDialog.jsx
+++ b/src/components/vm/vmEditDescriptionDialog.jsx
@@ -1,0 +1,82 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import cockpit from 'cockpit';
+import React, { useState } from 'react';
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
+import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form";
+import { Modal } from "@patternfly/react-core/dist/esm/components/Modal";
+import { TextArea } from "@patternfly/react-core/dist/esm/components/TextArea";
+
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { useDialogs } from 'dialogs.jsx';
+
+import { isObjectEmpty } from '../../helpers.js';
+import { domainSetDescription } from '../../libvirtApi/domain.js';
+
+const _ = cockpit.gettext;
+
+export const EditDescriptionDialog = ({ vm }) => {
+    const Dialogs = useDialogs();
+    const [description, setDescription] = useState(vm.inactiveXML.description || "");
+    const [error, dialogErrorSet] = useState({});
+
+    async function onSubmit() {
+        try {
+            await domainSetDescription(vm, description);
+            Dialogs.close();
+        } catch (exc) {
+            dialogErrorSet({
+                dialogError: cockpit.format(_("Failed to set description of VM $0"), vm.name),
+                dialogErrorDetail: exc.message
+            });
+        }
+    }
+
+    return (
+        <Modal position="top" variant="small" isOpen onClose={Dialogs.close}
+               title={cockpit.format(_("Edit description of VM $0"), vm.name)}
+               footer={
+                   <>
+                       <Button variant='primary'
+                               id="edit-description-dialog-confirm"
+                               onClick={onSubmit}>
+                           {_("Save")}
+                       </Button>
+                       <Button variant='link' onClick={Dialogs.close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }>
+            <Form onSubmit={e => {
+                e.preventDefault();
+                onSubmit();
+            }}
+            isHorizontal>
+                {!isObjectEmpty(error) && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
+                <FormGroup label={_("Description")}
+                           fieldId="edit-description-dialog-description">
+                    <TextArea id='edit-description-dialog-description'
+                               value={description}
+                               onChange={(_, value) => setDescription(value)} />
+                </FormGroup>
+            </Form>
+        </Modal>
+    );
+};
--- a/src/libvirt-xml-parse.js
+++ b/src/libvirt-xml-parse.js
@@ -245,6 +245,8 @@ export function parseDomainDumpxml(connectionName, domXml, objPath) {
     const metadataElem = getSingleOptionalElem(domainElem, "metadata");
 
     const name = domainElem.getElementsByTagName("name")[0].childNodes[0].nodeValue;
+    const uuid = domainElem.getElementsByTagName("uuid")[0].childNodes[0].nodeValue;
+    const description = domainElem.getElementsByTagName("description")[0]?.childNodes[0]?.nodeValue;
     const id = objPath;
     const osType = osTypeElem.childNodes[0].nodeValue;
     const osBoot = parseDumpxmlForOsBoot(osBootElems);
@@ -292,7 +294,9 @@ export function parseDomainDumpxml(connectionName, domXml, objPath) {
 
     return {
         connectionName,
+        uuid,
         name,
+        description,
         id,
         osType,
         osBoot,
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -888,6 +888,33 @@ export function domainSendNMI({
     return call(connectionName, objPath, 'org.libvirt.Domain', 'InjectNMI', [0], { timeout, type: 'u' });
 }
 
+function shlex_quote(str) {
+    // yay, command line apis...
+    return "'" + str.replaceAll("'", "'\"'\"'") + "'";
+}
+
+async function domainSetXML(vm, option, values) {
+    const opts = { err: "message" };
+    if (vm.connectionName === 'system')
+        opts.superuser = 'try';
+
+    // We don't pass the arguments for virt-xml through a shell, but
+    // virt-xml does its own parsing with the Python shlex module. So
+    // we need to do the equivalent of shlex.quote here.
+
+    const args = [];
+    for (const key in values)
+        args.push(shlex_quote(key + '=' + values[key]));
+
+    await cockpit.spawn([
+        'virt-xml', '-c', `qemu:///${vm.connectionName}`, '--' + option, args.join(','), vm.uuid, '--edit'
+    ], opts);
+}
+
+export async function domainSetDescription(vm, description) {
+    await domainSetXML(vm, "metadata", { description });
+}
+
 export function domainSetCpuMode({
     name,
     connectionName,
--- a/test/check-machines-lifecycle
+++ b/test/check-machines-lifecycle
@@ -367,6 +367,27 @@ class TestMachinesLifecycle(machineslib.VirtualMachinesCase):
         b.wait_text("h2.vm-name", "test%")
         b.wait_not_present("#navbar-oops")
 
+    def testEditDescription(self):
+        b = self.browser
+
+        self.createVm("mac", running=False)
+        self.login_and_go("/machines")
+        self.waitPageInit()
+        self.goToVmPage("mac")
+
+        self.performAction("mac", "edit-description")
+
+        # Non-ascii chars, unbalanced quotes, backslash, multiple lines
+        desc1 = '"Döscrü\\ptiän \'\'\' كرة القدم'
+        desc2 = 'Second <b>line</b>'
+
+        b.set_input_text("#edit-description-dialog-description", desc1 + "\n" + desc2)
+        b.click("#edit-description-dialog-confirm")
+        b.wait_not_present("#edit-description-dialog-confirm")
+
+        b.wait_text(".vm-description p:nth-child(1)", desc1)
+        b.wait_text(".vm-description p:nth-child(2)", desc2)
+
     def testDelete(self):
         b = self.browser
         m = self.machine
--- a/src/components/vm/vmActions.jsx
+++ b/src/components/vm/vmActions.jsx
@@ -38,6 +38,7 @@ import { ConfirmDialog } from './confirmDialog.jsx';
 import { DeleteDialog } from "./deleteDialog.jsx";
 import { MigrateDialog } from './vmMigrateDialog.jsx';
 import { RenameDialog } from './vmRenameDialog.jsx';
+import { EditDescriptionDialog } from './vmEditDescriptionDialog.jsx';
 import { ReplaceSpiceDialog } from './vmReplaceSpiceDialog.jsx';
 import {
     domainCanDelete,
@@ -451,8 +452,20 @@ const VmActions = ({ vm, vms, onAddErrorNotification, isDetailsPage }) => {
                 {_("Rename")}
             </DropdownItem>
         );
-        dropdownItems.push(<Divider key="separator-rename" />);
     }
+
+    if (isDetailsPage) {
+        dropdownItems.push(
+            <DropdownItem key={`${id}-edit-description`}
+                          id={`${id}-edit-description`}
+                          onClick={() => Dialogs.show(<EditDescriptionDialog vm={vm} />)}>
+                {_("Edit description")}
+            </DropdownItem>
+        );
+    }
+
+    if (domainCanRename(state) || isDetailsPage)
+        dropdownItems.push(<Divider key="separator-rename" />);
 
     if (state !== undefined && domainCanDelete(state, vm.id)) {
         if (!vm.persistent) {


src/components/vm/vmDetailsPage.jsx
```diff
@@ -67,14 +67,15 @@ export const VmDetailsPage = ({
     const vmActionsPageSection = (
         <PageSection className="actions-pagesection" variant={PageSectionVariants.light} isWidthLimited>
             <div className="vm-top-panel" data-vm-transient={!vm.persistent}>
-                <h2 className="vm-name">{vm.name}</h2>
+                <h2 className="vm-name">{vm.title ? vm.title + " (" + vm.name + ")" : vm.name}</h2>
```
cockpituous, line 70: This added line is not executed by any test.

src/components/vm/vmDetailsPage.jsx
```diff
@@ -67,14 +67,15 @@ export const VmDetailsPage = ({
...
                 <VmActions vm={vm}
                            config={config}
                            onAddErrorNotification={onAddErrorNotification}
                            isDetailsPage />
                 <VmNeedsShutdown vm={vm} />
                 <VmUsesSpice vm={vm} />
             </div>
+            { vm.description && <p>{vm.description}</p> }
```
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -23,47 +23,59 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
...
-                        cockpit.location.go(["vm"], { ...cockpit.location.options, name: newName, connection: connectionName });
-                }, exc => {
-                    dialogErrorSet({ dialogError: cockpit.format(_("Failed to rename VM $0"), vmName), dialogErrorDetail: exc.message });
-                });
+        try {
+            await domainSetTitle({ vm, title: fullName });
+            await domainSetDescription({ vm, description });
+            if (name != vm.name)
+                await domainRename({ connectionName, id: vm.id, newName: name });
+            Dialogs.close();
```
cockpituous, line 56: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -23,47 +23,59 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
...
+
+            // If we are on the VMs details page change the URL to
+            // reflect the new name after the rename operation
+            // succeeded
+            if (name != vm.name && cockpit.location.path.length > 0)
+                cockpit.location.go(["vm"], { ...cockpit.location.options, name, connection: connectionName });
```
cockpituous, line 62: These 2 added lines are not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +90,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
                 {!isObjectEmpty(error) && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
-                <FormGroup label={_("New name")}
+                <FormGroup label={_("Name")}
                            fieldId="rename-dialog-new-name">
                     <TextInput id='rename-dialog-new-name'
-                               validated={submitted && !newName ? "error" : "default"}
-                               value={newName}
-                               onChange={(_, value) => setNewName(value)} />
-                    <FormHelper helperTextInvalid={submitted && !newName && _("New name must not be empty")} />
+                               validated={submitted && !name ? "error" : "default"}
```
cockpituous, line 96: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +90,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
+                               value={name}
+                               onChange={(_, value) => setName(value)} />
+                    <FormHelper helperTextInvalid={submitted && !name && _("Name must not be empty")} />
```
cockpituous, line 99: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +90,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
+                </FormGroup>
+                <FormGroup label={_("Full name")}
+                           fieldId="rename-dialog-full-name">
+                    <TextInput id='rename-dialog-full-name'
+                               value={fullName}
+                               onChange={(_, value) => setFullName(value)} />
```
cockpituous, line 105: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +90,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
+                </FormGroup>
+                <FormGroup label={_("Description")}
+                           fieldId="rename-dialog-description">
+                    <TextArea id='rename-dialog-description'
+                              value={description}
+                              onChange={(_, value) => setDescription(value)} />
```
cockpituous, line 111: This added line is not executed by any test.

src/libvirt-xml-parse.js
```diff
@@ -245,6 +245,8 @@ export function parseDomainDumpxml(connectionName, domXml, objPath) {
     const metadataElem = getSingleOptionalElem(domainElem, "metadata");
 
     const name = domainElem.getElementsByTagName("name")[0].childNodes[0].nodeValue;
+    const title = domainElem.getElementsByTagName("title")[0]?.childNodes[0]?.nodeValue;
+    const description = domainElem.getElementsByTagName("description")[0]?.childNodes[0]?.nodeValue;
```
cockpituous, line 249: These 2 added lines are not executed by any test.

src/components/vm/vmDetailsPage.jsx
cockpituous, line 70: This added line is not executed by any test.

src/components/vm/vmDetailsPage.jsx
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -23,47 +23,58 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
...
-                    if (cockpit.location.path.length > 0)
-                        cockpit.location.go(["vm"], { ...cockpit.location.options, name: newName, connection: connectionName });
-                }, exc => {
-                    dialogErrorSet({ dialogError: cockpit.format(_("Failed to rename VM $0"), vmName), dialogErrorDetail: exc.message });
-                });
+        try {
+            await domainSetMetadata(vm, { title: fullName, description });
+            if (name != vm.name)
+                await domainRename({ connectionName, id: vm.id, newName: name });
+            Dialogs.close();
```
cockpituous, line 55: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -23,47 +23,58 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
...
+
+            // If we are on the VMs details page change the URL to
+            // reflect the new name after the rename operation
+            // succeeded
+            if (name != vm.name && cockpit.location.path.length > 0)
+                cockpit.location.go(["vm"], { ...cockpit.location.options, name, connection: connectionName });
```
cockpituous, line 61: These 2 added lines are not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +89,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
                 {!isObjectEmpty(error) && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
-                <FormGroup label={_("New name")}
+                <FormGroup label={_("Name")}
                            fieldId="rename-dialog-new-name">
                     <TextInput id='rename-dialog-new-name'
-                               validated={submitted && !newName ? "error" : "default"}
-                               value={newName}
-                               onChange={(_, value) => setNewName(value)} />
-                    <FormHelper helperTextInvalid={submitted && !newName && _("New name must not be empty")} />
+                               validated={submitted && !name ? "error" : "default"}
```
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +89,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
+                               value={name}
+                               onChange={(_, value) => setName(value)} />
+                    <FormHelper helperTextInvalid={submitted && !name && _("Name must not be empty")} />
```
cockpituous, line 98: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +89,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
+                </FormGroup>
+                <FormGroup label={_("Full name")}
+                           fieldId="rename-dialog-full-name">
+                    <TextInput id='rename-dialog-full-name'
+                               value={fullName}
+                               onChange={(_, value) => setFullName(value)} />
```
cockpituous, line 104: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +89,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
+                </FormGroup>
+                <FormGroup label={_("Description")}
+                           fieldId="rename-dialog-description">
+                    <TextArea id='rename-dialog-description'
+                              value={description}
+                              onChange={(_, value) => setDescription(value)} />
```
cockpituous, line 110: This added line is not executed by any test.

src/libvirt-xml-parse.js
cockpituous, line 249: These 2 added lines are not executed by any test.

src/components/vm/vmDetailsPage.jsx
cockpituous, line 70: This added line is not executed by any test.

src/components/vm/vmDetailsPage.jsx
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -23,47 +23,58 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
...
+
+            // If we are on the VMs details page change the URL to
+            // reflect the new name after the rename operation
+            // succeeded
+            if (name != vm.name && cockpit.location.path.length > 0)
+                cockpit.location.go(["vm"], { ...cockpit.location.options, name, connection: vm.connectionName });
+        } catch (exc) {
+            dialogErrorSet({
+                dialogError: cockpit.format(_("Failed to rename VM $0"), vm.name),
+                dialogErrorDetail: exc.message
```
cockpituous, line 65: These 5 added lines are not executed by any test.

src/components/vm/vmRenameDialog.jsx
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
cockpituous, line 98: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
cockpituous, line 104: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
cockpituous, line 110: This added line is not executed by any test.

src/libvirt-xml-parse.js
cockpituous, line 249: These 2 added lines are not executed by any test.

src/components/vm/vmRenameDialog.jsx
cockpituous, line 65: These 4 added lines are not executed by any test.

src/components/vm/vmRenameDialog.jsx
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/vmRenameDialog.jsx
cockpituous, line 98: This added line is not executed by any test.

martinpitt:
[This looks a bit sad](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1644-d9dc7b92-20240522-052942-debian-stable/log.html#60-2), retrying for comparison.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -23,47 +23,57 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
...
 import { useDialogs } from 'dialogs.jsx';
 
 import { isObjectEmpty } from '../../helpers.js';
-import { domainRename } from '../../libvirtApi/domain.js';
+import { domainRename, domainSetMetadata } from '../../libvirtApi/domain.js';
 
 const _ = cockpit.gettext;
 
-export const RenameDialog = ({ vmName, vmId, connectionName }) => {
+export const RenameDialog = ({ vm }) => {
```
martinpitt, line 37: Note: This is convenient, but `vm` objects have a ton of properties and change all the time (like runtime metrics). This will re-render the dialog a lot.

src/libvirtApi/domain.js
```diff
@@ -890,6 +890,24 @@ export function domainSendNMI({
...
 }
 
+async function domainSetXML(vm, option, values) {
+    const opts = { err: "message" };
+    if (vm.connectionName === 'system')
+        opts.superuser = 'try';
+
+    const args = [];
+    for (const key in values)
+        args.push(key + '="' + values[key].replaceAll('"', '\\"') + '"');
```
martinpitt, line 900: There can be any arbitrary junk in the description, including more quotes, backspaces, etc. I think this isn't run through shell, so it shouldn't be that bad, but at the same time this means that we can't use `shlex.quote()` or similar. How much can we trust virt-xml's command line parser to not mess this up?

test/check-machines-lifecycle
```diff
@@ -349,22 +349,65 @@ class TestMachinesLifecycle(machineslib.VirtualMachinesCase):
...
+        b.set_input_text("#rename-dialog-new-name", "/")
+        b.click("#rename-dialog-confirm")
+        b.wait_in_text(".pf-v5-c-alert__description", "name / cannot contain '/'")
+
+        # Set full name and description and use non-ASCII characters
+
+        b.set_input_text("#rename-dialog-new-name", 'test%')
+        # This is an attempted injection for the virt-xml invocation
+        b.set_input_text("#rename-dialog-full-name", '"full name",blah=yo')
+        b.set_input_text("#rename-dialog-description", 'Döscrüptiän كرة القدم')
```
martinpitt, line 372: It would be good to test a multi-line description which also contains a `"` and a `'` to make sure that this goes through virt-xml/xml and back correctly.

src/components/vm/vmRenameDialog.jsx
```diff
@@ -78,13 +88,25 @@ export const RenameDialog = ({ vmName, vmId, connectionName }) => {
...
-                               validated={submitted && !newName ? "error" : "default"}
-                               value={newName}
-                               onChange={(_, value) => setNewName(value)} />
-                    <FormHelper helperTextInvalid={submitted && !newName && _("New name must not be empty")} />
+                               validated={submitted && !name ? "error" : "default"}
+                               value={name}
+                               onChange={(_, value) => setName(value)} />
+                    <FormHelper helperTextInvalid={submitted && !name && _("Name must not be empty")} />
+                </FormGroup>
+                <FormGroup label={_("Full name")}
```
martinpitt, line 99: I wonder if this column can become dynamic? I.e. only show that if at least one VM actually *has* a full name? We haven't gotten any request about this for many years until recently, so I don't think a lot of users will bother with giving names/descriptions. 

It also shows in the updated pixel tests: they all have empty full names, and there isn't a new pixel test which demonstrates the name or description.

martinpitt (changes requested):
Thanks! Depending on how much time you can/want to throw at this, I can be convinced to give a +1. But I'm not super happy about this just yet.

src/components/vm/vmRenameDialog.jsx
mvollmer: Does it? Components get rendered when the parent is rendered or when their state changes, not when a field of one of their props changes.

mvollmer: Any change to the redux store renders all of cockpit-machines, which is maybe not great, but this change here doesn't make that worse, I'd say.

martinpitt: > Components get rendered when the parent is rendered or when their state changes

.. or when their properties change (which is one of the two main functions of React).

> Any change to the redux store renders all of cockpit-machines

Right, fair enough.

src/libvirtApi/domain.js
mvollmer: Good catch, we need to also escape "\", but then we should be good... I'll check the virt-xml sources.

src/components/vm/vmRenameDialog.jsx
mvollmer: > I wonder if this column can become dynamic? 

Yes, I was thinking the same.

src/libvirtApi/domain.js
mvollmer: For reference: https://github.com/virt-manager/virt-manager/blob/main/virtinst/cli.py#L1192

test/check-machines-lifecycle
mvollmer: Yes.

mvollmer: Done.

src/components/vm/vmRenameDialog.jsx
mvollmer: Done.

src/libvirtApi/domain.js
mvollmer: Fixed.

martinpitt:
Yay for the [new pixel test](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1644-feb47bfe-20240522-141019-fedora-40/TestMachinesLifecycle-testRename-full-name-table-pixels.png)!

martinpitt:
Very cool, thanks! Looks good, just needs the new pixel test mopped up.

src/components/vm/vmRenameDialog.jsx
mvollmer: > > Components get rendered when the parent is rendered or when their state changes
> 
> .. or when their properties change (which is one of the two main functions of React).

Hmm. Let me take this as an opportunity to figure this out for myself...

My mental model is this: Properties change as a result of rendering. That's the only way they can change. Property changes do not cause rendering. All rendering is triggered by a change in _state_ in some component, and then the whole tree starting at that component is re-rendered.

Components in the tree can optionally stop their re-rendering with `React.memo` (or by deriving from React.PureComponent), but that is not the default and we don't use this here.

And because c-machines uses Redux, comparing objects with Object.is is correct and wont miss changes. So I guess all our components could use React.memo by default... Is there a way to switch that on? Or maybe it is already?

In any case, I don't see how changing from name/connection to store object would change how often the dialog is rendered.

https://www.debugbear.com/blog/react-rerenders

(There might be other rules where React itself by default decides to stop rendering at a node in the subtree, but I haven't found anything concrete about this, unfortunately. React is surprisingly wishy-washy about what the rules for rendering actually are... But it doesn't compare property values, I am pretty sure. We rely on that in many many places. There is some magic with updating the state... this.setState({}) used to reliably trigger a render, but it's harder with the `useState` hook...)

martinpitt (approved):
Ah, I was about to mop up the pixel update, but you already did it -- I didn't see the force-push here. Force-pushed a rebase, and re-running the tests to check the flakes.

LGTM, assuming green. Thanks!

martinpitt:
This still needs a proper release note, but otherwise looks good to land. Thanks!

garrett (dismissed):
I thought this was already discussed before and we specifically decided to not do it this way. (Note: it was different people working on cockpit-machines at that point in time.)

A VM is not a person with a first, middle, and last names. It does not have a "Full name".

If we want to get around the limitation for the name that serves as a type of identifier (ID) for the VM, then we should have "Name" be a display name and generate the machine's unique and limited (in characters) name based on the display name. I believe that's the intent for this, to work around an implementation detail where the name itself is restricted to various limits (mainly, which characters it can use).

It does not makes sense to have yet another field to do the exact same thing. We already have name for a way to identify the machine and description to explain what it does. Having both will be confusing and also take up extra space everywhere where there's a name, for no good reason.

---

So, instead of having the split here's how we might handle it instead:

Completely move the short ID name to the background, have it based on the display name, and always show the display name (treating the non-display name as an implementation detail). We could always set both the display name and ID name _or_ only set the display if it doesn't fit within the ID's restrictions.

When displaying VMs, we'd do something like "display name || ID name" where it would default to showing the display name (if it exists) and falling back to the ID name if a display name is not shown. And we could even do something where we transform the display name to the ID name and if they match, only show display name, else show "Display name (ID name)". I think we're already doing this (if it already exists) without a way to set it, right?

mvollmer:
> only retain the optional description?

Makes sense to me.

mvollmer:
@garrett's design also makes a lot of sense to me. :-)

mvollmer:
> > only retain the optional description?
> 
> Makes sense to me.

But does it then make sense to use the Rename dialog to set the description?

martinpitt:
Good point -- dang, looks like this is back to square one wrt. design then? :cry:

martinpitt:
That said, while "Rename" is not very discoverable, it saves the need for yet another kebab menu entry and dialog. Perhaps changing it to "Rename/Describe"?

garrett:
> That said, while "Rename" is not very discoverable, it saves the need for yet another kebab menu entry and dialog. Perhaps changing it to "Rename/Describe"?

To have both in one string, we'd probably want something like "Edit name and description" instead.

garrett:
To handle descriptions in the table, we can either have a column that shows up when there is a description in any VM (and not there if there is none), like this:

Board

Or we can make it optional but in the name field, as some subtext, like this:

Table _ Table (grid)

As horizontal space is usually a premium and it doesn't make sense to be able to sort on the description (the table should be sortable, but isn't — however, this is a different topic), it probably doesn't make sense to have it in its own column. Adding it below is probably the best option here.

What do you think?

martinpitt:
Agreed about not showing it in a separate column -- the PR also doesn't do that. It's currently shown in the VM details page (see screenshot in description). TBH I'd leave it out from the overview list -- it can be arbitrarily long (it's multi-line), and is a fringe case. I could imagine an (i) icon with a popup in the description though, that doesn't break the layout.

garrett:
Sure, keeping it the description in the details page only can make sense... but then we shouldn't have an action to change it on the list. That's an either/or thing. We already have a way to edit it on the details page — where you can see it.

So I guess that means there's nothing left of this PR we want to go in, and just want to keep the status quo? (Unless I'm missing something.)

src/components/vm/vmEditDescriptionDialog.jsx
```diff
@@ -0,0 +1,82 @@
...
+    const [error, dialogErrorSet] = useState({});
+
+    async function onSubmit() {
+        try {
+            await domainSetDescription(vm, description);
+            Dialogs.close();
+        } catch (exc) {
+            dialogErrorSet({
+                dialogError: cockpit.format(_("Failed to set description of VM $0"), vm.name),
+                dialogErrorDetail: exc.message
```
cockpituous, line 47: These 4 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
```diff
@@ -0,0 +1,82 @@
...
+                           {_("Save")}
+                       </Button>
+                       <Button variant='link' onClick={Dialogs.close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }>
+            <Form onSubmit={e => {
+                e.preventDefault();
+                onSubmit();
```
cockpituous, line 69: These 3 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
```diff
@@ -0,0 +1,82 @@
...
+            }}
+            isHorizontal>
+                {!isObjectEmpty(error) && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
```
cockpituous, line 72: This added line is not executed by any test.

mvollmer:
`virt-xml --metadata --update` doesn't work and thus we can't update the description of a running VM. But that seems unmotivated, it's just some text without connection to anything involved in actually running the VM.

But `virt-cml` can change the "offline", "stale", "persistent", whateverthisiscalled XML definition. We should probably read the description from that, and not from the "online", "running" XML.

mvollmer:
> We should probably read the description from that, and not from the "online", "running" XML.

Done. (It's called the "inactive" definition.)

src/components/vm/vmEditDescriptionDialog.jsx
```diff
@@ -0,0 +1,82 @@
...
+    const [error, dialogErrorSet] = useState({});
+
+    async function onSubmit() {
+        try {
+            await domainSetDescription(vm, description);
+            Dialogs.close();
+        } catch (exc) {
+            dialogErrorSet({
+                dialogError: cockpit.format(_("Failed to set description of VM $0"), vm.name),
+                dialogErrorDetail: exc.message
```
cockpituous, line 47: These 4 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
```diff
@@ -0,0 +1,82 @@
...
+                           {_("Save")}
+                       </Button>
+                       <Button variant='link' onClick={Dialogs.close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }>
+            <Form onSubmit={e => {
+                e.preventDefault();
+                onSubmit();
```
cockpituous, line 69: These 3 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
```diff
@@ -0,0 +1,82 @@
...
+            }}
+            isHorizontal>
+                {!isObjectEmpty(error) && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
```
cockpituous, line 72: This added line is not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 47: These 4 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 69: These 3 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 72: This added line is not executed by any test.

mvollmer:
[This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1644-beaf9ae2-20240716-093223-arch/log.html#88-1) is a new flake... let's see.

mvollmer:
> [This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1644-beaf9ae2-20240716-093223-arch/log.html#88-1) is a new flake... let's see.

One apparently can't ever use `set_val` with React components. Fixed by using set_input_text. It can in fact create multiple lines, but one has to use "\r" instead of "\n".

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 47: These 5 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 69: These 3 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 72: This added line is not executed by any test.

src/libvirtApi/domain.js
```diff
@@ -890,6 +890,33 @@ export function domainSendNMI({
...
+    // virt-xml does its own parsing with the Python shlex module. So
+    // we need to do the equivalent of shlex.quote here.
+
+    const args = [];
+    for (const key in values)
+        args.push(shlex_quote(key + '=' + values[key]));
+
+    await cockpit.spawn([
+        'virt-xml', '-c', `qemu:///${vm.connectionName}`, '--' + option, args.join(','), vm.uuid, '--edit'
+    ], opts);
```
cockpituous, line 913: This added line is not executed by any test.

src/libvirtApi/domain.js
```diff
@@ -890,6 +890,33 @@ export function domainSendNMI({
...
+}
+
+export async function domainSetDescription(vm, description) {
+    await domainSetXML(vm, "metadata", { description });
```
cockpituous, line 917: This added line is not executed by any test.

martinpitt:
@vollmer: Please don't use `\r` and `\n` in `set_input_text()` or `b.input_text()` any more, this isn't portable across browsers. Use `b.key("Enter")` or similar.

mvollmer:
> @vollmer: Please don't use `\r` and `\n` in `set_input_text()` or `b.input_text()` any more, this isn't portable across browsers. Use `b.key("Enter")` or similar.

Hmm, can we make it so that set_input_text("\n") does the right thing? I'll make a PR.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 47: These 4 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 69: These 3 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 72: This added line is not executed by any test.

mvollmer:
> > @vollmer: Please don't use `\r` and `\n` in `set_input_text()` or `b.input_text()` any more, this isn't portable across browsers. Use `b.key("Enter")` or similar.
> 
> Hmm, can we make it so that set_input_text("\n") does the right thing? I'll make a PR.

The test now dioes this:

```
        b.set_input_text("#edit-description-dialog-description", desc1, value_check=False, blur=False)
        b.key("Enter")
        b.set_input_text("#edit-description-dialog-description", desc2, value_check=False, append=True)
        b.wait_val("#edit-description-dialog-description", desc1 + "\n" + desc2)
```

It would be really nice to replace that with

```
        b.set_input_text("#edit-description-dialog-description", desc1 + "\n" + desc2)
```

Not only because it is less to type, but also because this is the natural first thing to try, and it is not at all obvious what to do once you figure out that it doesn't work.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 47: These 4 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 69: These 3 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 72: This added line is not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 47: These 4 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 69: These 3 added lines are not executed by any test.

src/components/vm/vmEditDescriptionDialog.jsx
cockpituous, line 72: This added line is not executed by any test.

mvollmer:
Whoops, dropped the ball on this, sorry.  Let's get this back on track.

based on the discussion above, this PR leaves "title" and "id" alone. It is only about the description now.

The description of a VM is shown on the details page, and there is an action in the menu to edit it.  The description of this PR is up-to-date with the code.

src/libvirtApi/domain.js
```diff
@@ -888,6 +888,33 @@ export function domainSendNMI({
     return call(connectionName, objPath, 'org.libvirt.Domain', 'InjectNMI', [0], { timeout, type: 'u' });
 }
 
+function shlex_quote(str) {
+    // yay, command line apis...
+    return "'" + str.replaceAll("'", "'\"'\"'") + "'";
```
jelly, line 893: What about https://docs.python.org/3/library/shlex.html#shlex.quote ?

test/check-machines-lifecycle
```diff
@@ -367,6 +367,27 @@ class TestMachinesLifecycle(machineslib.VirtualMachinesCase):
...
+        self.createVm("mac", running=False)
+        self.login_and_go("/machines")
+        self.waitPageInit()
+        self.goToVmPage("mac")
+
+        self.performAction("mac", "edit-description")
+
+        # Non-ascii chars, unbalanced quotes, backslash, multiple lines
+        desc1 = '"Döscrü\\ptiän \'\'\' كرة القدم'
+        desc2 = 'Second <b>line</b>'
```
jelly, line 382: I hope this does not get rendered as such :)

jelly (commented):
One small comment/question, otherwise looks good (!)

src/libvirtApi/domain.js
mvollmer: Yeah, what about? :-) This is supposed to be a JavaScript implementation of it. Is something missing?

test/check-machines-lifecycle
mvollmer: It gets rendered as text, not as HTML.

image

src/libvirtApi/domain.js
jelly: Mistaken this for a Python file, don't ask :)

jelly (approved):
Cool!

PR #1653, opened 2024-05-27
fmf: Drop F40 repo on RHEL 10, re-disable c10s packit test

martinpitt:
https://issues.redhat.com/browse/TFT-2564 currently adds that for
$reasons, and it's all too likely to install/cross-grade to Fedora
packages with `dnf` commands. This currently happens with libvirt
packages.

Remove the Fedora repo early in the test plan, so that our test
dependencies will actually be installed from CentOS, not Fedora.

----

See [recent run](https://artifacts.dev.testing-farm.io/cf6b1fce-a7a6-4118-b5f5-d81434e61b60/). This should at least fix the cross-grade of libvirt to f40, and maybe even the testCreateDownloadAnOS test regression.

### DIFF
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -1,5 +1,14 @@
 discover:
     how: fmf
+
+# counteract https://issues.redhat.com/browse/TFT-2564
+adjust:
+  prepare+:
+  - how: shell
+    script: |
+      rm --force --verbose /etc/yum.repos.d/fedora.repo
+    when: distro == centos-10
+
 execute:
     how: tmt
 
--- a/packit.yaml
+++ b/packit.yaml
@@ -36,7 +36,8 @@ jobs:
     packages: [cockpit-machines-centos]
     targets:
     - centos-stream-9
-    - centos-stream-10
+    # blocked on https://issues.redhat.com/browse/RHEL-29893
+    # - centos-stream-10
 
   - job: tests
     packages: [cockpit-machines-fedora]
@@ -51,7 +52,8 @@ jobs:
     trigger: pull_request
     targets:
       - centos-stream-9
-      - centos-stream-10
+      # blocked on https://issues.redhat.com/browse/RHEL-29893
+      # - centos-stream-10
 
   - job: copr_build
     trigger: release


martinpitt:
[First run](https://artifacts.dev.testing-farm.io/5be1a208-b549-4121-8f28-0791cb4c915a/) shows a truckload of installed f40 packages. Trying a suggestion from Karel.

martinpitt:
No joy.. I [asked for help in jira](https://issues.redhat.com/browse/TFT-2564?focusedId=24800690&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24800690)

martinpitt:
I have to take out the "no Fedora packages" strict check, as the [last run shows a few pre-installed f40 packages](https://artifacts.dev.testing-farm.io/295d95df-8f66-4d37-a0fa-95b44b84f9a6/) which are hard to get rid of. These really need to be sorted out between RHEL and TF. But the "disable repo early" approach now generally works.

martinpitt:
Meh, [what now](https://artifacts.dev.testing-farm.io/b3ed7538-5fd2-4f33-8ea9-491cad8fcf88/)? kernel-modules-extra doesn't contain the iptables modules any more? But probably because the running kernel is 6.9.0-0.rc2.1.el10.x86_64 but that's not what the installed kernel is. Sigh, this is just too broken, I'll re-disable the c10s packit test for now until https://issues.redhat.com/browse/RHEL-29893 is fixed.

plans/all.fmf
```diff
@@ -1,5 +1,14 @@
 discover:
     how: fmf
+
+# counteract https://issues.redhat.com/browse/TFT-2564
```
martinpitt, line 4: Note: With the second commit of re-disabling c10s tests this commit isn't strictly necessary any more. But I'd still like to keep it: It doesn't hurt, it was very non-trivial to figure out, and once the iptables bug gets fixed we'll want to re-enable c10s tests.

martinpitt:
The [rawhide failures](https://artifacts.dev.testing-farm.io/25a05cab-57a2-43fc-b81a-f00fde693249/) are unrelated, have happened for a few days already, and are already on the pilot board.

jelly (approved):

PR #1655, opened 2024-05-28
Add testing support for tumbleweed

Nykseli:
Depends on:
 - [x] https://github.com/cockpit-project/bots/pull/6437
 - [x] https://github.com/cockpit-project/cockpit/pull/20525
 - [x] #1656
 - [x] #1658

### DIFF
--- a/test/check-machines-storage-pools
+++ b/test/check-machines-storage-pools
@@ -429,7 +429,7 @@ class TestMachinesStoragePools(machineslib.VirtualMachinesCase):
         dev = self.add_ram_disk(2)
         m.execute(f"parted {dev} mklabel gpt")
 
-        if m.image == "arch":
+        if m.image == "arch" or m.image == "opensuse-tumbleweed":
             m.execute("mkdir -p /media/")
 
         StoragePoolCreateDialog(
--- a/test/check-machines-storage-pools
+++ b/test/check-machines-storage-pools
@@ -477,7 +477,7 @@ class TestMachinesStoragePools(machineslib.VirtualMachinesCase):
                 b.wait_not_present("#create-storage-pool-dialog")
 
         # Prepare a Volume Group to be used as source for the LVM pool
-        m.execute(f"wipefs -a {dev}; pvcreate {dev}; vgcreate vol_grp1 {dev}")
+        m.execute(f"wipefs -a {dev}; until pvcreate {dev}; do sleep .5; done; vgcreate vol_grp1 {dev}")
         b.go("/machines#/storages")
         b.enter_page("/machines")
 
@@ -698,7 +698,7 @@ class TestMachinesStoragePools(machineslib.VirtualMachinesCase):
         ).execute()
 
         # create disk for lvm-pool
-        m.execute(f"wipefs -a {dev}; pvcreate {dev}; vgcreate vol_grp1 {dev}")
+        m.execute(f"wipefs -a {dev}; until pvcreate {dev}; do sleep .5; done; vgcreate vol_grp1 {dev}")
         m.execute("virsh pool-define-as lvm-pool --type logical --source-name vol_grp1; virsh pool-start lvm-pool")
 
         StorageVolumeCreateDialog(
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -311,6 +311,11 @@ class VirtualMachinesCaseHelpers:
                 f"cp {self.vm_tmpdir}/server.crt /etc/ca-certificates/trust-source/anchors/server.crt",
                 "update-ca-trust"
             ]
+        elif "suse" in self.machine.image:
+            cmds += [
+                f"cp {self.vm_tmpdir}/server.crt /etc/pki/trust/anchors/server.crt",
+                "update-ca-certificates"
+            ]
         else:
             cmds += [
                 f"cp {self.vm_tmpdir}/server.crt /etc/pki/ca-trust/source/anchors/server.crt",
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -348,30 +348,33 @@ class VirtualMachinesCase(testlib.MachineCase, VirtualMachinesCaseHelpers, stora
     def setUp(self):
         super().setUp()
 
-        m = self.machine
+        for m in self.machines:
+            m = self.machines[m]
 
-        # We don't want nested KVM since it doesn't work well enough
-        # three levels deep.
-        #
-        # The first level is the VM allocated to us that runs our
-        # "tasks" container in certain environments, the second level
-        # is the VM started by testlib.py to run a given test, and the
-        # third level are the VMS started by the test itself.  In most
-        # environments, the "tasks" container runs on bare metal, and
-        # the VMs started here are on level 2.
-        #
-        # Using KVM on level 3 is significantly slower than software
-        # emulation, by something like a factor of 2 at least, and
-        # much worse on a machine with many VMs to the point that the
-        # kernel will trigger its NMI watchdog and the boot never
-        # finishes. So we switch it off by removing "/dev/kvm".
-        #
-        # Our environments where the VMs started by the tests are on
-        # level 2 don't have support for nested KVM. So "/dev/kvm"
-        # does not exist in the first place and we don't need to be
-        # careful to leave it in place.
-        #
-        m.execute("rm -f /dev/kvm")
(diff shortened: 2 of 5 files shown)


test/check-machines-multi-host-consoles
```diff
@@ -49,6 +49,10 @@ class TestMultiMachineVNC(VirtualMachinesCase):
         m2 = self.machine2
         m2_host = "10.111.113.2"
 
+        user = None
+        if "suse" in m.image:
+            user = "root"
```
martinpitt, line 54: This doesn't look c-machines specific. I figure this should rather go into https://github.com/cockpit-project/cockpit/blob/main/test/common/testlib.py#L931 ?

test/check-machines-storage-pools
```diff
@@ -477,7 +477,7 @@ class TestMachinesStoragePools(machineslib.VirtualMachinesCase):
                 b.wait_not_present("#create-storage-pool-dialog")
 
         # Prepare a Volume Group to be used as source for the LVM pool
-        m.execute(f"wipefs -a {dev}; pvcreate {dev}; vgcreate vol_grp1 {dev}")
+        m.execute(f"wipefs -a {dev}; sleep .5; pvcreate {dev}; vgcreate vol_grp1 {dev}")
```
martinpitt, line 480: Meh, this is begging for test flakes; how does it fail? could you rather do something like `until pvcreate ...; do sleep 0.5; done`?

test/machineslib.py
```diff
@@ -348,41 +348,44 @@ class VirtualMachinesCase(testlib.MachineCase, VirtualMachinesCaseHelpers, stora
...
+            # level 2 don't have support for nested KVM. So "/dev/kvm"
+            # does not exist in the first place and we don't need to be
+            # careful to leave it in place.
+            #
+            m.execute("rm -f /dev/kvm")
+
+            # Keep pristine state of libvirt
+            self.restore_dir("/var/lib/libvirt")
+            self.restore_dir("/etc/libvirt")
+            self.restore_dir("/home/admin/.local/share/libvirt/")
```
martinpitt, line 380: This only applies to nondestructive tests, which by definition can't use multiple VMs. It's an unnecessary waste for multi-VM destructive ones.

test/machineslib.py
```diff
@@ -348,41 +348,44 @@ class VirtualMachinesCase(testlib.MachineCase, VirtualMachinesCaseHelpers, stora
...
+
+            self.startLibvirt(m)
```
martinpitt, line 382: This also doesn't  seem necessary for all VMs.

I suppose you only want the /dev/kvm bit?

Can you please split out the "test: Run libvert setup against all vm's" into a separate commit, with the above cleanups and a "why" explanation? Also please fix the typos: libvert → libvirt, and vm's → VMs

test/machineslib.py
```diff
@@ -348,41 +348,44 @@ class VirtualMachinesCase(testlib.MachineCase, VirtualMachinesCaseHelpers, stora
...
+
+            self.addCleanup(m.execute, f"systemctl stop {self.getLibvirtServiceName()}")
+            if not hasMonolithicDaemon(m.image):
+                self.addCleanup(m.execute, "systemctl stop virtstoraged.service virtnetworkd.service")
 
-        self.addCleanup(m.execute, f"systemctl stop {self.getLibvirtServiceName()}")
-        if not hasMonolithicDaemon(m.image):
-            self.addCleanup(m.execute, "systemctl stop virtstoraged.service virtnetworkd.service")
+        m = self.machine
```
martinpitt, line 388: That looks unnecesssary now.

test/vm.install
```diff
@@ -7,6 +7,11 @@ if grep -q 'ID=debian' /usr/lib/os-release; then
     echo '* soft core unlimited' >> /etc/security/limits.conf
 fi
 
+# for opensuse images, allow dmesg to read /var/lib/libvirt/pxe-config
```
martinpitt, line 10: Certainly not dmesg, but dnsmasq? This looks like an OS bug workaround -- please file a suse bug and link it here as a comment, with a `# HACK:` prefix (our usual thing to grep for, we regularly want to clean these up).

martinpitt (changes requested):
Thanks! Looking forward to enabling this.

test/check-machines-multi-host-consoles
Nykseli: I'm happy to move that there. The API allows setting the username so this felt like the right place to do it.

martinpitt: Yes, but I mean "sudo asks for root password" is a general property of the opensuse image, so cockpit etc. tests are also going to need this anyway, no?

test/check-machines-storage-pools
SludgeGirl: Yep that's much better, I'll get it amended

test/check-machines-multi-host-consoles
Nykseli: You're correct. Created a PR for the change: https://github.com/cockpit-project/cockpit/pull/20525

test/machineslib.py
SludgeGirl: Got that removed out of the loop

test/machineslib.py
SludgeGirl: The aim of those changes were just to bring the two vm's into a closer state, I was a bit over zealous as you've pointed out. I've switch it out to just handle removing /dev/kvm

test/machineslib.py
SludgeGirl: It wasn't at the time, lower down still references m which changed definition however with my latest changes anyway this has been sorted

test/vm.install
SludgeGirl: My bad for the mis-wording. As for actual fix, the App Armour configs for SUSE only give dnsmasq (speaking in the context of libvirt) access to /var/lib/libvirt/dnsmasq from what I've seen pxe-config isn't an standard path and from digging around cockpit I haven't seen any references to this that would create this issue under production.

The path is only defined in tests here: https://github.com/cockpit-project/cockpit-machines/blob/ffbbf25e527d642508194c0ef993bde5c8359212/test/machinesxmls.py#L103 since it's a constant I thought this would be the more correct action to fix the issue. Still I'm happy to make any changes if you've got an alternative method you'd prefer for this

SludgeGirl:
Alright that should be everything responded to/fixed. Let me know if you want any other changes!

test/vm.install
martinpitt: Thanks for the explanation! I sent PR #1658, if that goes green and avoids that hack, let's do that instead? If not, ok.

martinpitt:
Thanks @SludgeGirl -- the commits so far would be acceptable for me. I'd like to replace the AppArmor hack (which looks a bit brittle) with #1658 if possible, but not a blocker.

I triggered the tests (opensuse-tumbleweed, but also all the others to check for regressions), let's hear the bots' opinion! :grin: (testing-farm rawhide failure is known and fixed in #1657, and the f-39 failure was a "WTH?" -- retried)

martinpitt:
Cool, [success](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1655-2d5893fe-20240529-092530-opensuse-tumbleweed/log.html)! Do you mind pulling in the change from #1658, drop the AppArmor commit, and we re-test with that? If it works, I think we'd both prefer that. Otherwise we land this version. Cheers!

martinpitt:
[works](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1655-75c7d1d4-20240529-105747-opensuse-tumbleweed/log.html)! I landed #1658 and took the liberty to rebase this branch.

martinpitt (approved):
Thanks @Nykseli and @SludgeGirl , nice work! :star2: After this lands, let's enable it in the testmap so that it won't break again.

martinpitt:
See https://github.com/cockpit-project/bots/pull/6442

martinpitt:
[yay rawhide](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-machines-1655/fedora-rawhide-x86_64/07503408-cockpit-machines/builder-live.log.gz) :exploding_head: 

But the build still worked in an earlier version in commit 2d5893fe85a935041e29888d424f285279e9d20c today, so I'm going to ignore that.

Nykseli:
@martinpitt Thanks a lot for all the help, once again! :)

PR #1658, opened 2024-05-29
test: Change PXE config path to be AppArmor compatible

martinpitt:
At least OpenSUSE's AppArmor config only allows libvirt to access /var/lib/libvirt/dnsmasq/. Use that standard path instead of our custom one.

See https://github.com/cockpit-project/cockpit-machines/pull/1655#discussion_r1618403695

### DIFF
--- a/test/machinesxmls.py
+++ b/test/machinesxmls.py
@@ -100,7 +100,7 @@ NETWORK_XML_PXE = """<network>
   <bridge name='virbr0' stp='on' delay='0'/>
   <mac address='52:54:00:53:7d:8e'/>
   <ip address='192.168.122.1' netmask='255.255.255.0'>
-    <tftp root='/var/lib/libvirt/pxe-config'/>
+    <tftp root='/var/lib/libvirt/dnsmasq'/>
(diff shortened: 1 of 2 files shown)


test/machinesxmls.py
```diff
@@ -100,7 +100,7 @@
   <bridge name='virbr0' stp='on' delay='0'/>
   <mac address='52:54:00:53:7d:8e'/>
   <ip address='192.168.122.1' netmask='255.255.255.0'>
-    <tftp root='/var/lib/libvirt/pxe-config'/>
+    <tftp root='/var/lib/libvirt/dnsmasq'/>
```
mvollmer, line 103: Ah, so dnsmasq is also a PXE and TFTP server? Then this makes sense.

mvollmer (approved):

test/machinesxmls.py
martinpitt: Yes, see https://dnsmasq.org/doc.html

martinpitt:
Proven to work in #1655

PR #1666, opened 2024-06-03
test: add a check that creating a snapshot with the same description …

yunmingyang:
…and different name

### DIFF
--- a/test/check-machines-snapshots
+++ b/test/check-machines-snapshots
@@ -305,6 +305,16 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
             expect_external=supports_external,
             xfail="name",
         ).execute()
+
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Hm, ok -- out of interest, what kind of bug does that check/reproduce? A failure there would be rather weird and deep inside of libvirt.

yunmingyang:
> Hm, ok -- out of interest, what kind of bug does that check/reproduce? A failure there would be rather weird and deep inside of libvirt.

So for now the description of the snapshot could be duplicated. Maybe, there will be some changes in the future by libvirt. Then, we don't know how they do that, maybe an error raised, or a hint, or just a output in the command results. I think that may cause some trouble in our project. Thus, this case may find the trouble early. @martinpitt

PR #1677, opened 2024-06-13
scripts: Fallback to IPv4 for spice if lo doesn't have an ipv6 addr

SludgeGirl:
Hey! Just a small patch to allow spice to fallback to IPv4 in cases where the user has disabled IPv6. I needed a bit of a break from the general Cockpit tests so thought it'd pick up a bug or two, but the tests are getting close!

### DIFF
--- a/src/scripts/install_machine.py
+++ b/src/scripts/install_machine.py
@@ -54,11 +54,23 @@ def get_graphics_capabilies(connection):
 
 
 def prepare_graphics_params(connection):
-    graphics_config = {
+    # Check to make sure loopback has an IPv6 address assigned
+    # if not default back to IPv4
+    # once we stop supporting Python 3.6, use this:
+    # ip_lo = subprocess.run(['ip', '-6', 'addr', 'show', 'dev', 'lo'], capture_output=True, text=True)
+    ip_lo = subprocess.run(['ip', '-6', 'addr', 'show', 'dev', 'lo'], stdout=subprocess.PIPE, universal_newlines=True)
+
+    if ip_lo.returncode == 0 and "inet6" in ip_lo.stdout:
         # Use ipv6 as workaround for https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621
-        'spice': {'listen': '::1'},
+        spice_address = '::1'
+    else:
+        spice_address = '127.0.0.1'
+
+    graphics_config = {
+        'spice': {'listen': spice_address},
         'vnc': {'listen': '127.0.0.1'}
     }
+
     try:
         # Configparser needs a default section
         with open("/etc/libvirt/qemu.conf", 'r') as f:
@@ -67,7 +79,7 @@ def prepare_graphics_params(connection):
         config = configparser.ConfigParser()
         config.read_string(config_string)
 
-        graphics_config['spice']['listen'] = config['dummy_section'].get('spice_listen', '::1')
+        graphics_config['spice']['listen'] = config['dummy_section'].get('spice_listen', spice_address)
         spice_password = config['dummy_section'].get('spice_password', None)
         if spice_password is not None:
             graphics_config['spice']['password'] = spice_password


src/scripts/install_machine.py
```diff
@@ -54,11 +54,20 @@ def get_graphics_capabilies(connection):
...
 
 def prepare_graphics_params(connection):
+    # Use ipv6 as workaround for https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621
+    spice_address = '::1'
+    # Check to make sure loopback has an IPv6 address assigned
+    # if not default back to IPv4
+    ipv6_loopback = subprocess.check_output(['ip', '-6', 'addr'])
+
+    if not "lo:" in str(ipv6_loopback):
+        spice_address = '127.0.0.1'
```
martinpitt, line 64: Can you please do `ip -6 addr show dev lo`? And make this robust against failures of the call? The `ip` utility may not be installed, or perhaps some other weird thing. We don't want the script to crash then.

martinpitt (changes requested):
Thanks! The general approach looks fine, but let's robustify this a bit.

src/scripts/install_machine.py
SludgeGirl: Yep! Thanks for the review, I've just reworked it quite a bit to be more robust and check against ifconfig if ip doesn't exist. This should be alot more robust

martinpitt:
Interesting this [breaks on rawhide](https://artifacts.dev.testing-farm.io/25a99126-da33-45da-a19f-0ae964f25e4c/):

```
    self.assertEqual(spice_listen, spice[0].attrib['listen'])
AssertionError: '::1' != '127.0.0.1'
- ::1
+ 127.0.0.1
```

So this overrides an explicit configuration in `/etc/libvirt/qemu.conf`, which we should never do. This should only change the implicit default. The code tries to do that, but I think I see the problem.

src/scripts/install_machine.py
```diff
@@ -54,11 +54,35 @@ def get_graphics_capabilies(connection):
...
 
 def prepare_graphics_params(connection):
+    # Use ipv6 as workaround for https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621
+    spice_address = '::1'
+
+    # Check to make sure loopback has an IPv6 address assigned
+    # if not default back to IPv4
+    try:
+        if subprocess.run(['ip'], shell=True).returncode == 0:
+            ipv6_loopback = subprocess.check_output(['ip', '-6', 'addr', 'show', 'dev', 'lo'])
```
martinpitt, line 64: Please avoid `shell=True` and running this command twice. You can use [subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run), it defaults to `check=False` and you can check the error code for success and then use the output. You should also use `text=True` and drop the `str()` below

src/scripts/install_machine.py
```diff
@@ -54,11 +54,35 @@ def get_graphics_capabilies(connection):
...
+
+            if not "lo:" in str(ipv6_loopback):
```
martinpitt, line 66: Pretty sure you mean "inet6" here? You asked for `lo` already.

src/scripts/install_machine.py
```diff
@@ -54,11 +54,35 @@ def get_graphics_capabilies(connection):
...
+                spice_address = '127.0.0.1'
+        elif subprocess.run(['ifconfig'], shell=True).returncode == 0:
+            ipv6_loopback = subprocess.check_output(['ifconfig', 'lo'])
```
martinpitt, line 69: Ugh, `ifconfig` -- do you really need that for OpenSUSE? This is so old, I'd really rather not call that as well.

src/scripts/install_machine.py
```diff
@@ -54,11 +54,35 @@ def get_graphics_capabilies(connection):
...
+
+            if not "inet6" in str(ipv6_loopback):
+                spice_address = '127.0.0.1'
+        else:
+            # All previous attempts to find lo assignment have failed
+            # throw an exception and handle this as unrecoverable
+            raise Exception
+    except Exception:
+        # We've unrecoverably failed, lets assume worst case
```
martinpitt, line 78: No bare `Exception`s, please. If you want to use `check=True`, then catch `subprocess.CalledProcessError`.

src/scripts/install_machine.py
```diff
@@ -54,11 +54,35 @@ def get_graphics_capabilies(connection):
...
+        spice_address = '127.0.0.1'
```
martinpitt, line 79: This changes the default, and I suppose that's why the [rawhide test fails](https://artifacts.dev.testing-farm.io/25a99126-da33-45da-a19f-0ae964f25e4c/)?

I suggest something like

```py
ip_lo = subprocess.run(ip, ...)
if ip_lo.returncode == 0 and "inet6" in ip_lo.stdout:
    spice_address = "::1"
else:
   spice_address = "127.0.0.1"
```

that's nice and simple.

martinpitt (changes requested):

src/scripts/install_machine.py
SludgeGirl: Just got those updated, thanks I'll keep that in mind for future

src/scripts/install_machine.py
SludgeGirl: Since we've got the -6 flag, this still works as intended, as if there's no IPv6 assignment lo: wouldn't be displayed since it doesn't show the device if there's no ip assigned, that being said inet6 feels like a better bit of text to search for so I've updated that with this set of changes

SludgeGirl:
I've got this reworked to be inline with your suggestion, let me know if there's any other changes you'd like, thanks for your patience with this one!

martinpitt (approved):
Thanks, and sorry for the delay! This looks very clean and nice now. Triggered tests.

martinpitt:
Ugh, this [breaks RHEL 8 hard](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1677-fbf43483-20240621-073108-rhel-8-10/log.html). I have a look.

src/scripts/install_machine.py
```diff
@@ -54,11 +54,21 @@ def get_graphics_capabilies(connection):
 
 
 def prepare_graphics_params(connection):
-    graphics_config = {
+    # Check to make sure loopback has an IPv6 address assigned
+    # if not default back to IPv4
+    ip_lo = subprocess.run(['ip', '-6', 'addr', 'show', 'dev', 'lo'], capture_output=True, text=True)
```
martinpitt, line 59: Ah, haha: This doesn't yet work on Python 3.6:

   TypeError: __init__() got an unexpected keyword argument 'capture_output'"}"

I'll push a fix.

martinpitt (approved):

src/scripts/install_machine.py
SludgeGirl: Ohhhh gotcha, I'll keep that in mind for future and keep a better eye on when features were introduced. Sorry! Bare with me while I'm still getting to grips with it all

SludgeGirl:
Thanks for your help with this!

src/scripts/install_machine.py
martinpitt: Don't worry -- this happens to all of us *all the time*. That's why we have tests to tell us where we are wrong :grin:

PR #1681, opened 2024-06-16
build.js: support outputting esbuild metadata file

jelly:
ESbuild supports outputting a metadata file which can be used to view bundle size information using https://esbuild.github.io/analyze/


jelly:
Interesting note, we have two copies of `xterm.js` in our `node_modules` one from PF and one separate.

martinpitt (approved):

PR #1684, opened 2024-06-17
snapshots: Ignore limitations of external format on RHEL

mvollmer:
Internal snapshots are not fully supported on RHEL, so we assume that the limitations of external snapshots on RHEL are fixed.


mvollmer:
(This is not a serious proposal.)

PR #1694, opened 2024-06-25
Show snapshot creation date as time delta

martinpitt:
We use `timeformat.distanceToNow()` in most places in cockpit. Use that
for consistency rather than showing a fuzzified absolut/relative time
stamp mix with date-fn's formatRelative(). For more precision, show the
absolute creation time in a tooltip.

This weans cockpit-machines off the date-fns API, which we try to get
rid of (https://github.com/cockpit-project/cockpit/issues/20653).

-----

Current main:

snapshots-main

This PR:

snap-pr

### DIFF
--- a/src/components/vm/snapshots/vmSnapshotsCard.jsx
+++ b/src/components/vm/snapshots/vmSnapshotsCard.jsx
@@ -26,9 +26,10 @@ import { CheckIcon, InfoAltIcon } from '@patternfly/react-icons';
 
 import cockpit from 'cockpit';
 import { useDialogs, DialogsContext } from 'dialogs.jsx';
+import * as timeformat from 'timeformat';
 import { ListingTable } from "cockpit-components-table.jsx";
 
-import { vmId, localize_datetime, vmSupportsExternalSnapshots } from "../../../helpers.js";
+import { vmId, vmSupportsExternalSnapshots } from "../../../helpers.js";
 import { CreateSnapshotModal } from "./vmSnapshotsCreateModal.jsx";
 import { DeleteResourceButton } from '../../common/deleteResource.jsx';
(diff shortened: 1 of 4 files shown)


garrett:
I guess the layout issues in the first screenshot aren't addressed, and it's just coincidence that it's not so broken in the second screenshot?

martinpitt:
@garrett Right, this changes no layout at all. I just took the screenshots at different resolutions because reasons (mostly because firefox screenshots don't catch tooltips, so I did a full-screen snapshot). This is just about the time format.

src/components/vm/snapshots/vmSnapshotsCard.jsx
```diff
@@ -16,17 +16,20 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
+
 import React from 'react';
 
-import cockpit from 'cockpit';
```
KKoukiou, line 21: This can be very nicely enforced and autofixed globally as well. I do it in anaconda-webui, see https://github.com/rhinstaller/anaconda-webui/blob/main/.eslintrc.json#L32

KKoukiou (commented):
No objection on the time related changed.

garrett (approved):
I have problems with the UI, including consistency issues, layout wrapping oddness, and others...

...but for the string in this PR? That's an improvement. :+1:

martinpitt:
@jelly As Katerina and Garrett already acked this, I'll leave the review to you -- I think you had some criticism?

jelly (approved):

jelly:
> @jelly As Katerina and Garrett already acked this, I'll leave the review to you -- I think you had some criticism?

I thought Garrett would have objections, if its fine with him lets merge it.

PR #1697, opened 2024-06-26
package.json: drop deep-equal

jelly:
Machines does not deep_equal anywhere in its code or in anything imported from `pkg/lib`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "@patternfly/react-table": "5.3.3",
     "@patternfly/react-tokens": "5.3.1",
     "date-fns": "3.6.0",
-    "deep-equal": "2.2.3",
     "prop-types": "15.8.1",
     "react": "18.3.1",
     "react-dom": "18.3.1",


martinpitt (approved):
The [effective diff](https://github.com/cockpit-project/node-cache/compare/17ac5ee28ae2e513ccb1dc205019347608ff0c66..6b9e1b3237817f9b2f3f1d85319bde9aefb33daf) just makes it a devDependency, as it's a transient dep of eslint or something. But conceptually this is right, and should avoid some dependabot noise. Thanks!

jelly:
> The [effective diff](https://github.com/cockpit-project/node-cache/compare/17ac5ee28ae2e513ccb1dc205019347608ff0c66..6b9e1b3237817f9b2f3f1d85319bde9aefb33daf) just makes it a devDependency, as it's a transient dep of eslint or something. But conceptually this is right, and should avoid some dependabot noise. Thanks!

Some day maybe eslint will drop it :crossed_fingers:  but indeed this mostly helps us reduce updates.

PR #1700, opened 2024-06-27
test: avoid double waits

jelly:
`wait_in_text` already waits on the element so `wait_visible` is unneeded.


martinpitt (approved):
:broom: Thanks!

PR #1701, opened 2024-06-28
Debug flakes shutdown

jelly:


jelly:
Touching `machineslib.py` does not make tests re-run 3 times. Lets touch one test file for science.

jelly:
@martinpitt please take a look, as changing only `machineslib.py` does not trigger re-tries I still have a debug commit so you can see that the change did in fact make CI less flaky :)

jelly:
@martinpitt requested some investigation if this is due to too many parallel runs, soon we have `--amplify` to easily check this

martinpitt:
@jelly If waiting longer helps, by all means go for it! It's plausible if that is a "soft" shutdown, i.e. the OS actually has to shut down (as opposed to just killing the VM). It's all emulated, so the inner VMs are sloooow :snail:

jelly:
> @jelly If waiting longer helps, by all means go for it! It's plausible if that is a "soft" shutdown, i.e. the OS actually has to shut down (as opposed to just killing the VM). It's all emulated, so the inner VMs are sloooow 🐌

I did some more testing in https://github.com/cockpit-project/cockpit-machines/pull/1705  too much amplification makes the VM crash, like [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1705-33f76784-20240703-074528-ubuntu-stable/log.html#81-2). So I'd say this is good to merge?

jelly:
Still fails,this needs a deeper dive:

https://cockpit-logs.us-east-1.linodeobjects.com/pull-1701-72252da6-20240703-090406-debian-stable/TestMachinesLifecycle-testBasic-debian-stable-127.0.0.2-2901-FAIL.png

PR #1707, opened 2024-07-03
test: openssh 9.8 by default disabled DSA key generation

jelly:
By default DSA key support is now not enabled during compilation, Arch did not re-enable DSA support so can't create the keys anymore.

RHEL 8.10 can create ecdsa keys, so switching from DSA to ECDSA should work in all other supported test distributions.

---

This is required for https://github.com/cockpit-project/bots/pull/6570


martinpitt (approved):
Sure! I don't know about ecdsa, but probably we should move to ed25519 next? Anyway, this is good for now. Thanks!

jelly:
> Sure! I don't know about ecdsa, but probably we should move to ed25519 next? Anyway, this is good for now. Thanks!

Yeah, that crossed my mind, but maybe that is too modern for RHEL-8? (In Cockpit we still use ECDSA FYI).

jelly:
m.execute("ln -s /etc/apparmor.d/usr.sbin.libvirtd /etc/apparmor.d/disable/usr.sbin.libvirtd")

Unrelated failure in debian-testing.

PR #1711, opened 2024-07-05
test: Move from 'coreduo' to 'qemu64' custom CPU model

martinpitt:
"coreduo" is not supported any more in CentOS/RHEL 10.

---

Our c10s tests [started to fail](https://artifacts.dev.testing-farm.io/48f7438b-0ad5-4c74-8989-14023b23b81a/) a few days ago.

### DIFF
--- a/test/check-machines-settings
+++ b/test/check-machines-settings
@@ -249,15 +249,15 @@ class TestMachinesSettings(machineslib.VirtualMachinesCase):
         # Choose manually a CPU model
         b.click("#vm-subVmTest1-cpu button")
         b.wait_visible("#machines-cpu-modal-dialog")
-        b.select_from_dropdown("#cpu-model-select-group select", "coreduo")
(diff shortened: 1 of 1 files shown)


martinpitt:
Meh, so [kvm64](https://artifacts.dev.testing-farm.io/677dd446-fd32-42ed-ac4c-adb8bce81bea/) does not work either. I have some real trouble reproducing this locally, as our centos-10 image is too old/damaged.

martinpitt:
Ah, [here](https://artifacts.dev.testing-farm.io/1462c256-dc32-4f3b-8fa4-a25d4a44c5ca/work-network69rlxj3k/plans/all/network/execute/data/guest/default-0/test/browser/network-1/data/TestMachinesSettings-testCPUModel-centos-10-localhost-22-FAIL-2.html) we have an ugly, but working interactive HTML dump, so "qemu64" should work.

jelly (approved):
Nice, I suspected this was an issue but couldn't get a CentOS 10 image working.

PR #1718, opened 2024-07-11
fmf: Disable SELinux with broken RHEL 10 policy version

martinpitt:
This completely breaks libvirt and libvirt-dbus.

See https://issues.redhat.com/browse/RHEL-46893

---

See https://artifacts.dev.testing-farm.io/6bf72f5d-1950-41e2-a151-76f11ce6301d/ . This also broke yesterday's RHEL 10 gating tests.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -26,6 +26,9 @@ fi
 rpm -q selinux-policy cockpit-bridge cockpit-machines
 rpm -qa | grep -E 'virt|qemu' | sort
 
+# basic libvirt selftests
+virt-host-validate qemu || echo "failed with code $?"
+
 # allow test to set up things on the machine
 mkdir -p /root/.ssh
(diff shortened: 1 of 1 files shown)


martinpitt:
Meh, [still fails](https://artifacts.dev.testing-farm.io/7489b186-0682-48f7-9897-4460c1fc69b6/), but "only" two tests, not all of them.

martinpitt:
Ah, that was a flake, good. 

In downstream RHEL 9.5 gating, all tests fail with
```
  File "/source/test/check-machines-lifecycle", line 343, in testRename
    self.waitPageInit()
  File "/source/test/machineslib.py", line 38, in waitPageInit
    virtualization_enabled = "PASS" in self.machine.execute("virt-host-validate qemu | grep 'Checking for hardware virtualization'")
```
It's unclear why, so let's debug this. I added another commit for that.

jelly (approved):

PR #1719, opened 2024-07-11
manifest: drop content-security-policy

jelly:
The default content-security-policy of cockpit-bridge is enough for files it already includes `img-src: 'self' data:` and `frame-src` is not needed as we don't have or want any iframes or frames in machines.

---

Same as https://github.com/cockpit-project/cockpit-files/pull/646

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -24,8 +24,6 @@
         ]
      }
   },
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Cheers!

PR #1727, opened 2024-07-18
test: add checks for avoiding a timing issue, that is the path dropdown list needs more time to load

yunmingyang:


martinpitt:
Hello @yunmingyang ! This looks a bit dubious -- to understand the failure, do you have a link to the failed run? Or can  you copy the output and the screenshot from the failure?

yunmingyang:
@martinpitt I found this during downstream aarch64 testing, I am not sure whether it could be reproduced, but let me have a try later

test/check-machines-filesystems
```diff
@@ -60,8 +60,12 @@ class TestMachinesFilesystems(machineslib.VirtualMachinesCase):
 
         b.set_input_text("#vm-subVmTest1-filesystems-modal-mountTag", "")
         b.set_input_text("#vm-subVmTest1-filesystems-modal-source-group input", "/tmp")
-        b.wait_visible("#vm-subVmTest1-filesystems-modal-source")
+        # In some hosts with high latency, the path dropdown list needs some time to load, the check will fail if it is too quick.
+        # Thus, add more check to ensure that the loading is finished
+        b.wait_in_text("#vm-subVmTest1-filesystems-modal-source", "/tmp")
```
martinpitt, line 65: This is what I'm sceptical about: `set_input_text` already waits until the input has the expected value, so this is entirely redundant. Either that existing wait in testlib.py isn't enough -- then we should fix it there for all cases. Or it is redundant, then this is nothing more than a hidden "sleep(0.1)", and the real fix is to wait for something else.

test/check-machines-filesystems
```diff
@@ -60,8 +60,12 @@ class TestMachinesFilesystems(machineslib.VirtualMachinesCase):
...
         b.click("#vm-subVmTest1-filesystems-modal-source button:nth-child(1)")
+        b.wait_not_present("#vm-subVmTest1-filesystems-modal-source")
+        b.wait_attr_contains("#vm-subVmTest1-filesystems-modal-source-group input", "value", "/tmp")
```
martinpitt, line 68: Can you please explain this? The "click" could really do with a comment what it does, is that the help icon button for "More info for source path field"? But why would clicking it make the whole "Source path" input go away? And then *after* you wait for it to go away you check one of its attributes?

(Note: it may well be right: It is just absolutely impossible for me to understand it without being intimately familiar with what this does)

martinpitt (changes requested):
Thanks!

yunmingyang:
TestMachinesFilesystems-testBasicSystemConnection
Hi @martinpitt, I could also reproduce it during s390x test, but I think maybe there are some security information in *.js.log and *.html, like IP address and firmware, Thus, I just put the picture here. You could see there are two errors shown, but it should just 1 error, as "Source path" is already not empty.

test/check-machines-filesystems
yunmingyang: Yeah, set_input_text already waits until the input has the expected value. But, there will be a dropdown list for file path prompt as the input there is a FileAutoComplete. This line aims to make sure the input text is shown in the dropdown list. We don't check the dropdown list in set_input_text. And it seems that, if I don't add this line, the dropdown list is still there after b.click("#vm-subVmTest1-filesystems-modal-source button:nth-child(1)"). Here is also a picture if I remove this line.
TestMachinesFilesystems-testBasicSystemConnection-dropdown-list

test/check-machines-filesystems
yunmingyang: So I just want to make sure the dropdown list disappears(b.wait_not_present) and the input accepts the text correctly(b.wait_attr_contains). Then, after some new trying, I think b.wait_attr_contains is not necessary as set_input_text has already done this, but according to the previous comment I left, I think b.wait_not_present is helpful to ensure that the dropdown list has disappeared. Do you think that makes sense?

test/check-machines-filesystems
martinpitt: Sorry for the late answer, I've been sick for two weeks :cry: Yes, I understand the "wait for /tmp to be in dropdown" approach. But the wait_in_text() is not specific enough for that -- it covers the *whole* input group, i.e. including the input line. Doesn't it also catch *that* `/tmp` there? I.e. my question is, this selector needs to be more specific to the select, with `.pf-v5-c-menu` or similar.

test/check-machines-filesystems
martinpitt: Right, the `wait_attr_contains` is the same as `.val()`, and set_input_text already does the .wait_val() for this. The wait_not_present() strategically makes sense, I just have *no* idea what `-modal-source` actually is -- is that really specifically the *dropdown menu* part of the file selector? The code says

    <FileAutoComplete id={`${idPrefix}-modal-source`}

which should be the text `<input>` and sometimes the menu (which can open or close), and this is specifically meant for the *menu* part.

That's what I mean with "please explain this" -- and I don't mean "here in github", but "please add a comment".  These selectors are not understandable (which isn't your fault, but it still needs comments to make sense).

Thanks!

Please also rebase this, the PR is quite old by now.

test/check-machines-filesystems
yunmingyang: I checked with cockpit-machines-316-1.fc40.noarch. #vm-subVmTest1-filesystems-modal-source doesn't include the input line, and the input line is actual in another div, and the id of the input line is "pf-select-toggle-id-26-select-typeahead", which is automatically generated, I think. Are there some changes about the input line and the expand list in the newest vision?

test/check-machines-filesystems
yunmingyang: OK, I will rebase and add some comments

test/check-machines-filesystems
martinpitt: No, there are no changes. The main thing is that the selectors are not understandable, and could use some comments what they actually refer to.

Thanks for the explanation! Yes, please rebase, and let's get this in.

yunmingyang:
Re-push:

- Add more annotation for explaining
- Remove the b.wait_attr_contains since set_input_text had already done this

martinpitt (approved):
Thank you!

PR #1724, opened 2024-07-15
Fix race condition in testCreateDownloadAnOS() on tumbleweed, disable broken SELinux on rawhide

martinpitt:
This test [almost always](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6612-f8f545a2-20240713-052557-opensuse-tumbleweed-cockpit-project-cockpit-machines/log.html) [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1720-67dc5b2a-20240715-070654-opensuse-tumbleweed/log.html) on tumbleweed, but never *ever* locally. I even tried to reproduce the exact series of nondestructive tests:
```
TEST_OS=opensuse-tumbleweed test/common/run-tests -tv TestMachinesStoragePools.testStoragePoolsDeletion TestMachinesSettings.testBootOrder TestMachinesNetworks.testNetworkState TestMachinesNICs.testNICDelete TestMachinesHostDevs.testHostDevicesList TestMachinesFilesystems.testBasicSessionConnection TestMachinesDisks.testDetachDisk TestMachinesCreate.testDisabledCreate TestMachinesCreate.testCreateImportDisk TestMachinesCreate.testCreateDownloadAnOS
```

The [weather report](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit-machines&days=7&test=%2Ftest%2Fcheck-machines-create+TestMachinesCreate.testCreateDownloadAnOS) also shows that this fails exclusively on tumbleweed, in more than 77% of runs.

Turning debugging to 11.

I'll give this a little bit of investigation, but only so much -- if I can't figure it out quickly, I'll disable the test for now.

@Nykseli @Lunarequest @SludgeGirl FYI

PR #1729, opened 2024-07-18
test: selinux-policy-40.13.5-1.el10.noarch is broken as well

martinpitt:
Rewrite this using a `case` as that's syntactically easier.

---

See https://artifacts.dev.testing-farm.io/be2e002f-fb32-4cfb-8a55-3f2b2a870dc9/ - I kept the old version as that's still in some RHEL 10 composes.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -22,14 +22,15 @@ if rpm -q amazon-ec2-utils; then
     udevadm trigger /dev/nvme*
 fi
 
-# HACK: https://issues.redhat.com/browse/RHEL-46893
-if [ "$(rpm -q selinux-policy)" = "selinux-policy-40.13.4-1.el10.noarch" ] ; then
-    setenforce 0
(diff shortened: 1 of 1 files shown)


martinpitt:
argh yes, copr is kaputt today. My build in https://src.fedoraproject.org/rpms/cockpit-files/pull-request/10 also hasn't been picked up for many hours.

martinpitt:
OK, that's better -- now it "only" [fails testCreateDownloadAnOS](https://artifacts.dev.testing-farm.io/6d021de0-9d78-4c5b-aa30-b6454ca2c624/) (for a different reason), not all tests any more due to SELinux. So that's at least an improvement. I'll look at that new failure next. It strongly reminds me of #1724, presumably there's a new libvirt or virt-install version which is trickling into development distros now.

jelly (approved):
Blegh, thanks for the workarounds!

PR #1730, opened 2024-07-18
Improvements for external snapshots of running machines.

mvollmer:
- We only ask for the directory of the memory snapshot, not for the full filename. That way we don't have to worry about what to do when the user changes the name and we it makes more sense when we talk about how much space is available "at this location". Also, people don't have to worry about the filenames for disk snapshots, why would they worry about them for the memory part?
 - The default for the memory save location is also taken from previous snapshots. That way, people only have to change it for the first snapshot when they are not happy with putting it next to the primary disk.
 - The snapshot name does not need to include the VM name. Snapshots are already scoped to VMs.
 - The memory save file gets the extension ".save", to mirror what "virsh managedsave" is doing.

image

 - [ ] get the actual available space
 - [ ] warn when it is not enough
 - [ ] tests
 - [ ] coverage


mvollmer:
Previous version of the dialog:

image

mvollmer:
It's not yet clear to me how to compute the estimate for the maximum size of the memory save file.

In my experiments, the file size corresponds pretty closely with the memory usage shown by Cockpit. A snapshot taken while compiling the kernel consistently used 2 GB on disk, for a VM with 2 GB RAM. A snapshot of a freshly booted machine that showed 500 MB RAM usage (out of 2 GB) in Cockpit used 450 MB for the memory snapshot.

So I guess we can use the current memory usage as the estimate. Or to be conservative, we can use the total RAM.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
-    // Choosing a default path where memory snapshot should be stored might be tricky. Ideally we want
-    // to store it in the same directory where the primary disk (the disk which is first booted) is stored
-    // If howver no such disk can be found, we should fallback to libvirt's default /var/lib/libvirt
+function getDefaultMemoryLocation(vm) {
+    // If we find an existing external snapshot, use it's memory path
+    // as the default. Otherwise, try to find the primary disk and use
+    // it's location. If that fails as well, use a reasonable hard
+    // coded value.
+
+    for (const s of vm.snapshots.sort((a, b) => b.creationTime - a.creationTime)) {
```
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+        if (s.memoryPath)
+            return dirname(s.memoryPath);
```
cockpituous, line 80: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
                                                               d.device.source.file);
     if (devices.length > 0) {
         const primaryDiskPath = devices[0].device.source.file;
         const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        return directory;
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory/";
```
cockpituous, line 93: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory/";
```
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
 
-const MemoryPathRow = ({ onValueChanged, memoryPath, validationError }) => {
+const MemoryLocationRow = ({ onValueChanged, memoryLocation, validationError, available, needed }) => {
+    let info = "";
+    let info_variant = "default";
+
+    if (available) {
+        info = cockpit.format(_("$0 available at this location."), formatWithBestUnit(available));
+        if (available * 0.9 < needed)
+            info_variant = "warning";
```
cockpituous, line 108: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser: "try" })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 138: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+            // channel has already logged the error
+                callback(null);
```
cockpituous, line 140: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -139,29 +192,38 @@ export class CreateSnapshotModal extends React.Component {
...
-                description,
-                isExternal,
-                memoryPath: isExternal && vm.state === "running" && memoryPath,
-            })
+            let mpath = null;
+            if (isExternal && vm.state === "running" && memoryLocation) {
+                mpath = memoryLocation;
+                if (mpath[mpath.length - 1] != "/")
+                    mpath = mpath + "/";
+                mpath = mpath + vm.name + "." + name + ".save";
```
cockpituous, line 214: These 5 added lines are not executed by any test.

src/helpers.js
```diff
@@ -145,6 +145,12 @@ export function convertToUnitVerbose(input, inputUnit, outputUnit) {
     return result;
 }
 
+export function formatWithBestUnit(input, inputUnit) {
+    const best = convertToBestUnit(input, inputUnit || units.B);
+    const decimals = best.value >= 100 ? 0 : best.value >= 10 ? 1 : 2;
```
cockpituous, line 150: This added line is not executed by any test.

src/helpers.js
```diff
@@ -907,3 +913,11 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
 
     return true;
 }
+
+export function dirname(path) {
+    const i = path.lastIndexOf("/");
+    if (i < 0)
+        return null;
```
cockpituous, line 920: These 4 added lines are not executed by any test.

src/helpers.js
```diff
@@ -907,3 +913,11 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
...
+    else
+        return path.substr(0, i);
```
cockpituous, line 922: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 80: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 93: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 108: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 138: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 140: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 214: These 5 added lines are not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -175,19 +237,25 @@ export class CreateSnapshotModal extends React.Component {
...
             <Form onSubmit={e => e.preventDefault()} isHorizontal>
                 <NameRow name={name} validationError={validationError.name} onValueChanged={this.onValueChanged} />
                 <DescriptionRow description={description} onValueChanged={this.onValueChanged} />
                 {isExternal && vm.state === 'running' &&
-                    <MemoryPathRow memoryPath={memoryPath} onValueChanged={this.onValueChanged}
-                                   validationError={validationError.memory} />}
+                    <MemoryLocationRow memoryLocation={memoryLocation} onValueChanged={this.onValueChanged}
+                                       validationError={validationError.memory}
+                                       available={available}
+                                       needed={(vm.rssMemory || vm.currentMemory) * 1024} />}
```
cockpituous, line 258: This added line is not executed by any test.

src/helpers.js
cockpituous, line 150: This added line is not executed by any test.

src/helpers.js
cockpituous, line 920: These 4 added lines are not executed by any test.

src/helpers.js
cockpituous, line 922: This added line is not executed by any test.

mvollmer:
@garrett, what do you think of this?

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,79 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
-    // Choosing a default path where memory snapshot should be stored might be tricky. Ideally we want
-    // to store it in the same directory where the primary disk (the disk which is first booted) is stored
-    // If howver no such disk can be found, we should fallback to libvirt's default /var/lib/libvirt
+function getDefaultMemoryLocation(vm) {
+    // If we find an existing external snapshot, use it's memory path
+    // as the default. Otherwise, try to find the primary disk and use
+    // it's location. If that fails as well, use a reasonable hard
+    // coded value.
+
+    for (const s of vm.snapshots.sort((a, b) => b.creationTime - a.creationTime)) {
```
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,79 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+        if (s.memoryPath)
+            return dirname(s.memoryPath);
```
cockpituous, line 80: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,79 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
                                                               d.device.source.file);
     if (devices.length > 0) {
         const primaryDiskPath = devices[0].device.source.file;
         const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        return directory;
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory/";
```
cockpituous, line 93: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,79 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory/";
```
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,79 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
 
-const MemoryPathRow = ({ onValueChanged, memoryPath, validationError }) => {
+const MemoryLocationRow = ({ onValueChanged, memoryLocation, validationError, available, needed }) => {
+    let info = "";
+    let info_variant = "default";
+
+    if (available) {
+        info = cockpit.format(_("$0 available at this location."), formatWithBestUnit(available));
+        if (available * 0.9 < needed)
+            info_variant = "warning";
```
cockpituous, line 108: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,79 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser: "try" })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 139: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,79 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+            // channel has already logged the error
+                callback(null);
```
cockpituous, line 141: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -139,29 +193,38 @@ export class CreateSnapshotModal extends React.Component {
...
-                name,
-                description,
-                isExternal,
-                memoryPath: isExternal && vm.state === "running" && memoryPath,
-            })
+            let mpath = null;
+            if (isExternal && vm.state === "running" && memoryLocation) {
+                mpath = memoryLocation;
+                if (mpath[mpath.length - 1] != "/")
+                    mpath = mpath + "/";
```
cockpituous, line 214: This added line is not executed by any test.

src/helpers.js
cockpituous, line 920: These 4 added lines are not executed by any test.

src/helpers.js
cockpituous, line 922: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
-    // Choosing a default path where memory snapshot should be stored might be tricky. Ideally we want
-    // to store it in the same directory where the primary disk (the disk which is first booted) is stored
-    // If howver no such disk can be found, we should fallback to libvirt's default /var/lib/libvirt
+function getDefaultMemoryLocation(vm) {
+    // If we find an existing external snapshot, use it's memory path
+    // as the default. Otherwise, try to find the primary disk and use
+    // it's location. If that fails as well, use a reasonable hard
+    // coded value.
+
+    for (const s of vm.snapshots.sort((a, b) => b.creationTime - a.creationTime)) {
```
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
                                                               d.device.source.file);
     if (devices.length > 0) {
         const primaryDiskPath = devices[0].device.source.file;
         const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        return directory;
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory/";
```
cockpituous, line 93: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory/";
```
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
 
-const MemoryPathRow = ({ onValueChanged, memoryPath, validationError }) => {
+const MemoryLocationRow = ({ onValueChanged, memoryLocation, validationError, available, needed }) => {
+    let info = "";
+    let info_variant = "default";
+
+    if (available) {
+        info = cockpit.format(_("$0 available at this location."), formatWithBestUnit(available));
+        if (available * 0.9 < needed)
+            info_variant = "warning";
```
cockpituous, line 108: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser: "try" })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 140: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+            // channel has already logged the error
+                callback(null);
```
cockpituous, line 142: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -139,29 +194,38 @@ export class CreateSnapshotModal extends React.Component {
...
-                name,
-                description,
-                isExternal,
-                memoryPath: isExternal && vm.state === "running" && memoryPath,
-            })
+            let mpath = null;
+            if (isExternal && vm.state === "running" && memoryLocation) {
+                mpath = memoryLocation;
+                if (mpath[mpath.length - 1] != "/")
+                    mpath = mpath + "/";
```
cockpituous, line 215: This added line is not executed by any test.

src/helpers.js
```diff
@@ -907,3 +913,11 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
 
     return true;
 }
+
+export function dirname(path) {
+    const i = path.lastIndexOf("/");
+    if (i < 0)
+        return "/";
```
cockpituous, line 920: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 93: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 140: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 142: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 215: This added line is not executed by any test.

src/helpers.js
cockpituous, line 920: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
     if (devices.length > 0) {
         const primaryDiskPath = devices[0].device.source.file;
-        const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        const directory = dirname(primaryDiskPath);
+        return directory;
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory";
```
cockpituous, line 93: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory";
```
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser: "try" })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 140: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+            // channel has already logged the error
+                callback(null);
```
cockpituous, line 142: This added line is not executed by any test.

src/helpers.js
```diff
@@ -907,3 +913,13 @@ export function vmSupportsExternalSnapshots(config, vm, storagePools) {
 
     return true;
 }
+
+export function dirname(path) {
+    const i = path.lastIndexOf("/");
+    if (i < 0)
+        return null;
```
cockpituous, line 920: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
     if (devices.length > 0) {
         const primaryDiskPath = devices[0].device.source.file;
-        const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        const directory = dirname(primaryDiskPath);
+        return directory;
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory";
```
cockpituous, line 93: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory";
```
cockpituous, line 95: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser: "try" })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 140: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+            // channel has already logged the error
+                callback(null);
```
cockpituous, line 142: This added line is not executed by any test.

src/helpers.js
```diff
@@ -907,3 +913,13 @@ export function vmSupportsExternalSnapshots(config, vm) {
...
```
cockpituous, line 920: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
 
-const MemoryPathRow = ({ onValueChanged, memoryPath, validationError }) => {
+const MemoryLocationRow = ({ onValueChanged, memoryLocation, validationError, available, needed }) => {
+    let info = "";
+    let info_variant = "default";
+
+    if (available) {
+        info = cockpit.format(_("$0 available at this location."), formatWithBestUnit(available));
+        if (available * 0.9 < needed)
+            info_variant = "warning";
```
martinpitt, line 108: Below you check `if (needed)` so it's optional. `Number < undefined` is always false, so without `needed` this will always be variant `default`. This seems correct to me, just a bit unclean to rely on such quirky semantics.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+    }
+    if (needed) {
+        info = info + " " + cockpit.format(_("About $0 are needed to save the memory state of the machine."),
+                                           formatWithBestUnit(needed));
+    }
+
     return (
-        <FormGroup id="snapshot-create-dialog-memory-path" label={_("Memory file")}>
+        <FormGroup id="snapshot-create-dialog-memory-location" label={_("Memory save location")}>
```
martinpitt, line 116: Question for @garrett: Is "save location" legit English? Also, "location" is a bit ambiguous, that could mean both full file path or directory.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
             <FileAutoComplete
-                onChange={value => onValueChanged("memoryPath", value)}
-                superuser="try"
+                onChange={value => onValueChanged("memoryLocation", value)}
+                value={memoryLocation}
                 isOptionCreatable
-                value={memoryPath} />
-            <FormHelper helperTextInvalid={validationError} />
+                onlyDirectories
```
martinpitt, line 121: You may also want `isOptionCreatable="true"`? (a bit unhappily named) -- It's passed as `isCreatable` to the `<Select>`, not sure what difference it makes.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -107,11 +151,12 @@ export class CreateSnapshotModal extends React.Component {
         // cut off seconds, subseconds, and timezone
         const now = new Date().toISOString()
                 .replace(/:[^:]*$/, '');
-        const snapName = props.vm.name + '_' + now;
+        const snapName = now;
```
martinpitt, line 154: The commit message mentions the "what": "The default snapshot name does not include the VM name", but not the "why". Is the VM name already included in the path anyway, also for disk snapshots? I find it quite important to have it in the *files*, so that admins can use shell tools to associate them to VMs. Of course the UI doesn't need to repeat the VM name in the snapshot name. 

Update to myself: It gets set in `mpath` in line 216 below.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -139,29 +194,38 @@ export class CreateSnapshotModal extends React.Component {
...
+                        snapshotCreate({
+                            vm,
+                            name,
+                            description,
+                            isExternal,
+                            memoryPath: mpath,
+                        }))
                     .then(() => {
-                        // VM Snapshots do not trigger any events so we have to refresh them manually
+                    // VM Snapshots do not trigger any events so we have to refresh them manually
```
martinpitt, line 228: Oopsie, please revert (where's our linters for such stuff?)

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -175,19 +239,25 @@ export class CreateSnapshotModal extends React.Component {
...
             <Form onSubmit={e => e.preventDefault()} isHorizontal>
                 <NameRow name={name} validationError={validationError.name} onValueChanged={this.onValueChanged} />
                 <DescriptionRow description={description} onValueChanged={this.onValueChanged} />
                 {isExternal && vm.state === 'running' &&
-                    <MemoryPathRow memoryPath={memoryPath} onValueChanged={this.onValueChanged}
-                                   validationError={validationError.memory} />}
+                    <MemoryLocationRow memoryLocation={memoryLocation} onValueChanged={this.onValueChanged}
+                                       validationError={validationError.memory}
+                                       available={available}
+                                       needed={(vm.rssMemory || vm.currentMemory) * 1024} />}
```
martinpitt, line 260: Did you measure this for a few instances, is that accurate? Or does it need some factor, either > 1 for some overhead, or even < 1 because it gets compressed or created as sparse file or so?

src/helpers.js
```diff
@@ -145,6 +145,12 @@ export function convertToUnitVerbose(input, inputUnit, outputUnit) {
     return result;
 }
 
+export function formatWithBestUnit(input, inputUnit) {
+    const best = convertToBestUnit(input, inputUnit || units.B);
+    const decimals = best.value >= 100 ? 0 : best.value >= 10 ? 1 : 2;
+    return cockpit.format("$0 $1", best.value.toFixed(decimals), best.unit);
+}
```
martinpitt, line 152: "meh". The existing converToBestUnit() should die, die, die: it is base-2 (and thus inconsistent with all other cockpit pages which are base-10), and just duplicates the logic of `cockpit.format_number()`. I see zero reason to not just use the latter for the dialog, and then we should make a separate PR to get rid of all these helpers in machines. You don't even use `inputUnit` from anywhere.

test/check-machines-snapshots
```diff
@@ -469,6 +482,101 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
         self.assertEqual("no", m.execute("virsh snapshot-info --domain subVmTest1 --snapshotname snapshotB | grep 'Current:' | awk '{print $2}'").strip())
         self.assertEqual("yes", m.execute("virsh snapshot-info --domain subVmTest1 --snapshotname snapshotC | grep 'Current:' | awk '{print $2}'").strip())
 
+    def testMemoryLocation(self):
```
martinpitt, line 485: It would be good to test this for system and user. The latter has a different hardcoded default, there's permissions and SELinux involved and such. I'm ok with that being a follow-up, as we don't test current snapshots for users either.

test/check-machines-snapshots
```diff
@@ -469,6 +482,101 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
...
+        # Check creation of new directories
+        b.set_input_text("#snapshot-create-dialog-memory-location input", "/var/lib/libvirt/new")
+        b.wait_text("#snapshot-create-dialog-memory-location li button", 'Create "/var/lib/libvirt/new"')
+        b.click("#snapshot-create-dialog-memory-location li button")
+
+        # Check available space detection
+        b.wait_in_text("#snapshot-create-dialog-memory-location .pf-v5-c-helper-text__item-text",
+                       "available at this location")
+        b.wait_in_text("#snapshot-create-dialog-memory-location .pf-v5-c-helper-text__item-text",
+                       "are needed to save the memory state of the machine.")
```
martinpitt, line 521: Please check that this has variant "default"

martinpitt (dismissed):
Thanks! This is a great usability improvement, and by and large this looks great to me. I have a bunch of nitpicks, and I'd wait for @garrett 's design review here as well.

This is significant enough to warrant a release note IMHO.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -17,7 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 import cockpit from "cockpit";
-import React from "react";
+import React from 'react';
```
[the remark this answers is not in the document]
mvollmer: Fair!

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +69,80 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
     const devices = getSortedBootOrderDevices(vm).filter(d => d.bootOrder &&
                                                               d.device.device === "disk" &&
                                                               d.device.type === "file" &&
                                                               d.device.source.file);
     if (devices.length > 0) {
         const primaryDiskPath = devices[0].device.source.file;
-        const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        const directory = dirname(primaryDiskPath);
+        return directory;
```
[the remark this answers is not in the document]
mvollmer: Done.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
mvollmer: Yes, good point.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
mvollmer: It was already there (and still is).

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
mvollmer: > It gets set in mpath in line 216 below.

Yes, exactly.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
mvollmer: Done!

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
martinpitt: Whoops!

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
mvollmer: > Did you measure this for a few instances, is that accurate?

I did measure it, but I was a while and I should do it again and record the results.

src/helpers.js
mvollmer: Yeah... I was sticking with existing practices, but I am happy to use the official functions. I didn't pay attention to the recent overhaul... would I just use `cockpit.format_bytes(input)` here?

test/check-machines-snapshots
mvollmer: Done. We have to check that it is not a warning.

src/helpers.js
martinpitt: Yes, that automatically selects the right unit, in base-10. That should be fine for this dialog.

test/check-machines-snapshots
```diff
@@ -469,6 +477,120 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
...
+
+        # Check that default comes from the previous snapshot
+        b.wait_val("#snapshot-create-dialog-memory-location input", snapshot_dir + "/new")
+
+        # Use a different location so that we have two snapshots and
+        # can test the sorting function that determines the most
+        # recent one. To check another edge case, we use "/" as the
+        # memory save location.
+
+        snapshot_dir2 = "/" if connection == "system" else "/home/admin/.local/share/libvirt/2"
```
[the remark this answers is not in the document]
mvollmer: I have added path-browserify.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
                                                               d.device.source.file);
     if (devices.length > 0) {
-        const primaryDiskPath = devices[0].device.source.file;
-        const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        return dirname(devices[0].device.source.file);
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory";
```
cockpituous, line 89: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory";
```
cockpituous, line 91: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser: "try" })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 136: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+                // channel has already logged the error
+                callback(null);
```
cockpituous, line 138: This added line is not executed by any test.

package.json
```diff
@@ -55,12 +56,12 @@
     "@xterm/addon-canvas": "0.7.0",
     "@xterm/xterm": "5.5.0",
     "dequal": "2.0.3",
+    "path-browserify": "^1.0.1",
```
martinpitt, line 59: No hats please, we want to pin down versions.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+                        variant={validationError ? "error" : info_variant} />
         </FormGroup>
     );
 };
 
+function get_available_space(path, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser: "try" })
```
martinpitt, line 131: Might be worth conditionalizing the "try" on the system connection, and drop it (or set to False) for the session one?

package.json
mvollmer: Yes, done.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
mvollmer: Yes, good point.

martinpitt (approved):
Cheers!

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, superuser, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 136: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
garrett: It's very odd sounding

garrett:
I'm trying to figure out the context of this. I think we talked about this before, but I can't find it. I don't know what this was like before compared to now, either. Trying to figure it out now, but it'll take some time.

garrett:
OK, so the current UI is this?

image

Where we have a bogus path to a file that doesn't work by default and is so long you can't see where it is?

garrett:
OK, now I've switched branches, did a git clean, and make devel-install, and I get this:

image

It's better, as I can actually see the location. But it still fails by default, which isn't good. I don't know why it's failing either, or what to do.

image

Why does it complain about this path not existing?

And shouldn't it be something like "Snapshot path" or "Snapshot directory" instead of "Memory save location"?

garrett:
System VMs seem better, though, as they at least have a working path:

image

mvollmer:
> But it still fails by default, which isn't good. I don't know why it's failing either, or what to do.

These are bugs in our FileAutocomplete component. There is usually some way of getting around them with creative clicking around.

It's not a regression in this PR, so we shouldn't block on it.

mvollmer:
> And shouldn't it be something like "Snapshot path" or "Snapshot directory" instead of "Memory save location"?

You tell us.

garrett:
> These are bugs in our FileAutocomplete component. There is usually some way of getting around them with creative clicking around.
> 
> It's not a regression in this PR, so we shouldn't block on it.

No, that's not a bug in the autocomplete. The path does not exist locally, but it should create it if that's what's specified. This is the default path for user session machines, which is what my point is, not the autocomplete.

garrett:
Instead of:

- 117 GB available at this location. About 1.09 GB are needed to save the memory state of the machine.

What about this?

- Each memory snapshot will use around 1 GB. Total space available: 117 GB.

Or, better yet, as disks will also take up space, we should split this out from the dropdown and have it below, saying how much disks and how much ram will be used.

- Space for this snapshot: 95 GB. Total space available: 117 GB.

I did have this, but I think it's too granular, as _who cares_ how the disk and memory is split? It's all going to be on the disk one way or another.

- Space for this snapshot: 95 GB (95 GB disk + 1 GB memory). Total space available: 117 GB.

I guess if the memory is stored somewhere other than the same storage device, it might matter more? But then, you're asking for problems. And if they're stored elsewhere, then we probably care about the space that will be used and the free space. And it matters when it's low or not enough, mostly. Otherwise, it's a footnote that doesn't matter too much, right? Although you'd probably want to know how much space each snapshot costs you?

mvollmer:
> saying how much disks and how much ram will be used.

Disk snapshots are implemented via a "copy on write" scheme. A snapshot of a disk starts out needing practically zero bytes, and then grows over time as the disk content at the time of taking the snapshot and the current content diverges.

Thus, disk snapshots don't take any appreciable amount of space at the time of creating the snapshot.

garrett:
> Thus, disk snapshots don't take any appreciable amount of space at the time of creating the snapshot.

Right, but this modal isn't for the first time only, and people update the software on their VMs... sooooo...

mvollmer:
> > Thus, disk snapshots don't take any appreciable amount of space at the time of creating the snapshot.
> 
> Right, but this modal isn't for the first time only, and people update the software on their VMs... sooooo...

Sooooo... you want Cockpit to show the current sizes of snapshots in the list of snapshots?

I think we can do that as a total, like "XXX GB disk space used for snapshots of this machine", but not for individual snapshots. I think it's hard to predict how much space you get back when deleting a snapshot, for example.

garrett:
> Sooooo... you want Cockpit to show the current sizes of snapshots in the list of snapshots?
>
> I think we can do that as a total, like "XXX GB disk space used for snapshots of this machine", but not for individual snapshots. I think it's hard to predict how much space you get back when deleting a snapshot, for example.

Oh, that's not what I was talking about, but that'd be neat and useful.

I was saying that each snapshot takes a certain size, and we need to check that there's space for it on disk, so we should also show it too. (You're already doing this somewhat.) But that snapshots _also_ take disk space, so we should show the total size.

I guess what you're saying is that whatever happens on disk happens transparently though, _when things diverge by deleting or changing things after a snapshot_, the disk grows? But it doesn't grow at the time of the snapshot? And that the only thing that gets larger _at the time of the snapshot_ would be the storage of RAM? So, effectively, at the time of snapshotting it would be _always_ be 0 GB disk + whatever the RAM usage is?

garrett:
Generally, it's better to prefer non-jargon text when possible and to avoid exposing implementation details.

However, if we are going to do this bonkers thing to let people shoot themselves in the foot by making snapshots with mandatory backing files in completely different locations, and we have very limited space due to Cockpit having left-side labeling and PatternFly still not having that flexible (so it forces words to wrap or causes big gaps, depending on the form), then I guess we might as well say "RAM state" or "RAM file"... instead of memory? libvirt calls these files "RAM state" too, according to https://wiki.libvirt.org/Snapshots.html

(Note: PF uses top labelling by default. We're the outliers. GNOME uses a mix, with labels on top for section groups and on the left for individual widgets.)

I'm not done yet, but I started to work on a mockup for this... you can already see how we are so space constrained with left side labelling in PatternFly, otherwise it'd wrap if we did want to say something like "Memory state directory". You can see the difference if we had top labelling (at least in places), like this:

image

And here's if we wanted longer text, it'd be something more like this:

image

...and some languages (I'm specifically thinking of German, but it's not the only one) are often longer than English, so we'd have the wrapping or too-long issues even when things are translated.

For example, here's a humorously bad German translation for that already long string :wink: :

image


(Yes, yet again, I'm asking for us to start adopting top labelling for modals. We should finally just do it, and port over as many as it makes sense. It wouldn't be horrible to mix and match where appropriate, but we should try to have most similar. Some modals probably would work better with side labelling versus top, but many work better with top labelling. The alternative would be to adopt subgrid (or use display: contents and adjust the grid) to fix PF's static widths for side labelling.)

garrett:
Previous post was basically me saying that our approach to labelling is really cramping what we can do with respect to friendly labelling of fields, with examples.

We can try to find something short, but we'll probably have to sacrifice the best option for the text and ignore the translations... or decide to use top labelling (and make an effort to switch Cockpit over in other places, where we can, outside of this PR naturally).

mvollmer:
> So, effectively, at the time of snapshotting it would be _always_ be 0 GB disk + whatever the RAM usage is?

Yes.

mvollmer:
> Generally, it's better to prefer non-jargon text when possible and to avoid exposing implementation details.

Yeah. We can also omit the "RAM state location" completely and always just use a sensible default.

mvollmer:
> I'm asking for us to start adopting top labelling for modals. 

Alright, let's do it! :-)

mvollmer:
> libvirt calls these files "RAM state" too, according to https://wiki.libvirt.org/Snapshots.html
> [...]
> if we did want to say something like "Memory state directory".

[Here](https://libvirt.org/kbase/domainstatecapture.html) is the high-level description of the various snapshotting things that libvirt can do, and it doesn't mention "RAM", only "memory", and more specifically calls it "guest memory".

Cockpit only uses "memory" in its UI and not "RAM".

So we should stick with "memory".

"Guest memory state file directory" or "Guest memory state file location" would be most descriptive (if a bit ridiculous) and with top labels we have the space.

I'll change it to "Memory state file location", and change this dialog to use top-labels. All others are done in #1804.

mvollmer:
> I'll change it to "Memory state file location", and change this dialog to use top-labels. All others are done in #1804.

Done, and updated the screenshot in the description.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
-    // Choosing a default path where memory snapshot should be stored might be tricky. Ideally we want
-    // to store it in the same directory where the primary disk (the disk which is first booted) is stored
-    // If however no such disk can be found, we should fallback to libvirt's default /var/lib/libvirt
+function getDefaultMemoryLocation(vm) {
+    // If we find an existing external snapshot, use it's memory path
+    // as the default. Otherwise, try to find the primary disk and use
+    // it's location. If that fails as well, use a reasonable hard
+    // coded value.
+
+    for (const s of vm.snapshots.sort((a, b) => b.creationTime - a.creationTime)) {
```
cockpituous, line 76: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@
...
+        if (s.memoryPath)
+            return dirname(s.memoryPath);
```
cockpituous, line 78: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@
...
                                                               d.device.source.file);
     if (devices.length > 0) {
-        const primaryDiskPath = devices[0].device.source.file;
-        const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        return dirname(devices[0].device.source.file);
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory";
```
cockpituous, line 89: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory";
```
cockpituous, line 91: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@
...
+    let info_variant = "default";
+
+    if (needed) {
+        info = cockpit.format(_("Memory snapshot will use about $0."),
+                              cockpit.format_bytes(needed));
+    }
+    if (available) {
+        info = info + " " + cockpit.format(_("Total space available: $0."), cockpit.format_bytes(available));
+        if (needed && available * 0.9 < needed)
+            info_variant = "warning";
```
cockpituous, line 108: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@
...
+function get_available_space(path, superuser, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 136: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@
...
+                // channel has already logged the error
+                callback(null);
```
cockpituous, line 138: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -175,19 +237,38 @@
...
     render() {
         const Dialogs = this.context;
         const { idPrefix, isExternal, vm } = this.props;
-        const { name, description, memoryPath } = this.state;
+        const { name, description, memoryLocation, available } = this.state;
         const validationError = this.onValidate();
 
         const body = (
-            <Form onSubmit={e => e.preventDefault()} isHorizontal>
+            <Form onSubmit={e => e.preventDefault()}>
```
cockpituous, line 264: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 89: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 91: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 136: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 138: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 264: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 89: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 91: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, superuser, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 136: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +67,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+                // channel has already logged the error
+                callback(null);
```
cockpituous, line 138: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -175,19 +237,38 @@ export class CreateSnapshotModal extends React.Component {
...
     render() {
         const Dialogs = this.context;
         const { idPrefix, isExternal, vm } = this.props;
-        const { name, description, memoryPath } = this.state;
+        const { name, description, memoryLocation, available } = this.state;
         const validationError = this.onValidate();
 
         const body = (
-            <Form onSubmit={e => e.preventDefault()} isHorizontal>
+            <Form onSubmit={e => e.preventDefault()}>
```
cockpituous, line 264: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 89: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 91: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 136: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 138: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 264: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 89: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 91: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 136: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 138: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 264: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +66,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
                                                               d.device.source.file);
     if (devices.length > 0) {
-        const primaryDiskPath = devices[0].device.source.file;
-        const directory = primaryDiskPath.substring(0, primaryDiskPath.lastIndexOf("/") + 1);
-        return directory + snapName;
+        return dirname(devices[0].device.source.file);
     } else {
         if (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION)
-            return "/var/lib/libvirt/memory/" + snapName;
+            return "/var/lib/libvirt/memory";
```
cockpituous, line 88: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +66,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
         else if (current_user)
-            return current_user.home + "/.local/share/libvirt/memory/" + snapName;
+            return current_user.home + "/.local/share/libvirt/memory";
```
cockpituous, line 90: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +66,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, superuser, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 135: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +66,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+                // channel has already logged the error
+                callback(null);
```
cockpituous, line 137: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -175,19 +236,38 @@ export class CreateSnapshotModal extends React.Component {
...
     render() {
         const Dialogs = this.context;
         const { idPrefix, isExternal, vm } = this.props;
-        const { name, description, memoryPath } = this.state;
+        const { name, description, memoryLocation, available } = this.state;
         const validationError = this.onValidate();
 
         const body = (
-            <Form onSubmit={e => e.preventDefault()} isHorizontal>
+            <Form onSubmit={e => e.preventDefault()}>
```
cockpituous, line 263: This added line is not executed by any test.

src/helpers.js
```diff
@@ -910,3 +910,13 @@ export function vmSupportsExternalSnapshots(config, vm) {
 export function vmHasVFIOHostDevs(vm) {
     return !!vm.hostDevices.find(hd => hd.driver === "vfio");
 }
+
+export function dirname(path) {
+    const i = path.lastIndexOf("/");
+    if (i < 0)
+        return ".";
```
cockpituous, line 917: This added line is not executed by any test.

src/helpers.js
```diff
@@ -910,3 +910,13 @@ export function vmSupportsExternalSnapshots(config, vm) {
...
+    else if (i == 0)
+        return "/";
```
cockpituous, line 919: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 88: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 90: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +66,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+function get_available_space(path, superuser, callback) {
+    if (!path)
+        callback(null);
+
+    cockpit.script(get_available_space_sh, [path], { superuser })
+            .then(output => {
+                const info = JSON.parse(output);
+                callback(info.free * info.unit);
+            })
+            .catch(exc => {
```
cockpituous, line 135: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -64,41 +66,78 @@ const DescriptionRow = ({ onValueChanged, description }) => {
...
+                // channel has already logged the error
+                callback(null);
```
cockpituous, line 137: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 263: This added line is not executed by any test.

src/helpers.js
cockpituous, line 917: This added line is not executed by any test.

src/helpers.js
cockpituous, line 919: This added line is not executed by any test.

garrett (dismissed):
I think this PR is fine to merge.

But we should consider if we want to move to top-labelling overall and have side labelling be the exception, or have top-labelling be the exception. Naturally, things should be internally consistent (that is: the same in the same context; for example: this modal should have all labels be the same way, like it does).

Considering PatternFly uses top labeling by default, that's another argument aside from strings, to use that.

test/check-machines-snapshots
```diff
@@ -469,6 +477,117 @@ class TestMachinesSnapshots(machineslib.VirtualMachinesCase):
...
+                       "Total space available: ")
+
+        # Go ahead and let it fail
+        b.click(".pf-v5-c-modal-box__footer button:contains(Create)")
+        b.wait_in_text("#vm-subVmTest1-create-snapshot-modal .pf-v5-c-alert",
+                       "No space left on device")
+        b.click(".pf-v5-c-modal-box__footer button:contains(Cancel)")
+        b.wait_not_present("#vm-subVmTest1-create-snapshot-modal")
+
+    def testMemoryLocationSession(self):
```
jelly, line 588: What about testing the system session?

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -139,27 +190,37 @@ export class CreateSnapshotModal extends React.Component {
...
-            })
+            let mpath = null;
+            if (isExternal && vm.state === "running" && memoryLocation) {
+                mpath = memoryLocation;
+                if (mpath[mpath.length - 1] != "/")
+                    mpath = mpath + "/";
+                mpath = mpath + vm.name + "." + name + ".save";
+            }
+            const superuser = (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION);
+            cockpit.spawn(["mkdir", "-p", memoryLocation], { superuser, err: "message" })
```
jelly, line 215: Technically `superuser` is not a boolean option but it 'try' or 'require' https://cockpit-project.org/guide/latest/cockpit-spawn.html

I think cockpit-ws still has code to convert true into "require" but depending on that fact feels wrong.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -139,27 +190,37 @@ export class CreateSnapshotModal extends React.Component {
...
-                vm,
-                name,
-                description,
-                isExternal,
-                memoryPath: isExternal && vm.state === "running" && memoryPath,
-            })
+            let mpath = null;
+            if (isExternal && vm.state === "running" && memoryLocation) {
+                mpath = memoryLocation;
+                if (mpath[mpath.length - 1] != "/")
```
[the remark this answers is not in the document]
jelly: Can be left out :)

test/check-machines-snapshots
mvollmer: Hmm, the `testMemoryLocation` test runs twice, once for the "system" connection and once for the "session" connection. Is there also a "system session" connection?

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
mvollmer: Ok! I wasn't sure which one is the deprecated way, and which one isn't. Thanks for clearing that up!

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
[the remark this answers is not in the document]
mvollmer: Good point!

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 88: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 90: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 135: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 137: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
```diff
@@ -139,27 +190,37 @@ export class CreateSnapshotModal extends React.Component {
...
-                memoryPath: isExternal && vm.state === "running" && memoryPath,
-            })
+            let mpath = null;
+            if (isExternal && vm.state === "running" && memoryLocation) {
+                mpath = memoryLocation;
+                if (!mpath.endsWith("/"))
+                    mpath = mpath + "/";
+                mpath = mpath + vm.name + "." + name + ".save";
+            }
+            const superuser = (vm.connectionName === LIBVIRT_SYSTEM_CONNECTION) ? "require" : false;
```
cockpituous, line 214: This added line is not executed by any test.

src/components/vm/snapshots/vmSnapshotsCreateModal.jsx
cockpituous, line 263: This added line is not executed by any test.

src/helpers.js
cockpituous, line 917: This added line is not executed by any test.

src/helpers.js
cockpituous, line 919: This added line is not executed by any test.

test/check-machines-snapshots
jelly: Ah this is just an "unusual" way of doing it, I suppose which confused me, in podman we just introduce a helper like `_testMemoryLocation()` anyway not a blocker

test/machineslib.py
```diff
@@ -410,7 +410,7 @@ def stop_all():
                       '    echo "$out" | grep -q "domain is not running"; '
                       "  fi; done")
             m.execute("runuser -l admin -c 'for d in $(virsh -c qemu:///session list --all --name); do "
-                      "virsh -c qemu:///session undefine $d; done'")
+                      "virsh -c qemu:///session undefine $d --snapshots-metadata; done'")
```
jelly, line 413: Doesn't the system session need this to then?

jelly:
Looks fairly good to me, but on ubuntu there are some test flakes?

test/machineslib.py
mvollmer: Apparently not.  For the system connection we use `virsh destroy`, but for the session connection we do `virsh undefine`. I am not sure why the `undefine` is needed. `/home/admin/.local/share/libvirt/` is restored after each test, doesn't that make the session machines go away? *shrug*

mvollmer:
> Looks fairly good to me, but on ubuntu there are some test flakes?

Yes, let's see.

test/machineslib.py
jelly: I'm going to leave this for an expert :)

Issue #764, opened 2022-07-19
Reword "Previously taken snapshots allow you to revert to an earlier state if something goes wrong"


martinpitt:
Additionally, "Previously taken snapshots allow you to revert to an earlier state if something goes wrong"?!?!?

But there are no snapshots that have been previously taken. It's an empty state!

This also needs rewording. Suggestion:

"Snapshots are used to revert a VM to a earlier state"

...although it's already been a long day and it's hot, so I'm second-guessing my brain. Regardless, I think these two suggestions (this and the one in the comment above) are improvements. We can talk about it tomorrow if you have improvements on them.

_Originally posted by @garrett in https://github.com/cockpit-project/cockpit-machines/issues/756#issuecomment-1189190283_

PR #1751, opened 2024-08-01
#764 issue: Reword "Previously taken snapshots allow you to.."

lakshayletsgo:


martinpitt:
Sorry, wrong button..

po/cs.po
```diff
@@ -2068,8 +2068,7 @@ msgstr "Délku předpony je třeba vyplnit"
 
 #: src/components/vm/snapshots/vmSnapshotsCard.jsx:236
 msgid ""
-"Previously taken snapshots allow you to revert to an earlier state if "
-"something goes wrong"
+"Snapshots are used to revert a VM to a earlier state"
```
martinpitt, line 2071: Please drop all the po changes. The original string changes, so we can't keep the translations as-is. Also, translations don't work that way, they are maintained in [weblate](https://translate.fedoraproject.org/projects/cockpit-machines/main/).

src/components/vm/snapshots/vmSnapshotsCard.jsx
```diff
@@ -233,7 +233,7 @@ export class VmSnapshotsCard extends React.Component {
                           gridBreakPoint='grid-lg'
                           variant="compact"
                           emptyCaption={_("No snapshots defined for this VM")}
-                          emptyCaptionDetail={_("Previously taken snapshots allow you to revert to an earlier state if something goes wrong")}
+                          emptyCaptionDetail={_("Snapshots are used to revert a VM to a earlier state"  ")}
```
martinpitt, line 236: Please remove the duplicate `"` and extra space.

Also, it is "a*n* earlier state"

martinpitt (changes requested):
Thanks! The string needs a review by @garrett , I'm not a native speaker myself.

garrett:
This is super weird in context:

 
image

Firstly, the empty states are very repetitive and shouldn't say "for this VM" everywhere. We'll have to handle that in another PR.

For Snapshots specifically, if we're even going to have sub text in the empty state, we should say something more like:

> No existing snapshots
> Create a snapshot to roll back to a previous state

But we can probably even drop the "No existing snapshots" and just have "_Create a snapshot to roll back to a previous state_". WDYT?

(Explanation: "defined" and "VM" are jargony; "this VM" is redundant on the context of that page, especially when there are multiple empty categories.)

garrett:
Alternatively, something closer to what you already have (at least for the second line):

> No snapshots exist
> Revert to an earlier state using snapshots

mvollmer:
> Firstly, the empty states are very repetitive and shouldn't say "for this VM" everywhere. We'll have to handle that in another PR.
> 
> For Snapshots specifically, if we're even going to have sub text in the empty state, we should say something more like:
> 
> > No existing snapshots
> > Create a snapshot to roll back to a previous state

I have done this in #2307.  This PR here seems to have stalled.

PR #1757, opened 2024-08-07
fmf: Disable SELinux on RHEL 10 on current policy version

martinpitt:
RHEL-46893 may or may not be fixed, but now
https://issues.redhat.com/browse/RHEL-49567 completely ruins the show.

----

Recently, the centos-10 test has failed like this: https://artifacts.dev.testing-farm.io/7b011b67-48ba-4890-84d7-0ed15978ee73/ Let's hope not too much else has broken since then :crossed_fingers:

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -27,6 +27,10 @@ case "$(rpm -q selinux-policy)" in
     selinux-policy-40.13.4-1.el10.noarch|selinux-policy-40.13.5-1.el10.noarch)
         setenforce 0 ;;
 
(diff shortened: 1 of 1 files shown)


martinpitt:
OK, this moved from "everything fails" to "one test fails" (TestMachinesCreate.testCreateDownloadAnOS). I'd call that an improvement. So I suggest to land this now, and I'll look at that test as a next step.

tomasmatus (approved):
> So I suggest to land this now, and I'll look at that test as a next step.

sounds good to me!

PR #1760, opened 2024-08-08
test: Fix failures for unexpected `virt-host-validate qemu`

martinpitt:
For a few months now, the RHEL downstream gating tests for cockpit-machines have all failed like this:

```
  File "/source/test/machineslib.py", line 38, in waitPageInit
    virtualization_enabled = "PASS" in self.machine.execute("virt-host-validate qemu | grep 'Checking for hardware virtualization'")
subprocess.CalledProcessError
```

The actual output shows

> QEMU: Checking if device '/dev/kvm' exists: FAIL

and no "hardware virtualization" at all, which makes the grep fail. Treat an absent line as "fail".


martinpitt:
OK, that's the second try after #1759, I accept defeat -- no Fedora COPR today..

mvollmer (approved):
Thanks!

PR #1761, opened 2024-08-11
test: Move from CDP to BiDi

martinpitt:
See https://github.com/cockpit-project/cockpit/pull/20832

Explicitly add a "glob" devDependency to follow suit with
https://github.com/cockpit-project/cockpit/commit/680decc155a
It was previously used implicitly through a transient dependency of
something else, but our esbuild po-plugin uses it explicitly.

This causes some minor pixel test noise, adjust the references.

 - [x] https://github.com/cockpit-project/cockpit/pull/20883

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
-COCKPIT_REPO_COMMIT = b5a98be3e0601b45cd8a557cde1c80b896a903fb # 322 + 19 commits
+COCKPIT_REPO_COMMIT = 83e1a5ac16d8eb8df60f90c8fa502a0d8bf21365 # 322 + 35 commits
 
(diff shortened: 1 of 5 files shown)


martinpitt:
```
make[1]: Leaving directory '/work/make-checkout-workdir'
✘ [ERROR] Could not resolve "@xterm/addon-canvas"
```

This is due to https://github.com/cockpit-project/cockpit/pull/20862 . We use c-c-terminal in serialConsole.jsx, so we need to follow suit here.

martinpitt:
Fixed in #1762, rebased on that.

martinpitt:
OK, rest are flakes and pixel refs

martinpitt:
Applying the [pixel test noise](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1761-de074263-20240813-101554-fedora-40/log.html)

tomasmatus (approved):
🥳

martinpitt:
That was the last one of ours!

PR #1762, opened 2024-08-12
package.json: Switch to @xterm/*

martinpitt:
These are the new names, and they have more recent releases.

See https://github.com/cockpit-project/cockpit/commit/c9096a71671be7 Follow suit.

----

Spotted in #1761 when trying to bump cockpit lib import.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,7 +34,7 @@ updates:
           - "types*"
       xterm:
         patterns:
-          - "xterm*"
+          - "@xterm*"
 
   - package-ecosystem: "github-actions"
     directory: "/"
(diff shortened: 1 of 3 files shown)


mvollmer:
Thanks!

martinpitt:
@mvollmer retrying c10 doesn't work, it's reliably broken (and I couldn't repro it with our bots image). That's on the pilot board.

mvollmer:
> @mvollmer retrying c10 doesn't work, it's reliably broken (and I couldn't repro it with our bots image). That's on the pilot board.

Ah, thanks!

mvollmer (commented):
Thanks!

mvollmer (approved):

PR #1763, opened 2024-08-12
test: Disable SELinux when RHEL-49567 is present

mvollmer:


mvollmer:
This makes centos-10 pretty green.

martinpitt (approved):
Thanks!

mvollmer:
Uff, breaks on non-rpm systems of course.

mvollmer:
> Uff, breaks on non-rpm systems of course.

Actually, it doesn't... :-/

mvollmer:
This has now reproduced the TestMachinesCreate.testCreateDownloadAnOS failure that we see on TF in our own CI.

PR #1765, opened 2024-08-13
test: Also wait for virt-install to finish on centos-10

mvollmer:


mvollmer:
Aha, TF centos-10-stream is green.

mvollmer:
The (guessed) fix makes it reliably green: https://cockpit-logs.us-east-1.linodeobjects.com/pull-1765-67f81cd5-20240813-112925-centos-10/log.html

martinpitt:
If that works, I'm all for it! I had to make it conditional back then as waiting timed out on several OSes. I've never fully understood how `virt-install` works -- sometimes it leaves the VM running, sometimes not, I found it rather unpredictable. Perhaps by now all OSes have a new enough virt-install.

Thanks!

martinpitt:
Yeah, what I feared -- this needs to stay conditionalized. Perhaps there's a better criterion, like the virt-install version? But for now just extending the OS list seems ok.

mvollmer:
> Yeah, what I feared -- this needs to stay conditionalized. Perhaps there's a better criterion, like the virt-install version? But for now just extending the OS list seems ok.

Ah, I see, bummer. So the virt-install process itself sticks around forever? I have _one_ (1) closer look and then I make this conditional.

mvollmer:
On some OSes the created VM crashes in the
kernel and virt-install never exits, on others, the BIOS can't find a bootable
device but virt-install does exit. *shrug*

Maybe we should make it so that virt-install can actually produce a working VM.

mvollmer:
> the BIOS can't find a bootable device but virt-install does exit.

The might be the state after a "successful" installation of our fake Fedora 28 OS. So virt-install is indeed done at this point.

mvollmer:
Oh, the flakes...

test/check-machines-create
```diff
@@ -1328,8 +1328,12 @@ vnc_password= "{vnc_passwd}"
...
                     self.assertIn(user_login, self.user_login)
 
-            # HACK: wait for virt-install to finish; this only races on tumbleweed
-            if self.machine.image == 'opensuse-tumbleweed' and self.create_and_run:
+            # HACK: on some OSes we need to wait for virt-install to
+            # exit; on others, it wont ever exit but that's fine as
+            # well.  (On some OSes the created VM crashes in the
+            # kernel, on others, the BIOS can't find a bootable
+            # device. Maybe that's the difference.)
+            if self.create_and_run and self.machine.image in ['opensuse-tumbleweed', 'centos-10']:
```
martinpitt, line 1336: I don't have proof that this affects rhel-10-0 (or not), as the last gating run in real RHEl 10 all failed due to SELinux. But it almost surely needs the same treatment. I triggered a rhel-10-0 bots run.

martinpitt: The [rhel-10-0 run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1765-7f262a27-20240814-011914-rhel-10-0/log.html) confirms this. I suggest to write this as

    self.machine.image.startswith(('opensuse-tumbleweed', 'centos-10', 'rhel-10'))

so that it also catches future rhel-10-* images.

mvollmer: I have done that, and also increased the global timeout. I think opensuse runs into it.

martinpitt (approved):
Thanks for your investigations! The comment clarifies it a little indeed! (For our poor future selves -- this *will* come back and bite us in the ... a place where it hurts)

mvollmer:
> The comment clarifies it a little indeed!

The installation is not expected to succeed, but I don't know how exactly it is expected to fail. On some OSes, a kernel is clearly started, so we must be downloading something? I think this is worth figuring out and robustifying, eventually.

martinpitt:
I reviewed the failures:
 - The virtqemud crashes in [c10](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1765-e041f378-20240814-102938-centos-10/log.html) and [r10](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1765-e041f378-20240814-102928-rhel-10-0/log.html) are very persistent. We need to report these and hopefully find a CLI reproducer. But unrelated to this PR.
 - [r10 testCopyStorageMigration](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1765-e041f378-20240814-102928-rhel-10-0/log.html#33) is a known naughty in f40, let's copy it -- but that'll be a bots PR.
 - [d-testing also fails testCreateDownloadAnOS](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1765-e041f378-20240814-102934-debian-testing/log.html#78-1)  -- given that it's also a "new" distro similar to tumbleweed and R10, my gut feeling is that we should extend the hack from this PR there. But let's do that in a follow-up.

So in summary, I think this is fine to land, it makes things incrementally better.

Issue #1777, opened 2024-08-21
cockpit-machines still carries libvirt host interface management dependencies


hw-claudio:
Hello,

as we noticed when trying to use cockpit-machines in more minimal OS builds,
the now very old cockpit commit (pre-split of cockpit-machines):

commit 3af2e19fb64bcfe7d4bd277e68c03de2817f85b8
Author: Simon Kobyda <42733240+skobyda@users.noreply.github.com>
Date:   Thu Aug 1 13:14:50 2019 +0200

    machines: Introduce interfaces into store (#12465)
    
    Closes #12465

introduces functionality to get, add and update Host network interfaces using libvirt.
Nowadays the trend seems to be to use the OS dedicated network tools to manage host networks, not libvirt.

Is this functionality really used in cockpit-machine ?
Initial testing would suggest that this feature is not really needed, and forces an unneeded chain of dependencies for modular OS where modular libvirt is chosen.

Any thoughts about removing the call in common.js:458 :
        interfaceGetAll({ connectionName }),
                
?

Any aspects that I missed? Thanks!

jfehlig:
It's my first look at the code, but AFAICT cockpit-machines doesn't do anything with the information it gathers from 'ListInterfaces', and subsequent 'GetXMLDesc' for each interface. It gets stored, but not updated or retrieved. Did I overlook the use of this information? Can someone describe how it is used?

IMO, cockpit-machines should remove the use of libvirt's virInterface* APIs. virt-manager long ago dropped their use. The libvirt community would like to deprecate them

https://listman.redhat.com/archives/libvir-list/2020-December/msg00183.html

The only interface backend common among all distros is the udev backend, which only implements read operations and likely returns inconsistent results across distros (and even distro versions).

At SUSE, we'd like to take advantage of libvirt's modularization efforts and provide a minimal installation that does not include the virtinterfaced daemon. Removing the use of these APIs in cockpit-machines, or making their use configurable, would make it more friendly with minimal (but sufficient) libvirt installations.

BTW, the virInterface* APIs are all about managing network interface devices on the host. As Claudio mentioned, distro tools (NetworkManager, netplan, etc) are better suited for that. Does cockpit already support the creation and management of host eth, vlan, bond, bridge, etc devices? If so, that's where libvirt's virInterface* APIs should be used. But again, I'd suggest dropping their use altogether since they provide no practical value.

martinpitt:
[...]

PR #1779, opened 2024-08-22
libvirtApi remove use of interfaceGetAll

Lunarequest:
this is in reference to #1777, `interfaceGetAll` is only used in one place and in local testing all tests still pass. `GetXMLDesc` would need a larger rework to drop usage but most of its usage should be fine

### DIFF
--- a/src/libvirtApi/common.js
+++ b/src/libvirtApi/common.js
@@ -58,9 +58,6 @@ import {
     timeout,
 } from "../libvirtApi/helpers.js";
 import {
-    interfaceGetAll,
-} from "../libvirtApi/interface.js";
-import {
     networkGet,
     networkGetAll,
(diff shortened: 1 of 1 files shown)


jfehlig:
Seems to me the file src/libvirtApi/interface.js can be removed too.

martinpitt:
Sure, thanks! This will need some more cleanup, but doing a first round of tests. (Note that nobody in the Cockpit team will be working tomorrow, so will revisit this on Monday)

martinpitt (approved):
Thanks! Let's land that, and I'll send a follow-up to clean it up more thoroughly.

martinpitt:
@Lunarequest I'll revert this. See the discussion in #1782 , this needs more work.

PR #1781, opened 2024-08-26
Add client operation logs for cockpit-machines

ohruka:
This feature provides cockpit-machines to output client-side operation logs.

**What it does**
This feature enables logging of guest operations performed using libvirt-dbus.
This log entry will contain information about the operation, including the date, guest uuid, and operation. 

**Example:**
When this feature is enabled, shutting down a guest via cockpit-machines will generate a log entry in the journal, regardless of whether the guest actually shut down. 
```
Aug 14 15:19:51 localhost cockpit-machines-logging[2650]:   Sender=:1.38  Destination=org.libvirt  Path=/org/libvirt/QEMU/domain/_ced394f8_d398_478b_bd4b_aefdb520ad90  Interface=org.libvirt.Domain  Member=Shutdown
```

**Default behavior:**
This feature is disabled by default. Users needing logging can enable it by manually enabling the service (cockpit-machines-logging.service):
```
ohruka@localhost:~/cockpit-machines$ systemctl status cockpit-machines-logging
○ cockpit-machines-logging.service - Cockpit machines logging
     Loaded: loaded (/usr/lib/systemd/system/cockpit-machines-logging.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: inactive (dead)
       Docs: man:cockpit-machines
ohruka@localhost:~/cockpit-machines$
```

### DIFF
--- a/packaging/cockpit-machines.spec.in
+++ b/packaging/cockpit-machines.spec.in
@@ -86,11 +86,23 @@ If "virt-install" is installed, you can also create new virtual machines.
 %make_install PREFIX=/usr
 appstream-util validate-relax --nonet %{buildroot}/%{_datadir}/metainfo/*
 
+# Add cockpit-machines-logging service
+install -Dm644 %{_sourcedir}/src/systemd/cockpit-machines-logging.service %{buildroot}%{_unitdir}/cockpit-machines-logging.service
+install -Dm755 %{_sourcedir}/src/bin/cockpit-machines-logging %{buildroot}%{_bindir}/cockpit-machines-logging
+
+%post
+%systemd_post cockpit-machines-logging.service
+
+%preun
+%systemd_preun cockpit-machines-logging.service
+
 %files
 %doc README.md
 %license LICENSE dist/index.js.LEGAL.txt dist/index.css.LEGAL.txt
 %{_datadir}/cockpit/*
 %{_datadir}/metainfo/*
+%{_unitdir}/cockpit-machines-logging.service
+%attr(0755, root, root) %{_bindir}/cockpit-machines-logging
 
 # The changelog is automatically generated and merged
 %changelog
--- a/src/bin/cockpit-machines-logging
+++ b/src/bin/cockpit-machines-logging
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+busctl monitor --match 'interface=org.libvirt.Domain'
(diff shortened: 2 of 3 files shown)


martinpitt:
> Currently, I believe that cockpit itself does not have logs that record "when and what operations were performed".

Some services do, many in the form of "audit" logs, others in the normal journal. For example, when you stop a service on the "Services" page, that gets an entry in the journal. Same for logging in and out of cockpit (that's the normal PAM logs for starting/stopping a user session. Or if you e.g. stop/start/change a network interface, you get NetworkManger journal logs. Creating a new user also gets logged, but e.g. deleting a Linux user doesn't. But we consider that a bug in PAM or `deluser` -- in our view it's really not Cockpit's responsibility to do this logging, as operations such as `deluser` are also run interactively from admins, or from Ansible, shell scripts, or any number of other places.

> If such a function to record across the board is posted as a patch, is there any room for discussion to incorporate it as a default feature of cockpit?

I'm afraid no, for the reasons above. Doing things in Cockpit, Terminal, Ansible etc. should all be interchangeable and be possible to do side by side. Logging such actions from Cockpit is simply the wrong layer. Logging actions needs to be done by the API we talk to (D-Bus service, Linux subsystem, administrative CLI program, etc.).

This is why I am very concerned about adding this to c-machines: It sets a bad example and raises expectations which we have to refuse in other places. It needs to be *absolutely crystal clear* that such a log is neither a security proof nor reliable -- it is a debugging tool, nothing else. 

> by narrowing down the output to interface=org.libvirt.Domain, basically only records of operations in the machines environment should be output

Yes, that's a given, but my "firehose" comment really applied to that restriction. It has all the read operations such as `GetXMLDesc`, `GetStats`, `ListDomainSnapshots` etc, or signals. So to be usable, this would need some post processing and knowledge which API calls are actually changing things.

That's why I'm not a fan of a plain `busctl`, and letting this run in a permanent service -- it'll accumulate gigabytes of logs which clutter the journal and make it much less useful for other tasks (admins look at `journalctl` very often) -- for that reason I also retract my request to log this into the journal -- it should really go into a separate file with configured log rotation. Sorry about the back and forth.

ohruka:
Thank you for your comments.
Based on the feedback we received, we've decided not to add the client operation logging feature to cockpit-machines.
Therefore, I'm closing this pull request.

PR #1782, opened 2024-08-26
Drop usage of virtinterfaced

### LINKED ISSUE #1777: cockpit-machines still carries libvirt host interface management dependencies

martinpitt:
virtinterfaced is being deprecated [1], and at least some distros want
to get rid of the dependency.

Replace it with reading `ip --json a` instead  to get all the interface
names.

Fixes #1777

[1] https://listman.redhat.com/archives/libvir-list/2020-December/msg00183.html

---

 - [x] Goes on top of PR #1785 for cleanliness
 - [x] Confirm with @KKoukiou that this is really legit

### DIFF
--- a/src/libvirtApi/interface.js
+++ b/src/libvirtApi/interface.js
@@ -21,55 +21,27 @@
  * Provider for Libvirt using libvirt-dbus API.
  * See https://github.com/libvirt/libvirt-dbus
  */
+import cockpit from 'cockpit';
+
 import store from '../store.js';
 
 import { updateOrAddInterface } from '../actions/store-actions.js';
-import { parseIfaceDumpxml } from '../libvirt-xml-parse.js';
-import { call, Enum, timeout } from './helpers.js';
 
-/*
- * Read properties of a single Interface
- *
- * @param {object} objPath interface object path
- * @param {string} connectionName
- */
-export async function interfaceGet({
-    id: objPath,
-    connectionName,
-}) {
-    const props = {};
+export async function interfaceGetAll() {
+    let ifaces = [];
 
     try {
-        const [resultProps] = await call(connectionName, objPath, 'org.freedesktop.DBus.Properties', 'GetAll',
-                                         ['org.libvirt.Interface'], { timeout, type: 's' });
-        /* Sometimes not all properties are returned; for example when some network got deleted while part
-            * of the properties got fetched from libvirt. Make sure that there is check before reading the attributes.
-            */
-        if ("Active" in resultProps)
-            props.active = resultProps.Active.v.v;
-        if ("MAC" in resultProps)
-            props.mac = resultProps.MAC.v.v;
-        if ("Name" in resultProps)
-            props.name = resultProps.Name.v.v;
-        props.id = objPath;
-        props.connectionName = connectionName;
-
-        const [xml] = await call(connectionName, objPath, 'org.libvirt.Interface', 'GetXMLDesc', [0], { timeout, type: 'u' });
-        const iface = parseIfaceDumpxml(xml);
-        store.dispatch(updateOrAddInterface(Object.assign(props, iface)));
+        const ipData = await cockpit.spawn(["ip", "--json", "a"], { err: "message" });
+        ifaces = JSON.parse(ipData);
     } catch (ex) {
-        console.log('listInactiveInterfaces action for path', objPath, ex.toString());
+        console.warn("Failed to get interfaces with ip command:", ex.toString());
     }
-}
 
-export async function interfaceGetAll({ connectionName }) {
-    const flags = Enum.VIR_CONNECT_LIST_INTERFACES_ACTIVE | Enum.VIR_CONNECT_LIST_INTERFACES_INACTIVE;
-
-    try {
-        const [ifaces] = await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'ListInterfaces', [flags], { timeout, type: 'u' });
-        await Promise.all(ifaces.map(path => interfaceGet({ connectionName, id: path })));
-    } catch (ex) {
-        console.warn('getAllInterfaces action failed:', ex.toString());
-        throw ex;
+    for (const iface of ifaces) {
+        store.dispatch(updateOrAddInterface({
+            name: iface.ifname,
+            MAC: iface.address,
+            Active: iface.operstate === "UP",
+        }));
     }
 }
--- a/test/check-machines-networks
+++ b/test/check-machines-networks
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
+import json
 import xml.etree.ElementTree as ET
 
 import machineslib
@@ -25,16 +26,12 @@ from machinesxmls import TEST_NETWORK2_CLASH_XML, TEST_NETWORK2_XML, TEST_NETWOR
 
 
 def getNetworkDevice(m):
-    net_devices_str = m.execute("virsh iface-list")
-    net_devices_str = net_devices_str.split("\n", 2)[2]  # Remove first 2 lines of table header
-    if net_devices_str not in ["\n", "\r\n"]:
-        device = net_devices_str.split(' ', 2)[1]  # Get the name of device, ignoring spacing before device string
-    else:  # If $virsh-iface list did not return any device, check virsh nodedev-list net:noh
-        net_devices_str = m.execute("virsh nodedev-list net")
-        net_devices_str = net_devices_str.split("\n", 2)[0]
-        device = net_devices_str.split("_", 2)[1]  # Ignore prefix (example: net_enp0s31f6_8c_16_45_5f_77_34)
+    # To replicate `virsh iface-list` behavior, filter active interfaces
+    # from ip command and get the first active one.
+    routes = json.loads(m.execute("ip --json a"))
+    active = list(filter(lambda r: r['operstate'] == "UP", routes))
 
-    return device
+    return active[0]["ifname"]
 
 
 @testlib.nondestructive
--- a/packaging/cockpit-machines.spec.in
+++ b/packaging/cockpit-machines.spec.in
@@ -43,7 +43,6 @@ Requires: libvirt-daemon-driver-qemu
 Requires: libvirt-daemon-driver-network
 Requires: libvirt-daemon-driver-nodedev
 Requires: libvirt-daemon-driver-storage-core
-Requires: (libvirt-daemon-driver-interface if virt-install)
 Requires: (libvirt-daemon-config-network if virt-install)
 Recommends: libvirt-daemon-driver-storage-disk
 %if 0%{?rhel}
--- a/src/libvirt-xml-parse.js
+++ b/src/libvirt-xml-parse.js
@@ -899,15 +899,6 @@ export function parseDumpxmlMachinesMetadataElement(metadataElem, name) {
     return subElems.length > 0 ? subElems[0].textContent : null;
 }
 
-export function parseIfaceDumpxml(ifaceXml) {
-    const retObj = {};
-    const ifaceElem = getElem(ifaceXml);
-
-    retObj.type = ifaceElem.getAttribute("type");
-
-    return retObj;
-}
-
 export function parseNetDumpxml(netXml) {
     const retObj = {};
     const netElem = getElem(netXml);
--- a/src/libvirtApi/common.js
+++ b/src/libvirtApi/common.js
@@ -455,7 +455,7 @@ export function getApiData({ connectionName }) {
     return Promise.allSettled([
         domainGetAll({ connectionName }),
         storagePoolGetAll({ connectionName }),
-        interfaceGetAll({ connectionName }),
+        interfaceGetAll(),
         networkGetAll({ connectionName }),
         nodeDeviceGetAll({ connectionName }),
         getNodeMaxMemory({ connectionName }),
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -53,7 +53,6 @@ sh -x test/vm.install
 
 if [ "${PLATFORM_ID:-}" != "platform:el8" ]; then
     # https://gitlab.com/libvirt/libvirt/-/issues/219
-    systemctl start virtinterfaced.socket
     systemctl start virtnetworkd.socket
     systemctl start virtnodedevd.socket
     systemctl start virtstoraged.socket
--- a/test/vm.install
+++ b/test/vm.install
@@ -7,11 +7,6 @@ if grep -q 'ID=debian' /usr/lib/os-release; then
     echo '* soft core unlimited' >> /etc/security/limits.conf
 fi
 
-if grep -q 'ID="opensuse' /usr/lib/os-release; then
-    # Make sure virtinterfaced.socket is enabled
-    systemctl enable --now virtinterfaced.socket
-fi
-
 systemctl enable cockpit.socket
 
 # don't force https:// (self-signed cert)


martinpitt:
Our bots images still have `libvirt-daemon-driver-interface`. I locally prepared an image without it, and started the whole test suite. That will take a while, will update this comment/PR status once its done. However, TF runs should already happen without that dep.

martinpitt:
On TF, [TestMachinesNetworks.testNetworksCreate](https://artifacts.dev.testing-farm.io/3148e21e-5490-472e-a731-4ccf0507d242/) fails. I also get that locally, plus a whole lot of other failures which don't look related to `virsh iface-list` (which is the one thing that "legitimately" breaks here). This also reflects in the [tumbleweed failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1782-cde6d887-20240826-105150-opensuse-tumbleweed/log.html#66-2) where this PR removes the "systemctl enable" special case.

Added fix.

test/check-machines-networks
```diff
@@ -25,16 +25,7 @@ from machinesxmls import TEST_NETWORK2_XML, TEST_NETWORK3_XML, TEST_NETWORK4_XML
...
-    net_devices_str = net_devices_str.split("\n", 2)[2]  # Remove first 2 lines of table header
-    if net_devices_str not in ["\n", "\r\n"]:
-        device = net_devices_str.split(' ', 2)[1]  # Get the name of device, ignoring spacing before device string
-    else:  # If $virsh-iface list did not return any device, check virsh nodedev-list net:noh
-        net_devices_str = m.execute("virsh nodedev-list net")
-        net_devices_str = net_devices_str.split("\n", 2)[0]
-        device = net_devices_str.split("_", 2)[1]  # Ignore prefix (example: net_enp0s31f6_8c_16_45_5f_77_34)
-
-    return device
+    return m.execute("ls /sys/class/net | grep -v ^lo | head -n1").strip()
```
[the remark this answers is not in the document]
jelly: As this is in a VM I suppose it works, I thought this was JS code.

src/libvirtApi/interface.js
```diff
@@ -1,75 +0,0 @@
...
-        if ("MAC" in resultProps)
-            props.mac = resultProps.MAC.v.v;
-        if ("Name" in resultProps)
-            props.name = resultProps.Name.v.v;
-        props.id = objPath;
-        props.connectionName = connectionName;
-
-        const [xml] = await call(connectionName, objPath, 'org.libvirt.Interface', 'GetXMLDesc', [0], { timeout, type: 'u' });
-        const iface = parseIfaceDumpxml(xml);
-        store.dispatch(updateOrAddInterface(Object.assign(props, iface)));
```
jelly, line 59: This needs to be cleaned up as well then

https://github.com/search?q=repo%3Acockpit-project%2Fcockpit-machines%20UPDATE_ADD_INTERFACE&type=code

jelly (changes requested):
Not convinced at all that "nothing uses this":

`getBootOrderDevices` uses vm.interfaces which seems to come from this reducer state if I am reading this correctly:

And used here:
```
            } else if (boot.type === "network") {
                const dev = ifaces[0];
                if (dev) {
                    ifaces.splice(0, 1);
                    devices.push({
                        device: dev,
                        bootOrder: i + 1, // bootOrder begins at 1
                        type: "network"
                    });
                }
            }
```

test/check-machines-networks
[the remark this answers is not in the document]
martinpitt: The file system order in /sys/class/net is "whatever", but `ls` sorts ASCIIbetically by default. I don't know the sorting order of `virsh iface-list`, but I can't imagine any profound technical reason why it would prefer wifi over ethernet? This smells rather arbitrary?

But this is only for that integration test which assumes it's running in our test VMs. Of course I'm happy to change the grepping/sorting if it helps you with running this test directly on your machine.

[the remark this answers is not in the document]
jelly: It prefers WiFi over ethernet as I have no ethernet connected but it is available. Basically proving your assumptions don't hold in a real setup

martinpitt:
> getBootOrderDevices uses vm.interfaces

Sure, but there's no doubt about a VM having a list of network interfaces. virtinterfaced is tracking the *host's* network interface. As far as I understand the [deprecation announcement](https://listman.redhat.com/archives/libvir-list/2020-December/msg00183.html), the main use case was

> functions that list current interfaces and get their current live status (i.e. based on sending/receiving netlink 
messages, not on the contents of the host network config files). Even on that front, the one commonly used example that I know know of is virt-manager, which had used virInterface*() to get a list of bridges/ethernets available for guest network connections

... and the only place which does that is that test helper `getInterface()`, which I changed to sysfs.

I could of course misunderstand all of this. @KKoukiou do you remember something about this?

test/check-machines-networks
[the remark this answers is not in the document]
martinpitt: I'm not sure how to interpret this. An interface does not need to be connected (and I guess in most cases it actually should *not* be) in order to assign it to a VM -- that VM itself will in most cases configure it by itself.

Also, this isn't a real setup anyway -- the old test code's approach of "just pick the first device that `virsh iface-list` shows" is also completely arbitrary.

We can change the `ls` to something more elaborate, like `ip` -- but it seems neither of us knows how `virsh iface-list` sorted its devices. Even though I think it's irrelevant, let me do some research what list ordering it uses.

src/libvirtApi/interface.js
jelly: And `updateOrAddInterface` is not called anymore with your PR, so I don't really understand how stuff in **this** PR works.

jelly:
> > getBootOrderDevices uses vm.interfaces
> 
> Sure, but there's no doubt about a VM having a list of network interfaces. virtinterfaced is tracking the _host's_ network interface. As far as I understand the [deprecation announcement](https://listman.redhat.com/archives/libvir-list/2020-December/msg00183.html), the main use case was
> 
> > functions that list current interfaces and get their current live status (i.e. based on sending/receiving netlink
> > messages, not on the contents of the host network config files). Even on that front, the one commonly used example that I know know of is virt-manager, which had used virInterface*() to get a list of bridges/ethernets available for guest network connections
> 
> ... and the only place which does that is that test helper `getNetworkDevice()`, which I changed to sysfs.
> 
> I could of course misunderstand all of this. @KKoukiou do you remember something about this?

The problem here is that I cannot read :)

`vm.interfaces` != `interfaces`. We obtain `interfaces` here https://github.com/cockpit-project/cockpit-machines/blob/main/src/app.jsx#L250 and pass it down to HostVmList here https://github.com/cockpit-project/cockpit-machines/blob/main/src/app.jsx#L340

And that component does not use `interfaces`. So kill that please, that should have been the first removal commit imo instead of removing the API call without explaining why it was unused.

Next up remove `interfaces` from combineReduces and the `interfaces` function in `reducers.js`.

jelly (commented):
Soo turns out it is not really **unused**

```
/**
 * Returns a list of potential physical devices suitable as network devices
 * by merging all network node devices and interfaces.
 *
 * @returns {array}
 */
export function getNetworkDevices() {
    const { nodeDevices, interfaces } = store.getState();
    const devs = [];

    nodeDevices.forEach(dev => {
        if (dev.capability.type === "net")
            devs.push(dev.capability.interface);
    });

    interfaces.forEach(iface => {
        devs.push(iface.name);
    });
```

In src/helpers.js used i:

```
const NetworkDeviceRow = ({ onValueChanged, dialogValues }) => {
    const devices = getNetworkDevices();
    return (
```

jfehlig:
> /**
>  * Returns a list of potential physical devices suitable as network devices
>  * by merging all network node devices and interfaces.
>  *
>  * @returns {array}
>  */

I assume this means host network devices that could be used by VMs, right? Most commonly, that means a bridge type device. For reference, I _think_ this is how virt-manager handles that (see lines 42-110)

https://github.com/virt-manager/virt-manager/blob/main/virtinst/devices/interface.py#L94

Disclaimer: I'm slightly familiar with virt-manager code, but far from an expert :-)

martinpitt:
@jfehlig Thanks for the pointer! But ` _host_default_bridge()` can't be the whole story -- creating networks is in no way restricted to bridges. But I suppose the equivalent functionality is this:

https://github.com/virt-manager/virt-manager/blob/1fef5d8661118a4f00a8689b323d275144b39dbe/virtManager/device/netlist.py#L131

jfehlig:
> @jfehlig Thanks for the pointer! But ` _host_default_bridge()` can't be the whole story -- creating networks is in no way restricted to bridges. 

The terminology is starting to confuse me. Let's clarify it :-).

https://libvirt.org/html/libvirt-libvirt-network.html
libvirt fully supports managing virtual networks via the virNetwork* set of APIs, backed by virtnetworkd service. There are many apps and users of this functionality. It works great and is well tested across various distros.

https://libvirt.org/html/libvirt-libvirt-interface.html
libvirt (woefully inadequately IMO) supports managing host network interface devices via the virInterface* set of APIs backed by virtinterfaced service. Beyond cockpit-machines, I'm not aware of any users of this functionality. virt-manager removed it long ago. It's definitely not well tested. The only virtinerfaced backend that works across distros is the udev one, which supports just a handful of read-only operations that likely return inconsistent results across distros.

https://libvirt.org/formatdomain.html#network-interfaces
libvirt supports managing virtual interfaces used by VMs via the `<interface>` XML config. To be useful, these need to be plugged into a network.

Ok, back to the topic. virt-manager can be used to create/manage libvirt virtual networks. It cannot be used to create/manage any host network interface devices. The virt-manager code I referenced is used to find a host network interface of type 'bridge', which will then be displayed as a possible network to attach virtual interface(s) of new VMs. By default, virt-manager prefers connecting virtual interfaces to a bridge, but the list of possible networks will include any active libvirt virtual networks too. If no bridge is found, virt-manager doesn't provide a way to create one. It simply excludes 'bridge device' from the list of possible networks.

> https://github.com/virt-manager/virt-manager/blob/1fef5d8661118a4f00a8689b323d275144b39dbe/virtManager/device/netlist.py#L131

IIUC, that code is used to populate the 'Network selection' list I mentioned above. E.g. the list shown on step 5 of 5 of the new VM wizard.

test/check-machines-networks
```diff
@@ -17,24 +17,17 @@
...
-    if net_devices_str not in ["\n", "\r\n"]:
-        device = net_devices_str.split(' ', 2)[1]  # Get the name of device, ignoring spacing before device string
-    else:  # If $virsh-iface list did not return any device, check virsh nodedev-list net:noh
-        net_devices_str = m.execute("virsh nodedev-list net")
-        net_devices_str = net_devices_str.split("\n", 2)[0]
-        device = net_devices_str.split("_", 2)[1]  # Ignore prefix (example: net_enp0s31f6_8c_16_45_5f_77_34)
-
-    return device
+def getNetworkDefaultRouteDevice(m):
+    routes = json.loads(m.execute("ip --json route show default"))
```
Nykseli, line 29: Is `show default` needed here? At least on fedora 40 and tumbleweed, the test will pass if you use `ip --json route`

martinpitt: Yes, but only by happenstance -- as soon as you have more than one route, and the default one isn't the first one, that will fail. I'd rather make it explicit.

martinpitt:
For the record -- this is way more involved than it initially looked. This was a quick attempt, but it apparently needs someone who actually has a clue about c-machines/libvirt. So it's back in my long queue of "fix someday", or "we need a c-machines maintainer again". :cry:

test/check-machines-networks
Nykseli: I see. That's a fair point

JanZerebecki: It seems the problem is that there is no default route on those test systems at this point. The previous code did not pick a specific interface, just any first one, not even enabled. On one system here the first virsh command would have picked "lo". Does routing to outside the host actually need to work? If yes, then which destination? As example 127.0.0.1 then `ip --json route get 127.0.0.1`. If not then maybe hard coding "lo" is fine.

Nykseli:
I looked into this a bit and created a draft for replicating the old behavior with the `ip` command: https://github.com/cockpit-project/cockpit-machines/pull/1839

test/check-machines-networks
martinpitt: Ah, indeed -- [it failed](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1782-f22e9226-20240827-042551-fedora-40/log.html#88), as I locally tested in an inadequate way.

My main issue is that I have not the slightest idea what this operation in the test actually does --  it creates a libvirt network with a `<dev>` node, which happens to be "eth0" or "eth1". I *assume* this creates a bridge which forwards its traffic to the given host device? But then "lo" doesn't make much sense, does it?

So indeed I'll just drop the `default` here, that'll also pick "eth0". I don't know what exactly this does, but at least it stays compatible :sweat_smile:

martinpitt:
Thanks @Nykseli ! I merged (conceptually, not git) the two MRs. I started with your commit, fixed up some things (like parsing MAC and Active properties), and added back the packaging updates. I also triggered a run against the tumbleweed refresh.

@jfehlig @KKoukiou does this look sensible to you now?

src/libvirtApi/interface.js
```diff
@@ -21,55 +21,27 @@
...
+        const ipData = await cockpit.spawn(["ip", "--json", "a"], { err: "ignore" });
+        ifaces = JSON.parse(ipData);
     } catch (ex) {
-        console.log('listInactiveInterfaces action for path', objPath, ex.toString());
+        console.warn("Failed to get interfaces with ip command", ex.toString());
```
cockpituous, line 37: This added line is not executed by any test.

src/libvirtApi/interface.js
```diff
@@ -21,55 +21,27 @@
...
-            props.active = resultProps.Active.v.v;
-        if ("MAC" in resultProps)
-            props.mac = resultProps.MAC.v.v;
-        if ("Name" in resultProps)
-            props.name = resultProps.Name.v.v;
-        props.id = objPath;
-        props.connectionName = connectionName;
-
-        const [xml] = await call(connectionName, objPath, 'org.libvirt.Interface', 'GetXMLDesc', [0], { timeout, type: 'u' });
-        const iface = parseIfaceDumpxml(xml);
```
jelly, line 58: Nothing should use parseIfaceDumpxml in src/libvirt-xml-parse.js anymore so please drop it.

jelly (changes requested):
Small cleanup request

src/libvirtApi/interface.js
martinpitt: Thanks for spotting! Done, and I also improved the error message when calling `ip` failed.

KKoukiou (commented):
This replacement is reasonable considering the virtinterfaced is so undermaintained and causes visible crashes. 

Let me know if you want me to review the code as well.

jelly (approved):
Can't tell if the debian failure is a flake or not, if not feel free to merge

martinpitt:
Our tests are all green. Seems packit's TF integration is AWOL today -- but I feel comfortable to ignore, the previous version was all green, and the diff was just that little code cleanup.

jfehlig:
Sorry for the late response, but I don't have anything to add anyhow other than thanks a lot for resolving this issue!

PR #1784, opened 2024-08-26
Remove all unused passed arguments to functions or components

jelly:
Should avoid unneeded re-renders and confusion.


martinpitt (approved):
Cheers!

PR #1785, opened 2024-08-26
Revert "libvirtApi remove use of interfaceGetAll"

martinpitt:
Reverts cockpit-project/cockpit-machines#1779

This needs more investigation. This functionality isn't properly covered by tests.


martinpitt:
@Lunarequest FYI. Let's do this in #1782

jelly (approved):

PR #1796, opened 2024-08-30
Add "Add TPM" VM action

Shotaro-Kawaguchi:
We implemented this feature because there was a use case where upgrading the Windows OS to Windows11 required additional TPM devices.

Selecting "Add TPM" will add a TPM device with default settings.
"Add TPM" is only displayed if no TPM device exists.

 - [x] https://github.com/cockpit-project/bots/pull/7064

----

## Machines: Action to add a TPM

This can help with upgrading old VMs to a newer guest OS that requires a TPM, such as Windows 11.

tpm_kebab

### DIFF
--- a/src/components/common/needsShutdown.jsx
+++ b/src/components/common/needsShutdown.jsx
@@ -77,6 +77,10 @@ export function needsShutdownWatchdog(vm) {
     return vm.persistent && vm.state === "running" && vm.inactiveXML.watchdog.action !== vm.watchdog.action;
 }
 
+export function needsShutdownTpm(vm) {
+    return vm.persistent && vm.state === "running" && vm.inactiveXML.hasTPM !== vm.hasTPM;
+}
+
 export function needsShutdownSpice(vm) {
     return vm.hasSpice !== vm.inactiveXML.hasSpice;
 }
@@ -120,6 +124,10 @@ export function getDevicesRequiringShutdown(vm) {
     // SPICE
     if (needsShutdownSpice(vm))
         devices.push(_("SPICE"));
+
+    // TPM
+    if (needsShutdownTpm(vm))
+        devices.push(_("TPM"));
 
     return devices;
 }
--- a/src/components/vm/vmActions.jsx
+++ b/src/components/vm/vmActions.jsx
@@ -58,6 +58,7 @@ import {
     domainSendNMI,
     domainShutdown,
     domainStart,
+    domainAddTPM,
 } from '../../libvirtApi/domain.js';
 import store from "../../store.js";
 
@@ -180,6 +181,13 @@ const onSendNMI = (vm) => domainSendNMI({ name: vm.name, id: vm.id, connectionNa
         })
     );
 });
+
+const onAddTPM = (vm, onAddErrorNotification) => domainAddTPM({ connectionName: vm.connectionName, vmName: vm.name })
+        .catch(ex => onAddErrorNotification({
+            text: cockpit.format(_("Failed to add TPM to VM $0"), vm.name),
+            detail: ex.message,
+            resourceId: vm.id,
+        }));
 
 const VmActions = ({ vm, vms, onAddErrorNotification, isDetailsPage }) => {
     const Dialogs = useDialogs();
@@ -466,6 +474,17 @@ const VmActions = ({ vm, vms, onAddErrorNotification, isDetailsPage }) => {
 
     if (domainCanRename(state) || isDetailsPage)
         dropdownItems.push(<Divider key="separator-rename" />);
+
+    if (!vm.hasTPM && !vm.inactiveXML?.hasTPM && vm.capabilities?.supportsTPM) {
+        dropdownItems.push(
+            <DropdownItem key={`${id}-add-tpm`}
+                          id={`${id}-add-tpm`}
+                          onClick={() => onAddTPM(vm, onAddErrorNotification)}>
+                {_("Add TPM")}
+            </DropdownItem>
+        );
+        dropdownItems.push(<Divider key="separator-add-tpm" />);
+    }
 
     if (state !== undefined && domainCanDelete(state, vm.id)) {
         if (!vm.persistent) {
--- a/src/components/vm/vmActions.jsx
+++ b/src/components/vm/vmActions.jsx
@@ -66,6 +66,7 @@ const _ = cockpit.gettext;
 
 const onStart = (vm, setOperationInProgress) => domainStart({ name: vm.name, id: vm.id, connectionName: vm.connectionName }).catch(ex => {
     setOperationInProgress(false);
+    console.warn("Failed to start VM", vm.name, ":", cockpit.message(ex));
     store.dispatch(
         updateVm({
             connectionName: vm.connectionName,
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -1082,13 +1082,5 @@ export async function domainReplaceSpice({ connectionName, id: objPath }) {
 
 export async function domainAddTPM({ connectionName, vmName }) {
     const args = ["virt-xml", "-c", `qemu:///${connectionName}`, "--add-device", "--tpm", "default", vmName];
-    return cockpit.spawn(args, { err: "message", superuser: connectionName === "system" ? "try" : null })
-            // RHEL 8 does not support --tpm default; arch (and likely other system setups) fails with
-            // unsupported configuration: TPM version '2.0' is not supported
-            .catch(ex => {
-                if (ex.message?.includes("Unknown TPM backend type") || ex.message?.includes("unsupported configuration"))
-                    console.log("Failed to add TPM:", ex);
-                else
-                    return Promise.reject(ex);
-            });
+    return cockpit.spawn(args, { err: "message", superuser: connectionName === "system" ? "try" : null });
 }
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -33,6 +33,11 @@ def hasMonolithicDaemon(image):
             image in ["arch"])
 
 
+# software TPM doesn't work on some OSes
+def hasBrokenTPM(image):
+    return image.startswith("rhel-8") or image == 'arch'
+
+
 class VirtualMachinesCaseHelpers:
     machine: testvm.Machine
 
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -2186,6 +2186,12 @@ vnc_password= "{vnc_passwd}"
             tpm_xml = m.execute("virsh dumpxml VmNotInstalled | xmllint --xpath /domain/devices/tpm -")
             self.assertIn('model="tpm-crb"', tpm_xml)
             self.assertIn('type="emulator"', tpm_xml)
+            # thus menu does not offer to add a TPM
+            b.click("#vm-VmNotInstalled-system-action-kebab")
+            b.wait_visible("#vm-VmNotInstalled-system-delete")
+            self.assertNotIn("Add TPM", b.text("ul.pf-v5-c-menu__list"))
+            b.click("#vm-VmNotInstalled-system-action-kebab")
+            b.wait_not_present("ul.pf-v5-c-menu__list")
 
         # Temporarily delete the OVMF binary and check the firmware options again
         if "fedora" in m.image or "rhel" in m.image or "centos" in m.image:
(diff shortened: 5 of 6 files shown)


martinpitt:
Thanks @Shotaro-Kawaguchi ! I had a very cursory look. I'm mostly interested in the use case -- so far we've treated that as an implementation detail, and osinfo-db was telling libvirt whether the target OS needs a TPM (like Windows) or can use a TPM (EFI BIOS), and automatically adds an appropriate one. In which cases does that not work? Every bit of UI addition is code to maintain, increases combinatorial explosion of testing, and is a mental load to users, so we don't want to do that lightly.

This needs a round of design review and updates -- the dialogs are way too technical, and give the admin no guidance at all. When to use "tis" or "crb", and what's the difference? What's the (dis)advantage of emulator vs. passthrough? Is there any reason to use "1.2" when "2.0" is available? and so on. This needs some (i) popups with documentation links, and if possible some clearer descriptions.

Also this needs integration tests for all code paths. testConfigureBeforeInstall() has some cursory check that a software TPM was added, so our test infrastructure supports it. testConfigureBeforeInstallBiosTPM() also covers it explicitly.

Unfortunately it will take our team very long to help with that as we don't have a current machines maintainer. So it would be good if you could work on tests yourself? (Happy to guide, of course)

Shotaro-Kawaguchi:
@martinpitt  Thank you for your response.

> Thanks @Shotaro-Kawaguchi ! I had a very cursory look. I'm mostly interested in the use case -- so far we've treated that as an implementation detail, and osinfo-db was telling libvirt whether the target OS needs a TPM (like Windows) or can use a TPM (EFI BIOS), and automatically adds an appropriate one. In which cases does that not work? Every bit of UI addition is code to maintain, increases combinatorial explosion of testing, and is a mental load to users, so we don't want to do that lightly.

A TPM device is required for Windows 11, but it is not needed for Windows 10 and older. Therefore, I believe that a TPM device will not be automatically added when a virtual machine is upgraded to Windows 11.
I think it is useful to be able to add a TPM device when upgrading from Windows 10 or older to Windows 11.

> Also this needs integration tests for all code paths. 

I will try to create the test code.

martinpitt:
> I think it is useful to be able to add a TPM device when upgrading from Windows 10 or older to Windows 11.

OK, I see that. But I still don't believe that we should throw all these technicalities to the user. The dialog allows 8 combinations, which is too much both for the user to decide on (and document) and also for testing. This needs to be reduced, automatically pick the most appropriate one (`virt-xml` should help with that), and the remaining options better explained. But ideally there should not be any options at all.

Thanks!

jelly:
For upgrading it seems sensible to add a TPM, I assume most users would want a `swtpm` (software tpm). Passing through a TPM sounds like extra hurdles (plus the host might use it?).

For UEFI, it seems virt-install these days already adds a TPM

https://github.com/virt-manager/virt-manager/blob/2da4884962d7a8d4e66e4d6f273a452fdd3f2d4f/virtinst/guest.py#L1081

garrett:
Agreed; this should work by default and shouldn't have to expose any configuration UI.

Creating a VM in UEFI mode should _ideally_ automatically create a software TPM compatible with Windows 11. (I'm not sure if it does. It seems like it might, according to @jelle's comment.) And if it's compatible with Windows 11, it should also be compatible with Linux too.

Are there any cases where having a software TPM that's compatible with Windows 11 isn't enough for either Windows or Linux?

jelly:
> Agreed; this should work by default and shouldn't have to expose any configuration UI.
> 
> Creating a VM in UEFI mode should _ideally_ automatically create a software TPM compatible with Windows 11. (I'm not sure if it does. It seems like it might, according to @jelle's comment.) And if it's compatible with Windows 11, it should also be compatible with Linux too.
> 
> Are there any cases where having a software TPM that's compatible with Windows 11 isn't enough for either Windows or Linux?

Well as the contributor noted, when upgrading there is no TPM added. Say you have an old Windows 10 machine you want to upgrade. Current workflow would be to detach the disk, and re-create a new virtual machine and specify windows 11 as OS.

garrett:
Should a UEFI VM without a TPM be considered a bug? Should we automatically add it? Or have an inline alert banner to add it, or something like that?

Shotaro-Kawaguchi:
@martinpitt If the implementation is such that pressing the add button adds the TPM device with all options set to default values, would it be possible to merge it?

martinpitt:
@Shotaro-Kawaguchi You are actually in a better position to answer that -- as you proposed and want the feature, you should also know how much configurability you need for your use case. From my point of view, I think that "defaults" would be a great start -- let virt-xml figure out the appropriate defaults, and add an integration test for adding and removal -- then we can land and refine with more options later if/when they become necessary. 

Thanks!

martinpitt:
@Shotaro-Kawaguchi We just discussed this in our meeting. We feel that this is a too rare use case to put this into the overview all the time. If the VM has a TPM, then we don't need to show anything, and for most existing VMs you don't need to change it either.

A good place for these is the kebab menu,  similar to the "Replace SPICE" functionality (which is a similar corner case):

dot menu

This could just grow an "Add TPM" action if the VM doesn't already have one. I don't *think* that you'd even need to remove it, or do you have such a use case?

Thanks!

Shotaro-Kawaguchi:
@martinpitt Thank you for your feedback.
Since there are no use cases to delete, we will place the "Add TPM" action in the kebab menu if the VM does not yet have a TPM.

Shotaro-Kawaguchi:
Changed to select "Add TPM" from the kebab menu.
Selecting "Add TPM" will add a TPM device with default settings.
"Add TPM" is only displayed if no TPM device exists.

tpm_kebab

martinpitt (dismissed):
Thanks @Shotaro-Kawaguchi ! This looks nice now (aside from some code cleanup and the missing "needs shutdown" integration). This now needs tests -- do you want to work on them, or need help?

src/components/common/needsShutdown.jsx
```diff
@@ -77,6 +77,10 @@ export function needsShutdownWatchdog(vm) {
     return vm.persistent && vm.state === "running" && vm.inactiveXML.watchdog.action !== vm.watchdog.action;
 }
 
+export function needsShutdownTpm(vm) {
```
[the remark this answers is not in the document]
Shotaro-Kawaguchi: Thank you for your feedback. The suggested changes have been implemented. The function is now called within `getDevicesRequiringShutdown()`.

src/libvirt-xml-parse.js
```diff
@@ -560,7 +561,25 @@ function parseDumpxmlForSpice(devicesElem) {
...
+
+        const tpm = {
+            model,
+            type,
+            version,
+            devicePath,
+        };
+        return { tpm, hasTPM: true };
+    }
+    return { tpm: {}, hasTPM: false };
```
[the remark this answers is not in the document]
martinpitt: With just returning the `tpm` object here or `null`, this will be fully compatible with a boolean check. This commit *breaks* this (and is generally harder to use), hence my proposal to keep the current behaviour of being able to interpret the `parseDumpxmlForTPM()` result as a boolean.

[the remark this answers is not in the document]
Shotaro-Kawaguchi: I understand. Thank you. I have made the corrections, so please check them.

Shotaro-Kawaguchi:
Thank you.
I will proceed with modifying the source code.

> This now needs tests -- do you want to work on them, or need help?
Since I don't yet have the know-how to create a test program, is it possible to get some help?

martinpitt:
@Shotaro-Kawaguchi No worries. @mvollmer or I can spend some time on this next week and complete it with tests. So are you happy with the functionality now?

Shotaro-Kawaguchi:
@martinpitt Thank you for your support.
Yes, I am happy with the current functionality.

martinpitt:
This was conflicty, so I  pushed a rebase. I want to get a first run of the tests to check for regressions. I'll work on a test today. Thanks!

src/components/vm/vmActions.jsx
```diff
@@ -181,6 +182,8 @@ const onSendNMI = (vm) => domainSendNMI({ name: vm.name, id: vm.id, connectionNa
     );
 });
 
+const onAddTPM = (vm) => domainAddTPM({ connectionName: vm.connectionName, vmName: vm.name });
```
cockpituous, line 185: This added line is not executed by any test.

src/components/vm/vmActions.jsx
```diff
@@ -467,6 +470,17 @@ const VmActions = ({ vm, vms, onAddErrorNotification, isDetailsPage }) => {
     if (domainCanRename(state) || isDetailsPage)
         dropdownItems.push(<Divider key="separator-rename" />);
 
+    if (!vm.tpm) {
+        dropdownItems.push(
+            <DropdownItem key={`${id}-add-tpm`}
+                          id={`${id}-add-tpm`}
+                          onClick={() => onAddTPM(vm)}>
```
cockpituous, line 477: This added line is not executed by any test.

src/libvirt-xml-parse.js
```diff
@@ -566,7 +568,25 @@ function parseDumpxmlForSpice(devicesElem) {
...
-    return devicesElem.getElementsByTagName('tpm').length > 0;
+    const tpmElems = devicesElem.getElementsByTagName('tpm');
+    if (tpmElems.length > 0) {
+        const tpmElem = tpmElems[0];
+        const model = tpmElem.getAttribute('model');
+        const backendElem = tpmElem.getElementsByTagName('backend')[0];
+        const type = backendElem.getAttribute('type');
+        const version = backendElem.getAttribute('version');
+        const devicePathElem = backendElem.getElementsByTagName('device')[0];
+        const devicePath = devicePathElem ? devicePathElem.getAttribute('path') : null;
```
cockpituous, line 579: This added line is not executed by any test.

martinpitt:
Next push: I added integration tests.

The code change is still more complicated than it needs to be: the changes to src/libvirt-xml-parse.js can be reverted, as the new info from it isn't actually being used anywhere.

src/components/common/needsShutdown.jsx
```diff
@@ -121,6 +125,10 @@ export function getDevicesRequiringShutdown(vm) {
     if (needsShutdownSpice(vm))
         devices.push(_("SPICE"));
 
+    // TPM
+    if (needsShutdownTpm(vm))
+        devices.push(_("TPM"));
```
cockpituous, line 130: This added line is not executed by any test.

src/components/vm/vmActions.jsx
```diff
@@ -467,6 +470,17 @@ const VmActions = ({ vm, vms, onAddErrorNotification, isDetailsPage }) => {
     if (domainCanRename(state) || isDetailsPage)
         dropdownItems.push(<Divider key="separator-rename" />);
 
+    if (!vm.hasTPM) {
```
mvollmer, line 473: Should we check `vm.capabilities.supportsTPM` here as well?

mvollmer (dismissed):
Thanks!

mvollmer:
> We feel that this is a too rare use case to put this into the overview all the time. If the VM has a TPM, then we don't need to show anything, and for most existing VMs you don't need to change it either.
> 
> A good place for these is the kebab menu, similar to the "Replace SPICE" functionality (which is a similar corner case):

I think it would be good to get some feedback after adding a TPM. With SPICE, you can see how the options for the console have changed. With TPM there is nothing. Maybe it is worth alerting the user when a VM does not have a TPM?

src/components/vm/vmActions.jsx
martinpitt: Yes, good idea -- works, done.

martinpitt:
> I think it would be good to get some feedback after adding a TPM. 

Yeah, I stumbled over that when writing the tests -- you can only tell by inspecting the XML and the menu entry going away. Maybe a transient alert with "TPM added successfully" or so?

> Maybe it is worth alerting the user when a VM does not have a TPM?

That's too noisy IMHO -- like, almost all existing VMs without a TPM don't need one.

src/components/common/needsShutdown.jsx
cockpituous, line 130: This added line is not executed by any test.

martinpitt: Right you are, coverage tests. I added a check for adding a TPM to a running VM, and test that "needs restart" indicator. This also uncovered a bug: The "Add TPM" action does not go away in this state, i.e. TPM got added to the inactive XML, but not yet the running one.

I fixed that bug and added tests for the "needs shutdown" and "add TPM while running" cases.

martinpitt:
When adding a TPM for a running VM, there is at least *some* visual feedback:

image

There's none for a non-running VM, other than the menu entry going away. However, given how much of a fringe feature that is, I think we can handle a possible extra alert as a follow-up?

martinpitt:
Ugh, thereis an [AppArmor problem](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1796-c3d518e5-20241104-080012-debian-stable/log.html) on Debian, and this doesn't work at all on [rhel 8](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1796-c3d518e5-20241104-080015-rhel-8-10/log.html#88) and [arch](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1796-c3d518e5-20241104-080005-arch/log.html#88). Back to yak shaving..

src/components/vm/vmActions.jsx
```diff
@@ -181,6 +182,13 @@ const onSendNMI = (vm) => domainSendNMI({ name: vm.name, id: vm.id, connectionNa
     );
 });
 
+const onAddTPM = (vm, onAddErrorNotification) => domainAddTPM({ connectionName: vm.connectionName, vmName: vm.name })
+        .catch(ex => onAddErrorNotification({
+            text: cockpit.format(_("Failed to add TPM to VM $0"), vm.name),
+            detail: ex.message,
+            resourceId: vm.id,
```
cockpituous, line 189: These 4 added lines are not executed by any test.

src/libvirtApi/domain.js
```diff
@@ -1082,13 +1082,5 @@ export async function domainReplaceSpice({ connectionName, id: objPath }) {
...
-    return cockpit.spawn(args, { err: "message", superuser: connectionName === "system" ? "try" : null })
-            // RHEL 8 does not support --tpm default; arch (and likely other system setups) fails with
-            // unsupported configuration: TPM version '2.0' is not supported
-            .catch(ex => {
-                if (ex.message?.includes("Unknown TPM backend type") || ex.message?.includes("unsupported configuration"))
-                    console.log("Failed to add TPM:", ex);
-                else
-                    return Promise.reject(ex);
-            });
+    return cockpit.spawn(args, { err: "message", superuser: connectionName === "system" ? "try" : null });
```
cockpituous, line 1085: This added line is not executed by any test.

martinpitt:
Now we are down to starting the VM failing o [fedora 41](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1796-87be6ab2-20241104-131854-fedora-41/log.html#88) and [RHEL 10](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1796-87be6ab2-20241104-131847-rhel-10-0/log.html#88). This is actually a [known issue](https://github.com/cockpit-project/bots/issues/6792), it just happens in a new test now, so we need to adjust the naughty.

src/components/vm/vmActions.jsx
```diff
@@ -181,6 +183,13 @@ const onSendNMI = (vm) => domainSendNMI({ name: vm.name, id: vm.id, connectionNa
...
```
cockpituous, line 190: These 4 added lines are not executed by any test.

src/libvirtApi/domain.js
cockpituous, line 1085: This added line is not executed by any test.

garrett (approved):
OK, this looks reasonable to me, if I understand the following correctly:

1. when creating a VM, Cockpit Machines automatically adds a TPM if it either needs or can benefit from a TPM
2. when there's an existing machine, it has the ability to add a TPM in the menu, as shown in the screenshot
3. when a TPM is added when the machine is not running, it adds it and everything's good, and it'll be there on next boot with no UI about TPMs or anything
4. when a TPM is added when the machine is running, it has the info bubble that the machine needs to be shut down when, and then the TPM will be there
5. when there's no TPM to add (as when a TPM already exists or when it doesn't make sense, like a VM in BIOS), there's no UI whatsoever (including: make sure there isn't a border above/below where the TPM entry would)

As long as everything on the list above is met and the screenshots reflect the current state, then I approve.

Thanks!

martinpitt:
@garrett Confirmed, all of these are true. There's not much UI impact here indeed (this was mostly wrestling with technicalities)

test/machineslib.py
```diff
@@ -33,6 +33,11 @@ def hasMonolithicDaemon(image):
             image in ["arch"])
 
 
+# software TPM doesn't work on some OSes
+def hasBrokenTPM(image):
+    return image.startswith("rhel-8") or image == 'arch'
```
jelly, line 38: The arch image needs `swtpm` should be easily fixable.

jelly (approved):
Thanks!

PR #1805, opened 2024-09-05
build.js: drop ip.js hack

jelly:
Since f27fc31c514be87f643153b4a3b64ba7300c6f34 we ported from the ip module to the ipaddr.js module. ip was a pure node.js module so required a hack to work.


mvollmer (approved):
Cool, thanks!

PR #1813, opened 2024-09-09
PF overrides cleanup

jelly:
Token scrollbar issue is no longer an issue:
Screenshot from 2024-09-09 17-29-08


garrett (approved):
Looks good. Works for me too!

Thanks!

PR #1814, opened 2024-09-10
test: drop selinux-policy workarounds

jelly:
Neither of these versions exist anymore in Rawhide and RHEL 10.


martinpitt (approved):
Thanks!

PR #1821, opened 2024-09-16
Remove undefined fontweight light

jelly:
PatternFly removed the light weight font variant so this was undefined for quite a while and thus has no UI change.

---

Screenshot from 2024-09-16 16-15-16

### DIFF
--- a/src/components/vm/vmDetailsPage.scss
+++ b/src/components/vm/vmDetailsPage.scss
@@ -18,7 +18,6 @@
   flex-wrap: wrap;
 
   .vm-name {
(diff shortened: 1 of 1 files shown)


garrett (approved):
When looking at the git blame, it's from before Cockpit Machines was split out to its own thing (when it was part of the Cockpit repo), so it's rather old. It was ported from PF4 to PF5 after that.

Perhaps this was even around from the PF3 days and has just been ported over even if it does nothing.

I can't see anything where it might be used, either in the original designs (spanning several years back, even in PF3 mockups) to what's in main.

 It should be fine to remove. Thanks!

PR #1827, opened 2024-09-20
Use cockpit.format_bytes() for displaying VM memory

martinpitt:
Eliminate convertToBestUnit() in favour of our standard API. Use the default precision of "3" for consistency with other pages, and adjust the test accordingly.

---

This is a first test balloon. There are dozens of `convertToUnit()` calls which should also be moved to `cockpit.format_bytes()`.

### DIFF
--- a/src/components/vm/vmUsageCard.jsx
+++ b/src/components/vm/vmUsageCard.jsx
@@ -23,12 +23,7 @@ import cockpit from 'cockpit';
 import { Flex, FlexItem } from "@patternfly/react-core/dist/esm/layouts/Flex";
 import { Progress, ProgressVariant } from "@patternfly/react-core/dist/esm/components/Progress";
 
-import {
-    logDebug,
-    units,
-    convertToBestUnit,
-    convertToUnit,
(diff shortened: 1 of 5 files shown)


src/components/vm/vmUsageCard.jsx
```diff
@@ -23,23 +23,18 @@ import cockpit from 'cockpit';
...
 
 class VmUsageTab extends React.Component {
     render() {
         const vm = this.props.vm;
 
-        const rssMem = vm.rssMemory ? vm.rssMemory : 0; // in KiB
-        const memTotal = vm.currentMemory ? vm.currentMemory : 0; // in KiB
-        const memRssBest = convertToBestUnit(rssMem, units.KiB);
-        const memTotalBest = convertToBestUnit(memTotal, units.KiB);
+        const memTotal = vm.currentMemory ? vm.currentMemory * 1024 : 0;
```
cockpituous, line 34: This added line is not executed by any test.

KKoukiou (approved):

PR #1839, opened 2024-09-26
Replace deprecated virtinterfaced calls with ip command

Nykseli:
The libvirt ListInterfaces dbus call is used to only to list out names of interfaces. On a test vm (Tumbleweed and Fedora) it's needed to get the virbr0 bridge interface.

This complexity can be reduced to simply use `ip --json a` to get all the interface names.

Note that this probably needs to be ran against https://github.com/cockpit-project/bots/pull/6909

This PR tries to solve the issues/concerns in: https://github.com/cockpit-project/cockpit-machines/pull/1782

### DIFF
--- a/src/libvirtApi/interface.js
+++ b/src/libvirtApi/interface.js
@@ -24,52 +24,19 @@
 import store from '../store.js';
 
 import { updateOrAddInterface } from '../actions/store-actions.js';
-import { parseIfaceDumpxml } from '../libvirt-xml-parse.js';
-import { call, Enum, timeout } from './helpers.js';
 
-/*
- * Read properties of a single Interface
- *
- * @param {object} objPath interface object path
- * @param {string} connectionName
- */
-export async function interfaceGet({
-    id: objPath,
-    connectionName,
-}) {
-    const props = {};
+export async function interfaceGetAll() {
+    let ifaces = [];
 
     try {
-        const [resultProps] = await call(connectionName, objPath, 'org.freedesktop.DBus.Properties', 'GetAll',
-                                         ['org.libvirt.Interface'], { timeout, type: 's' });
-        /* Sometimes not all properties are returned; for example when some network got deleted while part
-            * of the properties got fetched from libvirt. Make sure that there is check before reading the attributes.
-            */
-        if ("Active" in resultProps)
-            props.active = resultProps.Active.v.v;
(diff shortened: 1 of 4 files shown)


src/libvirtApi/interface.js
```diff
@@ -24,52 +24,19 @@
...
-        const [resultProps] = await call(connectionName, objPath, 'org.freedesktop.DBus.Properties', 'GetAll',
-                                         ['org.libvirt.Interface'], { timeout, type: 's' });
-        /* Sometimes not all properties are returned; for example when some network got deleted while part
-            * of the properties got fetched from libvirt. Make sure that there is check before reading the attributes.
-            */
-        if ("Active" in resultProps)
-            props.active = resultProps.Active.v.v;
-        if ("MAC" in resultProps)
-            props.mac = resultProps.MAC.v.v;
-        if ("Name" in resultProps)
```
martinpitt, line 52: I see these properties being used in other parts of the code. It shouldn't be too hard to get the MAC address and Active (aka. "UP") from the JSON output as well?

martinpitt (changes requested):
hey @Nykseli, thanks for working on this! This is still draft, so I'm not sure if I am supposed to comment yet, but the direction looks nice. Note that I can't run this PR against the tumbleweed refresh yet, as this doesn't yet change the spec. But if the libvirt experts agree to this solution, I'm happy to merge your and my #1782 into something which may actually work :grinning: 

Thanks!

Nykseli:
Feel free to merge this and https://github.com/cockpit-project/cockpit-machines/pull/1782 !

Note that I'm off next week but I'll work on it the week after that if needed :slightly_smiling_face:

src/libvirtApi/interface.js
Nykseli: Ah, must've missed them. But yes, they can be picked from the json output too

martinpitt: Done in #1782.

martinpitt:
@Nykseli Thanks! I updated #1782 now. I started with your commit, fixed up some things (like parsing `MAC` and `Active` properties), and added back the packaging updates. I'll also trigger a run against the tumbleweed refresh there.

jelly:
This can now be closed right?

Nykseli:
This can be closed since the code got merged in the other PR

PR #1845, opened 2024-10-02
test: reduce line length to 150

jelly:


mvollmer (dismissed):

mvollmer:
Why not all the way to 118?

jelly:
> Why not all the way to 118?

So just spend like 30 minutes reformatting code, this is sadly not an automatic job :)

Next PR will bring it down further.

tomasmatus (approved):
going all the way down to 118 sounds like a good first issue for someone :)

PR #1851, opened 2024-10-09
Fix disabled VM action dropdown item tooltip

KKoukiou:
When using `isDisabled` property in Buttons/Dropdowns/etc, their tooltips are not getting triggered.
We need to use `isAriaDisabled` instead of `isDisabled` in this case.

See more info: https://github.com/patternfly/patternfly-react/issues/5826


martinpitt (approved):
Ephraristo! There's plenty of precedence for `isAriaDisabled`, so this looks fine.

PR #1853, opened 2024-10-10
test: add a test that checking uptime when there is a invalid time in the log

yunmingyang:


yunmingyang:
It smells like a bug that there will be a uptime with wheel user if change starting up time to 0.5, and no uptime with libvirt user if change starting up time to 0.5

yunmingyang:
@martinpitt I am not sure, but it seems that distanceToNow will calculate the difference between the current date and 0.5, even though 0.5 is not a valid date. Thus, the uptime will be shown as xx years. I guess that the reason why the testBasicLibvirtUserUnprivileged is breaked is libvirt user doesn't have enough privilege for accessing /var/log/libvirt/qemu.

martinpitt:
@yunmingyang I see -- but it feels a bit sad to put such a broken time there. Maybe something more realistic, like an empty date, or "0", or a non-numeric word?

yunmingyang:
For a non-numeric word, I think there should be already a case there, https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-lifecycle#L247. For 0 and empty date and 0, I think they are different scenarios, maybe we could add them both. But decimal should also be a scenario for checking uptime parsing robustness. Any comments? @martinpitt

martinpitt:
Mostly just that I don't quite like cementing a bad behaviour in a test case -- let's rather just fix the bug and ignore that invalid date?

yunmingyang:
All right

mvollmer:
> Mostly just that I don't quite like cementing a bad behaviour in a test case -- let's rather just fix the bug and ignore that invalid date?

What bug is that?  Could someone file it?

Does qemu actually write things in the log that we can't parse, or is this all purely theoretical?

mvollmer:
> What bug is that? Could someone file it?

Ahh, #1871.  We should test for the scenario in the bug, I'd say.

PR #1861, opened 2024-10-14
test: use self.machine.write() where possible

jelly:


test/check-machines-create
```diff
@@ -519,12 +519,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                       create_and_run=True))
 
         # Set up the PXE server configuration files
-        cmds = [
-            "mkdir -p /var/lib/libvirt/dnsmasq",
-            f"echo \"{PXE_SERVER_CFG}\" > /var/lib/libvirt/dnsmasq/pxe.cfg",
-            "chmod 666 /var/lib/libvirt/dnsmasq/pxe.cfg"
-        ]
-        self.machine.execute("; ".join(cmds))
+        self.machine.write("/var/lib/libvirt/dnsmasq/pxe.cfg", PXE_SERVER_CFG, perm="0666")
```
martinpitt, line 522: This isn't already in vm_tmpdir (unlike the previous cases), so this one actually *should* use self.write_file(), so that the file is restored after the test.

test/check-machines-create
```diff
@@ -1605,8 +1600,8 @@ vnc_password= "{vnc_passwd}"
                 self._fakeFedoraTree()
 
             # Make sure we don't run into long DNS timeouts when trying to contact this server
-            self.machine.execute("echo >>/etc/hosts 127.0.0.42 mirror.i3d.net")
-            self.machine.execute("echo >>/etc/hosts ::1 mirror.i3d.net")
+            self.machine.write("/etc/hosts", "127.0.0.42 mirror.i3d.net", append=True)
+            self.machine.write("/etc/hosts", "::1 mirror.i3d.net", append=True)
```
martinpitt, line 1604: Same here, /etc/hosts isn't already auto-restored. Of course you can't call self.write_file() twice on the same file, so please join the two commands into one with appending both lines.

martinpitt (changes requested):
Sorry.. while we are at it, let's clean this up properly, before it bites us.

test/check-machines-create
jelly: Looking deeper into this it's more of a mess then.

```
        # Set up the PXE server configuration files
        self.write_file("/var/lib/libvirt/dnsmasq/pxe.cfg", PXE_SERVER_CFG, perm="0666")

        # Define and start a NAT network with tftp server configuration
        self.machine.write("/tmp/pxe-nat.xml", NETWORK_XML_PXE)
        self.machine.execute("virsh net-define /tmp/pxe-nat.xml; virsh net-start pxe-nat")
```

Another write() although to /tmp/ we should probably clean that up. And then the network define which is never undefined. So needs cleanup to:

```
virsh net-destroy 
virsh net-undefine --network
```

martinpitt: sorry -- this is /var/lib/libvirt, so this is fine.

test/check-machines-create
martinpitt: You mentioned in chat that /etc/hosts is already restored indirectly, via setup_mock_server(). That wasn't obvious, sorry.

martinpitt (approved):

PR #1862, opened 2024-10-15
Even smaller line length

jelly:


jelly:
FYI: with testing I found out that:

```
diff --git a/test/check-machines-disks b/test/check-machines-disks
index 7a835de9..3f6a26a1 100755
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -1162,9 +1162,9 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
         # Check that usage information can't be fetched since the pool is inactive
         b.wait_not_present(f"#vm-subVmTest1-disks-{pool_disk}-used")

-        # AppArmor doesn't like the non-standard path for our storage pools
-        if m.image in ["debian-testing"]:
-            self.allow_journal_messages(f'.* type=1400 .* apparmor="DENIED" operation="open" profile="libvirt.* name="{self.vm_tmpdir}.*')  # noqa: E501
+        # # AppArmor doesn't like the non-standard path for our storage pools
+        # if m.image in ["debian-testing"]:
+        #     self.allow_journal_messages(f'.* type=1400 .* apparmor="DENIED" operation="open" profile="libvirt.* name="{self.vm_tmpdir}.*')  # noqa: E501
```

Is no longer needed

test/machineslib.py
```diff
@@ -33,6 +34,8 @@ def hasMonolithicDaemon(image):
 
 
 class VirtualMachinesCaseHelpers:
+    machine: testvm.Machine
```
martinpitt, line 37: What's this bit for?

test/check-machines-create
```diff
@@ -1057,8 +1060,8 @@ vnc_password= "{vnc_passwd}"
                 query_result = f'{query_result}'
                 # check if there is os present in osinfo-query because it can be filtered out in the UI
                 # throws exception if grep fails
-                self.machine.execute(
-                    fr"osinfo-query os --fields=name | tail -n +3 | sed -e 's/\s*|\s*/|/g; s/^\s*//g; s/\s*$//g' | grep '{query_result}'")
+                sanitize_cmd = r"sed -e 's/\s*|\s*/|/g; s/^\s*//g; s/\s*$//g'"
+                m.execute(fr"osinfo-query os --fields=name | tail -n +3 | {sanitize_cmd} | grep '{query_result}'")
```
martinpitt, line 1064: This is ok, but I'd actually find it more readable to split the whole command into multiple lines at the `|`:
```
m.execute("cmd 1 |"
          "cmd 2 |" ..)
```

test/check-machines-disks
```diff
@@ -837,12 +839,13 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
...
 
-        # Detach the iscsi disk before reaching teardown because shutting of the domains would sometimes hang when iscsi disks are attached
+        # Detach the iscsi disk before reaching teardown because shutting of the domains would sometimes hang when
+        # iscsi disks are attached.
         m.execute("virsh detach-disk subVmTest1 --target vdb")
 
         # AppArmor doesn't like the non-standard path for our storage pools
         if m.image in ["debian-testing"]:
-            self.allow_journal_messages(f'.* type=1400 .* apparmor="DENIED" operation="open" profile="libvirt.* name="{self.vm_tmpdir}.*')
+            self.allow_journal_messages(f'.* type=1400 .* apparmor="DENIED" operation="open" profile="libvirt.* name="{self.vm_tmpdir}.*')  # noqa: E501
```
martinpitt, line 848: FTR: Agreed to explicitly breaking the max line length for these.

martinpitt (commented):
Thanks! One question and one suggestion

test/machineslib.py
jelly: I've typed the new helper function and `self.machine` was not typed so mypy complains.

test/check-machines-disks
jelly: We can also try to remove them, how do you feel about that? Would be a separate PR and maybe future work? (when this occurs again)

test/machineslib.py
martinpitt: Ah, this is a type, not a member declaration, my bad.

test/check-machines-disks
martinpitt: Yes, the test could just put them into the correct dir -- /var/lib/libvirt/images/ or so?

martinpitt: I.e. fine for a follow-up, doesn't need to happen here.

test/check-machines-create
jelly: You want me to do this in a follow up?

martinpitt: Up to you, not that important.

PR #1869, opened 2024-10-18
create: Add a "view more" option to list of operating systems

mvollmer:
This will show all operating systems, regardless of EOL or release dates.  Operating systems with an expired EOL date will be annotated accordingly in the list.

Fixes #509

Demo: https://youtu.be/JXuX2GRQjCM

- [x] test
- [x] demo
- [ ] design review

image

image

Note that filtering is based both on End-Of-Life and release dates. Thus, we can't strictly say "Show unsupported operating systems" since the hidden ones are not all unsupported, some are just old.

PR #1872, opened 2024-10-23
create: Show all operating systems

### LINKED ISSUE #509: Allow inclusion of EOL OSes in list
My main contract is to work on MDN web docs' [Browser Compat Data](https://github.com/mdn/browser-compat-data) project, which involves a lot of old browsers and testing in them.  Because of this, I end up needing to create VMs for older OSes like Windows XP.  I would love for there to be an option to include EOL OSes in the dropdown list with a checkbox, much like the `virt-manager` program, something like the following:
A screenshot showing the virtual machine creation dialog, with an added checkbox to toggle end of life OS releases

Perhaps EOL releases can then be represented with a warning or hourglass icon in the list:
A screenshot showing the list of operating systems, with an hourglass icon next to the entry for Windows XP

mvollmer:
This will show all operating systems, regardless of EOL or release dates, but the recommended ones will be at the top.  Operating systems with an expired EOL date will be annotated accordingly in the list.

The tests have been brought up-to-date with regard to what OSes are actually in the list and are downloadable.

Fixes #509

Demo: https://youtu.be/I9VCtm2T49U

- [x] https://github.com/cockpit-project/cockpit/pull/21183
- [x] #1923 
- [x] tests
- [x] design review
 
-------

Machines: Download and install unsupported and older operating systems

Cockpit now offers to download and install also unsupported and older operating systems, while still promoting supported and newer ones.

Screen Shot 2024-12-11 at 09 11 03

### DIFF
--- a/src/components/create-vm-dialog/createVmDialogUtils.js
+++ b/src/components/create-vm-dialog/createVmDialogUtils.js
@@ -17,12 +17,18 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
+import cockpit from 'cockpit';
+import React from 'react';
+import { ExclamationTriangleIcon, OutlinedClockIcon } from "@patternfly/react-icons";
+
 import {
     getTodayYearShifted,
 } from "../../helpers.js";
 
 import * as python from "python.js";
 import autoDetectOSScript from './autoDetectOS.py';
+
+const _ = cockpit.gettext;
 
 const ACCEPT_RELEASE_DATES_AFTER = getTodayYearShifted(-3);
 const ACCEPT_EOL_DATES_AFTER = getTodayYearShifted(-1);
@@ -74,6 +80,14 @@ export function filterReleaseEolDates(os) {
         (os.eolDate && compareDates(ACCEPT_EOL_DATES_AFTER, os.eolDate) < 0) ||
         (!os.eolDate && os.releaseDate && compareDates(ACCEPT_RELEASE_DATES_AFTER, os.releaseDate) < 0)
     );
+}
+
+export function getOSDescription(os) {
+    if (os.eolDate && compareDates(ACCEPT_EOL_DATES_AFTER, os.eolDate) < 0)
+        return <span><ExclamationTriangleIcon /> {cockpit.format(_("Vendor support ended $0"), os.eolDate)}</span>;
+    if (!os.eolDate && os.releaseDate && compareDates(ACCEPT_RELEASE_DATES_AFTER, os.releaseDate) < 0)
+        return <span><OutlinedClockIcon /> {cockpit.format(_("Released $0"), os.releaseDate)}</span>;
+    return null;
 }
 
 export function compareDates(a, b, emptyFirst = false) {
--- a/src/components/create-vm-dialog/createVmDialogUtils.js
+++ b/src/components/create-vm-dialog/createVmDialogUtils.js
@@ -31,7 +31,6 @@ import autoDetectOSScript from './autoDetectOS.py';
 const _ = cockpit.gettext;
 
 const ACCEPT_RELEASE_DATES_AFTER = getTodayYearShifted(-3);
-const ACCEPT_EOL_DATES_AFTER = getTodayYearShifted(-1);
 const RHSM_TOKEN = "rhsm-offline-token";
 
 export const URL_SOURCE = 'url';
@@ -74,16 +73,15 @@ export function getOSStringRepresentation(os) {
 }
 
 export function filterReleaseEolDates(os) {
-    // Filter out all OSes their EOL date exists and is olrder than allowed
-    // or their EOL date does not exist but their release date is too old
+    // Filter out all OSes with elapsed EOL, or that have been released too long ago
     return !(
-        (os.eolDate && compareDates(ACCEPT_EOL_DATES_AFTER, os.eolDate) < 0) ||
+        (os.eolDate && new Date(os.eolDate).getTime() < Date.now()) ||
         (!os.eolDate && os.releaseDate && compareDates(ACCEPT_RELEASE_DATES_AFTER, os.releaseDate) < 0)
     );
 }
 
 export function getOSDescription(os) {
-    if (os.eolDate && compareDates(ACCEPT_EOL_DATES_AFTER, os.eolDate) < 0)
+    if (os.eolDate && new Date(os.eolDate).getTime() < Date.now())
         return <span><ExclamationTriangleIcon /> {cockpit.format(_("Vendor support ended $0"), os.eolDate)}</span>;
     if (!os.eolDate && os.releaseDate && compareDates(ACCEPT_RELEASE_DATES_AFTER, os.releaseDate) < 0)
         return <span><OutlinedClockIcon /> {cockpit.format(_("Released $0"), os.releaseDate)}</span>;
--- a/src/components/create-vm-dialog/createVmDialog.jsx
+++ b/src/components/create-vm-dialog/createVmDialog.jsx
@@ -41,6 +41,7 @@ import { useInit } from "hooks.js";
 import cockpit from 'cockpit';
 import store from "../../store.js";
 import { MachinesConnectionSelector } from '../common/machinesConnectionSelector.jsx';
+import { TypeaheadSelect } from 'cockpit-components-typeahead-select';
 import { FormHelper } from "cockpit-components-form-helper.jsx";
 import { FileAutoComplete } from "cockpit-components-file-autocomplete.jsx";
 import {
@@ -76,6 +77,7 @@ import {
     correctSpecialCases,
     filterReleaseEolDates,
     getOSStringRepresentation,
+    getOSDescription,
     needsRHToken,
     isDownloadableOs,
     loadOfflineToken,
@@ -405,10 +407,8 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, 
 class OSRow extends React.Component {
     constructor(props) {
         super(props);
-        const IGNORE_VENDORS = ['ALTLinux', 'Mandriva', 'GNOME Project'];
         const osInfoListExt = this.props.osInfoList
                 .map(os => correctSpecialCases(os))
-                .filter(os => filterReleaseEolDates(os) && !IGNORE_VENDORS.find(vendor => vendor == os.vendor))
                 .sort((a, b) => {
                     if (a.vendor == b.vendor) {
                         // Sort OS with numbered version by version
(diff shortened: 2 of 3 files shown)


mvollmer:
This is a variant of #1869.

martinpitt:
Thanks! See discussion in #1869, I like both of them, I just can't quite make up my mind.

garrett:
Specifically pointing out some issues in this screenshot crop:

image

1. We have some problems, due to PF, where the sub-text and the headings are exactly the same.
   - In another issue on this topic, [I suggested a little bit of CSS to move the sub-text to the right](https://github.com/cockpit-project/cockpit-machines/issues/509#issuecomment-1031406722) (which might need to be different, if we went that route). Adding a warning triangle :warning: even in grey might help as an alternative even or in addition.
2. Some entries are grouped under old, but they don't show EOL, like AlmaLinux 8 there...
3. Is AlmaLinux 8 actually old/EOL? It's available for download on https://almalinux.org/get-almalinux/  — or does that mean 8.0 specifically (instead of all 8.x), as 8.10 is available for download?

mvollmer:
> Some entries are grouped under old, but they don't show EOL, like AlmaLinux 8 there...

AlmaLinux 8 was released on 2021-03-30 (according to osinfo anyway), which is more than 3 years ago. We should probably also show release dates in the "old" group:

image

> Is AlmaLinux 8 actually old/EOL? It's available for download on https://almalinux.org/get-almalinux/ — or does that mean 8.0 specifically (instead of all 8.x), as 8.10 is available for download?

There is only "AlmaLinux 8" and "AlmaLinux 9" in the osinfo list. According to https://wiki.almalinux.org/release-notes/, it is actually AlmaLinux 8.3 that was released on 30 Mar 2021. When you install "AlmaLinux 8", you get AlmaLinux 8.10.

So, yeah, it's a bit messy.

garrett:
This is probably the right approach, to use groups instead of the expander thingy. It's fewer interactions by default, and the grouping brings the important ones to the top when browsing _and_ searching. (The other approach would still need to do this when searching.) The biggest downside to this approach is having a tinier scrollbar, which is mitigated when filtering the list.

Now that the old ones without EOL have a date there too, I think the main remaining thing we need to consider is to differentiate them, and I think the warning icon would make sense. Are we able to add that in PatternFly directly? (Otherwise, we might have to resort to a little bit of CSS to add the icon, if not.)

mvollmer:
> Now that the old ones without EOL have a date there too, I think the main remaining thing we need to consider is to differentiate them, and I think the warning icon would make sense.

So a warning icon for those with EOL date, and no warning icon for those a release date?

mvollmer:
Like this?

image

mvollmer:
With description on the same line:

image

mvollmer:
The deprecated `Select` component has problems: It shows very poor search performance when isGrouped is used, and it somehow can't find entries that end in ")".

image

image

These bugs are not present without `isGrouped`. I have not tried the non-deprecared `Select`. I think porting to it just got more important.

martinpitt:
> With description on the same line:

Nice on desktops, but does that work reasonably on mobile? Or is it some flex magic that automatically breaks it into two lines then?

mvollmer:
The search performance of the deprecated Select seems to be so bad that we get test timeouts while typing into it: https://cockpit-logs.us-east-1.linodeobjects.com/pull-1872-0fae78ec-20241024-124831-rhel-9-6/log.html#79

garrett:
1. The to-the-side thing looks too noisy with the icons.
2. Can we make the icons the same color as the text? They're meant to call attention, but not that much attention.
3. Perhaps we could use the clock icon (`fa-clock`) for old releases too. Having an icon there does help differentiate the footer text from the group text, and having these two icon types helps to express what each one is at a glance (without having to read it fully).
4. Thanks for implementing these suggestions!

Quick screenshot mod to show the icons all the same color as the text (roughly). If PF doesn't let us do that, setting `color: inherit` might work, although it would probably need `fill: currentcolor` instead (as it's SVG).

image

With this version (below) and these mods (grey icons, and icons of some sort for all with the old/expired text), I'd consider it pretty much ready from a design standpoint, provided the UX works as expected.

mvollmer:
> With this version (below) and these mods (grey icons, and icons of some sort for all with the old/expired text), I'd consider it pretty much ready from a design standpoint, provided the UX works as expected.

Tada:

image

garrett:
It looks great overall! Thanks for picking this up and adapting to the requested changes!

However, I see a pretty big issue: The menu extends off the page and isn't scrollable. 

Screenshot:

image

It needs a max height and it needs to be scrollable. I thought PF provided this being set already? (If not, there's a way to achieve this with some CSS, but I don't _think_ it should be necessary?)

You _could_ do something like this:

```css
.pf-v5-c-select__menu {
  max-block-size: 50vh;
  overflow: auto;
}
```

(It should not be that _exact_ CSS, as it would need be tied to that specific component, as you don't want to affect all `pf-v5-c-select__menu`s on the page.)

But, again, I'm pretty sure this should be handled by PF already.

mvollmer:
> However, I see a pretty big issue: The menu extends off the page and isn't scrollable.

Yes, this is fixed by https://github.com/cockpit-project/cockpit/pull/21156. This PR needs to be rebased to pick it up.

Also, the recent port to TypeaheadSelect solves this with an addition to machines.sccs.

mvollmer:
> But, again, I'm pretty sure this should be handled by PF already.

We have a fix for this in `pkg/lib/patternfly/patternfly-5-overrides.scss`. Without that, dropdowns don't seem to scroll ever.

mvollmer:
I have ported this over to the [TypeaheadSelect template](https://www.patternfly.org/components/menus/select/react-templates#typeahead). It doesn't have grouping, but I was able to fake some headers, which is all we want.

image

image

mvollmer:
> So, the TypeaheadSelect template uses a different style for the "clear" button. This clashes now with our Autocomplete widget, which uses the deprecated Select. Compare the "Installation source" and "Operating system" fields:

I have addressed that by copying the templatze into our project and adjusting it slightly to look like the deprecated Select.

Alternatively, I could try to pull the same "headers instead of groups" trick with the deprecated Select. It's doable, I guess, but... I think it's better to move forward here incrementally. Next step would be to stop using the deprecated Select in the FileAutocomplete component, and then replace the rest with the SimpleSelect template.

PatternFly 6 looks very similar to PF 5, and it has the same templates. So the work outlined above will likely carry over to PF 6 just fine.

mvollmer:
Let's reboot this, and put it on top of the TypeaheadSelect work in cockpit itself.

mvollmer:
@garrett, here are some current screenshots.

Right after openeing the dropdown:

image

Scrolling down a bit to the second header:

image

Filtered with entries from both categories:

image

Filtered with only one category:

image

garrett (approved):
Looks great! Thanks!

src/components/create-vm-dialog/createVmDialogUtils.js
```diff
@@ -76,6 +82,14 @@ export function filterReleaseEolDates(os) {
     );
 }
 
+export function getOSDescription(os) {
+    if (os.eolDate && compareDates(ACCEPT_EOL_DATES_AFTER, os.eolDate) < 0)
+        return <span><ExclamationTriangleIcon /> {cockpit.format(_("Vendor support ended $0"), os.eolDate)}</span>;
```
martinpitt, line 87: Did you keep the ACCEPT_EOL_DATES_AFTER on purpose here? That's the code path where libosinfo knows the EOL date, so it looks like this would accept "EOL < 1 year ago" as "new". That seems misleading to me?

src/components/create-vm-dialog/createVmDialogUtils.js
```diff
@@ -76,6 +82,14 @@ export function filterReleaseEolDates(os) {
...
+    if (!os.eolDate && os.releaseDate && compareDates(ACCEPT_RELEASE_DATES_AFTER, os.releaseDate) < 0)
+        return <span><OutlinedClockIcon /> {cockpit.format(_("Released $0"), os.releaseDate)}</span>;
```
martinpitt, line 89: FTR, I'm ok with keeping that heuristic.

martinpitt (changes requested):
Niiice! This looks very clean now, thanks for your heavy lifting with the TypeaheadSelect.

Can we do one more round of discussing the EOL thing?

This is totally release-note worthy, adding the label. [This screenshot](https://private-user-images.githubusercontent.com/3228183/390753968-417e8f12-b8e0-4af0-a608-7fc7caa96e8f.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzI4OTgxMjEsIm5iZiI6MTczMjg5NzgyMSwicGF0aCI6Ii8zMjI4MTgzLzM5MDc1Mzk2OC00MTdlOGYxMi1iOGUwLTRhZjAtYTYwOC03ZmM3Y2FhOTZlOGYucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTEyOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDExMjlUMTYzMDIxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ODQwNWZiOWU1ZTRiMzA4NGMzZGNiNTRhYTJjNWM0M2MzMjA4ZDNkMTRmYjc4NmZhYTc3Yzg2MzQ2NjNjYTBmOCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.0FjBIkQruTliSPuCqDWYpKTZBzU64KLwEiJJyKBry_Q) would work quite well IMHO, it shows the important point.

Cheers!

src/components/create-vm-dialog/createVmDialogUtils.js
mvollmer: Yeah, good point. Let's change that to "EOL date is in the past".

mvollmer: Yes, that's what I did.

martinpitt (approved):
Yay! Thank you!

I hit the retry button on the flakes. :crossed_fingers:

PR #1875, opened 2024-10-25
dependabot: Keep PF at current major version and downgrade sass

jelly:
Ignore sass updates which introduced a warning about incompatible changes on every build and don't upgrade to a PF major version as this is done in Cockpit first.


martinpitt (approved):
Cheers!

PR #1876, opened 2024-10-25
SASS to CSS where possible

jelly:
Let's try to reduce our dependency on SASS where we can already use simple CSS, this also found some unused overrides which I dropped.


martinpitt:
Nice! For n00bs like me: sass and scss are not the same thing, at least originally. In particular, SASS is much more like YAML, uses indentation instead of curly braces. But we don't actually have sass files, they are all *.scss. So our `sass` compiler takes SCSS as input, and we don't use the actual sass syntax? Did I get this right?

garrett:
> Nice! For n00bs like me: sass and scss are not the same thing, at least originally. In particular, SASS is much more like YAML, uses indentation instead of curly braces. But we don't actually have sass files, they are all *.scss. So our sass compiler takes SCSS as input, and we don't use the actual sass syntax? Did I get this right?

No, SASS is _preprocessor_ that has two formats: the original `SASS` language (not to be confused with SASS itself) and `SCSS`, which is "SASS CSS". Both have the same additional functions on top of CSS, but the SASS format is similar to Python/YAML where indentation matters, and curlies and semicolons don't. SASS then added SCSS later (a long time ago now) that re-adds the curlies and semicolons, for drop-in compatibility with CSS (so people don't have to juggle two different formats).

jelly:
In cockpit I've changed the SASS plugin to only take .scss files into account, we should probably land this first and then run cockpit-lib-update and then merge this.


https://github.com/cockpit-project/cockpit/pull/21214

jelly:
Split off the removals https://github.com/cockpit-project/cockpit-machines/pull/1908

jelly:
> @jelly Does that change design in any way? (pixel tests are fine) We already agreed with @garrett on the general direction, to move to standard CSS where possible.

It should not, I requested @garrett 's review for double checking.

PR #1891, opened 2024-11-01
vm: Limit size of descriptions

### LINKED ISSUE #1889: Too long description of the VM will cause an error
Description of problem:
When setting a description for a VM, if setting a too long string, such as 3964928 * 's', there will be an error shown on the dialog, maybe we need to limit the length of the description 

cockpit-machines-322-11-g63ac3f46

mvollmer:
Descriptions longer than about 128KiB will crash the Cockpit session with a protocol error since they are included in the open message for a "spawn" channel.

Only use the first 32k (about) when setting them to avoid that.

Fixes #1889

### DIFF
--- a/src/libvirtApi/domain.js
+++ b/src/libvirtApi/domain.js
@@ -912,6 +912,11 @@ async function domainSetXML(vm, option, values) {
 }
 
 export async function domainSetDescription(vm, description) {
+    // The description will appear in a "open" message for a "spawn"
+    // channel, and excessive lengths will crash the session with a
+    // protocol error. So let's limit it to a reasonable length here.
+    if (description.length > 32000)
+        description = description.slice(0, 32000);
     await domainSetXML(vm, "metadata", { description });
 }
 


src/libvirtApi/domain.js
```diff
@@ -912,6 +912,12 @@ async function domainSetXML(vm, option, values) {
 }
 
 export async function domainSetDescription(vm, description) {
+    // The description will appear in a "open" message for a "spawn"
+    // channel, and excessive lengths will crash the session with a
+    // protocol error. So let's limit it to a reasonable length here.
+    console.log("DE", description.length);
+    if (description.length > 32000)
+        description = description.slice(0, 32000);
```
cockpituous, line 920: This added line is not executed by any test.

src/libvirtApi/domain.js
```diff
@@ -912,6 +912,12 @@ async function domainSetXML(vm, option, values) {
 }
 
 export async function domainSetDescription(vm, description) {
+    // The description will appear in a "open" message for a "spawn"
+    // channel, and excessive lengths will crash the session with a
+    // protocol error. So let's limit it to a reasonable length here.
+    console.log("DE", description.length);
```
jelly, line 918: That's a debug leftover I assume?

mvollmer: Yes! Thanks.

src/libvirtApi/domain.js
```diff
@@ -912,6 +912,11 @@ async function domainSetXML(vm, option, values) {
 }
 
 export async function domainSetDescription(vm, description) {
+    // The description will appear in a "open" message for a "spawn"
+    // channel, and excessive lengths will crash the session with a
+    // protocol error. So let's limit it to a reasonable length here.
+    if (description.length > 32000)
+        description = description.slice(0, 32000);
```
cockpituous, line 919: This added line is not executed by any test.

jelly (approved):

PR #1892, opened 2024-11-01
Smaller line length

jelly:


test/check-machines-create
```diff
@@ -174,22 +174,24 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
...
 
         # user login
+        errors = {"create-vm-dialog-user-login": "User login must not be empty when user password is set"}
         runner.checkDialogFormValidationTest(TestMachinesCreate.VmDialog(self, sourceType='cloud',
                                                                          storage_size=10, storage_size_unit='MiB',
                                                                          location=config.VALID_DISK_IMAGE_PATH,
                                                                          user_password="catsaremybestfr13nds",
                                                                          create_and_run=True),
-                                             {"create-vm-dialog-user-login": "User login must not be empty when user password is set"})
+                                                                        errors)
```
martinpitt, line 183: This indentation change is confusing -- it now looks like `errors` is an argument to `VmDialaog()`, but it's not -- it's the second argument to `checkDialogFormValidationTest`. Same for the instances below.

martinpitt (changes requested):
Nice, thank you!

test/check-machines-create
jelly: Good catch.

martinpitt (approved):
*squeeze* *squeeze* Thanks!

Issue #1890, opened 2024-11-01
The VM description doesn't display the '\n'


yunmingyang:
Description of problem:
When setting a description for a VM, if setting a '\n' in the description, like '
1

2
'. The \n is not shown on the page. But, if checking by 'virsh dumpxml ${vm}', the '\n' could be shown correctly in the <description> element

Version-Release number of selected components (if applicable):
cockpit-machines-322-11-g63ac3f46

mvollmer:
I can't reproduce this, and I am not sure I have followed your steps correctly.

If put this into the dialog:

![image](https://github.com/user-attachments/assets/e049bfc1-959e-4a94-b35d-1a99f295cdca)

I get this in the details page of the machine:

![image](https://github.com/user-attachments/assets/04b04aa8-49c5-4889-a3b3-3eff497014f0)

The two lines `1\n` and `2` are both shown, and the `\n` sequence is also rendered correctly.

Sequences like `\n` have no special meaning when you type them into the description dialog. Did you expect them to be interpreted and turned into a newline character? If you want multiple lines, just put multiple lines into the text field.

yunmingyang:
![Screenshot from 2024-11-04 13-48-51](https://github.com/user-attachments/assets/b7f952dd-c519-4cc5-b02c-0ecb0cf4f8fe)

yunmingyang:
Just like that, for \n, I mean typing enter @mvollmer

mvollmer:
> Just like that, for \n, I mean typing enter @mvollmer

Ahh, you mean an empty line! :-)

Yes, I can reproduce that. Empty lines are indeed omitted when displaying a description. Thanks for finding this!

mvollmer:
Fixed in https://github.com/cockpit-project/cockpit-machines/pull/1901

PR #1901, opened 2024-11-11
Improve rendering of VM descriptions

### LINKED ISSUE #1888: The description is not entirely displayed if the description is too long during setting the VM description
Description of problem:
When setting the VM description, if the string is too long, the description may not be entirely displayed

cockpit-machines-322-11-g63ac3f46

mvollmer:
Render the description as a single paragraph, with preserved line breaks and word breaking on overflow.

Fixes #1888, #1890.

### DIFF
--- a/src/machines.scss
+++ b/src/machines.scss
@@ -102,3 +102,8 @@
 #storage-pool-delete-modal span.pf-v5-c-check__body {
     margin-block-start: 0;
 }
+
+.vm-description {
+    white-space: pre-wrap;
+    overflow-wrap: break-word;
+}
--- a/src/components/vm/vmDetailsPage.jsx
+++ b/src/components/vm/vmDetailsPage.jsx
@@ -77,9 +77,9 @@ export const VmDetailsPage = ({
             </div>
             {
                 vm.inactiveXML.description &&
-                    <div className="vm-description">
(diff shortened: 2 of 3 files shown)


mvollmer:
Looks like this this, for example:

image

The last line has no spaces in it at all, and is broken in the middle of words to avoid overflow.

The second to last line is long but has spaces. It is broken between words to avoid overflow.

src/machines.scss
```diff
@@ -102,3 +102,8 @@
 #storage-pool-delete-modal span.pf-v5-c-check__body {
     margin-block-start: 0;
 }
+
+.vm-description {
+    white-space: pre-wrap;
+    overflow-wrap: break-word;
+}
```
garrett, line 109: We could also have a `hyphens: auto;` in there; it's relatively new-ish in baseline 2023 and will hyphenate things it needs to wrap with break-word.

https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

> The hyphens CSS property specifies how words should be hyphenated when text wraps across multiple lines. It can prevent hyphenation entirely, hyphenate at manually-specified points within the text, or let the browser automatically insert hyphens where appropriate.

garrett (approved):
Looks good!

The only suggestion I have is adding in `hyphenations: auto` to let the browser hyphenate when appropriate.

jelly (approved):

PR #1917, opened 2024-11-22
Dont show system session for user

jelly:
This throws an error when trying to create a system session vm as limited user.


test/check-machines-create
```diff
@@ -2787,6 +2787,25 @@ vnc_password= "{vnc_passwd}"
...
+        b.wait_visible("#connectionName-system")
+        b.click("#create-vm-dialog .pf-v5-c-button.pf-m-link")
+        b.wait_not_present("#create-vm-dialog")
+
+        b.drop_superuser()
+        self.waitPageInit()
+
+        b.click("#create-new-vm")
+        b.wait_visible("#create-vm-dialog")
+        b.wait_not_present("#connection")
```
martinpitt, line 2807: Do we have any other test which covers VM creation in unprivileged mode? If so, this would better fit there. If not, that's a major gap indeed, and this test should actually create the VM and check that it starts and boots fine (including the "wait for login prompt").

jelly: I believe this is confusingly what TestMachinesLifecycle does:

```
    def _testBasic(self, user=None, superuser=True, expect_empty_list=False, run_pixel_tests=False):
        b = self.browser
        m = self.machine

        # We want no initial watchdog
        args = self.createVm("subVmTest1", os="linux2016")

        self.login_and_go("/machines", user=user, superuser=superuser)
```

So we can drop this test

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1331,12 +1332,13 @@ class CreateVmModal extends React.Component {
 
         const detailsTab = (
             <>
+                {superuser.allowed &&
```
mvollmer, line 1335: Hmm, I think people can create polkit rules to allow non-priv users access to the "system" connection. In fact cockpit-machines always connects to both "system" and "session" on initialization and tolerates AccessDenied errors for non-priv users.

So I think we should make a new global flag that says "can't access system connection", and use that instead of `superuser.allowed` in this PR.  The place to set that flag is app.jsx:94, I guess.

jelly: Polkit is a disaster if you ask me, we have no way to detect it. But plumbing the state through should work. But looking at our code this is all a bit confusing I guess we look at `systemSocketInactive`?

jelly: This turns into a yakshave...

```
    } else if ((superuser.allowed && systemSocketInactive) || loadingResources) {
```

So seems my solution is "correct" :) But ofcourse we can rework this.

jelly: Or not, not sure why this even exists.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1056,7 +1056,7 @@ class CreateVmModal extends React.Component {
             validate: false,
             vmName: '',
             suggestedVmName: '',
-            connectionName: LIBVIRT_SYSTEM_CONNECTION,
+            connectionName: this.props.systemSocketAvailable ? LIBVIRT_SYSTEM_CONNECTION : LIBVIRT_SESSION_CONNECTION,
```
cockpituous, line 1059: This added line is not executed by any test.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1331,12 +1331,13 @@ class CreateVmModal extends React.Component {
...
         const detailsTab = (
             <>
+                {this.props.systemSocketAvailable &&
                 <MachinesConnectionSelector
                     id='connection'
                     connectionName={this.state.connectionName}
                     onValueChanged={this.onValueChanged}
                     loggedUser={loggedUser}
-                    showInfoHelper />
+                    showInfoHelper />}
```
cockpituous, line 1340: This added line is not executed by any test.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1059,7 +1059,7 @@ class CreateVmModal extends React.Component {
             validate: false,
             vmName: '',
             suggestedVmName: '',
-            connectionName: LIBVIRT_SYSTEM_CONNECTION,
+            connectionName: this.props.systemSocketAvailable ? LIBVIRT_SYSTEM_CONNECTION : LIBVIRT_SESSION_CONNECTION,
```
cockpituous, line 1062: This added line is not executed by any test.

cockpituous, line 1062: This added line is not executed by any test.

src/components/create-vm-dialog/createVmDialog.jsx
mvollmer: Right, and with the move virt-xml, being able to connect to the system session vuia D-Bus does not mean that you can successfully run virt-xml.

Let's keep the superuser.allowed check.

mvollmer: Oh, you changed it already....

mvollmer:
Sorry, I have to look at this again. Luckily I learned a thing or two about c-machines in the mean time.

src/components/create-vm-dialog/createVmDialog.jsx
mvollmer: Or does virt-xml call the D-Bus API?

mvollmer: No, but polkit works inside th actual libvirt daemon as well, and I expect virt-install, virt-xml etc to be subject to the same polkit rules as D-Bus clients.

I'll research this properly.

mvollmer (approved):
Thanks, I think this is the right thing to do.

Being able to connecting to the `qemu:///system` connection does indeed mean that `virt-install` and `virt-xml` are also able to work. For example, a limited user in the `libvirt` group can create and manage VMs on the system connection, using our mix of D-Bus, virt-install, and virt-xml.

Libvirt can [use polkit internally](https://libvirt.org/acl.html) (not via D-Bus) for fine grained access control. I have not tested it, but I expect that the D-Bus interface as well as virt-install and virt-xml are subject to it equally.

src/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -1334,12 +1334,13 @@ class CreateVmModal extends React.Component {
 
         const detailsTab = (
             <>
+                {this.props.systemSocketAvailable &&
```
mvollmer: Instead of completely hiding the connection selector, it should turn read-only and just say "User session" without any radio buttons.

jelly: Sure :) But I did something else woops just disabled both.

src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1219,7 +1220,7 @@ class CreateVmModal extends React.Component<CreateVmModalProps, CreateVmModalSta
             validate: false,
             vmName: '',
             suggestedVmName: '',
-            connectionName: LIBVIRT_SYSTEM_CONNECTION,
+            connectionName: this.props.systemSocketAvailable ? LIBVIRT_SYSTEM_CONNECTION : LIBVIRT_SESSION_CONNECTION,
```
cockpituous, line 1223: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-1917-009b3757-20250904-162346-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/create-vm-dialog/createVmDialog.jsx
mvollmer: It's complicated. There are two places to define policy, one for native policy for libvirtd that affects direct clients like virsh and virt-xml, and another one for the D-Bus API, which affects clients using that, like half of cockpit-machines.

This is annoying but it can be made to work, and we should not break people arbitrarily that do that.

mvollmer:
> but I expect that the D-Bus interface as well as virt-install and virt-xml are subject to it equally.

Hah, being young and naive...

src/components/create-vm-dialog/createVmDialog.jsx
mvollmer: Hmm. I have pushed a version without radio buttons.

mvollmer:
The tests don't want to go green... I'll have a look.

mvollmer:
Rebased. Now looking at the tests...

src/components/common/machinesConnectionSelector.tsx
```diff
@@ -99,16 +101,21 @@ export const MachinesConnectionSelector = ({
...
-                   name="connectionName"
-                   id="connectionName-system"
-                   label={rephraseUI("connections", LIBVIRT_SYSTEM_CONNECTION)} />
-            <Radio isChecked={connectionName == LIBVIRT_SESSION_CONNECTION}
-                   onChange={() => onValueChanged('connectionName', LIBVIRT_SESSION_CONNECTION)}
-                   name="connectionName"
-                   id="connectionName-session"
-                   label={rephraseUI("connections", LIBVIRT_SESSION_CONNECTION)} />
+            { isDisabled
+                ? rephraseUI("connections", connectionName)
```
cockpituous, line 105: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-1917-76ac05de-20250915-164354-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/machinesConnectionSelector.tsx
```diff
@@ -99,16 +101,21 @@ export const MachinesConnectionSelector = ({
...
+                : <>
+                    <Radio isChecked={connectionName === LIBVIRT_SYSTEM_CONNECTION}
+                        onChange={() => onValueChanged('connectionName', LIBVIRT_SYSTEM_CONNECTION)}
```
cockpituous, line 108: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-1917-76ac05de-20250915-164354-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/create-vm-dialog/createVmDialog.tsx
cockpituous, line 1223: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-1917-76ac05de-20250915-164354-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
Argghhh, dropped the ball.

src/app.tsx
```diff
@@ -70,6 +70,7 @@ export const App = () => {
     const [loadingResources, setLoadingResources] = useState(true);
     const [error, setError] = useState('');
     const [systemSocketInactive, setSystemSocketInactive] = useState(false);
+    const [systemSocketAvailable, setSystemSocketAvailable] = useState(false);
```
mvollmer, line 73: I think `systemSocketInactive == !systemSocketAvailabl` after initialization is done. Thus, we don't need systemSocketAvailable, really.

src/components/common/machinesConnectionSelector.tsx
cockpituous, line 108: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-1917-08d0dae6-20251120-084558-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
This got bulldozed over by my recent initialization etc changes. I'll start from scratch with my own PR.

PR #1922, opened 2024-11-26
test: reduce line length to 118

jelly:
This brings machines in line with our other projects line length configuration.


martinpitt (approved):
*squeeeeze* *nnnng* *thunk*

thanks!

PR #1929, opened 2024-11-28
test: Only run alert and logout cleanups if Cockpit is actually opened

martinpitt:
Some tests like `testAddDiskNFS` fail during setup [1], so the test teardown runs while the browser still shows a blank page.

Use the new `Browser.have_test_api()` from
https://github.com/cockpit-project/cockpit/pull/21349 instead of merely checking if the browser is running to clean that up.

[1] https://github.com/cockpit-project/bots/issues/7137

----

 - [ ] Land https://github.com/cockpit-project/cockpit/pull/21349
 - [ ] Pull that in with its official ref, instead of the one from my fork


martinpitt:
Oh yes, [this testAddDiskNFS failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1929-ebaa19d8-20241128-124800-fedora-41/log.html) looks a lot cleaner -- i.e. just the NFS crash, not testlib going crazy with trying to talk to a nonexisting Cockpit page.

jelly (approved):
Thanks

PR #1957, opened 2024-12-23
packaging: Depend on virt-install

martinpitt:
Our code increasingly uses `virt-xml`, which is not related to creating new machines but administering existing ones. That is also shipped in the `virt-install` package, so promote it from Recommends to Requires.

Leave the `virtInstallAvailable` runtime detection in place for the beiboot case.

Side issue in https://issues.redhat.com/browse/RHEL-61727


martinpitt:
@jelly Ugh @ [build failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1957-5c18c3cc-20241223-073544-arch/log.html) -- dependencies need to be present during package *build*? Is there some concept of "build dependencies" in arch?

mvollmer (approved):
Good catch, thanks!

PR #1972, opened 2025-01-13
consoles: Some redesign

mvollmer:
Demo: https://youtu.be/1DbJH9FRMsw

### DIFF
--- a/src/components/vm/consoles/consoles.css
+++ b/src/components/vm/consoles/consoles.css
@@ -18,6 +18,10 @@
      grid-template-rows: min-content 1fr;
 }
 
+.vm-console-main {
+    grid-area: main;
+}
+
 .vm-console-footer {
     grid-area: 3 / 1 / 4 / 3;
 }
--- a/src/components/vm/consoles/consoles.css
+++ b/src/components/vm/consoles/consoles.css
@@ -18,6 +18,10 @@
      grid-template-rows: min-content 1fr;
 }
 
+.vm-console-footer {
+    grid-area: 3 / 1 / 4 / 3;
+}
+
 .consoles-page-expanded .actions-pagesection .pf-v5-c-page__main-body {
      padding-block-end: 0;
 }
--- a/src/components/vm/consoles/consoles.jsx
+++ b/src/components/vm/consoles/consoles.jsx
@@ -27,6 +27,8 @@ import SerialConsole from './serialConsole.jsx';
 import Vnc from './vnc.jsx';
 import DesktopConsole from './desktopConsole.jsx';
 import { ReplaceSpiceDialog } from '../vmReplaceSpiceDialog.jsx';
+import { AddVNC } from './vncAdd.jsx';
+import { EditVNCModal } from './vncEdit.jsx';
 
 import {
     domainCanConsole,
@@ -41,17 +43,36 @@ const _ = cockpit.gettext;
 const VmNotRunning = ({ vm, vnc, spice }) => {
     const Dialogs = useDialogs();
 
+    function add_vnc() {
+        Dialogs.show(<AddVNC
+                         idPrefix="add-vnc"
+                         vm={vm} />);
+    }
+
+    function edit_vnc() {
+        Dialogs.show(<EditVNCModal
+                         idPrefix="edit-vnc"
+                         consoleDetail={vnc}
+                         vmName={vm.name}
+                         vmId={vm.id}
+                         connectionName={vm.connectionName} />);
+    }
+
     return (
         <div id="vm-not-running-message">
             <div>{_("Please start the virtual machine to access its console.")}</div>
             { vnc
                 ? <div>
                       <b>{_("VNC")}</b> {vnc.address}:{vnc.port}
-                      <Button variant="link">{_("Edit")}</Button>
+                      <Button variant="link" onClick={edit_vnc}>
+                          {_("Edit")}
+                      </Button>
                   </div>
                 : <div>
-                      <b>{_("VNC")}</b> {_("Not supported")}
-                      <Button variant="link">{_("Add support")}</Button>
+                      <b>{_("VNC")}</b> {_("not supported.")}
+                      <Button variant="link" onClick={add_vnc}>
+                          {_("Add support")}
+                      </Button>
                   </div>
             }
             { spice
@@ -142,9 +163,7 @@ class Consoles extends React.Component {
                                                   vmName={vm.name}
                                                   spawnArgs={domainSerialConsoleCommand({ vm, alias: pty.alias })} />))}
                 <Vnc type="VncConsole"
-                     vmName={vm.name}
-                     vmId={vm.id}
-                     connectionName={vm.connectionName}
+                     vm={vm}
                      consoleDetail={vnc}
                      onLaunch={() => this.onDesktopConsoleDownload('vnc')}
                      onAddErrorNotification={onAddErrorNotification}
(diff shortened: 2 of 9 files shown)


mvollmer:
> Some random thoughts while watching the video:
> 
>     * Let's present VNC port -1 as "automatic" or "default" in the selector.

In the dialogs? Yes, absolutely.

>     * Showing the listening address in the "VNC: ..." below the console is too misleading IMHO. It could say something like "VNC: hostname:port (listening on 127.0.0.1)", with replacing "hostname" with what cockpit is connected to (I think we can get this information somehow)

Yeah, but if the listen port is 127.0.0.1 for example, then telling people to connect to "dev.mvo.lan" isn't helping, no?

I think the best advice we can give is to run `virt-viewer -c ssh://<host> <machine>`. I think this will tunnel the VNC traffic over SSH.

>     * The "Launch viewer" button is very kaputt.. Since Firefox 134 it just downloads a file with a completely random name and no extension, and I think chromium is no better (called "download"). If that is hard to fix (I didn't check), let's drop that button; but then again, how hard can it be to give the download a proper file name? (I know, famous last words)

It works for me (if the network is setup properly and the listening address works for connecting as the client).  Firefox will download the file and you can click on it in the dropdown that opens. That will open the viewer.
 
>     * Nice idea to add the "Replace SPICE devices" to the spice console table row!

Alright, I wasn't sure! :-)
 
>     * What happened to the instructions how to install a viewer? We don't need/want them any more?

I can't say. I got lazy and didn't add them in my experiments, but we probably should keep them.

martinpitt:
> > * Showing the listening address in the "VNC: ..." below the console is too misleading IMHO. It could say something like "VNC: hostname:port (listening on 127.0.0.1)", with replacing "hostname" with what cockpit is connected to (I think we can get this information somehow)
> 
> Yeah, but if the listen port is 127.0.0.1 for example, then telling people to connect to "dev.mvo.lan" isn't helping, no?
> I think the best advice we can give is to run `virt-viewer -c qemu+ssh://<host>/system <machine>`. This will tunnel the VNC traffic over SSH.

Right. I think we should handle this specially -- with "0.0.0.0" or something else (rare, and benefit of the doubt) we give "cockpit-hostname:port", and with 127.0.0.1 the ssh tunneling. That should cover the common cases?

> It works for me (if the network is setup properly and the listening address works for connecting as the client). Firefox will download the file and you can click on it in the dropdown that opens. That will open the viewer.

Ah, that works even with these totally random file names? nice! I thought they had to be called `*.vv`.

mvollmer:
> > > * Showing the listening address in the "VNC: ..." below the console is too misleading IMHO. It could say something like "VNC: hostname:port (listening on 127.0.0.1)", with replacing "hostname" with what cockpit is connected to (I think we can get this information somehow)
> > 
> > 
> > Yeah, but if the listen port is 127.0.0.1 for example, then telling people to connect to "dev.mvo.lan" isn't helping, no?
> > I think the best advice we can give is to run `virt-viewer -c qemu+ssh://<host>/system <machine>`. This will tunnel the VNC traffic over SSH.
> 
> Right. I think we should handle this specially -- with "0.0.0.0" or something else (rare, and benefit of the doubt) we give "cockpit-hostname:port", and with 127.0.0.1 the ssh tunneling. That should cover the common cases?

I guess... We also need to use the right address in the *.vv file, but that can't do SSH tunneling... but it's of course much better to get a "connection refused" error for the correct address than to connect to 127.0.0.1 and maybe pick up the console of the wrong machine.

garrett:
I don't understand what this PR is trying to do...

1. This doesn't look like the mockups for the "new" design (it's _several years old_ at this point in time) of the console.
2. It doesn't fix any of the issues, such as having `.vv` on the filename (which was a regression, as it used to have it) so the file actually works; we need to specify the filename to the ["download" attribute to the anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download).
3. It also doesn't automatically adjust the firewall to allow non-localhosts to actually connect (so it's actually useful for our users, as it isn't at all right now, and won't be until we finally add firewall integration).
4. Spice is no longer supported, so it shouldn't be in the UI.
5. We shouldn't have a dropdown for selecting tabs; they should be "tabs" of some sort... in this case, [the toggle group component](https://v5-archive.patternfly.org/components/toggle-group), which does support overflow in case there are too many serial text connections... although in most cases, there will be 1 or perhaps 2, not 15 or 150. Dropdowns should generally not be used for selection unless there are 5 or more items and/or it is indeterminate with no other method that would work better. And dropdowns should never be used to change the view of a content; there are tabs and tab-like components which are better used instead.

What we need to do instead:

- Set aside some time to talk about what we need to do in the Machines page, as it has sat around mainly dormant for years, aside from fixups here and there. We can do some of this at the F2F.
- Dust off the mockups from years ago and retool them to address what we want to do in the Machines page.
- Actually fix the desktop viewer... or even consider removing it, if we don't want to actually make it work. Right now, it no longer has the correct file, so there's no association (so it does nothing), relies on a deprecated viewer (IIRC), and doesn't work for remote machines. In real life, nobody really runs Cockpit locally; it's meant for remote servers, so this feature has never worked. We _can_ make it work, but we should decide to fix and finish the implementation, if so.
- Fix the expand action so people can actually get use more real estate for their connection, instead of dealing with a scaled down view with a bunch of stuff around it.

Current state of the mockups... they still need work; we need to discuss it and decide what to do with various features, but here they are right now:


Machines-Console

Machines-Console-noVNC

There are other parts to the mockups too (showing extended information for remote connections, configuring the port, etc.), but they heavily need more work (as they were made before we dropped SPICE, for example). These are old designs done to address needs that people who use Cockpit Machines actually have, as a "quick" response to the issues they encountered. They were intended to get the ball rolling and to have a conversation about how to fix Cockpit Machines, but I haven't focused on them, as nobody actually picked them up until now.

I'm super happy you're thinking about this and working on this, but we need to make sure we don't "put the cart before the horse". We need to have a dicussion about what is needed, figure out how it should all fit together, work on a design, make new mockups to address this, and _then_ implement it. (Part of the problem of the Machines page right now is that it's a hodgepodge of various features shoved together over the years, with small updates here and there on top; we need to fix it and make it more cohesive and address the various conceptual and implementation issues.)

martinpitt:
> Spice is no longer supported, so it shouldn't be in the UI.

Clarification: It isn't supported in RHEL 9/10. Everywhere else (Fedora/Debian/Arch/Ubuntu) it still is.

mvollmer:
> However, I don't understand what this PR is trying to do...

This was just me digging into c-machines consoles in general, and trying to figure out how to make progress with the stuff in #1795, without introducing all the bugs it has right now.

Before doing this, I wasn't really aware how broken the "Launch viewer" button actually is, I have to admit.

mvollmer:
> 1. This doesn't look like the mockups for the "new" design (it's _several years old_ at this point in time) of the console.

Not yet :-)  Thanks for clarifying how multiple serial consoles should be handled.

garrett:
> Clarification: It isn't supported in RHEL 9/10. Everywhere else (Fedora/Debian/Arch/Ubuntu) it still is.

From what I previously understood: The primary developers of SPICE was Red Hat, and Red Hat decided to stop actively developing it. (Being packaged and shipped in a distro is not the same thing as _supported_ either commercially or by a community of coders, of course. Just because it exists in a distribution doesn't mean it's being developed or maintained, or "supported" in any sense of the word.)

There are a handful of repos at https://gitlab.freedesktop.org/spice/?sort=latest_activity_desc that have had updates as of weeks and months ago, at least. While SPICE is already rather feature complete, I do see some bugfixes in a few repos over the past few years. So that's great to see!

mvollmer:
I want to start over with this. Thanks for all the comments, I'll not forget them.

garrett:
> I want to start over with this. Thanks for all the comments, I'll not forget them.

Thanks for making this PR! It's rekindled and moved the conversation along for all of us! :+1:

PR #1975, opened 2025-01-14
test: Check VM lifecycle with a sysadm_u user

martinpitt:
By default, users are `unconfined_u`.
    
First step in https://bugzilla.redhat.com/show_bug.cgi?id=2328030

---

Of course we should also test `user_u`. I have the beginnings of a test locally, but it's currently completely broken, see the above bz. So let's postpone that test until SELinux gets fixed (and then add naughties for the older releases).


martinpitt:
This triggers a lot of flakes, but checking the [new test on rhel-10](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1975-6101efe8-20250114-145850-rhel-10-0/log.html#85), that could be real.

martinpitt:
Whoops, [needs skipping on tumbleweed](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1975-357376ef-20250115-065253-opensuse-tumbleweed/log.html#87)

mvollmer (approved):
Thanks!

PR #1978, opened 2025-01-16
package.json: ipaddr.js is a runtime dependency

jelly:
This was wrongly added as devDependencies while it is a runtime dependency.

---

```
src/components/networks/utils.js:import * as ipaddr from "ipaddr.js";
```


martinpitt (approved):
Oops, thanks for spotting!

PR #1979, opened 2025-01-16
test: wait until the VM is running before a poweroff

jelly:
On RHEL 10 there is seems to be a race where the machine is powered off when it isn't fully running causing an libvirt error which in turns causes the test to flake.

---

This should solve [this flake](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1976-86f02498-20250114-181748-rhel-10-0/log.html) fails 40% of the time. Which shows an error:

image


martinpitt (approved):
Thanks! This looks good, and the current failures are all in other tests (i.e. other flakes). Not sure what's up with TF..

PR #1982, opened 2025-01-17
consoles: Don't use listening address to connect to VNC

mvollmer:
The listening address is most likely 127.0.0.1 (libvirt default) or 0.0.0.0 (yolo config). Neither of these works remotely.

Let's tell the remote viewer some address that has a chance of working. More complicated setups, like a bastion host or connecting via SSH inside Cockpit, might not work because of networking routing challanges, but at least we get better error messages like "No route to host" etc.

Also, with the libvirt default of only listening on 127.0.0.1, no remote connection will work. But again, we now at least get the expected "Connection refused" error message.

Demo: https://youtu.be/zjQlyt3SzG4

-------------------

Machines: "Launch viewer" now uses correct address

The "Launch viewer" button in the "Console" card of a virtual machine would tell the remote viewer application to connect to the listening address of the VNC server. This is wrong when the server listens on 127.0.0.1 (libvirt default) or 0.0.0.0 (allow connections on all interfaces). Now Cockpit will tell the remote viewer application to connect to the same domain name or IP address that Cockpit itself is using to reach the virtual machine host.


src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,23 @@ class Consoles extends React.Component {
...
         const { vm } = this.props;
         // fire download of the .vv file
-        domainDesktopConsole({ name: vm.name, id: vm.id, connectionName: vm.connectionName, consoleDetail: vm.displays.find(display => display.type == type) });
+        const consoleDetail = vm.displays.find(display => display.type == type);
+
+        let address;
+        if (cockpit.transport.host == "localhost") {
+            const app = cockpit.transport.application();
+            if (app.startsWith("cockpit+="))
+                address = app.substr(9);
```
cockpituous, line 97: This added line is not executed by any test.

src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,23 @@ class Consoles extends React.Component {
...
+            else
+                address = window.location.hostname;
+        } else {
+            address = cockpit.transport.host;
+            const pos = address.indexOf("@");
+            if (pos >= 0)
+                address = address.substr(pos + 1);
```
cockpituous, line 104: These 4 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
cockpituous, line 97: This added line is not executed by any test.

src/components/vm/consoles/consoles.jsx
cockpituous, line 104: This added line is not executed by any test.

src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,25 @@ class Consoles extends React.Component {
...
         const { vm } = this.props;
         // fire download of the .vv file
-        domainDesktopConsole({ name: vm.name, id: vm.id, connectionName: vm.connectionName, consoleDetail: vm.displays.find(display => display.type == type) });
+        const consoleDetail = vm.displays.find(display => display.type == type);
+
+        let address;
+        if (cockpit.transport.host == "localhost") {
+            const app = cockpit.transport.application();
+            if (app.startsWith("cockpit+=")) {
+                address = app.substr(9);
```
cockpituous, line 97: These 2 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,25 @@ class Consoles extends React.Component {
...
+            } else {
+                address = window.location.hostname;
+            }
+        } else {
+            address = cockpit.transport.host;
+            const pos = address.indexOf("@");
+            if (pos >= 0) {
+                address = address.substr(pos + 1);
```
cockpituous, line 105: These 2 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
martinpitt: Hmm, our tests connect to 127.0.0.2:9091, from the PoV of the VM it's remote. But as far as I know this isn't about browser → ws, but ws → session, no? I.e. no "remote ssh" session?

martinpitt: I am assuming that `cockpit.transport.host == "localhost")` means "local `cockpit-session` session", as opposed to "ssh session via 'Connect to:'" (bastion host).

Then I don't understand this `if`. What would the bit after "cockpit+=" be if not "localhost"? It appears to me that we would only have a cockpit+ "redirect" if the transport host is *not* localhost. When would that ever fire?

src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,25 @@ class Consoles extends React.Component {
...
+            }
+        }
+
+        domainDesktopConsole({ name: vm.name, consoleDetail: Object.assign({}, consoleDetail, { address }) });
```
martinpitt, line 109: `{...consoleDetail, address}` reads a bit nicer IMHO, but up to you.

test/check-machines-consoles
```diff
@@ -254,5 +258,76 @@ fullscreen=0
                                   "Failed when connecting: Connection closed")
 
 
+class TestMultiMachinesConsoles(ConsoleCase):
+    provision = {  # noqa: RUF012
+        "machine": {"address": "10.111.113.1/20"},
+    }
```
martinpitt, line 264: This is still only one machine -- seems you could get this more cheaply with the already running nondestructive machine? `172.27.0.15` is reliable, a couple of cockpit tests refer to it as well.

test/check-machines-consoles
```diff
@@ -254,5 +258,76 @@ fullscreen=0
...
+        self.machine.write("/etc/cockpit/cockpit.conf", "[Session]\nWarnBeforeConnecting=false\n", append=True)
+
+        self.createVm(name, graphics="vnc")
+
+        # Direct login via SSH
+
+        b.open(f"/={my_ip}/machines")
+        b.set_val('#login-user-input', "admin")
+        b.set_val('#login-password-input', "foobar")
+        b.click("#login-button")
```
martinpitt, line 284: That should be `self.login_and_go("/machines", host=my_ip)`. I wonder why you don't' have to `m.start_cockpit()` here, it's usually required?

martinpitt (changes requested):
Thanks! Let's discuss this a little bit. My first two comments were not part of the review.

test/check-machines-consoles
```diff
@@ -254,5 +258,76 @@ fullscreen=0
...
+        b.click("#DesktopViewer button")
+        b.wait_not_present("#pf-v5-c-console__type-selector + .pf-v5-c-select__menu")
+
+        b.wait_in_text(".pf-v5-c-console__manual-connection dl > div:first-child dd", "127.0.0.1")
+        b.wait_in_text(".pf-v5-c-console__manual-connection dl > div:nth-child(2) dd", "5900")
+
+        self.waitViewerDownload("vnc", my_ip)
+
+        # Login from Shell via SSH
+
```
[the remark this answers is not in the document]
mvollmer: The new code in the PR handles that differently.. maybe it doesn't have to.

Bastion host has:

 - cockpit.transport.host == "localhost"
 - application == "cockpit+=address"

where "address" is the string typed into the login page, and visible in the browser URL as "/=address". But it's not in the frame URL.

SSH from the Shell has:

 - cockpit.transport.host == "address"

where "address" is the string typed into the "add remote host" dialog.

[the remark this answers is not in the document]
mvollmer: (I don't really understand our "application" business, tbh.)

test/check-machines-consoles
mvollmer: Right, I thought I'd seen something like that... Will that work also with TF?

test/check-machines-consoles
mvollmer: > I wonder why you don't' have to m.start_cockpit() here, it's usually required?

Good question... I guess cockpit is enabled by default somewhere already?

test/check-machines-consoles
mvollmer: I have made the new test destructive and use that address.

test/check-machines-consoles
mvollmer: I have added a call to "start_cockpit".

src/components/vm/consoles/consoles.jsx
mvollmer: Done, thanks!

src/components/vm/consoles/consoles.jsx
cockpituous, line 97: These 2 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
cockpituous, line 105: These 2 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,27 @@ class Consoles extends React.Component {
...
         // fire download of the .vv file
-        domainDesktopConsole({ name: vm.name, id: vm.id, connectionName: vm.connectionName, consoleDetail: vm.displays.find(display => display.type == type) });
+        const consoleDetail = vm.displays.find(display => display.type == type);
+
+        let address;
+        if (cockpit.transport.host == "localhost") {
+            const app = cockpit.transport.application();
+            if (app.startsWith("cockpit+=")) {
+                address = app.substr(9);
+                console.log("WARNING: app", address);
```
cockpituous, line 98: These 3 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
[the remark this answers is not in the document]
mvollmer: I don't understand this... these lines clearly run.

martinpitt (approved):
TF is still running, but it seems it already skips the console tests there (the 172.27.0.15 IP is specific to our bots images), and some tests already passed.

This is a significant improvement, let's get this in without more bikeshedding. Thank you!

src/components/vm/consoles/consoles.jsx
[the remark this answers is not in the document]
martinpitt: @mvollmer is investigating this in #2006, so I suppose we can land this PR.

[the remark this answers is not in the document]
mvollmer: Yeah, coverage collection must be buggy somehow... grr.

PR #2003, opened 2025-01-27
packaging: depend on virt-install on Debian

ptoscano:
Starting from version 1:5.0.0-2, the `virt-manager` source in Debian renames the `virtinst` binary package to `virt-install`, adding the old package as transitional dummy package.

Hence, tweak the Debian packaging to try to pull the new `virt-install` first, falling back to the old `virtinst`.

There is no behaviour change on older Debian & Debian-based versions, and in newer versions it makes it possible to uninstall the transitional `virtinst`.


martinpitt (approved):
Grazie!

PR #2005, opened 2025-01-27
consoles: Also wait for address when connecting to VNC

mvollmer:
If the inactive XML config contains a port but not a address, we would try to connect too early, before the address was actually filled in.

### DIFF
--- a/src/components/vm/consoles/vnc.jsx
+++ b/src/components/vm/consoles/vnc.jsx
@@ -68,7 +68,7 @@ class Vnc extends React.Component {
         }
 
         const { consoleDetail } = props;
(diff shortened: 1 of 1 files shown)


martinpitt:
[The](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2005-56a91774-20250127-103411-rhel-10-0/log.html) [failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2005-56a91774-20250127-103409-ubuntu-stable/log.html) look like flakes, retrying.

martinpitt (approved):
Thanks!

PR #2006, opened 2025-01-28
DBG - break line that coverage says is not executed

mvollmer:


src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,25 @@ class Consoles extends React.Component {
...
         const { vm } = this.props;
         // fire download of the .vv file
-        domainDesktopConsole({ name: vm.name, id: vm.id, connectionName: vm.connectionName, consoleDetail: vm.displays.find(display => display.type == type) });
+        const consoleDetail = vm.displays.find(display => display.type == type);
+
+        let address;
+        if (cockpit.transport.host == "localhost") {
+            const app = cockpit.transport.application();
+            if (app.startsWith("cockpit+=")) {
+                address = "wrong:" + app.substr(9);
```
cockpituous, line 97: These 2 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
```diff
@@ -88,7 +88,25 @@ class Consoles extends React.Component {
...
+            } else {
+                address = window.location.hostname;
+            }
+        } else {
+            address = cockpit.transport.host;
+            const pos = address.indexOf("@");
+            if (pos >= 0) {
+                address = address.substr(pos + 1);
```
cockpituous, line 105: These 4 added lines are not executed by any test.

src/components/vm/consoles/consoles.jsx
mvollmer: The test failure shows the wrong value that is computed here. Hmm. Maybe the coverage collection goes wrong across logouts?

src/components/vm/consoles/consoles.jsx
mvollmer: This is expected, the test aborts before reaching the part that executes this code.

PR #2008, opened 2025-01-30
consoles: Redesign

mvollmer:
[COCKPIT-1225](https://issues.redhat.com/browse/COCKPIT-1225)

Demo: https://www.youtube.com/watch?v=PHHoSxs70Ho

------------------------

### Machines: Graphical and text consoles have been improved

It is now easier to switch between these two console types, and also easier to launch a remote viewer. Cockpit can now add VNC and serial consoles to machines that are missing them, and can change the port and password of the VNC servers of individual machines.

image

PR #2015, opened 2025-02-04
WIP - Console resize and scale settings

mvollmer:
- [ ] #2008 
- [ ] drilldown support in c-c-dropdown
- [ ] ...


Venefilyn:
Just checked the new demo and it looks great!

mvollmer:
I have to think hard about how to test this...

src/components/vm/consoles/vnc.jsx
```diff
@@ -68,6 +69,18 @@ const Enum = {
...
 
+export class VncState extends ConsoleState {
+    constructor() {
+        super();
+        this.sizeMode = "none";
+    }
+
+    setSizeMode(val) {
+        this.sizeMode = val;
+        this.update();
```
cockpituous, line 80: These 3 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2015-3fe0b058-20250716-111222-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.jsx
```diff
@@ -205,12 +218,29 @@ const VncEditModal = ({ vm, inactive_vnc }) => {
...
+            <SimpleSelect
+                options={
+                    [
+                        { value: "none", content: _("No scaling or resizing") },
+                        { value: "local", content: _("Local scaling") },
+                        { value: "remote", content: _("Remote resizing") },
+                    ]
+                }
+                selected={state.sizeMode}
+                onSelect={val => state.setSizeMode(val)}
```
cockpituous, line 239: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2015-3fe0b058-20250716-111222-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.jsx
```diff
@@ -68,6 +69,18 @@ const Enum = {
     KEY_DELETE: 111,
 };
 
+export class VncState extends ConsoleState {
+    constructor() {
+        super();
+        this.sizeMode = "none";
```
martinpitt, line 75: Can we type this to be a enum-like, i.e.`"none" | "local" | "remote"` ? I suppose that first requires TS-ifying the whole new console/VNC code that just landed, so not blocking this PR.

src/components/vm/consoles/vnc.jsx
```diff
@@ -400,8 +433,8 @@ export class VncActive extends React.Component {
                           onInitFailed={this.onInitFailed}
                           onSecurityFailure={this.onSecurityFailure}
                           consoleContainerId={isExpanded ? "vnc-display-container-expanded" : "vnc-display-container-minimized"}
-                          resizeSession
-                          scaleViewport
```
martinpitt, line 404: now that I've seen the video, this old code really doesn't make sense at all :sweat:

martinpitt (commented):
Thanks @mvollmer for the demo! That was excellent to understand the problem and available knobs in libvirt/VNC.

I agree that all three modes make sense depending on the circumstances, so we need to let the user decide; and also to always do "local" in the compressed case. The UI looks fine and straightforward to me; I'd like to get @Venefilyn 's sign-off on that (next Monday?), but it seems they already looked at the demo/screenshot, so I don't expect too much pushback here.

Wrt. testing, ideas:
 * Are the `resizeSession` and `scaleViewport` properties visible in the DOM? Then we could programmatically ensure that the compressed view is always scaleViewport only and that the dropdown isn't visible, and that in the expanded view all three modes can be selected and cause the expected property values in the noVNC widget. It can also test that unexpanding switches back to "local". If these properties are not visible in the DOM, the test could at least go through all three modes to verify it doesn't crash or cause console errors.
 * Our alpine VM image does not change regularly, so I think we could add a pixel test of all three modes once its boot stabilizes (the `login:` prompt). The scaled modes may need some higher `abs_tolerance`, but it should at least be enough to prevent regressions with e.g. noVNC updates.

WDYT?

src/components/vm/consoles/vnc.jsx
mvollmer: Not really... :-)

src/components/vm/consoles/vnc.jsx
mvollmer: Yes, coming up!

mvollmer:
> Wrt. testing, ideas:
> 
>  * Are the `resizeSession` and `scaleViewport` properties visible in the DOM? 

Not directly. The canvas created by NoVNC changes some attributes, but not enough to distinguish all three modes.

>  * Our alpine VM image does not change regularly, so I think we could add a pixel test of all three modes once its boot stabilizes (the `login:` prompt). The scaled modes may need some higher `abs_tolerance`, but it should at least be enough to prevent regressions with e.g. noVNC updates.

Yes, we could do this. The framebuffer console does not resize itself properly, but it does react. In any case, we can look at `/sys/class/drm/card0-Vitual-1/modes` inside the Alpine machine and see the resize changes there. Can we run code inside the Alpine machines?

We do need a `virtio` video device however, the regular `vga` one doesn't react to monitor resizes.

martinpitt:
> The canvas created by NoVNC changes some attributes, but not enough to distinguish all three modes.

Dang. Two at least? :grin: 

> Can we run code inside the Alpine machines?

Not with our current test rig. The test could presumably boot it with our standard bots `cloud-init.iso` and then ssh in, though. Note that it all runs emulated, so starting ssh and logging in will probably take a while.

> We do need a virtio video device however

I don't know about that, but hopefully that's possible with adjusting machineslib.py `createVm`?

mvollmer:
> > The canvas created by NoVNC changes some attributes, but not enough to distinguish all three modes.
> 
> Dang. Two at least? 😁

It's actually enough!  See the test that I have just pushed.

> > Can we run code inside the Alpine machines?
> 
> Not with our current test rig. The test could presumably boot it with our standard bots `cloud-init.iso` and then ssh in, though. Note that it all runs emulated, so starting ssh and logging in will probably take a while.

Initially I thought I do need this, and I got it working.  ssh as admin with a "foobar" password works out of the box and is not slow. It's just annoying to get the IP of the Alpine guest and automate ssh with passwords.

I didn't need this in the end.

> > We do need a virtio video device however
> 
> I don't know about that, but hopefully that's possible with adjusting machineslib.py `createVm`?

We get one by default except on Arch, and I of course started the whole investigation on Arch...

mvollmer:
We could do pixel tests, but the VNC console of our Alpine guests is actually empty except for some messages from Grub and a blinking cursor. I think checking what NoVNC is doing in the DOM is enough.

test/check-machines-consoles
```diff
@@ -508,6 +508,77 @@ fullscreen=0
...
+        # request a remote resize and it is rejected, these attributes
+        # will not change.  Thus, we don't actually have to log into
+        # the guest and do stuff like looking at
+        # /sys/class/drm/<foo>/modes to see how big the actual
+        # framebuffer is.
+
+        # Local scaling is done by NoVNC by setting the width and
+        # height style properties of the canvas element.  The browser
+        # will then scale the content of the canvas to make it as
+        # large as requested by its style.
```
martinpitt, line 535: That's actually rather clever! I had expected noVNC to do the local scaling, but right, I suppose browsers are actually good at that.

test/check-machines-consoles
```diff
@@ -508,6 +508,77 @@ fullscreen=0
...
+        # width become equal to the UI.
+
+        b.select_PF("#vm-console-vnc-scaling", "Remote resizing")
+        wait_widths(lambda remote, local, ui: remote == ui and local == ui)
+
+        # When collapsing and expanding again, nothing should have
+        # changed.
+
+        b.click(".consoles-card button:contains(Compress)")
+        b.wait_visible("#vm-details")
```
martinpitt, line 582: Here it could assert `b.wait_not_present("#vm-console-vnc-scaling")`

martinpitt (commented):
Thanks for figuring out the test! This now has the right kind of coverage for my taste and looks rather robust. One suggestion for improving the test a bit.  Depending on your patience we can either wait until Monday for @Venefilyn to come back and have a look, or do that post-merge with a possible follow-up.

test/check-machines-consoles
mvollmer: Yes, done.

mvollmer:
There is a layout problem with the expanded console in a narrow browser: 

https://www.youtube.com/watch?v=B58G46LhepA

We could make the stuff in the header take less space, or we could fix our (my) horrible CSS. My plan is to change the look of the expanded console anyway when we make it "fullscreen", see  #2229. We could fix the CSS problem at that time as well.

The CSS problem: The NoVNC viewer size needs to be constraint from the outside, otherwise it just grows too large. In order to do that for the expanded view, I have just eyeballed the height of the header to be 70 pixels and used that in some CSS calculations.  When the header wraps, that calculation is wrong.  Yes, yes, I know.  There is probably some grid magic that can get this done.

src/components/vm/consoles/vnc.jsx
```diff
@@ -205,12 +218,30 @@ const VncEditModal = ({ vm, inactive_vnc }) => {
...
+                options={
+                    [
+                        { value: "none", content: _("No scaling or resizing") },
+                        { value: "local", content: _("Local scaling") },
+                        { value: "remote", content: _("Remote resizing") },
+                    ]
+                }
+                selected={state.sizeMode}
+                onSelect={val => state.setSizeMode(val)}
+            />
```
Venefilyn, line 241: Naming is AFAICT directly from NoVNC naming which is fine to me given we use it. 

If we can get an aria label that makes sense I'd also add in "Scaling mode" so the select is a bit more descriptive there.

PR #2016, opened 2025-02-04
consoles: Fix constant VNC reconnections when there is a password

mvollmer:
We would pass a new object to VncConsole.credentials on every render, which would trigger it to re-connect to the VNC server every time. This shows up as flicker in the UI, and will also eventually break the connection entirely, somehow.


martinpitt (approved):
Oops! Thanks! This feels like an use case for `useRef()`, but `this` should work fine as well.

mvollmer:
> Oops! Thanks! This feels like an use case for `useRef()`, but `this` should work fine as well.

It's a use case for `useDeepEqualMemo`, but this is unfortunately still a class component, not a function component.

PR #2019, opened 2025-02-05
consoles: Allow adding VNC

mvollmer:
Screenshots: https://github.com/cockpit-project/cockpit-machines/pull/2019#issuecomment-2636986750

- [x] move VncState out of vnc.jsx and rename it to ConsoleState or similar, it's not really VNC specific.
- [x] Handle machine without serial console.
- [ ] test / coverage

### DIFF
--- a/src/components/common/needsShutdown.jsx
+++ b/src/components/common/needsShutdown.jsx
@@ -85,6 +85,20 @@ export function needsShutdownSpice(vm) {
     return vm.hasSpice !== vm.inactiveXML.hasSpice;
 }
 
+export function needsShutdownVnc(vm) {
+    function find_vnc(v) {
+        return v.displays && v.displays.find(d => d.type == "vnc");
+    }
+
+    const active_vnc = find_vnc(vm);
+    const inactive_vnc = find_vnc(vm.inactiveXML);
+
+    if (inactive_vnc && !active_vnc)
+        return true;
+
+    return false;
+}
+
 export function getDevicesRequiringShutdown(vm) {
     if (!vm.persistent)
         return [];
@@ -124,6 +138,10 @@ export function getDevicesRequiringShutdown(vm) {
     // SPICE
     if (needsShutdownSpice(vm))
         devices.push(_("SPICE"));
+
+    // VNC
+    if (needsShutdownVnc(vm))
+        devices.push(_("VNC"));
 
     // TPM
     if (needsShutdownTpm(vm))
--- a/src/components/vm/consoles/consoles.jsx
+++ b/src/components/vm/consoles/consoles.jsx
@@ -16,10 +16,12 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import cockpit from 'cockpit';
 import { AccessConsoles } from "@patternfly/react-console";
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
+import { EmptyState, EmptyStateBody, EmptyStateFooter, EmptyStateActions } from "@patternfly/react-core/dist/esm/components/EmptyState";
 
 import SerialConsole from './serialConsole.jsx';
 import Vnc from './vnc.jsx';
@@ -27,18 +29,79 @@ import DesktopConsole from './desktopConsole.jsx';
 import {
     domainCanConsole,
     domainDesktopConsole,
-    domainSerialConsoleCommand
+    domainSerialConsoleCommand,
+    domainAddVnc,
 } from '../../../libvirtApi/domain.js';
 
 import './consoles.css';
 
 const _ = cockpit.gettext;
 
-const VmNotRunning = () => {
+const ConsoleEmptyState = ({ vm, onAddErrorNotification }) => {
+    const serials = vm.displays && vm.displays.filter(display => display.type == 'pty');
+    const inactive_vnc = vm.inactiveXML.displays && vm.inactiveXML.displays.find(display => display.type == 'vnc');
+
+    const [inProgress, setInProgress] = useState(false);
+
+    function add_vnc() {
+        setInProgress(true);
+        domainAddVnc(vm)
+                .catch(ex => onAddErrorNotification({
+                    text: cockpit.format(_("Failed to add VNC to VM $0"), vm.name),
+                    detail: ex.message,
+                    resourceId: vm.id,
+                }))
+                .finally(() => setInProgress(false));
+    }
+
+    let top_message = null;
+    let bot_message = _("Graphical support not enabled.");
+
+    if (serials.length == 0 && !inactive_vnc) {
+        bot_message = _("Console support not enabled.");
+    } else {
+        if (vm.state != "running") {
+            if (serials.length > 0 && !inactive_vnc) {
+                top_message = _("Please start the virtual machine to access its serial console.");
+            } else {
+                top_message = _("Please start the virtual machine to access its console.");
+            }
+        } else if (inactive_vnc) {
+            if (serials.length > 0) {
+                top_message = _("Please shut down and restart the virtual machine to access its graphical console.");
+            } else {
+                top_message = _("Please shut down and restart the virtual machine to access its console.");
+            }
+        }
+    }
+
     return (
-        <div id="vm-not-running-message">
-            {_("Please start the virtual machine to access its console.")}
-        </div>
+        <EmptyState>
+            { top_message &&
+                <EmptyStateBody>
+                    {top_message}
+                </EmptyStateBody>
+            }
+            { !inactive_vnc &&
+                <>
+                    <EmptyStateBody>
+                        {bot_message}
+                    </EmptyStateBody>
+                    <EmptyStateFooter>
+                        <EmptyStateActions>
+                            <Button
+                                variant="secondary"
+                                onClick={add_vnc}
+                                isLoading={inProgress}
+                                disabled={inProgress}
+                            >
+                                {_("Add VNC")}
+                            </Button>
+                        </EmptyStateActions>
+                    </EmptyStateFooter>
+                </>
+            }
+        </EmptyState>
     );
 };
 
@@ -81,8 +144,9 @@ class Consoles extends React.Component {
             return 'SerialConsole';
         }
 
-        // no console defined
-        return null;
+        // no console defined, but the VncConsole is always there and
+        // will instruct people how to enable it for real.
+        return 'VncConsole';
     }
 
     onDesktopConsoleDownload (type) {
@@ -116,7 +180,11 @@ class Consoles extends React.Component {
         const vnc = vm.displays && vm.displays.find(display => display.type == 'vnc');
 
         if (!domainCanConsole || !domainCanConsole(vm.state)) {
-            return (<VmNotRunning />);
+            return (
+                <div id="vm-not-running-message">
+                    <ConsoleEmptyState vm={vm} onAddErrorNotification={onAddErrorNotification} />
+                </div>
+            );
         }
 
         const onDesktopConsole = () => { // prefer spice over vnc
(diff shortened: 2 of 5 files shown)


src/components/common/needsShutdown.jsx
```diff
@@ -85,6 +85,20 @@ export function needsShutdownSpice(vm) {
...
+export function needsShutdownVnc(vm) {
+    function find_vnc(v) {
+        return v.displays && v.displays.find(d => d.type == "vnc");
+    }
+
+    const active_vnc = find_vnc(vm);
+    const inactive_vnc = find_vnc(vm.inactiveXML);
+
+    if (inactive_vnc && !active_vnc)
+        return true;
```
cockpituous, line 97: This added line is not executed by any test.

src/components/common/needsShutdown.jsx
```diff
@@ -125,6 +139,10 @@ export function getDevicesRequiringShutdown(vm) {
     if (needsShutdownSpice(vm))
         devices.push(_("SPICE"));
 
+    // VNC
+    if (needsShutdownVnc(vm))
+        devices.push(_("VNC"));
```
cockpituous, line 144: This added line is not executed by any test.

src/components/vm/consoles/vnc.jsx
```diff
@@ -48,6 +50,55 @@ const Enum = {
...
+export const VncState = ({ vm, inactive_vnc, onAddErrorNotification }) => {
+    const [inProgress, setInProgress] = useState(false);
+
+    function add_vnc() {
+        setInProgress(true);
+        domainAddVnc(vm)
+                .catch(ex => onAddErrorNotification({
+                    text: cockpit.format(_("Failed to add VNC to VM $0"), vm.name),
+                    detail: ex.message,
+                    resourceId: vm.id,
```
cockpituous, line 62: These 7 added lines are not executed by any test.

src/components/vm/consoles/vnc.jsx
```diff
@@ -48,6 +50,55 @@ const Enum = {
...
+                }))
+                .finally(() => setInProgress(false));
```
cockpituous, line 64: This added line is not executed by any test.

src/components/vm/consoles/vnc.jsx
```diff
@@ -48,6 +50,55 @@ const Enum = {
...
+        <EmptyState>
+            { vm.state != "running" &&
+                <EmptyStateBody>
+                    {_("Please start the virtual machine to access its console.")}
+                </EmptyStateBody>
+            }
+            { vm.state == "running" && inactive_vnc &&
+                <EmptyStateBody>
+                    {_("Please shutdown and restart the virtual machine to access its graphical console.")}
+                </EmptyStateBody>
```
cockpituous, line 77: These 3 added lines are not executed by any test.

src/components/vm/consoles/vnc.jsx
```diff
@@ -137,11 +200,11 @@ class Vnc extends React.Component {
                     id={cockpit.format("ctrl-alt-$0", keyName)}
                     key={cockpit.format("ctrl-alt-$0", keyName)}
                     onClick={() => {
-                        return domainSendKey({ connectionName, id: vmId, keyCodes: [Enum.KEY_LEFTCTRL, Enum.KEY_LEFTALT, Enum[cockpit.format("KEY_$0", keyName.toUpperCase())]] })
+                        return domainSendKey({ connectionName: vm.connectionName, id: vm.id, keyCodes: [Enum.KEY_LEFTCTRL, Enum.KEY_LEFTALT, Enum[cockpit.format("KEY_$0", keyName.toUpperCase())]] })
                                 .catch(ex => onAddErrorNotification({
-                                    text: cockpit.format(_("Failed to send key Ctrl+Alt+$0 to VM $1"), keyName, vmName),
+                                    text: cockpit.format(_("Failed to send key Ctrl+Alt+$0 to VM $1"), keyName, vm.name),
```
cockpituous, line 205: This added line is not executed by any test.

src/components/vm/consoles/vnc.jsx
```diff
@@ -137,11 +200,11 @@ class Vnc extends React.Component {
...
                                     detail: ex.message,
-                                    resourceId: vmId,
+                                    resourceId: vm.id,
```
cockpituous, line 207: This added line is not executed by any test.

src/libvirtApi/domain.js
```diff
@@ -1089,3 +1089,13 @@ export async function domainAddTPM({ connectionName, vmName }) {
     const args = ["virt-xml", "-c", `qemu:///${connectionName}`, "--add-device", "--tpm", "default", vmName];
     return cockpit.spawn(args, { err: "message", superuser: connectionName === "system" ? "try" : null });
 }
+
+export function domainAddVnc(vm) {
+    const args = ['virt-xml', '-c', `qemu:///${vm.connectionName}`, vm.uuid, '--add-device', '--graphics', 'vnc'];
+    const options = { err: "message" };
```
cockpituous, line 1095: These 3 added lines are not executed by any test.

src/libvirtApi/domain.js
```diff
@@ -1089,3 +1089,13 @@ export async function domainAddTPM({ connectionName, vmName }) {
...
+
+    if (vm.connectionName === "system")
+        options.superuser = "try";
```
cockpituous, line 1098: These 2 added lines are not executed by any test.

src/libvirtApi/domain.js
```diff
@@ -1089,3 +1089,13 @@ export async function domainAddTPM({ connectionName, vmName }) {
...
+
+    return cockpit.spawn(args, options);
```
cockpituous, line 1100: This added line is not executed by any test.

mvollmer:
This is heavy in `EmptyState` components.

Shut-off machine without VNC:

image

Shut-off machine with VNC:

image

Running machine without VNC:

image

Running machine after adding VNC while it was running:

image

Hitting "Add VNC" will immediately add the VNC device to the VM without intervening dialog. The "Add VNC" button will be disabled and show a spinner while that happens (but it is usually so fast that the spinner doesn't manage to do a full round.)

When there is an error during addint the VNC device, a toast alert appears with the error message.

martinpitt:
In essence this is a downsized version of #1973. I like the approach! It can bring an existing vm without VNC to the same safe default config as a newly created VM. So it's at least required and hopefully undisputed. It's not yet clear whether it's *sufficient* for our particular customer (waiting for an answer for that), but either way I agree it makes sense to split add and edit. Porting the tests is hopefully not too much effort.

src/components/vm/consoles/consoles.jsx
```diff
@@ -16,29 +16,92 @@
...
+
+    const [inProgress, setInProgress] = useState(false);
+
+    function add_vnc() {
+        setInProgress(true);
+        domainAddVnc(vm)
+                .catch(ex => onAddErrorNotification({
+                    text: cockpit.format(_("Failed to add VNC to VM $0"), vm.name),
+                    detail: ex.message,
+                    resourceId: vm.id,
```
cockpituous, line 52: These 4 added lines are not executed by any test.

cockpituous, line 52: These 4 added lines are not executed by any test.

garrett:
Hey, thanks for splitting this out! This is a feature we should definitely land as soon as we can; it's small, it's straightforward, and it makes sense. Judging from your screenshots, it already looks good too!

I'm currently working on some mockups for some small touchups, but I think what you have is almost ready.

I'll post something soon with some minor tweaks, but I don't think it'll require much.

garrett:
Penpot design:
https://design.penpot.app/#/workspace/8ec95363-4e2c-80a4-8002-cc84eb3713ac/8cbf1f60-8e7b-11ec-a2fe-9eafb5cb1e0f?page-id=2b800199-2e0a-80de-8005-b2220bbe0adf&layout=layers

Screenshot capture of all the mockups in one board (at the time of writing, this is the same as the link in Penpot above):

cockpit-machines-add-vnc

It's mostly the same, but with a few small adjustments.

You should read this as shut off for the left three, powered on for the right two sets. Going from top to bottom, it's the initial state, and the second (and third, which is just a color variant of the second) is what happens after you add the console feature and haven't yet rebooted.

The color variations are because I thought restart was purple, not teal. In checking out locally, after making the mockups, I see it's teal, so I added that as a variant. I think teal is too close to green and could be mistaken as "success". It's not orange or red, as it's not a warning or dangerous, and it's not green because it's not a success state, and it's not grey because we're using color to emphasize that there are changes (which is what we do elsewhere). So that basically leaves us with purple and teal. But we should be consistent, so it's teal in the scope of this PR.

These mockups are idealistic. We can prune it back to make it match up with the features in this PR by:

1. Removing the link styled buttons; only having text for the status
2. Not changing the fully featured console in a shutoff state (the full console running state is untouched)
3. Ignoring the newer style switcher, and keeping the dropdown for now (consider the widget as meaning "switcher of some sort goes here")
4. Using the teal icon color (usually known in PF as "cyan", as specified elsewhere)
5. Only concern ourselves with adding VNC (for now)

All of the above list could be considered as follow-ups at some point for the future, some sooner and some later. Some might be debatable, such as the link styled buttons (which brings in context, but is also redundant with the main actions) or handling adding the text console (which might be like in the mockup, might look different, or decided against). When there is VNC and no text, we can have the button to add text console similar to the graphical one inside of a "Text" tab, but I think that's outside the scope of this PR as well. (The design would be similar, but the text would be a little different.)

garrett (changes requested):
Here's a pruned-back version which removes all the parts of that list, so you can focus on the changes for this specific PR and not have to care about anything that might be coming for the future:

cockpit-machines-add-vnc-mini

@ Penpot: https://design.penpot.app/#/workspace/8ec95363-4e2c-80a4-8002-cc84eb3713ac/8cbf1f60-8e7b-11ec-a2fe-9eafb5cb1e0f?page-id=7a18c4a0-b2c0-80e9-8005-b24d8c683fd8

I tried to make this modified version quickly near the end of the workweek. If I removed too much, please refer to the previous mockup. Otherwise, I think this is mainly what's needed for this PR.

This should be a really small set of changes on top of what you already have here. What you have here is already great! Thanks for splitting this out and opening up this new PR!

mvollmer:
> I tried to make this modified version quickly near the end of the workweek. If I removed too much, please refer to the previous mockup. Otherwise, I think this is mainly what's needed for this PR.

Thanks! It's excellent.

I had considered putting a "shutdown" action into the console card, but rejected that idea because we currently do not have a "start" action either, and I feel that "shutdown" is a much too dangerous action to offer as the "obvious next step, don't even think about it" to the user. But there would be a confirmation dialog, and the expanded version of the console will likely not have the VM actions menu. So I agree that we should offer "start" and "restart" right in the card.

The action in the card will also avoid the issue that a mere "reboot" is not enough, we need a "shutdown" followed by "run". We don't have a dialog for that yet, unfortunately. I can make a variant of the "shutdown" dialog with a simple `s/Shut down/Shut down and restart/`. We should probably also remove the "Reboot" secondary action in it, since it will be the wrong thing to do in this case.

I would like to leave the toggle buttons, adding of a serial console, and changing the color for later.

Note that the current Console card never has a dropdown selector when the machine is shut off. Do you want it to have one even in the pared down version of your design? This question is also not yet decided for the big redesign in #2008, I think.

If yes, how would the card look when the machine is shut off and the serial console is selected? Also, would the dropdown selector contain the "Desktop viewer" tab? How would that look? (The Desktop viewer tab will go away, but we are not there yet in this PR.)

martinpitt:
Thanks for the design! Changing the console switcher has nothing to do with adding a default VNC console. My recommendation is to land this PR more or less as-is, to at least get some progress on this matter (it is a highly useful and self-contained improvement), and do the other redesigns in a follow-up. If @mvollmer and @garrett want to do it all in one go, fine for me -- but at some point I have the feeling that both of you need some intermediate success to get back some motiviation :sweat_smile:

mvollmer:
Continued in #2008

Issue #240, opened 2021-07-03
Cockpit VM set up help - Network issue


TheInfamousToTo:
![2021-07-03_16h41_40](https://user-images.githubusercontent.com/72304665/124356230-93658b80-dc1d-11eb-8166-af09c5866338.png)
Cockpit version: 239-1~bpo10+1
OS: Debian GNU/Linux 10 (buster)
Page: <services>

Wait for Network to be Configured

I'm new to cockpit and I was trying to install VM and I start getting this error "Wait for Network to be Configured"  and whenever I'm trying to boot the VM I'm getting this error "Failed to connect socket to '/var/run/libvirt/virtlogd-sock': Connection refused" 

screen shots 
![2021-07-03_16h40_13](https://user-images.githubusercontent.com/72304665/124356207-79c44400-dc1d-11eb-890f-3d51a8063c19.png)
![2021-07-03_16h41_08](https://user-images.githubusercontent.com/72304665/124356209-7b8e0780-dc1d-11eb-81bb-21df2c58081f.png)

skobyda:
I don't see how `systemd-networkd-wait-online.service` relates to a VM not being able to start. 
Does the service fail when you try to start VM? Or did you configure that service to be wantedBy a libvirtd.service  or virtlogd?

Can also check the state of virtlogd.service on your system? Run `systemctl status virtlogd`

KKoukiou:
I can recall a SElinux issue with the same logs (virtlogd-socket Connection refused). Check `systemctl status virtlogd' and the journal for related errors and please paste here for further investigation.

TheInfamousToTo:
Hi @skobyda , 
im not sure if they related or not but i got them on the same time and since this 2 errors appears im unable to run the VMs
Note: this is my first time to use cockpit so im very confused

TheInfamousToTo:
Hi @KKoukiou 
please find the result below 

root@NAS:~# systemctl status virtlogd
● virtlogd.service - Virtual machine log manager
   Loaded: loaded (/lib/systemd/system/virtlogd.service; indirect; vendor preset: enabled)
   Active: active (running) since Mon 2021-07-05 22:16:04 +03; 29s ago
     Docs: man:virtlogd(8)
           https://libvirt.org
 Main PID: 30733 (virtlogd)
    Tasks: 1 (limit: 4915)
   Memory: 2.2M
   CGroup: /system.slice/virtlogd.service
           └─30733 /usr/sbin/virtlogd

[...]

PR #2021, opened 2025-02-06
packit: Drop copr_build job redundancy, re-disable Fedora 42

martinpitt:
Similar to https://github.com/cockpit-project/cockpit-podman/pull/1993


martinpitt:
Too bad, our split build apparently doesn't work with the implied COPR targets. However, the auto-generated centos dist-git PRs are broken anyway https://github.com/packit/packit-service/issues/2436, we can revert #240 for now.

jelly (approved):

PR #2023, opened 2025-02-07
Revert "test: Skip TestMachinesDisks.testAddDiskNFS on RHEL 10"

martinpitt:
https://issues.redhat.com/browse/RHEL-67841 got silently fixed.

This reverts commit 423d16a23f12d4dd41b35951bbfcd958f9e6cbbf.

----

I was revisiting that bug while reporting https://bugzilla.redhat.com/show_bug.cgi?id=2344326 . I cannot reproduce the crash manually on current RHEL 10 any more, so let's give this a try in CI.


martinpitt:
OK, this is still very much alive. I updated the RHEL issue.

PR #2028, opened 2025-02-10
test: Robustify file download checks

mvollmer:
- We only want to look at fully downloaded files, but in some situations the tests would be fast enough to catch Firefox's *.part file before it gets renamed to the real name. Ignore those.

- Bring back the wait for the file be of the expected size. This seems to be necessary on Chromium in order to not look at partial files.

- Remove the downloaded file so that the next check will not pick it up by accident. The one test with that calls waitViewerDownload twice (testExternalConsoleMultiHost) does it with exactly the same arguments and will pass if the second call looks at the file from the first call.


martinpitt (approved):
Thank you!

PR #2029, opened 2025-02-10
test: Bring back TestMultiMachineVNC tests

mvollmer:
I have deleted them in 8a11a1896ebe61a7cecf39ed8ed9026d779aeb45 because I didn't understand them and thought they are now redundant.

But they are still useful: They test that the channel for the in-page VNC viewer uses the correct "host" parameter.

(Specifically, I thought the check at the end is looking at open listening ports, not established connections.)


martinpitt (approved):
Thanks! Assuming green.

PR #2030, opened 2025-02-10
consoles: Pass onSecurityFailure handler to VncConsole

mvollmer:
It is not optional and VncConsole will crash otherwise when there is a authentication failure.

Authentication failures should not happen since we pass the correct credentials to VncConsole. However, manually setting a empty password in the machine XML can trigger a authentication failure, and we should be prepared for everything, of course.


martinpitt (approved):
Cheers!

PR #2038, opened 2025-02-19
console: Better connection failure error messages

mvollmer:
- [ ] #2008

When a second remote viewer starts, our in-page viewer will be cleanly
disconnected.
    
But when we fail to establish a connection because of protocol
incompatabilities or authentication failures, we should make that
clear in the UI by showing the error and saying "Retry" instead of
"Connect".
    
Unfortunately, there is not enough information coming from NoVNC that
would allow us to detect that the connection failed because the VNC
server wants encryption. But we can guess by looking into qemu.conf.


src/components/vm/consoles/VncConsole.jsx
```diff
@@ -55,7 +55,7 @@ export const VncConsole = ({
     repeaterID = '',
     vncLogging = 'warn',
     consoleContainerId,
-    onDisconnected = () => {},
+    onDisconnected = (event) => {},
```
cockpituous, line 58: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-7e41038d-20250806-115005-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.jsx
```diff
@@ -73,12 +75,31 @@ export class VncState extends ConsoleState {
...
+    }
+
+    setDisconnected(clean) {
+        this.connected = false;
+        this.clean_disconnect = clean;
+        this.update();
+    }
+
+    setConnectionFailure(val) {
+        this.connection_failure = val;
```
cockpituous, line 101: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-7e41038d-20250806-115005-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.jsx
```diff
@@ -377,17 +398,18 @@ export class VncActive extends React.Component {
...
     }
 
     onInitFailed(detail) {
         console.error('VncConsole failed to init: ', detail, this);
     }
 
     onSecurityFailure(event) {
-        console.info('Security failure:', event?.detail?.reason || "unknown reason");
+        if (event.detail?.reason)
+            this.props.state.setConnectionFailure(event.detail.reason);
```
cockpituous, line 412: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-7e41038d-20250806-115005-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.jsx
```diff
@@ -419,6 +450,16 @@ export class VncActive extends React.Component {
             />
         );
 
+        const vnc_tls = qemu_conf.vnc_tls == "1";
+
+        function get_failure_reason() {
+            if (state.connection_failure)
+                return state.connection_failure;
```
cockpituous, line 457: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-7e41038d-20250806-115005-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.jsx
```diff
@@ -419,6 +450,16 @@ export class VncActive extends React.Component {
...
+            if (vnc_tls)
+                return _("VNC with TLS is not supported by the in-page viewer");
```
cockpituous, line 459: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-7e41038d-20250806-115005-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
Oh, nice coverage.

src/app.jsx
```diff
@@ -60,6 +62,26 @@ async function unknownConnectionName() {
         return ["system", "session"];
 }
 
+function parse_qemu_conf(text) {
+    const res = { };
+    for (const l of text.split('\n')) {
```
martinpitt: useFile() uses cockpit.file.watch() which I believe will return `null` for nonexistent files. That would cause a crash here. I suspect it needs to intercept `!text` here and return an empty `res`.

src/app.jsx
```diff
@@ -60,6 +62,26 @@ async function unknownConnectionName() {
...
+        const lt = l.trim();
+        if (lt.startsWith('#'))
+            continue;
```
martinpitt: Are in-line comments a thing in that file? `key = value # comment`?

src/app.jsx
```diff
@@ -60,6 +62,26 @@ async function unknownConnectionName() {
...
+        const pos = lt.indexOf('=');
+        if (pos > 0) {
```
martinpitt: AFAIUI there should only really be `key = val` or comments or empty lines. Suggest a console.warn() for unparseable lines to aid debugging, also for admins?

src/app.jsx
```diff
@@ -60,6 +62,26 @@ async function unknownConnectionName() {
...
+            const name = lt.substring(0, pos).trim();
+            let val = lt.substring(pos + 1).trim();
+            if (val[0] == '"' && val[val.length - 1] == '"')
+                val = val.substring(1, val.length - 1);
+            res[name] = val;
```
martinpitt: All of this, together with possibly in-line comments, really cry for a regexp match :wink: If you prefer this, it's fine.

Too bad we don't already have an existing qunit setup for machines, this subtle logic would be perfect for it.

src/components/vm/consoles/vnc.jsx
```diff
@@ -73,12 +75,31 @@ export class VncState extends ConsoleState {
     constructor() {
         super();
         this.sizeMode = "none";
+        this.clean_disconnect = true;
+        this.connection_failure = null;
```
martinpitt: Is this really orthogonal? It doesn't feel like it, but rather that `connection_failure == null` means either "connected" or "cleanly disconnected". And given that `setConnected()` unconditionally sets `clean_disconnect = true` even for `val == true`, this doesn't feel like an useful state split. Perhaps you want some kind of tristate here? `connected: bool | Error`?

src/app.jsx
martinpitt: The file clearly has `= "string"` and `= number` types. This parsing turns everything into strings. Need to check the subsequent commits how that matters.

src/components/vm/consoles/vnc.jsx
```diff
@@ -419,6 +443,16 @@ export class VncActive extends React.Component {
             />
         );
 
+        const vnc_tls = qemu_conf.vnc_tls == "1";
```
martinpitt: In qemu.conf this is a number; not sure if it can be *exactly* only 0 | 1, but perhaps it'd be a bit more robust to parse this as a boolean?

src/components/vm/consoles/vnc.jsx
```diff
@@ -419,6 +443,16 @@ export class VncActive extends React.Component {
...
+
+        function get_failure_reason() {
+            if (state.connection_failure)
+                return state.connection_failure;
+            if (vnc_tls)
+                return _("VNC with TLS is not supported by the in-page viewer");
```
martinpitt: Shouldn't that check come first? I.e. trying to connect with plain TCP to a TLS socket, doesn't that already cause some kind of "protocol error"?

test/check-machines-consoles
```diff
@@ -589,6 +589,98 @@ fullscreen=0
...
+        self.waitPageInit()
+        self.waitVmRow(name)
+        self.goToVmPage(name)
+
+        # VNC viewer should be connected
+        b.wait_visible(".vm-console-vnc canvas")
+
+        # Start external non-shared viewer
+        vnc_client_cmd = "python3 /tmp/headless-vnc-client.py 127.0.0.1 5900"
+        vnc_process = m.spawn(vnc_client_cmd, "vnc-disconnect-test")
```
martinpitt: So you can only ever have (at most) one connected VNC client to a server? But how come that the second connection will disconnect the first one and win, instead of failing with "busy, already connected"? Is that really how VNC works? That could use a comment.

test/check-machines-consoles
```diff
@@ -589,6 +589,98 @@ fullscreen=0
...
+        # Our viewer should disconnect cleanly
+        b.wait_in_text(".consoles-card", "Disconnected")
+
+        # Trying to reconnect should fail while the external viewer is running
+        b.click('.consoles-card button:contains("Connect")')
+        b.wait_in_text(".consoles-card", "Failed to connect")
+
+        # Stopping the external viewer should allow us to connect again
+        m.execute(f"kill {vnc_process}")
+        b.click('.consoles-card button:contains("Retry")')
```
martinpitt: That feels like a race -- the `.execute()` should at least have a `while pgrep; sleep` loop?

test/check-machines-consoles
```diff
@@ -589,6 +589,98 @@ fullscreen=0
...
+        # Change global password, but without restarting the VM. This
+        # means that Cockpit will use the new password when
+        # connecting, but the VM still expects the old.
+
+        m.execute("sed -i -e 's/vnc_password = \"foobar\"/vnc_password = \"barfoo\"/' /etc/libvirt/qemu.conf")
+
+        # Changing the password in qemu.conf immediatly triggers a
+        # reconnection of the VNC viewer, which fails as expected.
+
+        b.wait_in_text(".consoles-card", "Authentication failed")
```
martinpitt: That behaviour feels wrong to me. If the VM's VNC server doesn't watch the file for password changes, then neither should we IMO.

test/check-machines-consoles
```diff
@@ -589,6 +589,98 @@ fullscreen=0
...
+
+    def testVNCTLS(self):
+        b = self.browser
+        m = self.machine
+
+        # Install some random certs
+        m.upload(["files/pki/libvirt"], "/etc/pki/")
+        m.execute("chmod a+r /etc/pki/libvirt/*")
```
martinpitt: Hm, libvirt accepts world readable secret keys? ssh doesn't for example. Could the permissions be more realistic, to check an actual use case?

test/files/pki/libvirt/server-key.pem
```diff
@@ -0,0 +1,182 @@
+Public Key Info:
```
martinpitt: Whenever I did something like this, it triggered all kinds of security scanners "wah wah you publicly committed a secret key", in GitHub. Red Hat has something similar. Didn't you?

If you want to use a static key as opposed to creating them on the fly, please add a `.gitleaks.toml` similar to [Cockpit's](https://github.com/cockpit-project/cockpit/blob/main/.gitleaks.toml).

test/files/headless-vnc-client.py
```diff
@@ -0,0 +1,195 @@
...
+        try:
+            self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            self.socket.settimeout(self.timeout)
+            self.socket.connect((self.host, self.port))
+
+            # VNC protocol handshake
+            # 1. Server sends protocol version
+            protocol_version = self.socket.recv(12)
+            if not protocol_version.startswith(b'RFB '):
+                raise Exception(f"Invalid VNC protocol: {protocol_version}")
```
martinpitt: Please use more specific exceptions. `ValueError`?

test/files/headless-vnc-client.py
```diff
@@ -0,0 +1,195 @@
...
+            # 2. Send protocol version back
+            self.socket.send(protocol_version)
+
+            # 3. Receive security types
+            security_types_count = struct.unpack('!B', self.socket.recv(1))[0]
+            if security_types_count == 0:
+                # Server error
+                reason_length = struct.unpack('!I', self.socket.recv(4))[0]
+                reason = self.socket.recv(reason_length)
+                raise Exception(f"Server error: {reason.decode()}")
```
martinpitt: `OSError`?

test/files/headless-vnc-client.py
```diff
@@ -0,0 +1,195 @@
...
+                result_data = self.socket.recv(4)
+                if len(result_data) == 4:
+                    security_result = struct.unpack('!I', result_data)[0]
+                    if security_result != 0:
+                        # Try to read error message
+                        try:
+                            reason_length = struct.unpack('!I', self.socket.recv(4))[0]
+                            reason = self.socket.recv(reason_length)
+                            raise Exception(f"Security handshake failed: {reason.decode()}")
+                        except Exception:
```
martinpitt: This doesn't make sense. Also, `except Exception` is a no-no.

test/files/headless-vnc-client.py
```diff
@@ -0,0 +1,195 @@
...
+            framebuffer_height = struct.unpack('!H', self.socket.recv(2))[0]
+            self.socket.recv(16)  # pixel_format - Skip pixel format details
+            name_length = struct.unpack('!I', self.socket.recv(4))[0]
+            name = self.socket.recv(name_length).decode()
+
+            print(f"Connected to VNC server: {name} ({framebuffer_width}x{framebuffer_height})")
+            self.connected = True
+            return True
+
+        except Exception as e:
```
martinpitt: Same here. This hides `AttributeError`s and other clearly stupid stuff. For a test script we don't need opportunistic exception handling at all, it's better to just fail with a clear backtrace.

(FTR: I see this with claude/Copilot all the time. They are way too happy to catch `Exception`, which is doubly wrong in such cases).

test/files/headless-vnc-client.py
```diff
@@ -0,0 +1,195 @@
...
+            print(f"VNC connection failed: {e}")
+            if self.socket:
+                try:
+                    self.socket.close()
+                except Exception:
+                    pass
```
martinpitt: Here as well. This try/except should just be dropped, same for the others below.

test/check-machines-consoles
```diff
@@ -589,6 +589,98 @@ fullscreen=0
         b.wait_text("#vm-console-vnc-scaling", "Remote resizing")
         wait_widths(lambda remote, local, ui: remote == ui and local == ui)
 
+    def testNonSharedExternal(self):
+        b = self.browser
+        m = self.machine
+
+        # Upload the headless VNC client script to the test machine
+        m.upload(["files/headless-vnc-client.py"], "/tmp/headless-vnc-client.py")
```
martinpitt: This feeels like utter overkill. We already have a VNC client -- it's called "cockpit-machines". Can't this open a second browser window with an independent cockpit session (that's in fact `Browser`'s default mode of operation) to simulate concurrent VNC sessions?

test/check-machines-consoles
mvollmer: VNC has a "shared" flag in its initial messages. If it is set to false, the server will disconnect all other clients. If it is true, it will allow concurrent connections.

test/check-machines-consoles
mvollmer: Yeah, I think I agree...

test/check-machines-consoles
mvollmer: Cockpit always makes shared connections, we need a non-shared one.

martinpitt: Ah, that wasn't clear, sorry. I did a quick search, and at first sight noVNC doesn't support non-shared connections.

src/app.jsx
mvollmer: It also has `[ ... ]` lists of strings that can stretch over multiple lines.  It's its own little language: https://gitlab.com/libvirt/libvirt/-/blob/master/src/util/virconf.c

mvollmer: This also misses backslash-escapes in strings etc.

So, proposal: We ditch reading qemu.conf completely. Nobody has asked us to read passwords out of it, and we can just drop detecting TLS, which nobody has asked for either.

martinpitt: Counter-proposal: Set up two regexps which parse out *exactly* those two settings. We know the password is a single string, and TLS is a flag. This is much easier to handle than a generic parser, where we ignore most of its output anyway.

src/components/vm/consoles/VncConsole.jsx
```diff
@@ -56,7 +56,7 @@ export const VncConsole = ({
     vncLogging = 'warn',
     consoleContainerId,
     onConnected = (element) => {},
-    onDisconnected = () => {},
+    onDisconnected = (event) => {},
```
cockpituous, line 59: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-86db6651-20250808-063524-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/app.jsx
mvollmer: Yes, good catch!

src/app.jsx
mvollmer: Ok, deal! Using the regexps doesn't make it necessarily more correct, but it makes it clear that we don't even try to parse the file correctly.

src/components/vm/consoles/vnc.jsx
mvollmer: The complication here is that `connected` comes from the base class, and we are overriding the `setConnect` method here. `setConnected(false)` is always a clean disconnect without any failure reason, and `setConnected(true)` should reset the clean/failure so that we don't keep old values by accident when the next disconnect happens.

The failure reason is received asynchronously from the indication whether this was a clean disconnect (because NoVNC), so I kept them separate.

src/components/vm/consoles/vnc.jsx
mvollmer: Hmm, no. If NoVNC has given us a concrete failure  reason, we should show that.  Hopefully it already explains the TLS issues.

But unfortunately, NoVNC is silent in the TLS case, and then we take a guess.

mvollmer: I will see whether I can get more failure reasons out of NoVNC. I am looking at it more closely anyway now for TypeScripting.

test/check-machines-consoles
mvollmer: It should, actually. We can make a "test-only" secret setting that switches our viewer into non-shared mode. Or maybe we make the detached console always non-shared?

mvollmer: I don't think we should open the can of UI worms and expose shared/non-shared to the user.  But "Remote resizing" kinda yells "non-shared", no?

test/check-machines-consoles
mvollmer: I'll retool this and make a `read_qemu_conf` function that we use together with the "credentialsrequired" and "disconnected" events of NoVNC.

test/files/pki/libvirt/server-key.pem
mvollmer: > If you want to use a static key as opposed to creating them on the fly,

Creating one would be nicer, but I'll have to figure out how to do that. I, too, was suprised that just copying my own ones here worked...

test/files/headless-vnc-client.py
mvollmer: I am not sure I want to make this code any better than it is, tbh. I'd rather improve our own console to make non-shared connections.

src/app.jsx
mvollmer: Actually, the `null` will not be passed through the syntax parser, so the code accessing the result needs to handle nulls.

martinpitt: Ah, so cockpit.js is that clever, nice.

test/check-machines-consoles
martinpitt: That is truly evil.. Trying to connect non-shared to a server with already existing connections should just fail. Well, software! :man_shrugging:

test/check-machines-consoles
martinpitt: Right, see #2268 . That makes headless-vnc-client.py obsolete (we can add a test-only/secret way to make it non-shared, too), so you can ignore all the threads for that.

src/components/vm/consoles/VncConsole.jsx
```diff
@@ -51,14 +51,14 @@ export const VncConsole = ({
     scaleViewport = false,
     viewOnly = false,
     shared = false,
-    credentials,
     repeaterID = '',
     vncLogging = 'warn',
     consoleContainerId,
     onConnected = (element) => {},
-    onDisconnected = () => {},
+    onDisconnected = (event) => {},
```
cockpituous, line 58: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-5fe3bc4a-20250808-144213-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.jsx
```diff
@@ -404,17 +428,29 @@ export class VncActive extends React.Component {
...
+    async onDisconnected(event) { // server disconnected
+        console.info('Connection lost: ', event.detail);
+        let reason = this.props.state.failure_reason;
+        if (event.detail.clean)
+            reason = null;
+        else if (!reason) {
+            // This might be TLS.
+            const qemu_conf = await readQemuConf();
+            if (qemu_conf.vnc_tls)
+                reason = _("VNC with TLS is not supported by the in-page viewer");
```
cockpituous, line 440: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-5fe3bc4a-20250808-144213-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/helpers.ts
```diff
@@ -995,3 +995,32 @@ export function vmSupportsExternalSnapshots(config: { capabilities?: HypervisorC
...
+
+export async function readQemuConf(): Promise<QemuConf> {
+    const conf: QemuConf = {
+        vnc_tls: false,
+        vnc_password: null,
+    };
+
+    const text = await cockpit.file("/etc/libvirt/qemu.conf", { superuser: "try" }).read();
+    if (!text)
+        return conf;
```
cockpituous, line 1015: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-5fe3bc4a-20250808-144213-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
```diff
@@ -481,6 +511,8 @@
 
     onSecurityFailure(reason: string | undefined) {
         console.info('Security failure:', reason || "unknown reason");
+        if (reason)
+            this.props.state.setDisconnected(reason || _("Security failure"));
```
cockpituous, line 515: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-4e108bfc-20250915-111345-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
```diff
@@ -470,9 +489,20 @@
...
         console.info('Connection lost: ', clean ? "clean" : "unclean");
-        this.props.state.setConnected(false);
+        let reason = this.props.state.failure_reason;
+        if (clean)
+            reason = null;
+        else if (!reason) {
+            // This might be TLS.
+            const qemu_conf = await readQemuConf();
+            if (qemu_conf.vnc_tls)
+                reason = _("VNC with TLS is not supported by the in-page viewer");
```
cockpituous, line 501: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-4e108bfc-20250915-111630-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
cockpituous, line 515: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-4e108bfc-20250915-111630-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
cockpituous, line 501: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-4e108bfc-20250915-113610-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
cockpituous, line 515: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-4e108bfc-20250915-113610-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
cockpituous, line 501: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-1373d7e2-20250916-070332-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
cockpituous, line 515: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-1373d7e2-20250916-070332-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
Seems to break testGlobalVNCPassword?

src/components/vm/consoles/vnc.tsx
```diff
@@ -470,9 +489,20 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
...
```
cockpituous, line 501: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-5a05fe26-20250916-083706-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/consoles/vnc.tsx
```diff
@@ -481,6 +511,8 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
 
     onSecurityFailure(reason: string | undefined) {
         console.info('Security failure:', reason || "unknown reason");
+        if (reason)
+            this.props.state.setDisconnected(reason);
```
cockpituous, line 515: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-5a05fe26-20250916-083706-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
There is something weird going on when connecting with a wrong password.  NoVNC often reports error code 1006, which is "Abnormal Closure" for a web socket. Cockpit-ws never sends this code, afaics.

We expect a "clean" authentication failure reported by the VNC protocol. This happens after a few retries. With "ws-container", all we get is 1006s and never any authentication failure.

This might be a bug/race condition in NoVNC, maybe.

src/components/vm/consoles/vnc.tsx
```diff
@@ -56,12 +56,31 @@ const _ = cockpit.gettext;
...
 
 export class VncState extends ConsoleState {
-    sizeMode: VncSizeMode = "none";
+    sizeMode: VncSizeMode;
+    failure_reason: string | null;
+
+    constructor() {
+        super();
+        this.sizeMode = "none";
+        this.failure_reason = null;
```
martinpitt, line 65: shouldn't this also initialize `this.connected`, to have complete state exposed here, and not falling into some `undefined` trap? (it'll probably work fine, but..)

Also, FTR: I've made my peace with initializing at the class level, so if you prefer that, feel free to go back to that.

src/components/vm/consoles/vnc.tsx
```diff
@@ -470,9 +489,21 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
...
+        let reason = this.props.state.failure_reason;
+        if (clean)
+            reason = null;
+        else if (!reason) {
+            // This might be TLS.
+            const qemu_conf = await readQemuConf();
+            if (qemu_conf.vnc_tls) {
+                reason = _("VNC with TLS is not supported by the in-page viewer");
+            } else {
+                reason = _("Failed to connect");
```
martinpitt, line 507: `'s kaputt!! 🤷‍♂️`

src/components/vm/consoles/vnc.tsx
```diff
@@ -481,6 +512,8 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
 
     onSecurityFailure(reason: string | undefined) {
         console.info('Security failure:', reason || "unknown reason");
+        if (reason)
+            this.props.state.setDisconnected(reason);
```
martinpitt: This condition looks a bit dubious: If we get here, isn't that always a connection failure, even if we don't have a reason? I.e. should this unconditionally call setDisconnected()?

test/check-machines-consoles
```diff
@@ -723,6 +723,7 @@ fullscreen=0
         # cleanup expects self.browser to be logged in.
         self.login_and_go("/machines")
 
+    @testlib.skipWsContainer("Something broken with WebSockets....")
```
martinpitt: That also stops the already existing and working test. Please move that into an `if m.ws_container:` condition for the part that doesn't work. FTR, I don't see an immediate reason why it can't work, websocket is all between browser and cockpit-ws. That it runs in a container and is connected to a bridge over ssh shouldn't matter.

test/check-machines-consoles
```diff
@@ -744,6 +745,102 @@ fullscreen=0
         b.wait_visible(".vm-console-vnc canvas")
         self.assertIn("password=on", m.execute(f"grep ^-vnc /var/log/libvirt/qemu/{name}.log"))
 
+        # Change global password, but without restarting the VM. This
+        # means that Cockpit will use the new password when
+        # connecting, but the VM still expects the old.
+        m.execute("sed -i -e 's/vnc_password = \"barfoo\"/vnc_password = \"foobar\"/' /etc/libvirt/qemu.conf")
```
martinpitt: :warning: destructive!

`self.sed_file`?

test/check-machines-consoles
```diff
@@ -744,6 +745,102 @@ fullscreen=0
...
+        b.wait_in_text(".consoles-card", "Disconnected")
+
+        # Quite often the reconnection here fails with code 1006
+        # (Abnormal Closure), but retrying will give us the expected
+        # error message, usually on the second try.
+
+        b.click(".consoles-card button:contains(Connect)")
+        for _ in range(5):
+            try:
+                b.wait_in_text(".consoles-card", "Authentication failed")
```
martinpitt: This will take 5 * 30 s = 2.5 minutes at worst. Perhaps reduce the timeout of the wait_in_text? (`with b.wait_timeout(10):` or so)

test/check-machines-consoles
```diff
@@ -744,6 +745,102 @@ fullscreen=0
...
+                if "Failed to connect" in b.text(".consoles-card"):
+                    print("Unexpected error message. Retrying...")
+                    b.click(".consoles-card button:contains(Retry)")
+                else:
+                    raise
+
+        b.wait_in_text(".consoles-card", "Authentication failed")
+
+    @testlib.skipImage("cockpit-ws too old",
+                       "rhel-9-*", "centos-9-*", "ubuntu-*", "debian-trixie", "opensuse-tumbleweed")
```
martinpitt, line 783: Is that the `websocket-channel-close-fix` thing?

test/check-machines-consoles
```diff
@@ -744,6 +745,102 @@ fullscreen=0
...
+
+        # VNC viewer should be connected
+        b.wait_visible(".vm-console-vnc canvas")
+
+        # Start external non-shared viewer.  We use Cockpit itself, in
+        # "Remote resizing" mode.
+        b2 = self.new_browser()
+        cookie = b.cookie("cockpit")
+        b2.open(f"/cockpit/@localhost/machines/index.html#/vm/vnc?name={name}&connection=system", cookie=cookie)
+        with b2.wait_timeout(60):
```
martinpitt, line 810: This is a bit strange -- perhaps wait for the page initialization first (with a longer timeout), and then "normally" do the dropdown thing? (Not a veto, just thinking aloud)

test/check-machines-consoles
```diff
@@ -744,6 +745,102 @@ fullscreen=0
...
+            b2.select_PF("#vm-console-vnc-scaling", "Remote resizing")
+
+        # Our viewer should disconnect cleanly
+        b.wait_in_text(".consoles-card", "Disconnected")
+
+        # Trying to reconnect should fail while the external viewer is running
+        b.click('.consoles-card button:contains("Connect")')
+        b.wait_in_text(".consoles-card", "Failed to connect")
```
martinpitt, line 825: So noVNC doesn't give us something useful here, like "already connected" or "busy" or so? Too bad..

test/check-machines-consoles
```diff
@@ -744,6 +745,102 @@ fullscreen=0
...
+
+        # Stopping the external viewer should allow us to connect again
+        b2.go("#/")
+        b.click('.consoles-card button:contains("Retry")')
```
martinpitt: That looks like a race condition? I.e. `b2` needs to wait a bit until the disconnect actually happened. Is that visible with some `virsh` polling loop or similar? i.e. checking the server side for connected clients?

martinpitt (changes requested):
Thanks! Some loud thinking, but also a few blockers.

test/check-machines-consoles
martinpitt: Nevermind, we already restore /etc/libvirt/ wholesale.

src/components/vm/consoles/vnc.tsx
mvollmer: `this.connected` is inherited and initialized by the call to `super()` here.

src/components/vm/consoles/vnc.tsx
mvollmer: Yeah, but still better than just "Disconnected". :-)

src/components/vm/consoles/vnc.tsx
mvollmer: Good observation!

It happens to be so that we _also_ get a `onDisconnected` event, and I made it so that `onSecurityFailure` and `onDisconnected` can be called in any order. But a comment is certainly in order.

test/check-machines-consoles
mvollmer: > Please move that into an if m.ws_container: condition for the part that doesn't work. 

Yes, good point.

> FTR, I don't see an immediate reason why it can't work

Me neither, this will need some deepish debugging. It might be that we loose the last few bytes to a buffer somewhere when the VNC server closes the connection, or something like that.

test/check-machines-consoles
mvollmer: My thinking was that this modifies a file that has been installed via `self.write_file` earlier, and will already be restored. Is that not safe?

test/check-machines-consoles
mvollmer: Hmm, isn't the default timeout already 10s?

test/check-machines-consoles
mvollmer: yes.

src/components/vm/consoles/vnc.tsx
martinpitt: Ah!

src/components/vm/consoles/vnc.tsx
martinpitt: (this was just a silly joke)

test/check-machines-consoles
mvollmer: tbh, I copy/pasted the big timeout without thinking. I'll check this out.

test/check-machines-consoles
mvollmer: Correct.

test/check-machines-consoles
martinpitt: Right, let's not block the PR on that, a `FIXME:` comment is fine. Let's just not disable the existing checks.

test/check-machines-consoles
martinpitt: oh right, it's 15. I hope in most cases that will already work on the second iteration? then you can leave it as-is.

test/check-machines-consoles
mvollmer: Yes, good point.

src/components/vm/consoles/vnc.tsx
mvollmer: Ok, I made this unconditional.

test/check-machines-consoles
mvollmer: Yes, second or third.

test/check-machines-consoles
mvollmer: Right, in other places 60 seconds account for the time it takes the guest to properly set up the size of its VNC framebuffer. This happens sometime during boot.  Here we have to wait for the page to initialize itself, which is unfortunately slow.

We have `self.waitPageInit` for that, but that only works for `self.browser`.

I left this but added a comment with a link to https://issues.redhat.com/browse/COCKPIT-1302

test/check-machines-consoles
martinpitt: Ack, good 'nuff.

test/check-machines-consoles
mvollmer: I added a "VNC connection counter" and check the expected number at various points.

src/components/vm/consoles/vnc.tsx
```diff
@@ -481,6 +516,7 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
 
     onSecurityFailure(reason: string | undefined) {
         console.info('Security failure:', reason || "unknown reason");
+        this.props.state.setDisconnected(reason || _("Security failure"));
```
cockpituous, line 519: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2038-e5e6d0bb-20250917-080113-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
> Thanks for bearing with me -- happy botty, happy pitti!

Oh, no trouble at all! This is all excellent stuff!

PR #2048, opened 2025-03-03
helpers: Include filename when triggering download

### LINKED ISSUE #1952: No .vv file downloaded
### Explain what happens

1. Open cockpit
2. Go to Virtual Machines tab
3. Start a VM
4. Clic on Launch remote viewer

The file has no extension.
On Google Chrome : "download"
On Firefox : "FHfpQ1UH"
On Vivaldi : "file"

With No extension, open the file directly in the browser (to not download it) didn't work because the association is with extension.

### Version of Cockpit

cockpit-machines 325

### Where is the problem in Cockpit?

Unknown or not applicable

### Server operating system

Fedora

### Server operating system version

41

### What browsers are you using?

Firefox, Chrome, Other

### System log

_No response_

mvollmer:
By combining the two old ideas: Instead of deciding between clicking on a link (needs CSP), or loading a "data:" URL into a nested iframe (omits filename), we click on a link in a nested iframe.

This should make the "Launch viewer" button work with Chrome and in other contexts where the extension of the filename is needed to find the right external application.

Fixes #1952

-----------------

### Machines: Launching remote viewers works better now

The "Launch viewer" button in Cockpit Machines works by letting the browser download a small file that should be recognized by the browser to be of the specific `application/x-virt-viewer` type used to launch the virt-viewer application. This did not always work reliably since Cockpit would not pass a filename along with the content of the file. This has been fixed and the "Launch viewer" button should not work in all situations (where virt-viewer has been correctly installed).

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -296,37 +296,48 @@ export function rephraseUI(key, original) {
  * @param mimeType
  * @returns {*}
  */
-export function fileDownload({ data, _fileName = 'myFile.dat', mimeType = 'application/octet-stream' }) {
+export function fileDownload({ data, fileName = 'myFile.dat', mimeType = 'application/octet-stream' }) {
     if (!data) {
         console.error('fileDownload(): no data to download');
         return false;
     }
 
-    const a = document.createElement('a');
-    a.id = 'dynamically-generated-file';
-    a.href = `data:${mimeType},${encodeURIComponent(data)}`;
-    document.body.appendChild(a); // if not used further then at least within integration tests
+    // It is important to tell the filename to the browser, especially
+    // the extension. That makes mime-type detection work, which seems
+    // to be necessary for at least Chrome. It also produces files in
+    // the Download folder of the user that make sense later.
+    //
+    // So we want to trigger the download by clicking on a link like
+    // this:
+    //
+    //   <a href="data:<mime>,<data>" download="<filename.ext>" />
+    //
+    // This, however, needs a CSP of "frame-src data:" in our parent
+    // frame (the Shell), so that the browser allows 'navigating' to
+    // the "data:" URL in our iframe.
+    //
+    // But we can instead create our own custom iframe inside
+    // ourselves and put the link inside that. This nested iframe also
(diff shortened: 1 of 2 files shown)


mvollmer:
This does not need any change to the Shell CSP, but making that change might be nice nevertheless.

If the Shell had "frame-src 'self' data:", we could use a real `<a>` element as the "Launch viewer" button that the user really clicks on. The would make all this magic unnecessary. 

(We would just have to style it as a PF button, which is probably possible.)

mvollmer:
> If the Shell had "frame-src 'self' data:",

Also, "frame-src 'self' vnc:" would be welcome, to start external viewers directly via `vnc://...` links. But that needs changes in virt-viewer and is thus not a immediate option for us.

mvollmer:
Another alternative is to use a external channel, like the SOS reports. It allows us to control the filename via the Content-Disposition header. But that doesn't really make sense for data that exists in JavaScript and not in a file on the host.

src/helpers.js
```diff
@@ -296,37 +296,48 @@ export function rephraseUI(key, original) {
...
         window.setTimeout(() => document.body.removeChild(f), 333);
-    }
+    });
 
-    window.setTimeout(() => { // give test browser some time ...
-        logDebug('removing temporary A.HREF for filedownload');
-        document.body.removeChild(a);
-    }, 5000);
+    // Start the show
+    document.body.appendChild(f);
```
jelly, line 340: We should probably still clean this up? Ideally we would convert the "launch viewer" button into a `<a` link itself but react-console does not offer that only an `onDownload` callback

jelly (changes requested):
Can't we clean up the created iframe?

src/helpers.js
mvollmer: It is cleaned up, on line 336: `window.setTimeout(() => document.body.removeChild(f), 333);` Does that not work?

jelly: Oh sorry, read over that

jelly (approved):

PR #2049, opened 2025-03-03
console: Launch external SPICE viewer via spice:// URL

mvollmer:
Demo: https://youtu.be/KpNcTKLjhR4
Story: https://issues.redhat.com/browse/COCKPIT-1318

- [x] #2008
- [ ] Allow user choice of VNC via console.vv, SPICE via console.vv, and SPICE via link, somehow.
- [ ] Update popup to talk about the choice.

### DIFF
--- a/src/components/vm/consoles/common.tsx
+++ b/src/components/vm/consoles/common.tsx
@@ -17,7 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import cockpit from 'cockpit';
 
 import type { VM, VMGraphics } from '../../../types';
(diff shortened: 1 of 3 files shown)


Venefilyn:
Had a convo about buttons for launching here that is relevant to this PR:

https://github.com/cockpit-project/cockpit-machines/pull/2008#discussion_r2152614744

src/components/vm/consoles/common.tsx
```diff
@@ -64,8 +64,21 @@ export function connection_address() {
 }
 
 export function console_launch(vm: VM, consoleDetail: VMGraphics) {
-    // fire download of the .vv file
-    domainDesktopConsole({ name: vm.name, consoleDetail: { ...consoleDetail, address: connection_address() } });
+    if (consoleDetail.type == "spice") {
+        // Launch via direct link
+        const protocol = consoleDetail.tlsPort ? "spice+tls" : "spice";
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2049-1768a41c-20251020-073622-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
Here are some design thoughts:

 - We can use a "split menu button" to let the user chose the action in a dropdown that the main button will then perform. 
 - Instead of making the dropdown modify what the main button does, the dropdown items can do the action immediately. 
 
I don't know what the "best pattern" here is. In our case, we don't have a clear primary action and a set of secondary actions. It's purely user preference which action to perform, and the user will likely do the same thing every time, so the first behavior is probably best for us.

 - We can also move the choice of action from a dropdown into the big helper popup. There we can explain the choices better with more text.

 - In any case, we should remember the choice, maybe per machine, or maybe globally.

 - Instead of omitting impossible actions (such as omitting SPICE when the VM has no SPICE), we can also always offer all options and then dynamically add whatever is missing when the user actually launches the viewer.

Venefilyn:
> Instead of making the dropdown modify what the main button does, the dropdown items can do the action immediately.

IMO this is more common and involves less clicks. I don't think I've seen much for configuring what a button does based on a dropdown

PR #2052, opened 2025-03-06
test: fully type machineslib

jelly:
Typing machines own testhelper mixin makes it easier to write tests assisted with typehints.

### DIFF
--- a/test/check-machines-networks
+++ b/test/check-machines-networks
@@ -122,7 +122,8 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         # Transient network cannot be deleted
         b.wait_visible(f'li.pf-m-aria-disabled > #delete-network-test_network4-{connectionName}')
         b.click(f'#deactivate-network-test_network4-{connectionName}')  # Deactivate transient network
-        self.waitNetworkRow("test_network4", connectionName, False)  # Check it's not present after deactivation
+        self.waitNetworkRow("test_network4", connectionName,
+                            present=False)  # Check it's not present after deactivation
 
     def testNetworksCreate(self):
         b = self.browser
@@ -870,7 +871,7 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
                             f'#delete-network-test_network2-{connectionName}')
         b.wait_in_text(".pf-v5-c-modal-box__body .pf-v5-c-description-list", "test_network2")
         b.click(".pf-v5-c-modal-box__footer button:contains(Delete)")
-        self.waitNetworkRow("test_network2", connectionName, False)
+        self.waitNetworkRow("test_network2", connectionName, present=False)
 
         # Delete an active network
         b.wait_in_text(f"#network-test_network2_clash-{connectionName}-state", "active")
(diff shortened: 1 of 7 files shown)


martinpitt:
TypeError: VirtualMachinesCaseHelpers.waitVmRow() takes from 2 to 3 positional arguments but 4 were given

jelly:
> TypeError: VirtualMachinesCaseHelpers.waitVmRow() takes from 2 to 3 positional arguments but 4 were given

That's accuse of boolean trap!

    def waitVmRow(self, vmName: str, connectionName: str = 'system', *, present: bool = True) -> None:

test/check-machines-networks
```diff
@@ -122,7 +122,7 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         # Transient network cannot be deleted
         b.wait_visible(f'li.pf-m-aria-disabled > #delete-network-test_network4-{connectionName}')
         b.click(f'#deactivate-network-test_network4-{connectionName}')  # Deactivate transient network
-        self.waitNetworkRow("test_network4", connectionName, False)  # Check it's not present after deactivation
+        self.waitNetworkRow("test_network4", connectionName, present=False)  # Check it's not present after deactivation
```
martinpitt: This change should not be part of "Remove useless React fragment". Rebase error?

test/machineslib.py
```diff
@@ -19,29 +19,34 @@
...
     return image.startswith("rhel-8")
 
 
-class VirtualMachinesCaseHelpers:
+class VirtualMachinesCaseHelpers(unittest.TestCase):
     machine: testvm.Machine
+    browser: testlib.Browser
 
-    def waitPageInit(self):
+    def allow_journal_messages(self, *_patterns: str) -> None: ...
```
martinpitt, line 47: That's a method of testlib.py. Why does that need to be declared here? Or is this temporary until we pull in a typed testlib.py?

martinpitt (changes requested):
Thanks! Marius is on PTO now, so taking over the review.

test/check-machines-networks
jelly: Woops!

test/machineslib.py
```diff
@@ -19,29 +19,34 @@
...
 
 
 # software TPM doesn't work on some OSes
-def hasBrokenTPM(image):
+def hasBrokenTPM(image: str) -> bool:
     return image.startswith("rhel-8")
 
 
-class VirtualMachinesCaseHelpers:
+class VirtualMachinesCaseHelpers(unittest.TestCase):
```
martinpitt, line 43: This is wrong in two different ways:
 * Deriving from unittest.TestCase doesn't give you the `allow_journal_messages()` declaration.
 * This class was explicitly written to be a mix-in. Nothing else apart from `VirtualMachinesCase` uses it any more though

I rather suggest to just move everything into `VirtualMachinesCase` (which already derives from our `MachinesCase`).

martinpitt (changes requested):

test/machineslib.py
jelly: Just rebase on https://github.com/cockpit-project/cockpit-machines/pull/2059 and we should be good

martinpitt: That PR landed, unblocked

martinpitt (approved):
Nice!

PR #2055, opened 2025-03-07
Biome follow up fixes

jelly:


src/libvirtApi/domain.js
```diff
@@ -78,7 +78,7 @@ import { call, Enum, timeout, resolveUiState } from './helpers.js';
 import { CLOUD_IMAGE, DOWNLOAD_AN_OS, LOCAL_INSTALL_MEDIA_SOURCE, needsRHToken } from "../components/create-vm-dialog/createVmDialogUtils.js";
 
 export const domainCanConsole = (vmState) => vmState == 'running';
-export const domainCanDelete = (vmState, vmId) => true;
+export const domainCanDelete = (_vmState, _vmId) => true;
```
martinpitt, line 81: This is just silly. Can we eliminate this altogether?

martinpitt (approved):
Thanks! OK, but let's use the opportunity to clean up that silly constant function.

martinpitt (approved):
Dankjewel! :broom:

PR #2073, opened 2025-03-27
test: drop centos-8 conditional

jelly:
CentOS 8 is no longer supported.


martinpitt (approved):
Consider updating AI to learn that supported OSes are not per test :smirk:

PR #2080, opened 2025-04-02
Drop pf-m-redhat-font

jelly:
This no longer exists as class in PF since the PF 3->4 migration.

See also https://github.com/cockpit-project/cockpit/commit/517c8e05fed676a6bc7751a01debfdf4f644fe1b


tomasmatus (approved):
thanks, assuming green

Venefilyn (approved):

PR #2107, opened 2025-04-29
libvirtApi: Chunk nodeDeviceGetAll calls to nodeDeviceGet

SludgeGirl:
Systems with a large amount of pci devices will call udevadm alot as part of the nodeDeviceGetAll calls, this will then lead to a bunch of results crashing due to reaching the maximum of allowed open files

The system I tested this against had 430 devices listed by Qemu through the same method cockpit retrieves them:

```
$ dbus-send --system --print-reply --type=method_call --dest=org.libvirt /org/libvirt/QEMU org.libvirt.Connect.ListNodeDevices uint32:0 | wc -l
430
```

This then goes on to error with:

```
Apr 28 16:35:16 /usr/bin/cockpit-bridge[20468]: Failed to spawn ('udevadm', 'info', '--path', '/sys/devices/pci0000:fe/0000:fe:08.1'): [Errno 24] Too many open files
```

Chunking these results to a reasonable amount shouldn't slow things down too much and will help mitigate this issue

### DIFF
--- a/src/libvirtApi/nodeDevice.js
+++ b/src/libvirtApi/nodeDevice.js
@@ -66,7 +66,13 @@ export async function nodeDeviceGetAll({
 }) {
     try {
         const [objPaths] = await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'ListNodeDevices', [0], { timeout, type: 'u' });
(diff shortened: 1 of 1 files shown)


jelly (commented):
Thanks! The PR looks good, triggered the tests.

I wonder if we can instead combine udevadm calls, but that is something for a different PR :)

jelly (approved):

PR #2136, opened 2025-05-16
package: Add .ts and .tsx extensions to eslint runs

mvollmer:
The ./test/common/static-code check already checks .ts and .tsx files, so this does not affect what gets accepted in CI, but it's less surprising when "npm run eslint:fix" doesn't ignore TypeScript.


jelly (approved):
Sounds good

PR #2140, opened 2025-05-19
vm: Don't touch the "shareable" attribute of disks

### LINKED ISSUE #2011: Disks should not become shared without explicit user consent
Currently, when an user add an already-in-use RAW volume to another virtual machine, the disk is attached as a "shared writable" volume and, at the same time, such a change is scheduled for the first virtual machine also. Sharing a disk in this manner poses a significant risks as, when the two virtual  machines start, they will both try to open the underlying volume and corruption will quickly happen. The only warning the user see is a yellow/orange triangle and a (small) orange statement that the disk will become shared.

Changing a disk from "writeble" to "shared writable" should be done only after explicit user consent and never be easy to accidentally trigger. `virt-manager` does exactly that: it asks the user if the disk should *really* be added to another virtual machine and, after that, does *not* automatically marks the disk as shared. So user not only needs to confirm the add operation, but it has to enable the "shareable" flag on *both* machines before allowing them to start.

Please do the same for `cockpit-machines`. As a side note, the add disk via "custom path" form does *not* automatically change the "shareable" flag, so removing this behavior from the "use existing" disk form is going to present a more consistent (and much safer) behavior.

mvollmer:
This is deemed too dangerous to do it by default.

Fixes #2011

-----------------

### Machines: The "shareable" attribute of disks is no longer modified by Cockpit

By default, libvirt can attach a disk to multiple virtual machines, but if the disk allows writing, only a single one of these machines can be running at a given time. Writing to a disk concurrently from multiple VMs at the same time can easily lead to data corruption, and libvirt prevents this unless the "shareable" attribute of a disk is set. Previously, Cockpit would set this attribute without much fuzz and would thus invite data corruption. To avoid this, Cockpit no longer mark disks as "shareable" when attaching them to multiple VMs.

Moreover, the "Edit attributes" dialog no longer allows setting the "shareable" attribute either. We now believe this attribute is to dangerous to be exposed in such an easy way. If you really need it, you can use "virt-xml" on the command line. Cockpit will show disks with the "shareable" attribute as "Concurrently writable" in the UI.

### DIFF
--- a/src/components/vm/disks/vmDisksCard.jsx
+++ b/src/components/vm/disks/vmDisksCard.jsx
@@ -213,7 +213,7 @@ export const VmDisksCard = ({ vm, vms, disks, renderCapacity, supportedDiskBusTy
         if (renderAccess) {
             const access = (
                 <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }} id={`${idPrefixRow}-access`}>
-                    <FlexItem>{ disk.readonly ? _("Read-only") : disk.shareable ? _("Writeable and shared") : _("Writeable") }</FlexItem>
+                    <FlexItem>{ disk.readonly ? _("Read-only") : disk.shareable ? _("Concurrently writeable") : _("Writeable") }</FlexItem>
                     { needsShutdownDiskAccess(vm, disk.target) && <NeedsShutdownTooltip iconId={`${idPrefixRow}-access-tooltip`} tooltipId={`tip-${idPrefixRow}-access`} /> }
                 </Flex>
             );
--- a/src/components/vm/disks/diskAdd.jsx
+++ b/src/components/vm/disks/diskAdd.jsx
@@ -38,9 +38,9 @@ import { FormHelper } from 'cockpit-components-form-helper.jsx';
 
 import { FileAutoComplete } from 'cockpit-components-file-autocomplete.jsx';
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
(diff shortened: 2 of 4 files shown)


martinpitt (dismissed):
Thanks! Fine for me, I always found this rather suspicious anyway. But that now means that there's no UI for making a disk shared?

mvollmer:
> Thanks! Fine for me, I always found this rather suspicious anyway. But that now means that there's no UI for making a disk shared?

There actually is. The "Edit attributes" dialog has a "Writeable and shared" option, but only for "raw" volumes.

One could argue that this is still too misleading, and we should either remove that as well, or add copious warnings to that option.

src/components/vm/disks/vmDisksCard.jsx
```diff
@@ -213,7 +213,7 @@ export const VmDisksCard = ({ vm, vms, disks, renderCapacity, supportedDiskBusTy
         if (renderAccess) {
             const access = (
                 <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }} id={`${idPrefixRow}-access`}>
-                    <FlexItem>{ disk.readonly ? _("Read-only") : disk.shareable ? _("Writeable and shared") : _("Writeable") }</FlexItem>
+                    <FlexItem>{ disk.readonly ? _("Read-only") : disk.shareable ? _("Concurrently writeable") : _("Writeable") }</FlexItem>
```
cockpituous, line 216: This added line is not executed by any test.

martinpitt (approved):
Thanks!

mvollmer:
There is trouble with rhel-10-1, I'll investigate separately.

PR #2142, opened 2025-05-20
vm: Don't crash if parent of a host device is missing

mvollmer:
Host devices that are referenced by the "parent" attribute of other host devices are probably always reported by libvirt, but their retrieval by us might still fail.  Let's not crash in this case.

See https://issues.redhat.com/browse/RHEL-88405


jelly (approved):
Thanks!

PR #2147, opened 2025-05-22
api: Remove erroneous cockpit.assert inserted during TypeScripting

mvollmer:
It is in fact possible to reach this place with "file == null" and we shouldn't crash then.  The "Add disk" dialog does not have enough validation in it to prevent this case.


jelly (approved):
Any plans to address the validation?

PR #2150, opened 2025-05-22
nics: Allow managing source mode of directly attached network interfaces

mvollmer:
Virtual functions from SR-IOV devices might require "passthrough" to work, so let's surface the mode property of directly attached devices fully.

See https://issues.redhat.com/browse/RHEL-88407

Demo:  https://youtu.be/rhm3MDBZXV0

Screenshot:

image

------------

### Machines: The source mode for network interfaces can now be managed

Previously, Cockpit Machines would always use the "Bridge" mode for directly attached network interfaces. Some devices require other modes, however, and Cockpit Machines now let's you set them explicitly when adding or editing directly attached network interfaces.

### DIFF
--- a/src/components/vm/nics/nicBody.jsx
+++ b/src/components/vm/nics/nicBody.jsx
@@ -19,13 +19,12 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { Flex } from "@patternfly/react-core/dist/esm/layouts/Flex";
 import { FormGroup } from "@patternfly/react-core/dist/esm/components/Form";
 import { FormSelect, FormSelectOption } from "@patternfly/react-core/dist/esm/components/FormSelect";
+import { Radio } from "@patternfly/react-core/dist/esm/components/Radio";
 import { PopoverPosition } from "@patternfly/react-core/dist/esm/components/Popover";
 import { Content, ContentVariants } from "@patternfly/react-core/dist/esm/components/Content";
-import { ExternalLinkSquareAltIcon } from '@patternfly/react-icons';
 
 import { InfoPopover } from '../../common/infoPopover.jsx';
 
@@ -83,7 +82,7 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
     const virtualNetwork = [{
         name: 'network',
         desc: 'Virtual network',
-        detailHeadline: _("This is the recommended config for general guest connectivity on hosts with dynamic / wireless networking configs."),
+        detailHeadline: _("This is the recommended type for general guest connectivity on hosts with dynamic / wireless networking configs."),
         detailParagraph: _("Provides a connection whose details are described by the named network definition.")
     }];
     if (connectionName !== 'session') {
@@ -92,26 +91,13 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
             {
                 name: 'bridge',
                 desc: 'Bridge to LAN',
-                detailHeadline: _("This is the recommended config for general guest connectivity on hosts with static wired networking configs."),
+                detailHeadline: _("This is the recommended type for general guest connectivity on hosts with static wired networking configs."),
                 detailParagraph: _("Provides a bridge from the guest virtual machine directly onto the LAN. This needs a bridge device on the host with one or more physical NICs.")
             },
             {
                 name: 'direct',
                 desc: 'Direct attachment',
-                detailHeadline: _("This is the recommended config for high performance or enhanced security."),
-                detailParagraph: _("In the default 'vepa' mode, switching is offloaded to the external switch. If the switch is not VEPA-capable, communication between guest virtual machines, or between a guests and the host is not possible."),
-                externalDocs: (
-                    <Button isInline
-                            className='ct-external-docs-link'
-                            variant="link"
-                            component="a"
-                            icon={<ExternalLinkSquareAltIcon />}
-                            iconPosition="right"
-                            target="__blank"
-                            href="https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface">
-                        {_("more info")}
-                    </Button>
-                )
+                detailParagraph: _("This is the recommended type for high performance or enhanced security."),
             },
         ];
     } else {
@@ -170,10 +156,7 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
                                         {availableNetworkTypes.map(type => (<Content key={type.name}>
                                             <Content component={ContentVariants.h4}>{type.desc}</Content>
                                             <strong>{type.detailHeadline}</strong>
-                                            <p>
-                                                {type.detailParagraph}
-                                                {type.externalDocs}
-                                            </p>
+                                            <p>{type.detailParagraph}</p>
                                         </Content>))}
                                     </Flex>}
                            />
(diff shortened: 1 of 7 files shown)


src/components/vm/nics/nicBody.jsx
```diff
@@ -203,6 +203,30 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
                     </FormSelect>
                 </FormGroup>
             )}
+            {dialogValues.networkType == "direct" && (
+                <FormGroup fieldId={`${idPrefix}-source-mode`} label={_("Mode")} hasNoPaddingTop isInline>
+                    <Radio id={`${idPrefix}-source-mode-vepa`}
+                        name="mode-vepa"
+                        isChecked={dialogValues.networkSourceMode == "vepa"}
+                        label={_("VEPA")}
+                        onChange={() => onValueChanged('networkSourceMode', "vepa")} />
```
cockpituous, line 212: This added line is not executed by any test.

src/components/vm/nics/nicBody.jsx
```diff
@@ -203,6 +203,30 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
+                    <Radio id={`${idPrefix}-source-mode-bridge`}
+                        name="mode-bridge"
+                        isChecked={dialogValues.networkSourceMode == "bridge"}
+                        label={_("Bridge")}
+                        onChange={() => onValueChanged('networkSourceMode', "bridge")} />
```
cockpituous, line 217: This added line is not executed by any test.

src/components/vm/nics/nicBody.jsx
```diff
@@ -203,6 +203,30 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
+                    <Radio id={`${idPrefix}-source-mode-private`}
+                        name="mode-private"
+                        isChecked={dialogValues.networkSourceMode == "private"}
+                        label={_("Private")}
+                        onChange={() => onValueChanged('networkSourceMode', "private")} />
+                    <Radio id={`${idPrefix}-source-mode-passthrough`}
+                        name="mode-passthrough"
+                        isChecked={dialogValues.networkSourceMode == "passthrough"}
+                        label={_("Passthrough")}
+                        onChange={() => onValueChanged('networkSourceMode', "passthrough")} />
```
cockpituous, line 227: This added line is not executed by any test.

src/components/vm/nics/nicBody.jsx
```diff
@@ -203,6 +203,31 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
                     </FormSelect>
                 </FormGroup>
             )}
+            {dialogValues.networkType == "direct" && (
+                <FormGroup id={`${idPrefix}-source-mode`} label={_("Mode")} hasNoPaddingTop isInline
+                           data-value={dialogValues.networkSourceMode}>
+                    <Radio id={`${idPrefix}-source-mode-vepa`}
+                        name="mode-vepa"
+                        isChecked={dialogValues.networkSourceMode == "vepa"}
+                        label={_("VEPA")}
```
jelly, line 212: I don't think we can translate acronyms right?

jelly (commented):
Looks good, one small suggestion to translation string, we should imo also not translate VNC.

src/components/vm/nics/nicBody.jsx
mvollmer: Fixed.

jelly (approved):
Thanks!

Venefilyn:
> @Venefilyn, this has a bit of UI in it, if you are interested. Some open issues I can think of:
> 
>     * Should the "Mode" line maybe come after "Interface type" instead of after "Source"?

@mvollmer I think how it is currently is fine. We start with what we're creating, what entity to take from, what mode the entity should be mapped to. So I think it narrows down naturally here. But the other way around could work too, though this way seems better.

>     * Should we put a (?) icon on the "Mode" label with the link to the docs?

Yes! I think a (?) here would work. We could do tooltips on the checkbox texts too but we don't have that anywhere else so would be weird

mvollmer:
> @mvollmer I think how it is currently is fine. We start with what we're creating, what entity to take from, what mode the entity should be mapped to. So I think it narrows down naturally here. But the other way around could work too, though this way seems better.

Yes, agreed. Also, in the use case the mode is a consequence of which source you use. Some sources need one mode, other sources need another. So it also makes sense to select the mode after selecting the source.

> Yes! I think a (?) here would work. We could do tooltips on the checkbox texts too but we don't have that anywhere else so would be weird

Alright, I'll come up with something.

mvollmer:
> Alright, I'll come up with something.

Looks like this now:

image

The text is from the "more info" link that we had previously, https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface

src/components/vm/nics/nicBody.jsx
```diff
@@ -203,6 +186,65 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
+                                           {_("This feature attaches a virtual function of a SRIOV capable NIC directly to a VM without losing the migration capability. All packets are sent to the VF/IF of the configured network device. Depending on the capabilities of the device additional prerequisites or limitations may apply.")}
+                                       </Content>
+                                   </Content>}
+                           />
+                       }>
+                    <Radio id={`${idPrefix}-source-mode-vepa`}
+                        name="mode-vepa"
+                        isChecked={dialogValues.networkSourceMode == "vepa"}
+                        label="VEPA"
+                        onChange={() => onValueChanged('networkSourceMode', "vepa")} />
```
cockpituous, line 230: This added line is not executed by any test.

src/components/vm/nics/nicBody.jsx
```diff
@@ -203,6 +186,65 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
+                    <Radio id={`${idPrefix}-source-mode-bridge`}
+                        name="mode-bridge"
+                        isChecked={dialogValues.networkSourceMode == "bridge"}
+                        label={_("Bridge")}
+                        onChange={() => onValueChanged('networkSourceMode', "bridge")} />
+                    <Radio id={`${idPrefix}-source-mode-private`}
+                        name="mode-private"
+                        isChecked={dialogValues.networkSourceMode == "private"}
+                        label={_("Private")}
+                        onChange={() => onValueChanged('networkSourceMode', "private")} />
```
cockpituous, line 240: This added line is not executed by any test.

src/components/vm/nics/nicBody.jsx
```diff
@@ -203,6 +186,65 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
+                                           {_("Private")}
+                                       </Content>
+                                       <Content component={ContentVariants.p}>
+                                           {_("All packets are sent to the external bridge and will only be delivered to a target VM on the same host if they are sent through an external router or gateway and that device sends them back to the host. This procedure is followed if either the source or destination device is in \"Private\" mode.")}
+                                       </Content>
+                                       <Content component={ContentVariants.h4}>
+                                           {_("Passthrough")}
+                                       </Content>
+                                       <Content component={ContentVariants.p}>
+                                           {_("This feature attaches a virtual function of a SRIOV capable NIC directly to a VM without losing the migration capability. All packets are sent to the VF/IF of the configured network device. Depending on the capabilities of the device additional prerequisites or limitations may apply.")}
```
jelly, line 221: Very nitpicky it is commonly typed as `SR-IOV` https://en.wikipedia.org/wiki/Single-root_input/output_virtualization

mvollmer: :-) Tell that to https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface!

jelly (approved):
Thanks!

PR #2163, opened 2025-05-31
Add CPU and Memory usage metrics to Host VM List

DlpOne:
Currently, the VM list does not display resource usage metrics for  virtual machines. This makes it difficult to identify VMs that are consuming excessive CPU or memory.

This patch adds CPU and memory usage statistics for each VM in the VM list. Similar functionality is available in other virtualization management solutions and improves visibility and resource management.

HostVmListUsage

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -337,6 +337,8 @@ class AppActive extends React.Component {
                     networks={networks}
                     actions={vmActions}
                     resourceHasError={this.state.resourceHasError}
+                    onUsageStartPollingVMS={() => vms.filter(vm => !vm.isUi).forEach(vm => {usageStartPolling({name: vm.name, id: vm.id, connectionName: vm.connectionName })})}
+                    onUsageStopPollingVMS={() => vms.filter(vm => !vm.isUi).forEach(vm => {usageStopPolling({name: vm.name, connectionName: vm.connectionName })})}
                     onAddErrorNotification={this.onAddErrorNotification} />
                 }
                 {path.length > 0 && path[0] == 'storages' &&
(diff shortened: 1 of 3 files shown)


mvollmer:
Thanks, nice idea!

I can see three issues:

 1) Usage polling is switched on for all VMs when the host list is first shown, but not updated when the list changes. Any VM that appears later on that list will not have its usage polled. In fact, when cockpit-machines loads and shows the VM list, it is initially empty for me, and then filled with VMS. VMs that are added this way during loading do not get their usage polled.

 1) If there are "a lot" of machines in the list, polling can get expensive. We should put some limits in place to avoid excessive network traffic and server load because of polling.

 1) There is a bug in our existing `usageStartPolling` and `usageStopPolling` functions: If you call `usageStartPolling` too quickly after `usageStopPolling` for a given machine, you will get two polling loops. Thus, when you are on the details page for machine "foo" and then switch to the list of VMs, leaving the "foo" page will stop polling, but entering the list page will start it again and too quickyl. Now "foo" is being polled twice per polling interval.

@DlpOne, do you think you can work on these?  If the bug in 3) sounds a bit overwhelming to fix, I can have a go at that. Issue 1) will probably get solved at the same time.

mvollmer (changes requested):
See above.

mvollmer:
@DlpOne, this is what I have in mind: https://github.com/cockpit-project/cockpit-machines/pull/2183

The "WIP" commit in there shows how to enable polling for all machines on the Overview.

Still open:

 - We should only poll on the overview when there not "too many" machines. 20 seems like a good default limit, but we should make that configurable, like the polling interval itself.

 - Ideally, we should only poll the machines that are actually shown. Those that are scrolled out of view or are filtered out don't need to be polled. But that's significantly more work and I'd say we don't need to do that now.

mvollmer:
@DlpOne, I have added your commit to #2183, let's continue there, ok?

DlpOne:
@mvollmer  Your adjustments look much better. Yes, let's continue from there

PR #2170, opened 2025-06-04
[rhel-10-0] nics: Allow managing source mode of directly attached network interfaces

mvollmer:
Virtual functions from SR-IOV devices might require "passthrough" to work, so let's surface the mode property of directly attached devices fully.

See https://issues.redhat.com/browse/RHEL-88407

### DIFF
--- a/src/components/vm/nics/nicBody.jsx
+++ b/src/components/vm/nics/nicBody.jsx
@@ -26,6 +26,7 @@ import { FormSelect, FormSelectOption } from "@patternfly/react-core/dist/esm/co
 import { Popover, PopoverPosition } from "@patternfly/react-core/dist/esm/components/Popover";
 import { Text, TextContent, TextVariants } from "@patternfly/react-core/dist/esm/components/Text";
 import { ExternalLinkSquareAltIcon, OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import { Radio } from "@patternfly/react-core/dist/esm/components/Radio";
 
 import cockpit from 'cockpit';
 
@@ -202,6 +203,22 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
                                 value={defaultNetworkSource}>
                         {networkSourcesContent}
                     </FormSelect>
+                </FormGroup>
+            )}
+            {dialogValues.networkType == "direct" && (
+                <FormGroup id={`${idPrefix}-source-mode`} label={_("Mode")} hasNoPaddingTop isInline
+                           data-value={dialogValues.networkSourceMode}>
+                    {["vepa", "bridge", "private", "passthrough"].map(mode =>
+                        <Radio
+                            key={mode}
+                            id={`${idPrefix}-source-mode-${mode}`}
+                            name={`mode-${mode}`}
+                            isChecked={dialogValues.networkSourceMode == mode}
+                            // The label is not translated since the
+                            // documentation we link to is always in
+                            // English.
+                            label={<pre>{mode}</pre>}
+                            onChange={() => onValueChanged('networkSourceMode', mode)} />)}
                 </FormGroup>
             )}
         </>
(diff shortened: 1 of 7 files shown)


mvollmer:
We might want #2171 as well.

mvollmer:
> We might want #2171 as well.

Included.

src/components/vm/nics/nicBody.jsx
```diff
@@ -200,29 +201,19 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
-                                       </Text>
-                                       <Text component={TextVariants.h4}>
-                                           {_("Private")}
-                                       </Text>
-                                       <Text component={TextVariants.p}>
-                                           {_("All packets are sent to the external bridge and will only be delivered to a target VM on the same host if they are sent through an external router or gateway and that device sends them back to the host. This procedure is followed if either the source or destination device is in \"Private\" mode.")}
-                                       </Text>
-                                       <Text component={TextVariants.h4}>
-                                           {_("Passthrough")}
+                                           {_("The mode influences the delivery of packets.")}
```
martinpitt: I don't consider this change appropriate for z-stream. It is not a critical bug, and it adds new translatable strings. I know the previous commit does that too, but that at least has a better justification.

src/components/vm/nics/nicBody.jsx
```diff
@@ -81,7 +81,7 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
     const virtualNetwork = [{
         name: 'network',
         desc: 'Virtual network',
-        detailHeadline: _("This is the recommended config for general guest connectivity on hosts with dynamic / wireless networking configs."),
+        detailHeadline: _("This is the recommended type for general guest connectivity on hosts with dynamic / wireless networking configs."),
```
martinpitt: Please revert this bit and the same config → type change below. This needlessly breaks translations, and thus is not appropriate for a stable z-stream update.

src/components/vm/nics/nicBody.jsx
```diff
@@ -90,26 +90,13 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
-                            variant="link"
-                            component="a"
-                            icon={<ExternalLinkSquareAltIcon />}
-                            iconPosition="right"
-                            target="__blank"
-                            href="https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface">
-                        {_("more info")}
-                    </Button>
-                )
+                detailParagraph: _("This is the recommended type for high performance or enhanced security."),
```
martinpitt: Same here, please use "config" to re-use the existing string.

martinpitt (changes requested):
Hrm, hmm.. We need to lessen the translation regression here. Thanks!

src/components/vm/nics/nicBody.jsx
mvollmer: What is "this change" exactly?  I can remove gratuitous changes to the popups and restrict this PR to just adding the "Mode" line with the radios. Would that be acceptable?

martinpitt: The whole "Replace text about direct mode with link to external docs" commit. Note that RH has some contractual obligation to have some languages 100% translated. That's why we always do this post-dev-freeze dance (details in private if you are interested)

mvollmer: > The whole "Replace text about direct mode with link to external docs" commit. 

Note that the text that this commit removes is added by the first commit in this PR... :-) I'll redo the whole lot while minimizing UI text changes.

martinpitt: Ah, oops -- sorry, I read the commits individually.

martinpitt (approved):
That sir, is excellent! :100: The new "Mode" string is already present elsewhere. :+1:

mvollmer:
> The new "Mode" string is already present elsewhere. 👍

Yes, I checked!

(I have to say, I am a bit embarrassed by the chaotic, superficial stuff I am delivering here and on main... My only excuse is that I am not used anymore to getting my sh*t merged so easily... :-) I'll improve, promise.)

PR #2171, opened 2025-06-04
nics: Replace text about direct mode with link to external docs

mvollmer:
The text was too massive, jargony, ultimately not super helpful out of context, and we don't want to subject our translators to it. So let's just put the link to the libvirt documentation back.

image


mvollmer:
@Venefilyn, we might want this in 333.

PR #2173, opened 2025-06-05
nics: Don't translate direct attachement modes

mvollmer:
They are basically magic API values and the important thing is that they can be found in the documentation that we link to, which is in English.

(Also, when the mode is displayed in the list of interfaces, it isn't translated either.)

(Also, also, we do the same thing in the [RHEL 10.0 backport](https://github.com/cockpit-project/cockpit-machines/pull/2170), which can't have any new UI strings.)

### DIFF
--- a/src/components/vm/nics/nicBody.tsx
+++ b/src/components/vm/nics/nicBody.tsx
@@ -250,26 +250,17 @@ export const NetworkTypeAndSourceRow = ({
                                    </Content>}
                            />
                        }>
-                    <Radio id={`${idPrefix}-source-mode-vepa`}
-                        name="mode-vepa"
-                        isChecked={dialogValues.networkSourceMode == "vepa"}
-                        label="VEPA"
-                        onChange={() => onValueChanged('networkSourceMode', "vepa")} />
-                    <Radio id={`${idPrefix}-source-mode-bridge`}
-                        name="mode-bridge"
-                        isChecked={dialogValues.networkSourceMode == "bridge"}
-                        label={_("Bridge")}
-                        onChange={() => onValueChanged('networkSourceMode', "bridge")} />
-                    <Radio id={`${idPrefix}-source-mode-private`}
-                        name="mode-private"
-                        isChecked={dialogValues.networkSourceMode == "private"}
-                        label={_("Private")}
-                        onChange={() => onValueChanged('networkSourceMode', "private")} />
-                    <Radio id={`${idPrefix}-source-mode-passthrough`}
-                        name="mode-passthrough"
-                        isChecked={dialogValues.networkSourceMode == "passthrough"}
-                        label={_("Passthrough")}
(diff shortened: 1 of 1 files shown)


mvollmer:
@Venefilyn, apologies for the chaos with this feature. "speed of iteration beats quality of iteration" and all that, but even I am not happy with how much I missed along the way.

src/components/vm/nics/nicBody.jsx
```diff
@@ -215,26 +215,17 @@ export const NetworkTypeAndSourceRow = ({ idPrefix, onValueChanged, dialogValues
...
+                        <Radio
+                            key={mode}
+                            id={`${idPrefix}-source-mode-${mode}`}
+                            name={`mode-${mode}`}
+                            isChecked={dialogValues.networkSourceMode == mode}
+                            // The label is not translated since the
+                            // documentation we link to is always in
+                            // English.
+                            label={<pre>{mode}</pre>}
+                            onChange={() => onValueChanged('networkSourceMode', mode)} />)}
```
Venefilyn: Do we not want to keep the capitalization at least? So something like this?
```suggestion
                    {["VEPA", "Bridge", "Private", "Passthrough"].map(mode => {
                        const lowercaseMode = mode.toLowerCase();
                        return <Radio
                            key={lowercaseMode}
                            id={`${idPrefix}-source-mode-${lowercaseMode}`}
                            name={`mode-${lowercaseMode}`}
                            isChecked={dialogValues.networkSourceMode == lowercaseMode}
                            // The label is not translated since the
                            // documentation we link to is always in
                            // English.
                            label={<pre>{mode}</pre>}
                            onChange={() => onValueChanged('networkSourceMode', lowercaseMode)} />
                        })}
```

Venefilyn:
> @Venefilyn, apologies for the chaos with this feature. "speed of iteration beats quality of iteration" and all that, but even I am not happy with how much I missed along the way.

That makes sense, thanks for spotting it :pray:

src/components/vm/nics/nicBody.jsx
martinpitt: I suppose that doesn't work, as the `name` needs to be the actual lower-case API value. TBH I think the current approach is fine, "VEPA" isn't translated and "private" and "passthrough" could very well be translated.

Venefilyn:
Small rebase needed

src/components/vm/nics/nicBody.jsx
mvollmer: Yes, and I even put them into `<pre>` to make them look like code.

This is something clearly in the ["non-guided"](https://github.com/mvollmer/cockpit-design-elements/blob/main/elements.md) part of Cockpit, and we could hide it further, maybe with a fourth "Direct attachement with mode (advanced)" type. Or just say "Direct attachement (advanced)".

PR #2181, opened 2025-06-09
nics: Fix listing of host interfaces via NetworkManager

mvollmer:
The Python bridge does not fall back to emulating a ObjectManager with introspection if it is told to watch a path namespace. So we need to get our proxies from the path where the NetworkManager has its object manager, which is "/org/freedesktop".


jelly (approved):
Thanks!

PR #2200, opened 2025-06-23
Cloud create flake

jelly:
- [x] #2206

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1250,9 +1250,9 @@ vnc_password= "{vnc_passwd}"
 
         def createAndVerifyVirtInstallArgsCloudInit(self):
             if self.create_and_run:
-                self.browser.click(".pf-v6-c-modal-box__footer button:contains(Create and run)")
+                self.browser.click(".pf-v6-c-modal-box__footer #create-and-run")
             else:
-                self.browser.click(".pf-v6-c-modal-box__footer button:contains(Create and edit)")
+                self.browser.click(".pf-v6-c-modal-box__footer #create-and-edit")
             self.browser.wait_not_present("#create-vm-dialog")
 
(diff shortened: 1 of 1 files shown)


jelly:
This is a disaster @martinpitt, another cascading effect of flakes or issues with BiDi clicks not getting registered. I don't see that I can fix all the other flakes in this PR this week.

martinpitt:
@jelly As discussed -- let's land https://github.com/cockpit-project/cockpit/pull/22153 and then sprinkle `b.chromium_fake_mouse = True` over the problematic tests until they get green, then fix them one by one.

test/check-machines-create
```diff
@@ -1250,9 +1250,9 @@ vnc_password= "{vnc_passwd}"
 
         def createAndVerifyVirtInstallArgsCloudInit(self):
             if self.create_and_run:
-                self.browser.click(".pf-v6-c-modal-box__footer #create-and-run")
+                self.browser.mouse(".pf-v6-c-modal-box__footer #create-and-run", "click", y=0)
             else:
-                self.browser.click(".pf-v6-c-modal-box__footer #create-and-edit")
+                self.browser.mouse(".pf-v6-c-modal-box__footer #create-and-edit", "click", y=0)
```
martinpitt, line 1255: This [does not seem to flake on firefox](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit-machines&days=7&test=%2Ftest%2Fcheck-machines-create+TestMachinesCreate.testCreateCloudBaseImage), so I made this blocking on #2206 and then we can instead use the `chromium_fake_mouse` flag for this function.

martinpitt:
As per chat discussions, taking over. Replaced with #2208

PR #2202, opened 2025-06-25
nics: ignore domain not found error caused by domain being removed

Nykseli:
If user is in the VMDetails page during migration, the VM might get deleted before domainInterfaceAddresses is done processing. This race condition can end up causing a confusing view where the UI shows a redundant "Domain not found" error, and the regular "VM doesn't exist" page at the same time

The confusing page looks something like this
error-page

There might be a better way to get around the race condition but it seems like it would require a bigger rewrite

### DIFF
--- a/src/components/vm/nics/vmNicsCard.tsx
+++ b/src/components/vm/nics/vmNicsCard.tsx
@@ -301,8 +301,15 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
         domainInterfaceAddresses({ connectionName: this.props.vm.connectionName, objPath: this.props.vm.id })
                 .then(domifaddressAllSources => {
                     const allRejected = !domifaddressAllSources.some(promise => promise.status == 'fulfilled');
+                    // If user is in the VMDetails page during migration, the VM might get deleted
+                    // before domainInterfaceAddresses is done processing. This race condition can end up
+                    // causing a confusing view where the UI shows a redundant "Domain not found error"
+                    // and the regular VM doesn't exist page, at the same time.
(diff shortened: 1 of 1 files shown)


mvollmer:
Thanks, looks good!

Maybe we shouldn't ever show a notification for a non-existing VM.

mvollmer:
I have rebased this to catch up with our CI changes.

mvollmer:
Dang, this disappeared from my radar... sorry. I'll finish this.

mvollmer (approved):
Thanks!

mvollmer:
Rebased again.

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -301,8 +301,15 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
+                    // If user is in the VMDetails page during migration, the VM might get deleted
+                    // before domainInterfaceAddresses is done processing. This race condition can end up
+                    // causing a confusing view where the UI shows a redundant "Domain not found error"
+                    // and the regular VM doesn't exist page, at the same time.
+                    const domainNotFound = domifaddressAllSources.some(promise =>
+                        promise.status === 'rejected' && promise.reason?.message.startsWith("Domain not found:")
+                    );
 
-                    if (allRejected) {
+                    if (allRejected && !domainNotFound) {
```
cockpituous, line 312: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2202-23ded52b-20250902-074305-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

PR #2211, opened 2025-07-03
test: Stop installing EOL CentOS 7

martinpitt:
In RHEL 8.10's osinfo-db, CentOS 7 is marked as EOL on 2024-06-30. As this version of cockpit-machines shows OSes up to (EOL + 1 year), CentOS 7 dropped out of the list a few days ago.

Similarly, RHEL 8.{1,2} have been EOL for quite a while. Move everything to RHEL 8.10, which is supported for many years still -- and once it isn't, we can mothball this rhel-8 branch.

---

Fixes [this time bomb](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7942-ca1a84a1-20250703-094309-rhel-8-10-cockpit-project-cockpit-machines-rhel-8/log.html) from https://github.com/cockpit-project/bots/pull/7942


martinpitt:
Fixed packit config while I was at it.

test/check-machines-create
```diff
@@ -85,7 +85,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                             location=config.TREE_URL,
                                                             memory_size=128, memory_size_unit='MiB',
                                                             storage_pool=NO_STORAGE,
-                                                            os_name=config.CENTOS_7,
+                                                            os_name=config.RHEL_8_10,
```
jelly, line 88: This commit derivates from `main`. I think we should also drop CENTOS_7 in `main`

martinpitt: We can do that, but we do offer EOL releases in main, so it's a different kind of change.

I just still need to fix [this regression](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2211-1f92038b-20250703-105805-rhel-8-10/log.html#13).

martinpitt:
[meh](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2211-1f92038b-20250703-105805-rhel-8-10/log.html#13)

> Creation of VM subVmTestCreate3 failed
> No image available for RHEL 8.10 (x86_64)

Why on earth would *current* rhel 8.10 not have an image for 8.10, but one for 8.1???

martinpitt:
Fixed the regression, and also unbroke the reposchutz workflow (for the future, it runs from target branch)

test/check-machines-create
martinpitt: Sent #2212 for the corresponding main change.

jelly (approved):

jelly:
Triggered the bots refresh, not sure if I need /rhel-8 to specify the branch?

PR #2212, opened 2025-07-03
test: Stop installing EOL CentOS 7

martinpitt:
In osinfo-db, CentOS 7 is marked as EOL on 2024-06-30. We still show EOL versions, but we didn't in the rhel-8 branch yet, and in general we should not rely on outdated OSes. Move the test to RHEL 8.10.

Similarly, RHEL 8.1 has been EOL for quite a while. Move everything to RHEL 8.10, which is supported for many years still.

---

Forward port of #2211. It's not broken on main, but probably better to keep them in sync.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -91,7 +91,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                             location=config.TREE_URL,
                                                             memory_size=128, memory_size_unit='MiB',
                                                             storage_pool=NO_STORAGE,
-                                                            os_name=config.CENTOS_7,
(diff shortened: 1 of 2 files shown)


martinpitt:
Ah meh, [this does not work](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2212-9ba7b1a1-20250703-124857-debian-testing/log.html):

> ERROR    The requested volume capacity will exceed the available pool space when the volume is fully allocated. (10240 M requested capacity > 6315 M available)

So apparently 8.10 bumped the disk size requirement, and that doesn't fit into our test VMs anymore. Well, this isn't important.

PR #2224, opened 2025-07-16
libvirtApi: Use String() to turn error objects into text

mvollmer:
Checking whether they are an instance of "Error" was wrong. Our cockpit.ProcessError and the unexported DBusError are not derived from Error.

This was done with Cursor, using these prompts:

- Find places that use "instanceof Error" in a catch clause.

- In all these places, remove the "instanceof" check and use "String()" to safely convert the error to a string.

- There is an error in snapshot.ts, `ex.name` can not be replaced with "String()". Explicitly check whether the "name" field exists and keep the original comparison.

### DIFF
--- a/src/libvirtApi/domain.ts
+++ b/src/libvirtApi/domain.ts
@@ -518,7 +518,7 @@ export async function domainDelete({
         await call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [flags], { timeout, type: 'u' });
     } catch (ex) {
         // Transient domains get undefined after shut off
-        if (!live || !(ex instanceof Error && ex.message.includes("Domain not found")))
+        if (!live || !String(ex).includes("Domain not found"))
             throw ex;
     }
 }
@@ -883,13 +883,11 @@ export async function domainGet({
 
         await snapshotGetAll({ connectionName, domainPath: objPath });
     } catch (ex) {
-        if (ex instanceof Error) {
-            // "not found" is an expected error, as this runs on Stopped/Undefined events; so be quiet about these
-            if (ex.message.startsWith("Domain not found"))
-                logDebug(`GET_VM: domain ${connectionName} ${objPath} went away, undefining: ${ex.toString()}`);
-            else
-                console.warn(`GET_VM failed for ${objPath}, undefining: ${ex.toString()}`);
-        }
+        // "not found" is an expected error, as this runs on Stopped/Undefined events; so be quiet about these
+        if (String(ex).startsWith("Domain not found"))
+            logDebug(`GET_VM: domain ${connectionName} ${objPath} went away, undefining: ${String(ex)}`);
+        else
+            console.warn(`GET_VM failed for ${objPath}, undefining: ${String(ex)}`);
         // but undefine either way -- if we  can't get info about the VM, don't show it
         store.dispatch(undefineVm({ connectionName, id: objPath }));
     }
@@ -902,8 +900,7 @@ export async function domainGetAll({ connectionName } : { connectionName: Connec
         store.dispatch(deleteUnlistedVMs(connectionName, [], objPaths));
         await Promise.all(objPaths.map(path => domainGet({ connectionName, id: path })));
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_ALL_VMS action failed:', ex.toString());
+        console.warn('GET_ALL_VMS action failed:', String(ex));
         throw ex;
     }
 }
--- a/src/libvirtApi/interface.ts
+++ b/src/libvirtApi/interface.ts
@@ -34,8 +34,7 @@ export async function interfaceGetAll(): Promise<void> {
         const ipData = await cockpit.spawn(["ip", "--json", "a"], { err: "message" });
         ifaces = JSON.parse(ipData);
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn("Failed to get interfaces with ip command:", ex.toString());
+        console.warn("Failed to get interfaces with ip command:", String(ex));
     }
 
     for (const iface of ifaces) {
--- a/src/libvirtApi/network.ts
+++ b/src/libvirtApi/network.ts
@@ -105,8 +105,7 @@ export async function networkGetAll({
         const [objPaths]: [string[]] = await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'ListNetworks', [0], { timeout, type: 'u' });
         await Promise.all(objPaths.map(path => networkGet({ connectionName, id: path })));
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_ALL_NETWORKS action failed:', ex.toString());
+        console.warn('GET_ALL_NETWORKS action failed:', String(ex));
         throw ex;
     }
 }
@@ -148,8 +147,7 @@ export async function networkGet({
         const network = parseNetDumpxml(xml);
         store.dispatch(updateOrAddNetwork(Object.assign(props, network), !!updateOnly));
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_NETWORK action failed for path', objPath, ex.toString());
+        console.warn('GET_NETWORK action failed for path', objPath, String(ex));
     }
 }
 
--- a/src/libvirtApi/nodeDevice.ts
+++ b/src/libvirtApi/nodeDevice.ts
@@ -67,8 +67,7 @@ export async function nodeDeviceGet({
             store.dispatch(updateOrAddNodeDevice(deviceXmlObject));
         }
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_NODE_DEVICE action failed for path', objPath, ex.toString());
+        console.warn('GET_NODE_DEVICE action failed for path', objPath, String(ex));
     }
 }
 
@@ -87,8 +86,7 @@ export async function nodeDeviceGetAll({
             await Promise.all(objPathsChunked.map(path => nodeDeviceGet({ connectionName, id: path })));
         }
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_ALL_NODE_DEVICES action failed:', ex.toString());
+        console.warn('GET_ALL_NODE_DEVICES action failed:', String(ex));
         throw ex;
     }
 }
--- a/src/libvirtApi/snapshot.ts
+++ b/src/libvirtApi/snapshot.ts
@@ -137,12 +137,10 @@ export async function snapshotGetAll({
             snaps: snaps.sort((a, b) => Number(a.creationTime) - Number(b.creationTime))
         }));
     } catch (ex) {
-        if (ex instanceof Error) {
-            if (ex.name === 'org.freedesktop.DBus.Error.UnknownMethod')
-                logDebug("LIST_DOMAIN_SNAPSHOTS action failed for domain", domainPath, ", not supported by libvirt-dbus");
-            else
-                console.warn("LIST_DOMAIN_SNAPSHOTS action failed for domain", domainPath, ":", JSON.stringify(ex), "name:", ex.name);
-        }
+        if (ex && typeof ex === 'object' && 'name' in ex && ex.name === 'org.freedesktop.DBus.Error.UnknownMethod')
+            logDebug("LIST_DOMAIN_SNAPSHOTS action failed for domain", domainPath, ", not supported by libvirt-dbus");
+        else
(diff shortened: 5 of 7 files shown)


mvollmer:
Does any of you puny humans insist on reviewing this as well?

src/libvirtApi/domain.ts
```diff
@@ -518,7 +518,7 @@ export async function domainDelete({
         await call(connectionName, objPath, 'org.libvirt.Domain', 'Undefine', [flags], { timeout, type: 'u' });
     } catch (ex) {
         // Transient domains get undefined after shut off
-        if (!live || !(ex instanceof Error && ex.message.includes("Domain not found")))
+        if (!live || !String(ex).includes("Domain not found"))
```
cockpituous, line 521: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.ts
```diff
@@ -883,13 +883,11 @@ export async function domainGet({
...
-        if (ex instanceof Error) {
-            // "not found" is an expected error, as this runs on Stopped/Undefined events; so be quiet about these
-            if (ex.message.startsWith("Domain not found"))
-                logDebug(`GET_VM: domain ${connectionName} ${objPath} went away, undefining: ${ex.toString()}`);
-            else
-                console.warn(`GET_VM failed for ${objPath}, undefining: ${ex.toString()}`);
-        }
+        // "not found" is an expected error, as this runs on Stopped/Undefined events; so be quiet about these
+        if (String(ex).startsWith("Domain not found"))
+            logDebug(`GET_VM: domain ${connectionName} ${objPath} went away, undefining: ${String(ex)}`);
```
cockpituous, line 888: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.ts
```diff
@@ -883,13 +883,11 @@ export async function domainGet({
...
+        else
+            console.warn(`GET_VM failed for ${objPath}, undefining: ${String(ex)}`);
```
cockpituous, line 890: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.ts
```diff
@@ -902,8 +900,7 @@ export async function domainGetAll({ connectionName } : { connectionName: Connec
         store.dispatch(deleteUnlistedVMs(connectionName, [], objPaths));
         await Promise.all(objPaths.map(path => domainGet({ connectionName, id: path })));
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_ALL_VMS action failed:', ex.toString());
+        console.warn('GET_ALL_VMS action failed:', String(ex));
```
cockpituous, line 903: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/interface.ts
```diff
@@ -34,8 +34,7 @@ export async function interfaceGetAll(): Promise<void> {
         const ipData = await cockpit.spawn(["ip", "--json", "a"], { err: "message" });
         ifaces = JSON.parse(ipData);
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn("Failed to get interfaces with ip command:", ex.toString());
+        console.warn("Failed to get interfaces with ip command:", String(ex));
```
cockpituous, line 37: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/network.ts
```diff
@@ -105,8 +105,7 @@ export async function networkGetAll({
         const [objPaths]: [string[]] = await call(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'ListNetworks', [0], { timeout, type: 'u' });
         await Promise.all(objPaths.map(path => networkGet({ connectionName, id: path })));
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_ALL_NETWORKS action failed:', ex.toString());
+        console.warn('GET_ALL_NETWORKS action failed:', String(ex));
```
cockpituous, line 108: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/nodeDevice.ts
```diff
@@ -87,8 +86,7 @@ export async function nodeDeviceGetAll({
             await Promise.all(objPathsChunked.map(path => nodeDeviceGet({ connectionName, id: path })));
         }
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_ALL_NODE_DEVICES action failed:', ex.toString());
+        console.warn('GET_ALL_NODE_DEVICES action failed:', String(ex));
```
cockpituous, line 89: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/snapshot.ts
```diff
@@ -137,12 +137,10 @@ export async function snapshotGetAll({
...
         }));
     } catch (ex) {
-        if (ex instanceof Error) {
-            if (ex.name === 'org.freedesktop.DBus.Error.UnknownMethod')
-                logDebug("LIST_DOMAIN_SNAPSHOTS action failed for domain", domainPath, ", not supported by libvirt-dbus");
-            else
-                console.warn("LIST_DOMAIN_SNAPSHOTS action failed for domain", domainPath, ":", JSON.stringify(ex), "name:", ex.name);
-        }
+        if (ex && typeof ex === 'object' && 'name' in ex && ex.name === 'org.freedesktop.DBus.Error.UnknownMethod')
+            logDebug("LIST_DOMAIN_SNAPSHOTS action failed for domain", domainPath, ", not supported by libvirt-dbus");
```
cockpituous, line 141: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/snapshot.ts
```diff
@@ -137,12 +137,10 @@ export async function snapshotGetAll({
...
+        else
+            console.warn("LIST_DOMAIN_SNAPSHOTS action failed for domain", domainPath, ":", String(ex));
```
cockpituous, line 143: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/storagePool.ts
```diff
@@ -127,8 +127,7 @@ export async function storagePoolGet({
         }
         store.dispatch(updateOrAddStoragePool(Object.assign({}, dumpxmlParams, props), !!updateOnly));
     } catch (ex) {
-        if (ex instanceof Error)
-            console.warn('GET_STORAGE_POOL action failed for path', objPath, ex.toString());
+        console.warn('GET_STORAGE_POOL action failed for path', objPath, String(ex));
```
cockpituous, line 130: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173902-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/libvirtApi/domain.ts
martinpitt, line 521: This puny human isn't quite sure how that works -- `String()` does [string coercion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#string_coercion), which isn't documented to recognize exceptions or `.message` fields. Exceptions don't have a `.toPrimitive()` function either. So it's a mystery to me *why* this works.

So I asked Copilot about it, and it claims:

> .toString() method: If there's no [Symbol.toPrimitive], JavaScript calls the object's .toString() method.
>     The built-in Error types (Error, TypeError, etc.) have a toString() method that returns a message like "TypeError: message".

But again this isn't really backed up by documentation.

I just found a [precedent on the login page](https://github.com/cockpit-project/cockpit/blob/decde81f18e0fe1ac5383b8de975a8c768e66c2e/pkg/static/login.js#L18). It's very rare though, mostly we use `cockpit.message` or `.toString().

However, I just tried `try { x = undefined.foo } catch (e) { console.log(String(e)) }` in Firefox and node, and it seems to work. I'd still feel better about `ex.toString()` though.

martinpitt:
Fun, I fixed [this flake](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2224-ee84b41e-20250716-173900-ubuntu-2204/log.html#61-2) in https://github.com/cockpit-project/cockpit/pull/22215 -- if anyone feels like reviewing. In the meantime, retrying.

src/libvirtApi/domain.ts
mvollmer: Upps, yes, I should have explained this. I wasn't sure whether `String(x)` is guaranteed to call `x.toString()` when `x` is an object, and Google didn't turn up anything that would explicitly answer exactly this. But MDN had a link to the ECMAScript spec, which did answer this via a couple of levels of abstract operations: Yes, `String(x)` will call `x.toString()` for objects.

So if you are okay with replacing `x.message` with `x.toString()`, then going further and using `String(x)` should also be acceptable.

All Error-like objects in cockpit.js define a `toString` method that returns their `message` field: BasicError, ProcessError, and DBusError.

martinpitt: Yes, I am okay with it, just was a bit unsure.

mvollmer: Cursor says:

# Does String call toString on objects?
_Exported on 7/17/2025 at 09:35:31 GMT+3 from Cursor (1.2.4)_

---

**User**

Does `String(x)` call `x.toString()` when x is an object?

---

**Cursor**

Yes, `String(x)` does call `x.toString()` when `x` is an object, but it's part of a more complex conversion process defined by the ECMAScript specification.

Here's how `String(x)` works when `x` is an object:

1. **First, it checks for `Symbol.toPrimitive`**: If the object has a `Symbol.toPrimitive` method, it calls that method with the hint `"string"`

2. **Then it tries `toString()`**: If there's no `Symbol.toPrimitive` method (or it doesn't return a primitive), it calls the object's `toString()` method

3. **Falls back to `valueOf()`**: If `toString()` doesn't return a primitive value, it tries `valueOf()`

4. **Throws if all fail**: If neither method returns a primitive, it throws a `TypeError`

Here are some examples:

```javascript
// Standard object - calls toString()
const obj = { name: "test" };
String(obj); // "[object Object]" (calls obj.toString())

// Custom toString()
const customObj = {
  toString() {
    return "custom string";
  }
};
String(customObj); // "custom string"

// With Symbol.toPrimitive (takes precedence)
const objWithSymbol = {
  toString() {
    return "from toString";
  },
  [Symbol.toPrimitive](hint) {
    return hint === "string" ? "from toPrimitive" : "other";
  }
};
String(objWithSymbol); // "from toPrimitive"
```

So yes, `toString()` is typically called, but it's not always the first or only method invoked during the string conversion process.

```javascript
// Standard object - calls toString()
const obj = { name: "test" };
String(obj); // "[object Object]" (calls obj.toString())

// Custom toString()
const customObj = {
  toString() {
    return "custom string";
  }
};
String(customObj); // "custom string"

// With Symbol.toPrimitive (takes precedence)
const objWithSymbol = {
  toString() {
    return "from toString";
  },
  [Symbol.toPrimitive](hint) {
    return hint === "string" ? "from toPrimitive" : "other";
  }
};
String(objWithSymbol); // "from toPrimitive"
```

PR #2231, opened 2025-07-18
console: Use icons for header actions

mvollmer:
- [x] #2015 

This saves space at the cost of reduced discoverability...

image


mvollmer:
The antenna for "Disconnect" is especially undiscoverable, I'd say. Maybe we can move it into the settings menu?  That would mean that the collapsed view would have a settings menu (with a cog) that has a single action in it.

mvollmer:
We could also remove the "Disconnect" action from the expanded view. There is no point in disconnecting while you are working with the console. Disconnecting is useful when you are looking at details of a machine and are not interested in the console, but the VNC connection somehow interferes with other viewers.

mvollmer:
@Venefilyn, what do you think of this?  I am quite unsure what to do here, if anything.

I dislike the big "Disconnect" button especially. I don't think it has a lot of utility.

Also, with the fix in https://github.com/cockpit-project/cockpit-machines/pull/2248, a wrapping header is no longer broken, so there is no urgency to make the header content smaller.

Venefilyn:
> The antenna for "Disconnect" is especially undiscoverable, I'd say. Maybe we can move it into the settings menu? That would mean that the collapsed view would have a settings menu (with a cog) that has a single action in it.

Agree, could possibly use this [FA icon](https://fontawesome.com/icons/arrow-right-from-bracket?f=classic&s=solid). Antenna is more of a stream/connection of data. I wouldn't associate it with connect/disconnect a VNC

> We could also remove the "Disconnect" action from the expanded view. There is no point in disconnecting while you are working with the console. Disconnecting is useful when you are looking at details of a machine and are not interested in the console, but the VNC connection somehow interferes with other viewers.

Maybe, but it would shift the UI around even more if we remove it.

mvollmer:
Yeah, let's leave things as they are.

PR #2237, opened 2025-07-22
tests: fix tooltip not showing up in firefox flake testCreateDownloadAnOS

tomasmatus:


tomasmatus:
so this is not a race between checking if tooltip is visible and `mouseleave` event.

`mouseenter` -> checks for tooltip -> `mouseleave` when the dialog is closed. All in the correct order

tomasmatus:
here's a failed test log with verbose output:

```
INFO:bidi.command:← input.performActions({'context': '12884901891', 'actions': [{'id': 'pointer-490', 'type': 'pointer', 'parameters': {'pointerType': 'mouse'}, 'actions': [{'type': 'pointerMove', 'x': 0, 'y': 0, 'origin': {'type': 'element', 'element': {'type': 'node', 'sharedId': '8eb39e81-7ecf-4f84-95d6-b69f2537cea0', 'value': {'nodeType': 1, 'localName': 'button', 'namespaceURI': 'http://www.w3.org/1999/xhtml', 'childNodeCount': 1, 'attributes': {'id': 'create-and-edit', 'aria-disabled': 'true', 'class': 'pf-v6-c-button pf-m-secondary pf-m-aria-disabled pf-m-progress', 'type': 'button', 'data-ouia-component-type': 'PF6/Button', 'data-ouia-safe': 'true', 'data-ouia-component-id': 'OUIA-Generated-Button-secondary-5'}, 'shadowRoot': None}}}}]}]}) [id 491]
> info: mouseenter [object HTMLSpanElement] [object MouseEvent]
> info: mouseleave [object HTMLSpanElement] [object MouseEvent]
INFO:bidi.command:→ {}
INFO:bidi.command:← script.evaluate({'expression': 'ph_wait_cond(() => ph_is_present("#create-and-edit-disabled-tooltip"), 15000, null)', 'awaitPromise': True, 'target': {'context': '12884901891'}}) [id 492]
INFO:bidi.command:← script.evaluate({'expression': '!!window.ph_find', 'awaitPromise': True, 'target': {'context': '12884901891'}}) [id 493]
INFO:bidi.command:→ {'realm': 'b92fa374-744f-43e8-80da-1b08db09bb3f', 'type': 'success', 'result': True}
INFO:bidi.command:← script.evaluate({'expression': 'ph_is_present("#button.alert-link.more-button")', 'awaitPromise': True, 'target': {'context': '12884901891'}}) [id 494]
INFO:bidi.command:→ {'realm': 'b92fa374-744f-43e8-80da-1b08db09bb3f', 'type': 'success', 'result': False}
```

so it is a race between `mouseleave` and check if tooltip is present

tomasmatus:
with this well positioned `time.sleep` I get 100% reproducible flake. I have a feeling firefox might not like that we set the mouse enter event to x:0 y:0 coordinates of the element. [When I set](https://github.com/cockpit-project/cockpit/blob/5ae1ed6bc5a9763d35c2c2169856d4372f830356/test/common/testlib.py#L556) the `pointerMove` bidi event to use x:5, y:5 test starts to pass again (locally)

tomasmatus:
let's try similar approach as in https://github.com/cockpit-project/cockpit-files/pull/1116

tomasmatus:
all of the amplifies test runs went green, this seems to be the solution 🥳 
https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2237-9ad71364-20250723-080122-fedora-42-firefox/log.html#88

test/check-machines-create
```diff
@@ -1496,6 +1496,8 @@ vnc_password= "{vnc_passwd}"
 
                 if self.is_unattended:
                     b.wait_visible("#create-and-edit[aria-disabled=true]")
+                    b.mouse("#create-and-edit", "mouseleave")
+                    b.wait_not_present("#create-and-edit-disabled-tooltip")
```
jelly, line 1502: Two notes:

* Git commits dont support markdown syntax and the log urls expire after X days so it isn't worth having them in the commit message.
* The fix talks about the tooltip not showing, but the added changes hide the tooltip, can you explain why that fixes it? :)

jelly (changes requested):

test/check-machines-create
tomasmatus: > can you explain why that fixes it? :)

honestly? I dont know :) but it made it stable again. Either this somehow works or setting x,y coordinates for the mouse click.

martinpitt: This might be a case of "level" vs. "edge" triggering? Maybe the mouse already was hovering in that button position from a previous test? I.e. the tooltip gets shown when the mouse actually enters the button from outside, not if it was already in that place when the dialog opens.

jelly: Yup, this might be the case. I looked at anything clicking/hovering `create-and-edit` and couldn't find anything.

As this doesn't hurt anything except it being pretty akward lets get this merged with a "HACK" comment

jelly (changes requested):

test/check-machines-create
tomasmatus: sounds good, I added the HACK comment and fixed commit message

tomasmatus:
rhel-8-10 PXE test keeps on failing. The test is flaky and when it selects an interface in the dialog the order of interfaces is "random". When `eth0` is used the test works, when `veth2242eb9e` the VM is not created and fails with `ERROR error creating macvtap interface macvtap0@veth82218750 device or resource busy`

jelly:
> rhel-8-10 PXE test keeps on failing. The test is flaky and when it selects an interface in the dialog the order of interfaces is "random". When `eth0` is used the test works, when `veth2242eb9e` the VM is not created and fails with `ERROR error creating macvtap interface macvtap0@veth82218750 device or resource busy`

The flake fix landed and now it is green! Great success!

Issue #680, opened 2022-04-21
Please add full screen option for VNC


gaionim:
As said from mzamot in [#8392 ](https://github.com/cockpit-project/cockpit/issues/8392#issuecomment-566170769)

Having a full screen button or the option of opening the VNC in a different tab would be really nice

I'll try to do it by myself, but I'm new to react/js :-(

skobyda:
I believe it could be achieved. Our vnc display is basically a single react component <VncConsole />
We can either make the VNC display component to stretch to the 100% height of the browser.
Or we truly make it fullscreen, there seem to be some npm libraries that can lock the whole component to the fullscreen. 

I'm now not sure if it should be a replacement for our already existing "Expand" option.
Or maybe just an alternative, offerend next to "Expand", something like:
![Screenshot from 2022-04-21 14-43-19](https://user-images.githubusercontent.com/42733240/164460524-c3a58efe-1afd-4d45-87a7-6db42be65c0c.png)


@garrett WDYT?

skobyda:
It would also have some fitting icon next to "Fullscreen" just like expand option has an icon.

garrett:
You can already expand VNC in Cockpit-Machines.

![Screenshot 2022-04-21 at 14-25-01 Virtual Machines - garrett@Bolt](https://user-images.githubusercontent.com/10246/164457668-3e18becb-8b28-4c70-bc6c-fc15f9dd5319.png)

Then click expand

![image](https://user-images.githubusercontent.com/10246/164457810-baeab886-cd0d-4c28-98b5-0aa5bdd49e4d.png)


And you'll have a larger view.

![Screenshot 2022-04-21 at 14-26-07 Virtual Machines - garrett@Bolt](https://user-images.githubusercontent.com/10246/164457686-277143a1-de73-49b2-b00f-c0a108efb332.png)

However, the bar is still at the top and the navigation is still there on the side.

I guess you'd prefer a pop-out (with minimal Cockpit UI) or fullscreen option instead? (I'm not sure doing true fullscreen is possible.)

garrett:
I think we'd want to make a minimal version or fullscreen (if possible) replace the current expand functionality.

gaionim:
My naif idea was to add the alternative to open in new tab/window only the single component and delagate to the browser the fullscreen

garrett:
[...]

PR #2229, opened 2025-07-18
consoles: Make expanded console use the whole browser

mvollmer:
- [ ] https://github.com/cockpit-project/cockpit/pull/21670
- [x] Make "expand button" a real link.
- [x] Fix bottom white padding
- [x] Demo
- [x] Fix https://www.youtube.com/watch?v=B58G46LhepA, maybe by constructing the whole thing from scratch instead of re-styling a card.

Demo: https://youtu.be/FsBE4eQo0As

Fixes: #680

Issue #965, opened 2023-02-25
RFE: Option for VNC console in separate browser window


dowdle:
I applaud cockpit-machines for offering an embedded VNC display... but so many other projects are providing a separate window with a full-screen option via noVNC.  I would love to have a pleasingly usable desktop experience.

garrett:
Thanks for your feedback! I agree, we should have this.

Right now, we do have the "expand" button, which gives you a bit more space, but it's not quite the same. It should have even more space. However, middle clicking on the link doesn't work, so you can't open it in a new window.

Also we do have the ability to open up a native app, but it does need improving too (ports and firewall might get in the way or need more configuring, however). One of the blockers is the lack for an ssh option in the `.vv` file. It seems like an omission upstream, so we're waiting on that to be fixed to be able to support it. That's an issue with virt-viewer and it's filed @ https://gitlab.com/virt-viewer/virt-viewer/-/issues/86. More information about the native app support (including this and much more) is on the issue @ https://github.com/cockpit-project/cockpit-machines/issues/553.

Meanwhile, I'll also try to raise the priority of us working on the VNC component, so we can hopefully work on a fix for this sooner.

garrett:
I was considering closing this as a duplicate of https://github.com/cockpit-project/cockpit-machines/issues/680, but I think this captures an idea that the other one doesn't:

- opening the expanded view in a separate tab currently isn't even possible

We should _at least_ start by making middle clicks and control clicks work, to make the VNC viewer work in another tab (or window).

dowdle:
garrett,

One thing you completely side-stepped was my mention of noVNC.  noVNC is packaged in Fedora and is used by several dozen other projects like Proxmox VE, TrueNAS Scale, well... pretty much any virtualization related product that offers a web-based interface.  I'm hoping that there isn't anything in Cockpit's design that would hinder noVNC integration... but I'm not a developer myself.  While the integration might take some effort, all of the features that noVNC comes with would mean, I assume, less effort in having to create those from scratch.

Anyhoo, something to consider.  Having a small region that in the existing window is really a non-starter.

Thanks.

garrett:
@dowdle: Cockpit-Machines actually uses noVNC, as Cockpit-Machines uses PatternFly React Console, which in turn uses noVNC:
https://www.npmjs.com/package/@patternfly/react-console?activeTab=dependencies

When you said "desktop experience", I thought you meant a desktop VNC client.

[...]

PR #2241, opened 2025-07-23
consoles: Enable fullscreen functionality

Venefilyn:
Having a way to fullscreen the VNC console would further improve the
usability of the VNC console. To start with this adds a fullscreen
button and makes sure that the VNC canvas doesn't cause the header to
wrap or disappear when scrollbars appear.

Fixes: #680
Related-to: https://issues.redhat.com/browse/COCKPIT-977
Signed-off-by: Freya Gustavsson <freya@venefilyn.se>

### DIFF
--- a/src/components/vm/consoles/consoles.jsx
+++ b/src/components/vm/consoles/consoles.jsx
@@ -21,7 +21,7 @@ import cockpit from 'cockpit';
 import { StateObject } from './state';
 import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { Card, CardBody, CardHeader, CardTitle } from '@patternfly/react-core/dist/esm/components/Card';
-import { ExpandIcon, CompressIcon } from "@patternfly/react-icons";
+import { ExpandIcon, CompressIcon, UnknownIcon } from "@patternfly/react-icons";
 import { ToggleGroup, ToggleGroupItem } from '@patternfly/react-core/dist/esm/components/ToggleGroup';
 import { Split, SplitItem } from "@patternfly/react-core/dist/esm/layouts/Split/index.js";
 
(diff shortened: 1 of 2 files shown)


mvollmer:
@Venefilyn, I think previously we had decided that we don't want to use the `requestFullscreen` API because it eats ESC. Also, people want a "console in a window" as well: #965. This means removing the Cockpit Shell UI but keep the brower UI.

I have written down some notes in https://issues.redhat.com/browse/COCKPIT-977

As to this PR:

 - Having both "Expand" and "Fullscreen" buttons feels confusing at first. Clicking "Fullscreen" and "Exit fullscreen" includes expanding/compressing the card, but doing that individually is still possible. Is that intended? For example, clicking "Fullscreen" and then "Compress" gives a fullscreen version of the details page...
 - There is some bug with the label: it says "Fullscreen" when in fullscreen and "Exit fullscreen" when not in fullscreen.

src/components/vm/consoles/consoles.jsx
```diff
@@ -227,6 +227,32 @@ export const ConsoleCard = ({ state, vm, config, onAddErrorNotification, isExpan
...
+    if (document.fullscreenEnabled) {
+        actions.push(
+            <Button
+                id="vnc-fullscreen"
+                key="vnc-fullscreen"
+                variant="link"
+                onClick={() => {
+                    const isFullscreened = document.fullscreenElement;
+                    if (!isFullscreened) {
+                        document.getElementById("virtual-machines-page")?.requestFullscreen();
```
mvollmer, line 240: We had previously ruled out using the browser "Fullscreen API": https://github.com/cockpit-project/cockpit-machines/issues/680#issuecomment-1109913793

Are the arguments against it not valid?

I propose to replace this PR here with #2229

mvollmer:
I think this is obsolete, right?

PR #2248, opened 2025-07-29
console: Fix layout bug when header wraps

mvollmer:
Demo of the bug: https://www.youtube.com/watch?v=B58G46LhepA

### DIFF
--- a/src/components/vm/consoles/consoles.css
+++ b/src/components/vm/consoles/consoles.css
@@ -55,9 +55,8 @@
     }
 
     .consoles-card .pf-v6-c-card__body {
-        /* HACK - better done with a Flex instead of a Card? */
-        max-block-size: calc(100% - 77px);  /* 100% - card header height */
         padding-inline-start: 0;
         padding-inline-end: 0;
+        overflow: auto;
     }
 }


mvollmer:
I did not find any other way... If someone figures out a way to do this with CSS only, that would be good.

Venefilyn:
@mvollmer how does the layout wrap bug look like?

mvollmer:
> @mvollmer how does the layout wrap bug look like?

Upps, I thought I had attached the demo, apparently not. Here: https://www.youtube.com/watch?v=B58G46LhepA

mvollmer:
The pixel changes are unexpected, I'll have a look. The card header is taller now.

mvollmer:
> The pixel changes are unexpected, I'll have a look. The card header is taller now.

In fact, the card header should remain at 77px tall, but seems to be 85px now...

Venefilyn:
@mvollmer I had this change in my PR to fix this. Is that enough or is there an edge case I don't get? See the CSS file with overflow 

https://github.com/cockpit-project/cockpit-machines/pull/2241/files

mvollmer:
> @mvollmer I had this change in my PR to fix this. Is that enough or is there an edge case I don't get? See the CSS file with overflow
> 
> https://github.com/cockpit-project/cockpit-machines/pull/2241/files

The layout bug happens with the expanded console, your CSS addition only applies to the compressed one in the Details page. So it unfortunately doesn't help here.

Can you demonstrate the problem that your CSS is fixing?

Venefilyn:
> > @mvollmer I had this change in my PR to fix this. Is that enough or is there an edge case I don't get? See the CSS file with overflow
> > [#2241 (files)](https://github.com/cockpit-project/cockpit-machines/pull/2241/files)
> 
> The layout bug happens with the expanded console, your CSS addition only applies to the compressed one in the Details page. So it unfortunately doesn't help here.
> 
> Can you demonstrate the problem that your CSS is fixing?

Ah you're right. Somehow I tested it and it worked but I didn't test it well enough cause it really wasn't fixed lmao. 

However! It does fix the solution if applied to the right element

Video of the fix in action, I applied `overflow: auto` to `.pf-v6-c-card__body`. To apply this to console I would add a new class through `className` to the <CardBody> that would add `overflow: auto` -- if it doesn't exist already in a modifier somewhere..

[Kooha-2025-07-30-13-50-00.webm](https://github.com/user-attachments/assets/0d9465e7-ab3d-4a90-a984-2f3a02604309)

mvollmer:
> However! It does fix the solution if applied to the right element

Uff, that's a big relief!  I don't understand why it works and would have never figured this out!

(Does it work because now the body is a ["Block formatting context"](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Block_formatting_context)?)

mvollmer:
> (Does it work because now the body is a ["Block formatting context"](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Block_formatting_context)?)

Maybe, maybe not! :-)

Other ways to get a new BFC like `display: flow-root` do not fix this bug, but all `overflow` values that are documented to start a BFC do fix it, and all values that don't start a BFC do not fix it.

For example, `overflow: scroll` also works, without adding any additional scrollbars that I could find.

Magic.

mvollmer:
@Venefilyn, PTAL.  Also, of course, please close this and open a PR in your name if you prefer.

PR #2252, opened 2025-07-30
test: fix unstable testCreatePXE

tomasmatus:
Set interface as the dropdown order is different across reboots. This test was flaky on rhel-8-10/ws-container, see log. As the selection of interfaces was random the test used to fail when interface macvtap0@veth2242eb9e got selected with
`ERROR error creating macvtap interface macvtap0@veth82218750 device or resource busy` libvirt error.

https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2237-171d7320-20250730-113257-rhel-8-10-ws-container/log.html

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -537,9 +537,20 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
 
         m.execute("virsh net-destroy default; virsh net-undefine default")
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Dekuji! This seems to work, remaining failures are unrelated (retried one round).

martinpitt (approved):

PR #2268, opened 2025-08-08
consoles: Make a non-shared connection when asking remote to resize

mvollmer:
It doesn't feel like a good idea to let multiple clients each try to resize the remote end.

This also gives us a easy way to make non-shared connections in our tests, which we will use in the future.

- [x] Fix the bug that keeps vnc connections open until the session ends. I think the novnc websocket connection is correctly closed, must be something in the bridge. Was in ws: https://github.com/cockpit-project/cockpit/pull/22310
- [x] https://github.com/cockpit-project/cockpit/pull/22376
- [x] Tests
- [x] Wait for #22376 to be released into at least one of our images.

-----

### Machines: Cockpit can make exclusive VNC connections

A VNC viewer can tell the VNC server whether or not it wants exclusive access to it, or whether it is okay that multiple viewers display and interact with the same console. Until now, Cockpit has always allowed shared access when it connects.

Now, Cockpit will request exclusive access when it is in the "Remote resizing" mode, to avoid having multiple viewers make conflicting resize requests.

The feature requires the web server from Cockpit 346.

### DIFF
--- a/src/components/vm/consoles/vnc.tsx
+++ b/src/components/vm/consoles/vnc.tsx
@@ -505,6 +505,29 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
             />
         );
 
+        let scaleViewport = true;
+        let resizeSession = false;
+        if (isExpanded) {
+            scaleViewport = (state.sizeMode == "local");
+            resizeSession = (state.sizeMode == "remote");
+        }
+
+        // Older versions of Cockpit would erroneously never close the
+        // remote end of a external channel when the local end was
+        // closed by the browser.  When this bug is present, all
+        // connections to the VNC server that we make here will stay
+        // open forever.  This is especially bad with non-shared
+        // connections: As long as it is open, no other connections
+        // can be made, not even other non-shared ones.
+        //
+        // Thus, we only make non-shared connections when this bug is
+        // fixed, which cockpit-ws will tell us via the transport
+        // capabilities.
+
+        const caps = cockpit.transport.options.capabilities;
+        const bug_is_fixed = Array.isArray(caps) && caps.includes("websocket-channel-close-fix");
+        const shared = !(resizeSession && bug_is_fixed);
+
         return (
             <>
                 { state.connected
@@ -513,15 +536,15 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
                           port={window.location.port || (encrypt ? '443' : '80')}
                           path={path}
                           encrypt={encrypt}
-                          shared
                           onConnected={this.onConnected}
                           onDisconnected={this.onDisconnected}
                           getCredentials={this.getCredentials}
                           onInitFailed={this.onInitFailed}
                           onSecurityFailure={this.onSecurityFailure}
                           consoleContainerId={isExpanded ? "vnc-display-container-expanded" : "vnc-display-container-minimized"}
-                          scaleViewport={!isExpanded || state.sizeMode == "local"}
-                          resizeSession={!!isExpanded && state.sizeMode == "remote"}
+                          scaleViewport={scaleViewport}
+                          resizeSession={resizeSession}
+                          shared={shared}
                     />
                     : <div className="vm-console-vnc">
                         <EmptyState>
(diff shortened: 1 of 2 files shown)


mvollmer:
fedora-43 now has cockpit 346, which has https://github.com/cockpit-project/cockpit/pull/22376.

test/check-machines-consoles
```diff
@@ -685,6 +685,37 @@ fullscreen=0
...
+        b.wait_visible(".vm-console-vnc canvas")
+
+        # But if we switch the detached console into "Remote
+        # resizing", the embedded one will be disconnected again
+        # because the detached now uses a non-shared connection. (On
+        # images that have the necessary cockpit-ws fix.)
+
+        b2.select_PF("#vm-console-vnc-scaling", "Remote resizing")
+
+        cockpit_version = int(m.execute('(/usr/libexec/cockpit-ws --version'
```
mvollmer: We can simplify this by finding the list of images that don't have 346. That's probably just rhel-9 and centos-9, and those will never get it.

mvollmer:
I think tumbleweed and ubuntu-stable will get 346 pretty soon, no? We could wait for them, then we have a stable list of images that don't have the fix, I think.

martinpitt:
@mvollmer u-stable will never get 346. That's only in backports (once it gets approved), but we need to support the shipped version, too.

mvollmer:
Ubuntu is hard to get green in general. I blame the [Browser.open bidi timeout.](https://github.com/cockpit-project/cockpit/pull/22438).

mvollmer:
@martinpitt, what do you think? Should we check the cockpit version to figure out what cockpit is going to do, or have a list of OSes?

I think the list is better and it is probably very stable once opensuse-tumbleweed has Cockpit 346.

src/components/vm/consoles/vnc.tsx
```diff
@@ -505,6 +505,29 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
...
+        // connections: As long as it is open, no other connections
+        // can be made, not even other non-shared ones.
+        //
+        // Thus, we only make non-shared connections when this bug is
+        // fixed, which cockpit-ws will tell us via the transport
+        // capabilities.
+
+        const caps = cockpit.transport.options.capabilities;
+        const bug_is_fixed = Array.isArray(caps) && caps.includes("websocket-channel-close-fix");
+        const shared = !(resizeSession && bug_is_fixed);
```
martinpitt, line 529: That's  a good idea in general, but what happens to an existing connection? If it starts out with "resize server" and you switch it to e.g. "local scaling", it won't automatically become shared. Worse, if you start with "local scaling" and change to "resize server", it will stay private. Or does that reconnect internally?

test/check-machines-consoles
martinpitt: Yes, agreed to keep a shrinking static OS l ist. This is quite a lot of error prone version checking here. Or if you prefer a runtime check, perhaps check `b.eval_js('cockpit.transport.options.capabilities')`?

test/check-machines-consoles
```diff
@@ -699,12 +699,7 @@ fullscreen=0
 
         b2.select_PF("#vm-console-vnc-scaling", "Remote resizing")
 
-        cockpit_version = int(m.execute('(/usr/libexec/cockpit-ws --version'
-                                        ' || /usr/lib/cockpit/cockpit-ws --version'
-                                        ' || echo "Version: 346")'
-                                        ' | grep ^Version: | cut -d" " -f2'))
-
-        if cockpit_version >= 346:
+        if not m.image.startswith(("rhel-9", "centos-9", "ubuntu-", "debian-trixie", "opensuse-")):
```
martinpitt: ah yes, this looks better. But for clarity, perhaps spell out "opensuse-tumbleweed", just in case we get SLES at some point. We'll have to do a lockstep machines/bots update when tumbleweed updates, but we gain robustness.

src/components/vm/consoles/vnc.tsx
mvollmer: It will reconnect.  Every change to the scale/resize settings will lead to a reconnection, because of the way the React wrapper is written. (That wrapper is now in our control, so we could improve that.)

test/check-machines-consoles
mvollmer: Roger.

PR #2275, opened 2025-08-11
force ASCII output for `virt-host-validate`

Moroshima:
I use `zh_CN.UTF-8` as my computer's locale, which makes the output of `virt-host-validate` look like

```
QEMU: Checking for hardware virtualization                                 : 通过
```

instead of

```
QEMU: Checking for hardware virtualization                                 : PASS
```

This prevents cockpit-machine from passing the hardware virtualization support check. So we should set `LANG=C.UTF-8` to ensure `virt-host-validate` outputs in ASCII, making the 'grep' check for hardware virtualization reliable regardless of system locale.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -103,7 +103,7 @@ export const App = () => {
         (async () => {
             try {
                 const hardwareVirtCheck = await cockpit.script(
-                    "virt-host-validate qemu | grep 'Checking for hardware virtualization'");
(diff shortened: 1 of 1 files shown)


mvollmer:
Thanks!

(Technically, the commit message is slightly incorrect when saying "ASCII". We are now forcing the "C" locale, which is the bit that causes the output to be in English. It's not really the character encoding that matters here, it was UTF-8 all the time.  :-)

PR #2284, opened 2025-08-19
test: Adjust for new debian-trixie image

mvollmer:
- We need the same AppArmor hack as in debian-testing in TestMachinesSnapshots.

 - The VSock hotplug bug is also in debian-trixie.


mvollmer:
We could make a naughty for the VSock hotplug bug.

test/check-machines-settings
```diff
@@ -955,9 +955,9 @@ class TestMachinesSettings(machineslib.VirtualMachinesCase):
         b.click("#vm-subVmTest1-system-run")
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
 
-        # Because of bug in debian-testing, hotplug of vsock device fails
+        # Because of bug in Debian, hotplug of vsock device fails
         # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1033872
-        if m.image not in ["debian-testing", "debian-stable"]:
+        if not m.image.startswith("debian-"):
```
allisonkarlitskaya: I think I'd slightly prefer you to add `debian-trixie` here... one day this might get fixed and we'll end up having dead code on all Debian versions forever and ever...

mvollmer: yes, I was thinking the same...

allisonkarlitskaya (approved):

PR #2292, opened 2025-08-25
api: Preserve passwords when updating domain XML

mvollmer:
Add VIR_DOMAIN_XML_SECURE flag to GetXMLDesc calls that modify XML before calling DomainDefineXML to ensure sensitive data like passwords are preserved during domain updates.

🤖 Generated with [Claude Code](https://claude.ai/code) for $2.04

### DIFF
--- a/src/libvirtApi/domain.ts
+++ b/src/libvirtApi/domain.ts
@@ -339,7 +339,7 @@ export async function domainChangeBootOrder({
     connectionName: ConnectionName,
     devices: BootOrderDevice[],
 }): Promise<void> {
-    const [domXml] = await call<[string]>(connectionName, objPath, 'org.libvirt.Domain', 'GetXMLDesc', [Enum.VIR_DOMAIN_XML_INACTIVE], { timeout, type: 'u' });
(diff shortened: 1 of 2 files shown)


mvollmer:
The alternative is to replace all these 5 places with invocations of virt-xml. I was going to do that any way, but now I know why and where to start.

martinpitt (approved):
Thanks! Did claude write the test as well?

mvollmer:
> Thanks! Did claude write the test as well?

The skeleton, but it didn't work.  Can't blame Claude, it had no way to run the test.

(Claude used a password with nine characters but VNC only allows eight! And it got the dialog manipulations wrong, and some details about how virsh dumpxml quotes things.)

martinpitt:
>  Can't blame Claude, it had no way to run the test.

FTR, that works quite well, you just have to tell it how. I usually do something like

> Check your work with `toolbox run --container cockpit test/check-foo TestBar $RUNC` 

(and run a `./build -wrc` on the side). I haven't literally tested that with cockpit, but with `pytest` and `ansible` for system-roles, but I see no reason why it wouldn't work with Cockpit?

PR #2295, opened 2025-08-28
build.js: don't copy assets when the build is cancelled

jelly:
When the build is cancelled in watch mode, `dist` might not exist so copying to it will always fail and show a long useless traceback. Unfortunately esbuild errors don't give much information so we can't match the exact cancelled error.


martinpitt (approved):
Thanks!

PR #2303, opened 2025-09-02
NICs: Show all network IPs

Venefilyn:
When browsing a VM's network interfaces you would not be able to see
when a VM had multiple different IP addresses of the same type within
one interface. AKA having a secondary IP address.

Fixing this was simple as we already get the IP addresses from libvirt
and can just iterate over correctly. Instead of assuming interfaces
can only have one IPv4 and one IPv6 address we now make no distinction
on how many you can have per interface.

Fixes: https://issues.redhat.com/browse/RHEL-89015
Signed-off-by: Freya Gustavsson <freya@venefilyn.se>


src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -250,16 +250,28 @@ interface VmNetworkTabProps {
     onAddErrorNotification: (notification: Notification) => void;
 }
 
-interface IpAddress {
+enum IpAddressVersion {
+    v4 = 0,
+    v6 = 0
```
mvollmer: Both are `= 0`?

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -250,16 +250,28 @@ interface VmNetworkTabProps {
...
+}
+
+interface IpAddressType<T extends string> {
+    type: T,
```
mvollmer: I think `type: "inet" | "inet6";` should be enough. No need for the generic. Furthermore, we can simply use `type: IpAddressVersion`, no?

mvollmer:
Hah, I was afraid it's more complicated than this! :-) I left some premature comments, just ignore them if it's too early.

src/components/vm/nics/vmNicsCard.tsx
Venefilyn: Fixed!

src/components/vm/nics/vmNicsCard.tsx
Venefilyn: Fixed it now!

Venefilyn:
> Hah, I was afraid it's more complicated than this! :-) I left some premature comments, just ignore them if it's too early.

@mvollmer It pretty much works right now! Updated the original comment. I also figured out you can add secondary IPs fairly easily using `ip` too. Gonna see what tests failures are like

```bash
alpine-127-0-0-2-2201:~# ip address add 192.168.122.242/24 dev eth0
alpine-127-0-0-2-2201:~#
```

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -250,16 +250,26 @@ interface VmNetworkTabProps {
     onAddErrorNotification: (notification: Notification) => void;
 }
 
+enum IpAddressVersion {
+    v4,
```
mvollmer: Sine the values of these enums are defined by the libvirt API, I think we should spell them out here.

`v4 = 0`, `v6 = 1`.

Can we even use TypeScript enums?

Venefilyn: Ahh thought it was just internal assignment. Yeah we can do `v4 = 0` etc.

test/check-machines-nics
```diff
@@ -140,35 +144,40 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
...
+            # With medium layout this variable width of the columns makes pixel tests references of
+            # the table different each test run
+            skip_layouts=["medium", "rtl"],
+        )
 
         # Start vm and wait until kernel is booted
         testlib.wait(lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.122. | wc -l"),
                      delay=3)
         b.reload()
+        testlib.sit()
```
mvollmer: Honk!

test/check-machines-nics
```diff
@@ -60,21 +60,25 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         b.wait_in_text("#vm-subVmTest1-network-1-source", "default")
         b.wait_in_text('#vm-subVmTest1-networks td[data-label="Source"]', "TAP device")
         b.wait_in_text("#vm-subVmTest1-network-1-tapdevice", "vnet0")
-        b.wait_in_text("#vm-subVmTest1-network-1-ipv4-address", "192.168.122.")
+        b.wait_in_text("#vm-subVmTest1-network-1-ipv4-address-0", "192.168.122.")
```
mvollmer, line 63: Do we actually test a second IP now? We should.

Executing commands inside the guest is tricky. We could fish it's IP out of the page and SSH into it. Or use the serial console.

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -440,49 +447,59 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
             {
                 name: _("IP address"),
                 value: (network, networkId) => {
-                    const ips = this.state.ips.filter(ip => ip.mac === network.mac);
+                    const ipInterfaces = this.state.ips.filter(ip => ip.mac === network.mac);
 
-                    const ip: { inet?: string, inet6?: string } = {};
+                    const inetIps: IpAddress["ip"][] = [];
```
mvollmer: I think `string` instead of `IpAddress["ip"]` is just fine here, and less confusing.

test/check-machines-nics
Venefilyn: Woops, fixed

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -440,49 +447,59 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
-                        // Various sources (agent,lease,arp) may have only partial information about interface
-                        // (e.g. one source may only know ipv4 address, the other may only know ipv6 address)
-                        // Therefore, look at each source to get all available information
-                        if (!ip.inet && a.ip.inet)
-                            ip.inet = a.ip.inet;
-                        if (!ip.inet6 && a.ip.inet6)
-                            ip.inet6 = a.ip.inet6;
+                    ipInterfaces.sort((a) => a.source === "agent" ? -1 : 1).forEach(ipInterface => {
+                        ipInterface.ips.forEach(inet => {
+                            // Separate out IPv4 and IPv6 and don't add duplicates
```
mvollmer: (grr, github ate a really nice comment...)

I still get duplicates of sorts:

image

I think we should deduplicate based on the address alone, without the prefix tacked onto it.  Also, we should prefer the "lease" source over "arp", i.e., sort them agent -> lease -> arp.

mvollmer:
Nice!

The guest needs to run the guest agent in order to see multiple IPs reliably.  That's fine.
There is probably a way to get a second address into the arp cache, but I don't know it.

@Venefilyn, if you feel up for it, we could clean up the weird return type of domainInterfaceAddresses at the same time.  All the decoding and sorting and deduplication could be done by that function. But we can leave it as it as as well.

src/components/vm/nics/vmNicsCard.tsx
Venefilyn: Do we want to remove the subnet mask suffix completely? I mean it isn't super relevant but it would remove information still.

And to double check, you want the list to be top to bottom from agent, lease, arp?

mvollmer: > Do we want to remove the subnet mask suffix completely? I mean it isn't super relevant but it would remove information still.

Good question. `virsh domifadr` outputs it, so I think we should, too:

```
# virsh domifaddr foo
 Name       MAC address          Protocol     Address
-------------------------------------------------------------------------------
 vnet0      52:54:00:cb:01:48    ipv4         192.168.122.44/24
```

mvollmer: > And to double check, you want the list to be top to bottom from agent, lease, arp?

No, this should be the order to resolve duplicates. If the same IP comes from agent, lease, and arp, agent should win. (The code does this already with the `sort` call). When it comes from lease and arp, lease should win. (That is missing.)
The reason is that the agent has the best information and arp has worst (i.e., no prefix).

Venefilyn: Ack so if we have the following then we'd only end up adding `192.168.122.44/24` to the list. So check for duplicates for just the IP address without submask? 
```
agent:
  - 192.168.122.44/24
lease:
  - 192.168.122.44/0
```

mvollmer: Correct.

But also, in reality, both agent and lease will give us a good prefix, and it is arp that will report a bogus 0.

```
agent:
  - 192.168.122.44/24
lease:
  - 192.168.122.44/24
arp:
 -  192.168.122.44/0
```

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -440,49 +453,71 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
+                    const inet6Ips: string[] = [];
+
+                    // Setup a preferred order of IP/SUBNET-MASK we want to add.
+                    // Helps ensure that the IP subnet mask are taken from the right source.
+                    // agent > lease > arp
+                    ipInterfaces.sort((a, b) => {
+                        if (a.source === "agent")
+                            return -1;
+                        else if (a.source === "lease" && b.source !== "agent")
+                            return -1;
```
cockpituous, line 468: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2303-0f9a39db-20250905-160146-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -440,49 +453,71 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
+                        return 1;
+                    }).forEach(ipInterface => {
+                        ipInterface.ips.forEach(inet => {
+                            // If match was not found we default to `inet.ip`
+                            const ip = inet.ip.match(regexpIpExcludeSubnetMask)?.[0] || inet.ip;
```
cockpituous, line 473: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2303-0f9a39db-20250905-160146-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -440,49 +453,71 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
+
+                            // Separate out IPv4 and IPv6 and don't add duplicate IPs.
+                            if (inet.type === IpAddressVersion.v4 && !inetIps.some(inetIp => inetIp.startsWith(ip))) {
+                                inetIps.push(inet.ip);
+                            } else if (inet.type === IpAddressVersion.v6 && !inet6Ips.some(inet6Ip => inet6Ip.startsWith(ip))) {
```
cockpituous, line 478: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2303-0f9a39db-20250905-160146-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

test/check-machines-nics
```diff
@@ -140,26 +144,53 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
...
+        # This differs depending on local and cloud testing
+        iface = "$(ip -br address show | grep '192.168.122.' | cut -d ' ' -f 1)"
+        # Add a secondary IP which should show
+        secondary_ip = "192.168.123.100/24"
+        self.machine.execute(
+            "virsh qemu-agent-command subVmTest1 "
+            '\'{"execute": "guest-exec", "arguments":'
+            '{ "path": "ip", "arg": [ "address","add","' + secondary_ip + '", "dev", "' + iface + '" ], '
+            '"capture-output": true }}\''
+        )
```
Venefilyn: Can't run tests locally without rebooting atm so won't continue today. But this did work locally but doesn't seem to work on the host. Had a different solution before that didn't work either, and now with this cleaner approach I also struggle to get it to add the IP.

```bash
cockpit-machines on  nics/capture-all-ips [!] via  v22.18.0 via 🐍 v3.13.7 
❯ ip -br address show | grep '10.35.30.' | cut -d ' ' -f 1
enp5s0
```

I suspect that the interface is correct but not sure why this doesn't work whatsoever. Gonna give it another try by waiting for the IP to appear

src/components/vm/nics/vmNicsCard.tsx
cockpituous, line 478: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2303-3a2d4f06-20250905-195746-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -440,49 +453,71 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
+
+                    // Setup a preferred order of IP/SUBNET-MASK we want to add.
+                    // Helps ensure that the IP subnet mask are taken from the right source.
+                    // agent > lease > arp
+                    ipInterfaces.sort((a, b) => {
+                        if (a.source === "agent")
+                            return -1;
+                        else if (a.source === "lease" && b.source !== "agent")
+                            return -1;
+                        return 1;
```
mvollmer: It's not obvious that this is a well defined comparison function. For example, it violates the "Reflexive" property and returns `1` or `-1` for `a == b`. It's also very hard to change if we ever want to do that.

A better approach is to first map the source to a number, and then compare the numbers.

```
const source_order = {
   "agent": 1,
   "lease": 2,
   "arp": 3,
};

...sort((a, b) => source_order(a)  - source_order(b))
```

src/components/vm/nics/vmNicsCard.tsx
mvollmer: I find it strange that we have to use a regexp to take a string apart that we have created a few lines earlier.  It is always better to avoid custom parsing, and parsing with custom regexps specifically. What about keeping ip and prefix separate until we are done collating them?

We could add a `prefix` field to `IpAddress` and work with `IpAddress` instances until the very end.

```
const inetIps: IpAddress[] = [];
...

<DescriptionListDescription id={`${id}-network-${networkId}-ipv4-address-${index}`}>
  {ip.ip} / {ip.prefix}
</DescriptionListDescription>
```

test/check-machines-nics
mvollmer: Oh, I didn't know the agent can execute arbitrary code... neat!  And luckily we explicitly install the agent on the Alpine images!  (I forgot we did that, tbh.)

We should probably wrap this up and add it as a function to machineslib.py.

test/check-machines-nics
```diff
@@ -140,26 +144,56 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
...
+
+        testlib.wait(
+            lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.123.100 | wc -l"), delay=3
+        )
+        # Add a secondary IP with different subnet mask, it should not show given the IP itself
```
mvollmer: Does it make sense to add the same ip with different netmasks, from a networking point of view?

Venefilyn: Don't think so. At least like we do it here. Wanna remove it?

mvollmer: > Wanna remove it?

Only if it causes trouble, I'd say. If this works for testing our de-duplicating, then I think it's fine to keep it.

test/check-machines-nics
```diff
@@ -140,26 +144,56 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
...
+            skip_layouts=["medium", "rtl"],
+        )
 
         # Start vm and wait until kernel is booted
         testlib.wait(lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.122. | wc -l"),
                      delay=3)
+
+        # To add more IPs we need to know which interface to add them to.
+        # This differs depending on local and cloud testing
+        iface = "$(ip -br address show | grep '192.168.122.' | cut -d ' ' -f 1)"
```
mvollmer: This needs to actually run, no? I don't think guest-exec does shell substitution...

test/check-machines-nics
```diff
@@ -140,26 +144,56 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
...
+        secondary_ip = "192.168.123.100/24"
+        self.machine.execute(
+            "virsh qemu-agent-command subVmTest1 "
+            '\'{"execute": "guest-exec", "arguments":'
+            '{ "path": "ip", "arg": [ "address","add","' + secondary_ip + '", "dev", "' + iface + '" ], '
+            '"capture-output": true }}\''
+        )
+
+        testlib.wait(
+            lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.123.100 | wc -l"), delay=3
```
mvollmer: Only the agent will report the second address, so I think this needs to use `--source agent`. (`virsh domifaddr` only uses one source, and defaults to "lease", I think.)

test/check-machines-nics
mvollmer: It depends on the order that the agent reports these addresses, no?  If the agent reports "192.168.123.100/0" first, we will show it instead of "192.168.123.100/24".

Thinking about this, the correct thing would actually be to show both "192.168.123.100/0" _and_ 
"192.168.123.100/24", since they come from the same source.

So we could change the logic to see which sources we have, and then pick only one source and report what it says, unfiltered. I think this is the correct approach, and actually simpler, no?  Sorry for not seeing this earlier, I was too focused on incrementally changing the existing code. What do you think about this?

mvollmer: So, hmm, let's just remove this adding of a the same IP twice with different netmasks.  The de-duplication is already tested since we get agent, lease, and arp reports in the test.

test/check-machines-nics
mvollmer: I got the test to pass with these changes:

```diff
diff --git a/test/check-machines-nics b/test/check-machines-nics
index 803d27667..1f53e0b41 100755
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -169,9 +169,7 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         testlib.wait(lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.122. | wc -l"),
                      delay=3)
 
-        # To add more IPs we need to know which interface to add them to.
-        # This differs depending on local and cloud testing
-        iface = "$(ip -br address show | grep '192.168.122.' | cut -d ' ' -f 1)"
+        iface = "eth0"
         # Add a secondary IP which should show
         secondary_ip = "192.168.123.100/24"
         self.machine.execute(
@@ -182,8 +180,9 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         )
 
         testlib.wait(
-            lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.123.100 | wc -l"), delay=3
+            lambda: "1" in self.machine.execute("virsh domifaddr --source agent subVmTest1 | grep 192.168.123.100 | wc -l"), delay=3
         )
+
         # Add a secondary IP with different subnet mask, it should not show given the IP itself
         # is already added
         secondary_ip_hidden = "192.168.123.100/0"

```

I don't see why the interface would have different names locally and in CI. It's determined by the kernel/OS inside the guest, no?

test/check-machines-nics
Venefilyn: Ack will do. Gonna see about getting the tests to work though.. currently been struggling to get things to work correctly

Venefilyn: Currently working on something like this but @mvollmer mentioned in the meeting today that he got it to work. Just push yours up instead and we go with that instead of mine below :pray: 

```diff
diff --git a/test/check-machines-nics b/test/check-machines-nics
index 803d2766..006435b8 100755
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -171,15 +171,33 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
 
         # To add more IPs we need to know which interface to add them to.
         # This differs depending on local and cloud testing
-        iface = "$(ip -br address show | grep '192.168.122.' | cut -d ' ' -f 1)"
+        # iface = "$(ip -br address show | grep '192.168.122.' | cut -d ' ' -f 1)"
+        qemu_pid_response = json.loads(
+            self.machine.execute(
+                "virsh qemu-agent-command subVmTest1 "
+                '\'{"execute": "guest-exec", "arguments": '
+                '{ "path": "ip", "arg": [ "-br", "address", "show" ], '
+                '"capture-output": true }}\''
+            )
+        )
+        iface = self.machine.execute(
+            "virsh qemu-agent-command subVmTest1 "
+            '\'{"execute": "guest-exec-status", "arguments": '
+            '{ "pid": ' + str(qemu_pid_response["return"]["pid"]) + " ], "
+            '"capture-output": true }}\' '
+            # Parse JSON response, get the out-data if it exists, decode base64 and get the interface name.
+            "| jq -r '.return.\"out-data\"' | base64 --decode | grep '192.168.122.' | cut -d ' ' -f 1)"
+        )
+
         # Add a secondary IP which should show
         secondary_ip = "192.168.123.100/24"
         self.machine.execute(
             "virsh qemu-agent-command subVmTest1 "
-            '\'{"execute": "guest-exec", "arguments":'
+            '\'{"execute": "guest-exec", "arguments": '
             '{ "path": "ip", "arg": [ "address","add","' + secondary_ip + '", "dev", "' + iface + '" ], '
             '"capture-output": true }}\''
         )
+        testlib.sit()
 
         testlib.wait(
             lambda: "1" in self.machine.execute("virsh domifaddr subVmTest1 | grep 192.168.123.100 | wc -l"), delay=3
@@ -189,7 +207,7 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         secondary_ip_hidden = "192.168.123.100/0"
         self.machine.execute(
             "virsh qemu-agent-command subVmTest1 "
-            '\'{"execute": "guest-exec", "arguments":'
+            '\'{"execute": "guest-exec", "arguments": '
             '{ "path": "ip", "arg": [ "address","add","' + secondary_ip_hidden + '", "dev", "' + iface + '" ], '
             '"capture-output": true }}\''
         )
```

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -328,20 +350,16 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
                                         iface[2].forEach(ifAddress => {
-                                            // type == 0 -> ipv4
-                                            // type == 1 -> ipv6
-                                            const type = ifAddress[0] == 0 ? 'inet' : 'inet6';
-
-                                            if (ifAddress.length && ifAddress[0] <= 1) {
+                                            if (ifAddress.length && ifAddress[0] in IpAddressVersion) {
                                                 // 0 cell -> type, 1 cell -> address, 2 cell -> prefix
-                                                address.ip[type] = ifAddress[1] + '/' + ifAddress[2];
+                                                address.ips.push({ type: ifAddress[0], ip: ifAddress[1] + '/' + ifAddress[2] });
```
mvollmer: I meant: we shouldn't combine `ifAddress[1]` and `ifAddress[2]` here into a single string, just to parse it apart later with the regexp.

```
address.ips.push({ type: ifAddress[0], ip: ifAddress[1], prefix: ifAddress[2] });
```

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -440,49 +452,67 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
...
+                        agent: 1,
+                        lease: 2,
+                        arp: 3
+                    };
+                    ipInterfaces.sort((a, b) => source_order[a.source] - source_order[b.source]).forEach(ipInterface => {
+                        ipInterface.ips.forEach(inet => {
+                            // Separate out IPv4 and IPv6 and don't add duplicate IPs.
+                            if (inet.type === IpAddressVersion.v4 && !inetIps.some(inetIp => inetIp.ip == inet.ip)) {
+                                inetIps.push(inet);
+                            } else if (inet.type === IpAddressVersion.v6 && !inet6Ips.some(inet6Ip => inet6Ip.ip == inet.ip)) {
```
cockpituous, line 473: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2303-6a39774f-20250912-115612-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
Please don't just hit the button here. This needs squashing and/or rebasing.

PR #2305, opened 2025-09-03
networks: Fix spacing of helper text in "Create" dialog

mvollmer:
The "IP configuration" FormGroup had the isStack property set, which in turn would change the top margin of FormHelperText inside that FormGroup from 4 pixels to -4 pixels.

The isStack property does not seem to have any other effect, so let's remove it.

Fixes: #1164


mvollmer:
Before: 

image

After:

image

mvollmer:
The `isStack` property increases the spacing between the fields inside "IP configuration" form group, but I'd say that's undesirable so removing it is a double win.

PR #2306, opened 2025-09-03
test: Check temporary "UI only" VMs

mvollmer:
These special VM objects only exist temporarily while a new, real VM is being defined. In order to test them, we add support for artificially creating permanent versions of them during automated testing.


mvollmer:
Heh, we could have a pixel test now!

src/mock.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import type { UIVM } from './types';
+
+import store from './store.js';
+import { addUiVm } from './actions/store-actions.js';
+
+function testMockAddUiVm(name: string) {
```
martinpitt, line 25: But what does this prove? The integration test checks this mock, not the actual "init" VM. That should stay around for more than a second, so can't we just test the real thing?

mvollmer: This tests the rendering of this transitional page, nothing else.

mvollmer: > That should stay around for more than a second, so can't we just test the real thing?

That is the question. I was assuming "no". It stay around for the time that `virt-install` needs (about 0.5 seconds locally) plus some initial downloading for "Download an OS". But I don't think we actually have any test that actually downloads, no?

PR #2307, opened 2025-09-03
details: Simplify empty state texts

mvollmer:
Avoid the repetitive and redundant "for this VM" and use active voice for explaining snapshots.

Fixes: #764

### DIFF
--- a/src/components/vm/snapshots/vmSnapshotsCard.tsx
+++ b/src/components/vm/snapshots/vmSnapshotsCard.tsx
@@ -267,8 +267,8 @@ export class VmSnapshotsCard extends React.Component<VmSnapshotsCardProps> {
             <ListingTable aria-label={`VM ${vm.name} Snapshots Cards`}
                           gridBreakPoint='grid-lg'
                           variant="compact"
-                          emptyCaption={_("No snapshots defined for this VM")}
-                          emptyCaptionDetail={_("Previously taken snapshots allow you to revert to an earlier state if something goes wrong")}
+                          emptyCaption={_("No snapshots")}
+                          emptyCaptionDetail={_("Snapshots allow you to revert to an earlier state")}
                           columns={columnTitles}
                           rows={rows} />
         );
--- a/src/components/vm/disks/vmDisksCard.tsx
+++ b/src/components/vm/disks/vmDisksCard.tsx
@@ -291,7 +291,7 @@ export const VmDisksCard = ({
     return (
         <ListingTable variant='compact'
                 gridBreakPoint='grid-lg'
-                emptyCaption={_("No disks defined for this VM")}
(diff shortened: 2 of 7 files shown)


src/components/vm/snapshots/vmSnapshotsCard.tsx
```diff
@@ -267,8 +267,8 @@ export class VmSnapshotsCard extends React.Component<VmSnapshotsCardProps> {
             <ListingTable aria-label={`VM ${vm.name} Snapshots Cards`}
                           gridBreakPoint='grid-lg'
                           variant="compact"
-                          emptyCaption={_("No snapshots defined for this VM")}
-                          emptyCaptionDetail={_("Previously taken snapshots allow you to revert to an earlier state if something goes wrong")}
+                          emptyCaption={_("No snapshots")}
+                          emptyCaptionDetail={_("Create a snapshot to roll back to a previous state")}
```
Venefilyn:
```suggestion
                          emptyCaptionDetail={_("Create a snapshot to allow rollback to a previous state")}
```

martinpitt: I agree that "allow" should be in here. But English grammar demands that it's always "allow SOMEONE to ...". Maybe "snapshot for rolling back" works?

Venefilyn: > But English grammar demands that it's always "allow SOMEONE to ...".

Not necessarily needed with English grammar. That sentence is valid.

WDYT about
```suggestion
                          emptyCaptionDetail={_("Create a snapshot to rollback to.")}
```

```suggestion
                          emptyCaptionDetail={_("Snapshot current state to enable rollback to it.")}
```

```suggestion
                          emptyCaptionDetail={_("Creating a snapshot will allow you to rollback to it.")}
```

martinpitt: To be honest they all feel too unwieldy. Dangling preposition in the first, too many "to"s in the second, etc. I also changed my mind about "allow". The current PR version is fine, or we just drop that string all together -- it just feels like filler text.

 Argh bikeshedding..

Venefilyn: Question is if we want helper text like this when we have empty state too. Do we need to say anything about rolling back

mvollmer:
@martinpitt, @Venefilyn, I have now just removed some words from the old snapshot explanation. Is that a clear improvement?

martinpitt (approved):
I like these texts! Nice and simple, useful, and no nonsense.

Venefilyn (approved):
Much better!

mvollmer:
Yeeha!

PR #2317, opened 2025-09-11
disks: Use correct domainstat property for the "Used" column

### LINKED ISSUE #1423: Wrong USED disk space reported

mvollmer:
We want the actual space used in the host filesystem, as reported by "du".  This is either the "allocation" property or the "physical" property, depending on whether the VM is running or not.

Fixes #1423

### DIFF
--- a/test/check-machines-disks
+++ b/test/check-machines-disks
@@ -284,7 +284,7 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
         self.wait_for_disk_stats("subVmTest1", "vda")
         if b.is_present("#vm-subVmTest1-disks-vda-used"):
             b.wait_in_text("#vm-subVmTest1-disks-vda-used", "GiB")
-            self.assertRegex(b.text("#vm-subVmTest1-disks-vda-used"), r"^(0|0\.0|0\.01) GiB$")
+            self.assertRegex(b.text("#vm-subVmTest1-disks-vda-used"), r"^(0|0\.0|0\.09) GiB$")
 
         # Test add disk by external action
         m.execute("qemu-img create -f raw /var/lib/libvirt/images/image3.img 128M")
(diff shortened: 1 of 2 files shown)


mvollmer:
Should we file a libvirt bug?  I guess...

test/check-machines-disks
```diff
@@ -284,7 +284,7 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
         self.wait_for_disk_stats("subVmTest1", "vda")
         if b.is_present("#vm-subVmTest1-disks-vda-used"):
             b.wait_in_text("#vm-subVmTest1-disks-vda-used", "GiB")
-            self.assertRegex(b.text("#vm-subVmTest1-disks-vda-used"), r"^(0|0\.0|0\.01) GiB$")
+            self.assertRegex(b.text("#vm-subVmTest1-disks-vda-used"), r"^(0|0\.0|0\.09) GiB$")
```
martinpitt, line 287: Weird to see the number go *up*, I had expected "down". But :man_shrugging:

mvollmer: Well, we don't really know what `allocation` means for a running machine... doesn't seem to have anything to do with the image itself. Maybe it is "the highest block written since the start of the machine"? It's pretty random if you ask me.

mvollmer:
> That is https://issues.redhat.com/browse/COCKPIT-1089 right?

Yes.

PR #2323, opened 2025-09-16
Fix: hostadd allow device form to overflow on X-axis

Lunarequest:
If user has a host with large vendor ids, or pci-e ids the hostadd card may overflow and be cut off for example on a host with > 100 pci-e lanes. This adds a div with overflowX set to auto so if there is a overflow it can overflow and add a scroll bar. If there is a better way of doing this in patternfly I couldn't find one


martinpitt:
Thanks for fixing this, and sorry for the delay. I triggered tests now, could be that this affects some pixel tests.

martinpitt (approved):

PR #2326, opened 2025-09-17
nics: Don't render "0" when there are no IP addresses

mvollmer:
We render "Unknown" when there are no IP addresses of any IP version, but if there are IPv4 addresses but no IPv6 addresses, the code for the IPv6 addresses would render a "0". (And the other way around.)

### DIFF
--- a/src/components/vm/nics/vmNicsCard.tsx
+++ b/src/components/vm/nics/vmNicsCard.tsx
@@ -487,7 +487,7 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
 
                     return (
                         <DescriptionList isHorizontal isFluid>
-                            {inetIps.length && <DescriptionListGroup>
+                            {inetIps.length > 0 && <DescriptionListGroup>
                                 {inetIps.map((ip, index) => (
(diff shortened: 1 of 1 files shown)


mvollmer:
> Classic, I wish there was a linter to catch this, I did it multiple times in the past.

Yeah, and I knew about this gotcha of course, but missed it this time... yes, linter rule would be great.

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -499,7 +499,7 @@ export class VmNetworkTab extends React.Component<VmNetworkTabProps, VmNetworkTa
                                     </React.Fragment>
                                 ))}
                             </DescriptionListGroup>}
-                            {inet6Ips.length && <DescriptionListGroup>
+                            {inet6Ips.length > 0 && <DescriptionListGroup>
```
allisonkarlitskaya, line 502: On the other hand, would it hurt us to drop the conditional entirely and always have the `<DescriptionListGroup>`, possibly empty?

mvollmer: It would probably not hurt, but this here implements the original intention correctly, so I would go with that.

PR #2334, opened 2025-09-26
api: Renovate virt-xml invocations

mvollmer:
There are two main ideas here:

 - Using virt-xml is good and we should do more of it, and make it easy, and do it properly.

 - Functions like domainAttachIface are mostly obfuscation. They have a large complicated API surface that is gratuitously different from virt-xml. And they are only called in one place anyway. Better to have all the hair related to a particular API topic next to the UI code for it.

PR #2340, opened 2025-09-30
api: Use virtXmlEdit to replace Spice graphics

mvollmer:
- [ ] #2334


mvollmer:
Only ubuntu-2204 and ubuntu-2404 have too old versions of virt-xml. We could simply not offer spice conversion there, since it's probably not important for ubuntu.

src/components/vm/vmReplaceSpiceDialog.tsx
```diff
@@ -77,9 +77,9 @@ export const ReplaceSpiceDialog = ({
                 // convert them serially, to avoid hammering libvirt with too many parallel requests
                 for (const sel of spiceVMs)
                     if (selected.includes(selectionId(sel)))
-                        await domainReplaceSpice({ connectionName: sel.connectionName, id: sel.id });
+                        await virtXmlEdit(sel, "convert-to-vnc", 1, {});
```
martinpitt: Oh, that exists finally! However, still fails on Ubuntu LTS, so maybe it's not there yet?

I'm ok with not offering that feature on 22.04 and 24.04 any more, but it should have some feature test then?

martinpitt (commented):
Maybe you can split this up by moving to the new virtXmlEdit() API first (which should work everywhere), and then moving to `convert-to-vnc` separately?

mvollmer:
> Maybe you can split this up by moving to the new virtXmlEdit() API first (which should work everywhere), and then moving to `convert-to-vnc` separately?

Hmm, do you mean that we should replace the explicit XML modifications in the `replaceSpice` function with some `virt-xml --edit` invocation? I don't think that's straighforward, if at all possible.

src/libvirtApi/common.ts
```diff
@@ -391,6 +392,19 @@ async function storagePoolStopOrUndefine(connectionName: ConnectionName, poolPat
     }
 }
 
+async function getVirtXmlCapabilities() {
+    try {
+        const help_text = await cockpit.spawn(["virt-xml", "--help"]);
+        store.dispatch(setVirtXmlCapabilities(
+            {
+                convert_to_vnc: help_text.includes("--convert-to-vnc"),
```
martinpitt, line 400: Ugh, but what can you do..

martinpitt:
@mvollmer [this failure](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2340-abb9dd09-20251010-135202-ubuntu-2404/log.html) is persistent after a retry. This may need adjustment?

src/libvirtApi/common.ts
mvollmer: Yes, agreed.

mvollmer:
> @mvollmer [this failure](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2340-abb9dd09-20251010-135202-ubuntu-2404/log.html) is persistent after a retry. This may need adjustment?

Yes, thanks!

src/libvirtApi/common.ts
```diff
@@ -391,6 +392,19 @@ async function storagePoolStopOrUndefine(connectionName: ConnectionName, poolPat
...
+            }
+        ));
+    } catch (exc) {
+        console.error("Failed to query virt-xml capabilities:", String(exc));
```
cockpituous, line 404: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2340-e5f64a02-20251013-054745-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

PR #2343, opened 2025-10-02
test: the console now forcefully disconnects on rhel-9-8

jelly:
rhel-9-8 now has Cockpit > 346 which fixed the websocket channel close behaviour, see:

https://github.com/cockpit-project/cockpit/commit/3ef9f72a29608b97d66a6073166ed6b09d452b6f


martinpitt (approved):
Thanks! Works of course, but I figure you might want to merge this with #2344? Although, we can land r98 first and then unconflict that.

PR #2354, opened 2025-10-14
consoles: Make "Expand"/"Compress" a real link

mvollmer:
So that people can bookmark it and open it in a new tab, just like the "Detach" link.

This is not at all important, imo.  Bookmarking the "Detach" link or opening it in a new tab is probably always preferable for VNC. But for consistency and for getting the serial console into a tab this might be an improvement.


mvollmer:
@bookwar, you wanted to have a look at this, right? Should we do this? Personally, I don't really care either way.

mvollmer:
Let's not bother...

PR #2355, opened 2025-10-14
usage: Show actual used guest memory, not qemu process size

mvollmer:
Demo: https://youtu.be/OSnwfRvcl5w
Demo for first commit: https://youtu.be/uplVsdTWopw

### Machines: Memory usage now shows numbers reported by the guest

The memory usage bars now show the same numbers as tools like "top" and "free" when run inside the guest. Specifically, the memory usage shown by Cockpit is based on the "MemAvailable" number of the guest kernel.

To implement this, Cockpit will run `virsh dommemstat <vm name> --period 10 --life`. If that is not acceptable, please let us know.

PR #2359, opened 2025-10-16
vm: Configure balloon polling

mvollmer:
Demo: https://youtu.be/jB3q_VGdp60

- [ ] #2355 
- [ ] Tests

### DIFF
--- a/src/components/vm/vmDetailsPage.tsx
+++ b/src/components/vm/vmDetailsPage.tsx
@@ -173,7 +173,7 @@ export const VmDetailsPage = ({
             id: `${vmId(vm.name)}-usage`,
             className: 'usage-card',
             title: _("Usage"),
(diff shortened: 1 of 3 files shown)


mvollmer:
One idea is to actually just switch on the polling for a running VM while Cockpit is interested in the usage data. We would never modify the inactive configuration of a VM.

This might interfere with other tools (or other Cockpit sessions) that do the same since there is no "reference counting" of how often the polling was switched on. Also, we might forget to switch it off when the session ends abruptly.

Maybe it's fine to simply always switch it on when needed and then just leave it on until the machine is shut down.

mvollmer:
> One idea is to actually just switch on the polling for a running VM while Cockpit is interested in the usage data.

Yeah, let's start with that until people complain. Done in #2355.

PR #2383, opened 2025-10-29
app: Improve initializing expensive state

mvollmer:
The OsInfoList is retrieved only when the "Create VM" dialog is opened for the first time.

The virt-xml capabilities are initialized in the background.


mvollmer:
For me, this saves about 1 second or about 80% of the time until `AppActive` is rendered.

src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1247,7 +1251,7 @@ class CreateVmModal extends React.Component<CreateVmModalProps, CreateVmModalSta
             this.autodetectOSPromise = autodetectOS(installMedia);
             this.autodetectOSPromise.then(resJSON => {
                 const res = JSON.parse(resJSON.trim());
-                const osEntry = this.props.osInfoList.filter(osEntry => osEntry.id == res.os);
+                const osEntry = this.state.osInfoList.filter(osEntry => osEntry.id == res.os);
```
martinpitt, line 1254: Doesn't this have to be guarded against waiting for osInfoList to finish loading? Is `onOsAutodetect()` only used in the non-"Loading" version of the dialog? That's not visible in the diff, but an assertion might be in order?

mvollmer: > Is onOsAutodetect() only used in the non-"Loading" version of the dialog?

Yes, it's used when the user starts inputting things into the dialog fields.

PR #2386, opened 2025-10-30
 vm: Load snapshots in the background

mvollmer:


mvollmer:
This is another half second or so for me, when there is one VM with one snapshot.

src/app.tsx
```diff
@@ -111,8 +111,8 @@ export const App = () => {
                     if (connectionName == "system")
                         setSystemSocketInactive(true);
                 }
-                setLoadingResources(false);
             }));
+            setLoadingResources(false);
```
martinpitt, line 115: Ironically that makes initialization *slower*. I'm not actually sure if that was a bug -- it seems nice to be able to e.g. interact with session images already even though system ones are still loading?

No veto, just a thought.

martinpitt (dismissed):

mvollmer:
Oh, `snapshots == undefined` means "not supported". PR needs to be fixed for that.

src/app.tsx
mvollmer: Well, maybe, but if this is designed, the logic behind it is weak.

Once the session VMs are loaded, the system VMs will appear incrementally and make your VM list dance like Travolta while you try to interact with your sessions VMs. And if that is okay, why not make the session VMs appear incrementally as well?

martinpitt: Yeah, good point.

src/libvirtApi/snapshot.ts
```diff
@@ -144,7 +144,7 @@ export async function snapshotGetAll({
         store.dispatch(updateDomainSnapshots({
             connectionName,
             domainPath,
-            snaps: undefined,
+            snaps: false,
```
cockpituous, line 147: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2386-cd287e64-20251030-080645-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/snapshots/vmSnapshotsCard.tsx
```diff
@@ -267,8 +268,8 @@ export class VmSnapshotsCard extends React.Component<VmSnapshotsCardProps> {
             <ListingTable aria-label={`VM ${vm.name} Snapshots Cards`}
                           gridBreakPoint='grid-lg'
                           variant="compact"
-                          emptyCaption={_("No snapshots")}
-                          emptyCaptionDetail={_("Snapshots allow you to revert to an earlier state")}
+                          emptyCaption={vm.snapshots ? _("No snapshots") : <Spinner />}
```
martinpitt, line 271: Shouldn't that be specific to `undefined` now? `vm.snapshots === undefined ? <Spinner /> : _("No snapshots")`

We don't want to show a spinner when detection finished but there are no snapshots.

Self-answer: line 202 has `vm.snapshots !== false`, so it can't be `false` here.

mvollmer: > Self-answer: line 202 has `vm.snapshots !== false`, so it can't be `false` here.

Yes, but good point! This kind of distributed non-trivial logic is indeed questionable.

PR #2390, opened 2025-11-03
test: work around rhel-8-10 space issue

jelly:
TestMachinesCreate.testCreateDownloadRhel fails in a new image refresh of rhel-8-10 error'ing out about not enough space being available (10G being required for rhel 8-10). However the test tries to validate the `offline_token`, apparently semantics for validation changed a bit making this become an issue. Therefore always configure the 500M volume so we actually test the token validation and not lack of free space on our images.

```
The requested volume capacity will exceed the available pool space when
the volume is fully allocated. (10240 M requested capacity > 10074 M
available)
```

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -2407,6 +2407,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                                           os_short_id=config.RHEL_8_1_SHORTID,
                                                                           offline_token="invalid_token",
                                                                           offline_token_autofilled=False,
+                                                                          location=config.VALID_DISK_IMAGE_PATH,
                                                                           create_and_run=True),
                                               "Error",
                                               True)


mvollmer:
The failure in question was on the `rhel-8` branch, not on `main`, no?

mvollmer:
> The failure in question was on the `rhel-8` branch, not on `main`, no?

And the failing part of the test is this, as far as I can tell.

```
        runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='os',
                                                      os_name=config.RHEL_8_10,
                                                      os_short_id=config.RHEL_8_10_SHORTID,
                                                      offline_token="valid_token",
                                                      create_and_run=True,
                                                      connection="session"),
                          check_env_empty=False)
```

The part of the test that your are modifying here has already passed when [this ](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-8392-95993d78-20251101-101406-rhel-8-10-cockpit-project-cockpit-machines-rhel-8/log.html) failure happens.

mvollmer (changes requested):
We need to fix the `rhel-8` branch, not `main`.

jelly:
> We need to fix the `rhel-8` branch, not `main`.

Right! This makes it more confusing

jelly:
Moved too https://github.com/cockpit-project/cockpit-machines/pull/2391

PR #2392, opened 2025-11-04
test: Fake OSInfo resource minimums for all architectures

mvollmer:
Just faking the first doesn't seem to be enough for the new rhel-8-10 image.

- [x] Figure out which one cockpit is actually using, because it sure isn't the "x86_64" one.


mvollmer:
Cockpit uses the first element returned by `os.get_minimum_resources` and the test used to modify the first `<resources>` entry in the XML file. This used to be the same entry, apparently, but for some reason it's probably no longer the case in the new rhel-8-10.

I think Cockpit should look for a specific architecture when looking up resource requirements, but for now this fix here is sufficient, I'd say.

jelly (approved):
Thanks!

mvollmer:
Right, we need this also on the rhel-8 branch.

PR #2395, opened 2025-11-05
test: temporarily install gobject-introspection on Fedora

jelly:
This is no longer pulled in via python3-gobject-base after the following change.

https://src.fedoraproject.org/rpms/pygobject3/c/6a4422642c746e124fc6b0bcecd40b74e4217510?branch=rawhide

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -6,6 +6,12 @@ cd "${0%/*}/../.."
 
 . /usr/lib/os-release
 
+
+# HACK: workaround https://bugzilla.redhat.com/show_bug.cgi?id=2412451
+if [ "$ID" = fedora ]; then
+    dnf install -y gobject-introspection
+fi
+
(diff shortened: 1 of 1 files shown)


test/browser/browser.sh
```diff
@@ -6,6 +6,12 @@ cd "${0%/*}/../.."
 
 . /usr/lib/os-release
 
+
+# HACK: workaround https://bugzilla.redhat.com/show_bug.cgi?id=2412451
+if [ "$ID" = fedora ]; then
+    dnf install -y gobject-introspection
```
mvollmer, line 12: I would have put it into `main.fmf` in the `require:` sections, but maybe a temporary workaround is better here? No strong opinion.

jelly: Right, but that involves copy pasting things 3 times for each scenario :( This should be super temporary .. hopefully :)

Issue #1210, opened 2023-09-14
Vm creation fails with: failed to initialize spice server (Ubuntu 22.04)


tgwaste:
### Explain what happens

ERROR internal error: process exited while connecting to monitor: 2023-09-13T07:02:14.596717Z qemu-system-x86_64: warning: Spice: ../server/reds.cpp:2512:reds_init_socket: getaddrinfo(::1,5913): Address family for hostname not supported 2023-09-13T07:02:14.596749Z qemu-system-x86_64: warning: Spice: ../server/reds.cpp:3387:do_spice_init: Failed to open SPICE sockets 2023-09-13T07:02:14.596760Z qemu-system-x86_64: failed to initialize spice server Domain installation does not appear to have been successful. If it was, you can restart your domain by running: virsh --connect qemu:///system start ubuntu-desktop otherwise, please restart your installation.

### Version of Cockpit

cockpit-machines     298-1~bpo22.04.1 -- every thing else 300.1-1~bpo22.04.1

### Where is the problem in Cockpit?

Services

### Server operating system

Ubuntu

### Server operating system version

22.04

### What browsers are you using?

Firefox

### System log

```shell
I used this same .iso a month or so ago and the vm created fine.
Tried a few more isos and got the same error.
Tried rebooting and still no good.
I am completely unable to create any new vm.
```

tgwaste:
Any ideas here? I mean this pretty much breaks the fundamental purpose of cockpit.

martinpitt:
@skobyda Any idea about that spice error? That sounds like a missing package perhaps, or some customization to enable/disable spice?

We run tests on Ubuntu 22.04, it does work in general. @tgwaste as a first step, please give us the output of

    dpkg -l '*qemu*' '*libvirt'

tgwaste:
```
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                           Version                                Architecture Description
+++-==============================-======================================-============-======================================================================
ii  ipxe-qemu                      1.21.1+git-20220113.fbbdc3926-0ubuntu1 all          PXE boot firmware - ROM images for qemu
[...]

PR #2398, opened 2025-11-07
create: Never add SPICE graphics to new machines

mvollmer:
Cockpit itself does not benefit from it, and SPICE potentially creates suprising problems.

See #1210 and https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621

----

### Machines: New virtual machines don't get SPICE graphics anymore

Cockpit used to add both VNC and SPICE graphics when creating new virtual machines. Now only VNC graphics are added. We have made this change since Cockpit itself does not benefit from SPICE graphics and we think that opening the SPICE port for a running machine should be intentional on part of the user and not happen silently and surprisingly.

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -2481,6 +2481,9 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
             if m.image.startswith("debian") or m.image.startswith("ubuntu"):
                 # on Debian, spice graphics is built in, spicevmc channels are a plugin
                 self.assertNotIn("spicevmc", caps)
+            elif m.image.startswith("rhel-8-"):
+                # On RHEL 8, domain capabilities are unaffected for some reason.
+                pass
             else:
                 # in Fedora/RHEL we don't expect any remaining capability
                 self.assertNotIn("spice", caps)
@@ -2504,6 +2507,9 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
             if expect_vnc:
                 with b.wait_timeout(60):
                     b.wait_visible(".vm-console-vnc canvas")
+            else:
+                # Without VNC, the serial console is displayed by default. Switch to "Graphical".
+                b.click(".consoles-card .pf-v6-c-toggle-group button:contains(Graphical)")
 
             if not expect_vnc and not expect_spice:
                 b.wait_not_present(".vm-console-footer button.pf-m-plain")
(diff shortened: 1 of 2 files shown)


mvollmer:
@martinpitt, what do you think about this?

Alternatively, we could tell the (few) people that really have trouble with spice to uninstall some packages so that libvirt no longer claims spice support.

But I was surprised that c-machines agressively gives you both VNC and SPICE (without asking) for each new machine.  Virt-manager gives you only SPICE (without asking), so it would make sense for c-machines to only add VNC, since we have a inline viewer for it.

mvollmer:
So the two arguments against adding SPICE by default are:

 - We are unsure of the maintenance status of Spice, and RHEL for example already got rid of it.
 - Opening ports gratuitously, even when restricted to 127.0.0.1, is never a good idea.

@allisonkarlitskaya even suggest that we should use Unix domain sockets for VNC instead of a TCP socket.  That would be nice if we are only concerned about Cockpit itself. But we do have the "Launch viewer" button, and we know that people rely on external viewers together with cockpit-machines, so I think we are better off just keeping the libvirt defaults, and keep calling virt-install with just `--graphics vnc`.  Fewer surprises.

We should make sure that c-machines still works when people change the global libvirt config to use unix domain sockets for VNC (is that possible?), or change an individual VM to use them.

We could add UI to configure all this in detail, but meh.

snuggles4553:
This proposal gets my vote. I've commented on this in [#1210](https://github.com/cockpit-project/cockpit-machines/issues/1210#issuecomment-3502186081)

test/check-machines-create
```diff
@@ -2565,59 +2568,68 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
...
-                b.click(f"#vm-{vmCockpit}-system-action-kebab")
-                b.wait_not_present("ul.pf-v6-c-menu__list")
-                if "Replace SPICE" not in menu:
-                    break
+            menu = b.text("ul.pf-v6-c-menu__list")
+            b.click(f"#vm-{vmCockpit}-system-action-kebab")
+            b.wait_not_present("ul.pf-v6-c-menu__list")
+            if "Replace SPICE" not in menu:
+                break
             else:
```
mvollmer: *honk* python indentation alarm!

mvollmer: Fixed.

snuggles4553:
Isn't the menu option called `Replace SPICE devices` rather than `Replace SPICE`? If so, the unit test should be adjusted to match this reality.

Snippet from `./src/components/vm/vmActions.tsx`:
```
    if (vm.inactiveXML?.hasSpice && canReplaceSpice()) {
        dropdownItems.push(
            <DropdownItem key={`${id}-replace-spice`}
                          id={`${id}-replace-spice`}
                          onClick={() => Dialogs.show(<ReplaceSpiceDialog vm={vm} vms={vms} />)}>
                {_("Replace SPICE devices")}
            </DropdownItem> 
        );
        dropdownItems.push(<Divider key="separator-spice" />);
    }
```

mvollmer:
> Isn't the menu option called `Replace SPICE devices` rather than `Replace SPICE`? If so, the unit test should be adjusted to match this reality.

I don't think we need any adjustments. This aspect has not changed, neither in the code nor in the tests.

snuggles4553:
> > Isn't the menu option called `Replace SPICE devices` rather than `Replace SPICE`? If so, the unit test should be adjusted to match this reality.
> 
> I don't think we need any adjustments. This aspect has not changed, neither in the code nor in the tests.

Ah. I just quickly glanced over this piece in the unit test:

```
        self.assertNotIn("Replace SPICE",
                         b.text("ul.pf-v6-c-menu__list"))
```

Hence, I saw it used the text `Replace SPICE` rather than `Replace SPICE devices`. But there's indeed no problem, because the `assertNotIn` method checks whether the `Replace SPICE` text is present in the menu, rather than performing a full match. Just something that caught my attention.

All good! :+1:

mvollmer:
> Just something that caught my attention.

Alright, thanks for checking!

PR #2412, opened 2025-11-21
On demand VM loading

mvollmer:
Demo: https://www.youtube.com/watch?v=Yo2WxxF1V8A

Follow ups:

- Move more init stuff into AppState, like the check for hardware virtualization.
- Refactor pages more, so that we don't repeat `<Page></Page>` so much. Some EmpyStates are not in a Page, actually.

PR #2418, opened 2025-11-24
store: Fix deleteUnlistedVMs action

mvollmer:
The action was dispatched with "vmIds", but the reducer was looking at "vmIDs".  As a result, each call to domainGetAll() would temporarily remove all VMs.

Fix that by using the same spelling in both places.

### DIFF
--- a/src/reducers.ts
+++ b/src/reducers.ts
@@ -286,9 +286,9 @@ function vms(state: VM[] | undefined, action): VM[] {
                         (action.transientOnly && vm.persistent)));
     }
     case DELETE_UNLISTED_VMS: {
-        if (action.vmIDs)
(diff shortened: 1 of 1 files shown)


mvollmer:
This is difficult to test, I think. I see if I can find a place in the code that calls domainGetAll while on the page for a single machine. With the bug, that page should disappear briefly and lose its state (like open dropdowns).

mvollmer:
The "Migrate VM" dialog calls domainGetAll, but there the machine is supposed to disappear, I think. This bug is easier to provoke in #2412, where the "Add Disk" dialog might call domainGetAll during opening, and the machine briefly disappearing will mess up the state such that the dialog doesn't open at all.

PR #2429, opened 2025-11-28
workflows: sync cockpit-lib-update workflow with starter-kit

jelly:
The starter-kit workflow never required a deploy_key for cloning the repo as we create the pull request via GitHub token in tasklib.

### DIFF
--- a/.github/workflows/cockpit-lib-update.yml
+++ b/.github/workflows/cockpit-lib-update.yml
@@ -6,11 +6,10 @@ on:
   workflow_dispatch:
 jobs:
   cockpit-lib-update:
-    environment: self
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
-      statuses: write
+      contents: write
     steps:
       - name: Set up dependencies
         run: |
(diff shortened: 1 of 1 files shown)


martinpitt:
> we create the pull request via GitHub token in tasklib.

Yes ok, but it still needs to push the origin branch? Can you re-post this from an origin branch so that we can test-run the workflow from that?

martinpitt (approved):
Ah, you mentioned a test run in https://github.com/cockpit-project/cockpit-podman/pull/2345 -- so I figure it does the push via token-authenticated http instead of ssh. Sure -- does that help us get rid of the "self" deploy key?

jelly:
> Ah, you mentioned a test run in [cockpit-project/cockpit-podman#2345](https://github.com/cockpit-project/cockpit-podman/pull/2345) -- so I figure it does the push via token-authenticated http instead of ssh. Sure -- does that help us get rid of the "self" deploy key?

Yes with token authenticated http instead of ssh. That indeed gets us rid of the "self" deploy key. We still need that key for the PO/POT workflow

PR #2435, opened 2025-12-02
nics: Improve the dialog experience with new framework components

mvollmer:
Demo:  https://youtu.be/VSb6anWw-tc

This uses the obvious porcelain components for checkboxes, dropdown selects and radio buttons. At the same time, the dialog experience is improved in a number of ways:

 - Instead of being read-only with a tooltip, the mac field is now disabled with a helper text.  The readonly state is rare and communicates less well than the disabled state. Tooltips (that open on hover) are not mobile friendly and generally awkward. There is enough space in the dialog for a inline explanation.

 - The "interface type" is selected with radio buttons instead of a dropdown. There are only three types and there is enough space to show them all. Instead of a info popover with questionably typography and content, the choices now have brief helper text to explain their purpose.

 - When a interface type has no sources, the type is disabled, instead of keeping it enabled and showing an empty list of sources, which would be a dead-end.

 - When editing a interface, the current source is always in the "Source" dropdown, even if it doesn't currently exist. Consequently, the current type is always enabled.

 - When adding a interface and there are no sources at all, the dialog explains that and only offers "Cancel".

image

image

image

Fixes: [COCKPIT-1761](https://redhat.atlassian.net/browse/COCKPIT-1761)

[COCKPIT-1761]: https://redhat.atlassian.net/browse/COCKPIT-1761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/vm/nics/nicAdd.tsx
+++ b/src/components/vm/nics/nicAdd.tsx
@@ -10,7 +10,6 @@ import { useDialogs } from 'dialogs';
 import type { VM } from '../../../types';
 import type { AvailableSources } from './vmNicsCard';
 
-import { Checkbox } from "@patternfly/react-core/dist/esm/components/Checkbox";
 import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form";
 import {
     Modal, ModalBody, ModalFooter, ModalHeader
@@ -19,19 +18,20 @@ import { Radio } from "@patternfly/react-core/dist/esm/components/Radio";
 import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput";
 
 import {
-    NetworkTypeAndSourceValue, NetworkTypeAndSourceRow,
-    init_NetworkTypeAndSourceRow, validate_NetworkTypeAndSourceRow,
+    NetworkTypeAndSourceValue, NetworkTypeAndSourceRow, init_NetworkTypeAndSourceRow,
     NetworkModelRow,
     PortForwardsValue, NetworkPortForwardsRow, validate_PortForwards,
     dialogPortForwardsToInterface,
 } from './nicBody.jsx';
 import { virtXmlHotAdd, domainGet, domainIsRunning } from '../../../libvirtApi/domain.js';
+
 import { appState } from '../../../state';
 
 import {
-    useDialogState, DialogField, DialogError,
+    useDialogState_async, DialogState, DialogField, DialogError,
     DialogErrorMessage, DialogHelperText,
-    DialogActionButton, DialogCancelButton
+    DialogActionButton, DialogCancelButton,
+    DialogCheckbox,
 } from 'cockpit/dialog';
 
 import './nic.css';
@@ -130,18 +130,11 @@ const PermanentChange = ({
     // down only. Enable permanent change of the domain.xml
 
     return (
-        <FormGroup
-            fieldId={field.id()}
-            label={_("Persistence")}
-            hasNoPaddingTop
-        >
-            <Checkbox
-                id={field.id()}
-                isChecked={field.get()}
-                label={_("Always attach")}
-                onChange={(_event, checked) => field.set(checked)}
-            />
-        </FormGroup>
+        <DialogCheckbox
+            field_label={_("Persistence")}
+            checkbox_label={_("Always attach")}
+            field={field}
+        />
     );
 };
 
@@ -164,7 +157,7 @@ export const AddNIC = ({
 }) => {
     const Dialogs = useDialogs();
 
-    function init(): AddNICValues {
+    async function init(): Promise<AddNICValues> {
         return {
             model: "virtio",
             type_and_source: init_NetworkTypeAndSourceRow(vm, null, availableSources),
@@ -174,9 +167,8 @@ export const AddNIC = ({
         };
     }
 
-    function validate() {
+    function validate(dlg: DialogState<AddNICValues>) {
         validate_NetworkMacRow(dlg.field("mac"));
-        validate_NetworkTypeAndSourceRow(dlg.field("type_and_source"));
         if (dlg.values.type_and_source.type == "user")
             validate_PortForwards(dlg.field("portForwards"));
     }
@@ -227,34 +219,37 @@ export const AddNIC = ({
         }
     }
 
-    const dlg = useDialogState(init, validate);
+    const dlg = useDialogState_async(init, validate);
 
-    const defaultBody = (
-        <>
-            <Form onSubmit={e => e.preventDefault()} isHorizontal>
-                <NetworkTypeAndSourceRow field={dlg.field("type_and_source")} />
+    let defaultBody: React.ReactNode;
+    if (dlg instanceof DialogState) {
+        defaultBody = (
+            <>
+                <Form onSubmit={e => e.preventDefault()} isHorizontal>
+                    <NetworkTypeAndSourceRow field={dlg.field("type_and_source").at(dlg.values.type_and_source)} />
 
-                <NetworkModelRow
-                    field={dlg.field("model")}
-                    osTypeArch={vm.arch}
-                    osTypeMachine={vm.emulatedMachine}
-                />
+                    <NetworkModelRow
+                        field={dlg.field("model")}
+                        osTypeArch={vm.arch}
+                        osTypeMachine={vm.emulatedMachine}
+                    />
 
-                <NetworkMacRow field={dlg.field("mac")} />
+                    <NetworkMacRow field={dlg.field("mac")} />
 
-                { domainIsRunning(vm.state) && vm.persistent &&
-                    <PermanentChange field={dlg.field("permanent")} />
-                }
-            </Form>
-            { dlg.values.type_and_source.type == "user" &&
-                vm.capabilities.interfaceBackends.includes("passt") &&
-                <Form>
-                    <br />
-                    <NetworkPortForwardsRow field={dlg.field("portForwards")} />
+                    { domainIsRunning(vm.state) && vm.persistent &&
+                        <PermanentChange field={dlg.field("permanent")} />
+                    }
                 </Form>
-            }
-        </>
-    );
+                { dlg.values.type_and_source.type == "user" &&
+                    vm.capabilities.interfaceBackends.includes("passt") &&
+                    <Form>
+                        <br />
+                        <NetworkPortForwardsRow field={dlg.field("portForwards")} />
+                    </Form>
+                }
+            </>
+        );
+    }
 
     return (
         <Modal
--- a/src/components/vm/nics/nicBody.tsx
+++ b/src/components/vm/nics/nicBody.tsx
@@ -13,12 +13,8 @@ import type { AvailableSources } from './vmNicsCard';
 
 import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { EmptyState, EmptyStateBody } from "@patternfly/react-core/dist/esm/components/EmptyState";
-import { Flex } from "@patternfly/react-core/dist/esm/layouts/Flex";
 import { FormGroup, FormFieldGroup, FormFieldGroupHeader } from "@patternfly/react-core/dist/esm/components/Form";
 import { FormSelect, FormSelectOption } from "@patternfly/react-core/dist/esm/components/FormSelect";
-import { Radio } from "@patternfly/react-core/dist/esm/components/Radio";
-import { PopoverPosition } from "@patternfly/react-core/dist/esm/components/Popover";
-import { Content, ContentVariants } from "@patternfly/react-core/dist/esm/components/Content";
 import { ExternalLinkSquareAltIcon, TrashIcon } from '@patternfly/react-icons';
 import { Grid } from "@patternfly/react-core/dist/esm/layouts/Grid";
 
@@ -27,9 +23,11 @@ import { InfoPopover } from '../../common/infoPopover.jsx';
 import cockpit from 'cockpit';
 
 import {
+    DialogError,
     DialogField,
     DialogTextInput,
-    DialogHelperText,
+    DialogDropdownSelect, DialogDropdownSelectOption,
+    DialogRadioSelect, DialogRadioSelectOption
 } from 'cockpit/dialog';
 
 import './nic.css';
@@ -45,68 +43,83 @@ export const NetworkModelRow = ({
     osTypeArch: optString,
     osTypeMachine: optString,
 }) => {
-    const availableModelTypes: { name: string, desc?: string }[] = [
-        { name: 'virtio', desc: 'Linux, perf' },
-        { name: 'e1000e', desc: 'PCI' },
-        { name: 'e1000', desc: 'PCI, legacy' },
-        { name: 'rtl8139', desc: 'PCI, legacy' }];
-    const defaultModelType = field.get();
+    const availableModelTypes = [
+        { value: 'virtio', label: 'virtio (Linux, perf)' },
+        { value: 'e1000e', label: 'e1000e (PCI)' },
+        { value: 'e1000', label: 'e1000 (PCI, legacy)' },
+        { value: 'rtl8139', label: 'rtl8139 (PCI, legacy)' }];
 
     if (osTypeArch == 'ppc64' && osTypeMachine == 'pseries')
-        availableModelTypes.push({ name: 'spapr-vlan' });
+        availableModelTypes.push({ value: 'spapr-vlan', label: 'spapr-vlan' });
 
     return (
-        <FormGroup fieldId={field.id()} label={_("Model")}>
-            <FormSelect
-                id={field.id()}
-                onChange={(_event, val) => field.set(val)}
-                data-value={defaultModelType}
-                value={defaultModelType}
-            >
-                {availableModelTypes
-                        .map(networkModel => {
-                            return (
-                                <FormSelectOption value={networkModel.name} key={networkModel.name}
-                                                  label={networkModel.name + ' (' + networkModel.desc + ')'} />
-                            );
-                        })}
-            </FormSelect>
-        </FormGroup>
+        <DialogDropdownSelect
+            label={_("Model")}
+            field={field}
+            options={availableModelTypes}
+        />
     );
 };
 
-interface NetworkTypeDescription {
-    name: string,
-    desc: string,
-    detailHeadline?: React.ReactNode,
-    detailParagraph?: React.ReactNode,
-    externalDocs?: React.ReactNode,
-    disabled?: boolean,
+type NetworkType = "network" | "bridge" | "direct" | "user";
+
+export function isSupportedNetworkType(type: string): type is NetworkType {
+    return type == "network" || type == "bridge" || type == "direct" || type == "user";
 }
 
-function getAvailableNetworkTypes(vm: VM, availableSources: AvailableSources) {
+interface NetworkTypeDescription {
+    type: NetworkType,
+    desc: string,
+    detailHeadline?: React.ReactNode,
+    excuse: string | null,
+}
+
+interface AvailableSourcesForType {
+    network: string[];
+    bridge: string[];
+    direct: string[];
+    user: string[];
+}
+
+function getAvailableNetworkTypes(vm: VM, availableSources: AvailableSourcesForType) {
     let availableNetworkTypes: NetworkTypeDescription[] = [];
 
     // { name: 'ethernet', desc: 'Generic ethernet connection' }, Add back to the list when implemented
     const virtualNetwork: NetworkTypeDescription[] = [{
-        name: 'network',
+        type: 'network',
         desc: 'Virtual network',
-        detailHeadline: _("This is the recommended type for general guest connectivity on hosts with dynamic / wireless networking configs."),
-        detailParagraph: _("Provides a connection whose details are described by the named network definition.")
+        detailHeadline: _("For general guest connectivity on hosts with dynamic / wireless networking configs."),
+        excuse: availableSources.network.length == 0 ? _("No virtual networks") : null,
     }];
     if (vm.connectionName !== 'session') {
         availableNetworkTypes = [
             ...virtualNetwork,
             {
-                name: 'bridge',
+                type: 'bridge',
                 desc: 'Bridge to LAN',
-                detailHeadline: _("This is the recommended type for general guest connectivity on hosts with static wired networking configs."),
-                detailParagraph: _("Provides a bridge from the guest virtual machine directly onto the LAN. This needs a bridge device on the host with one or more physical NICs.")
+                detailHeadline: _("For general guest connectivity on hosts with static wired networking configs."),
+                excuse: availableSources.bridge.length == 0 ? _("No bridge devices") : null,
             },
             {
-                name: 'direct',
+                type: 'direct',
                 desc: 'Direct attachment',
-                detailParagraph: _("This is the recommended type for high performance or enhanced security."),
+                detailHeadline: (
+                    <>
+                        {_("Advanced interface type for high performance or enhanced security.")}
+                        {"\n"}
+                        <Button
+                            isInline
+                            variant="link"
+                            component="a"
+                            icon={<ExternalLinkSquareAltIcon />}
+                            iconPosition="right"
+                            target="__blank"
+                            href="https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface">
+                            {_("More info")}
+                        </Button>
+                    </>
+                ),
+                excuse: availableSources.direct.length == 0 ? _("No network devices") : null,
             },
         ];
     } else {
@@ -114,9 +127,10 @@ function getAvailableNetworkTypes(vm: VM, availableSources: AvailableSources) {
         if (availableSources.network.length > 0) {
             availableNetworkTypes = [
                 {
-                    name: 'user',
+                    type: 'user',
                     desc: 'Userspace stack',
-                    detailParagraph: _("Provides a virtual LAN with NAT to the outside world.")
+                    detailHeadline: _("Provides a virtual LAN with NAT to the outside world."),
+                    excuse: null,
                 },
                 ...virtualNetwork,
             ];
@@ -136,12 +150,12 @@ function getNetworkSource(network: VMInterface): optString {
 }
 
 export interface NetworkTypeAndSourceValue {
-    type: string;
+    type: NetworkType;
     source: string;
     mode: string;
 
     _availableTypes: NetworkTypeDescription[];
-    _availableSourcesForType: Record<string, string[]>;
+    _availableSourcesForType: AvailableSourcesForType;
 }
 
 export function init_NetworkTypeAndSourceRow(
@@ -149,33 +163,54 @@ export function init_NetworkTypeAndSourceRow(
     network: VMInterface | null,
     availableSources: AvailableSources
 ): NetworkTypeAndSourceValue {
-    const _availableSourcesForType: Record<string, string[]> = {
+    const _availableSourcesForType: AvailableSourcesForType = {
         network: availableSources.network,
         direct: Object.keys(availableSources.device)
                 .filter(dev => availableSources.device[dev].type != "bridge"),
         bridge: Object.keys(availableSources.device)
-                .filter(dev => availableSources.device[dev].type == "bridge")
+                .filter(dev => availableSources.device[dev].type == "bridge"),
+        user: [],
     };
 
-    const _availableTypes = getAvailableNetworkTypes(vm, availableSources);
+    const _availableTypes = getAvailableNetworkTypes(vm, _availableSourcesForType);
 
-    let type: string;
+    let type: NetworkType | "";
     let source: string;
     let mode: string;
 
     if (network) {
+        if (!isSupportedNetworkType(network.type))
+            throw new DialogError(cockpit.format(_("Network interfaces of type \"$0\" can not be edited here."), network.type));
+
         type = network.type;
         source = getNetworkSource(network) || "";
         mode = (type == "direct" ? (network.source.mode || "") : "bridge");
-    } else {
-        type = vm.connectionName == "session" ? "user" : "network";
+
+        // Make sure the current network source is available, no
+        // matter what.
+        const avail = _availableSourcesForType[network.type];
+        const src = getNetworkSource(network) || "";
+        if (!avail.includes(src))
+            avail.push(src);
+    } else if (vm.connectionName == "session") {
+        type = "user";
         source = "";
+        mode = "";
+    } else {
+        type = "";
+        source = "";
+        for (const t of _availableTypes) {
+            if (!t.excuse) {
+                type = t.type;
+                source = _availableSourcesForType[type][0];
+                break;
+            }
+        }
         mode = "bridge";
     }
 
-    const available = _availableSourcesForType[type] || [];
-    if (!available.includes(source))
-        source = available.length > 0 ? available[0] : "";
+    if (!type)
+        throw new DialogError(_("There are no sources for virtual network interfaces on this host. You might want to create a virtual network."));
 
     return {
         type,
@@ -193,145 +228,69 @@ export const NetworkTypeAndSourceRow = ({
     field: DialogField<NetworkTypeAndSourceValue>,
 }) => {
     const {
-        type, source, mode,
+        type,
         _availableTypes, _availableSourcesForType
     } = field.get();
 
-    let sourceValue = source;
-    let networkSourcesContent: React.ReactNode;
-    let networkSourceEnabled: boolean = true;
+    const sources = _availableSourcesForType[type];
+    const networkSourcesContent: DialogDropdownSelectOption<string>[] = sources.sort().map(networkSource => {
+        return {
+            value: networkSource,
+            label: networkSource
+        };
+    });
 
-    if (["network", "direct", "bridge"].includes(type)) {
-        const sources = _availableSourcesForType[type] || [];
-        if (sources.length > 0) {
-            networkSourcesContent = sources.sort().map(networkSource => {
-                return (
-                    <FormSelectOption value={networkSource} key={networkSource}
-                                      label={networkSource} />
-                );
-            });
-        } else {
-            if (type === "network")
-                sourceValue = _("No virtual networks");
-            else
-                sourceValue = _("No network devices");
-
-            networkSourcesContent = (
-                <FormSelectOption value='empty-list' key='empty-list'
-                                  label={sourceValue} />
-            );
-            networkSourceEnabled = false;
-        }
-    }
-
-    function setNetworkType(type: string) {
-        field.sub("type").set(type);
+    function updateType(type: NetworkType) {
         const sources = _availableSourcesForType[type] || [];
         field.sub("source").set(sources.length > 0 ? sources[0] : "");
+    }
+
+    function makeTypeOption(type: NetworkTypeDescription): DialogRadioSelectOption<NetworkType> {
+        return {
+            value: type.type,
+            label: type.desc,
+            excuse: type.excuse,
+            explanation: type.detailHeadline,
+        };
     }
 
     return (
         <>
             { _availableTypes.length > 0 &&
-                <FormGroup fieldId={field.sub("type").id()}
+                <DialogRadioSelect<NetworkType>
                     label={_("Interface type")}
-                    labelHelp={
-                        <InfoPopover aria-label={_("Interface type help")}
-                            position={PopoverPosition.bottom}
-                            enableFlip={false}
-                            bodyContent={
-                                <Flex direction={{ default: 'column' }}>
-                                    {_availableTypes.map(type => (
-                                        <Content key={type.name}>
-                                            <Content component={ContentVariants.h4}>{type.desc}</Content>
-                                            <strong>{type.detailHeadline}</strong>
-                                            <p>{type.detailParagraph}</p>
-                                        </Content>))}
-                                </Flex>
-                            }
-                        />
-                    }>
-                    <FormSelect id={field.sub("type").id()}
-                        onChange={(_event, value) => setNetworkType(value)}
-                        data-value={type}
-                        value={type}>
-                        {_availableTypes
-                                .map(networkType => {
-                                    return (
-                                        <FormSelectOption value={networkType.name} key={networkType.name}
-                                    isDisabled={networkType.disabled || false}
-                                    label={networkType.desc} />
-                                    );
-                                })}
-                    </FormSelect>
-                </FormGroup>
+                    field={field.sub("type", updateType)}
+                    options={_availableTypes.map(makeTypeOption)}
+                />
             }
             {["network", "direct", "bridge"].includes(type) && (
-                <FormGroup fieldId={field.sub("source").id()} label={_("Source")}>
-                    <FormSelect id={field.sub("source").id()}
-                        onChange={(_event, val) => field.sub("source").set(val)}
-                                isDisabled={!networkSourceEnabled}
-                                data-value={sourceValue}
-                                value={sourceValue}>
-                        {networkSourcesContent}
-                    </FormSelect>
-                    <DialogHelperText field={field.sub("source")} />
-                </FormGroup>
+                <DialogDropdownSelect
+                    label={_("Source")}
+                    field={field.sub("source")}
+                    options={networkSourcesContent}
+                />
             )}
             {type == "direct" && (
-                <FormGroup id={field.sub("mode").id()} label={_("Mode")} hasNoPaddingTop isInline
-                    data-value={mode}
-                    labelHelp={
-                        <InfoPopover
-                            aria-label={_("Mode help")}
-                            position={PopoverPosition.bottom}
-                            enableFlip={false}
-                            bodyContent={
-                                <Content>
-                                    <Content component={ContentVariants.p}>
-                                        {_("The mode influences the delivery of packets.")}
-                                    </Content>
-                                    <Content component={ContentVariants.p}>
-                                        <Button isInline
-                                            variant="link"
-                                            component="a"
-                                            icon={<ExternalLinkSquareAltIcon />}
-                                            iconPosition="right"
-                                            target="__blank"
-                                            href="https://libvirt.org/formatdomain.html#direct-attachment-to-physical-interface">
-                                            {_("More info")}
-                                        </Button>
-                                    </Content>
-                                </Content>}
-                        />
-                    }>
-                    {["vepa", "bridge", "private", "passthrough"].map(m =>
-                        <Radio
-                            key={m}
-                            id={field.sub("mode").id(m)}
-                            name={`mode-${m}`}
-                            isChecked={mode == m}
+                <DialogRadioSelect
+                    label={_("Mode")}
+                    isInline
+                    field={field.sub("mode")}
+                    options={
+                        [
                             // The label is not translated since the
                             // documentation we link to is always in
                             // English.
-                            label={<pre>{m}</pre>}
-                            onChange={() => field.sub("mode").set(m)} />)}
-                </FormGroup>
+                            { value: "vepa", label: <pre>vepa</pre> },
+                            { value: "bridge", label: <pre>bridge</pre> },
+                            { value: "private", label: <pre>private</pre> },
+                            { value: "passthrough", label: <pre>passthrough</pre> },
+                        ]
+                    }
+                />
             )}
         </>
     );
 };
-
-export function validate_NetworkTypeAndSourceRow(
-    field: DialogField<NetworkTypeAndSourceValue>,
-) {
-    const val = field.get();
-    if (val._availableTypes.length > 0)
-        field.sub("source").validate(v => {
-            if (v == "")
-                return _("No sources available");
-        });
-}
 
 export interface DialogComplexPortForward {
     kind: "complex";
--- a/src/components/vm/nics/vmNicsCard.tsx
+++ b/src/components/vm/nics/vmNicsCard.tsx
@@ -40,7 +40,8 @@ const getNetworkDevices = async (): Promise<Record<string, NetworkDevice>> => {
 
         const devs: Record<string, NetworkDevice> = {};
         for (const dev of output.trim().split('\n')) {
-            devs[dev] = {};
+            if (dev)
+                devs[dev] = {};
         }
 
         const bridges = await cockpit.spawn(["ip", "-j", "link", "show", "type", "bridge"], { err: "message" });
--- a/test/check-machines-networks
+++ b/test/check-machines-networks
@@ -579,13 +579,13 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         b.click("#vm-subVmTest1-network-2-edit-dialog")
 
         # Change network model type of a running domain
-        b.select_from_dropdown(d.field("model"), "e1000e")
+        d.set_DropdownSelect("model", "e1000e")
         # Wait for the dialog warning to appear
         b.wait_visible("#vm-subVmTest1-network-2-edit-dialog-idle-message")
         # Change network type and source of a running domain
-        b.wait_val(d.field("type_and_source.type"), "network")
-        b.wait_val(d.field("type_and_source.source"), "default")
-        b.select_from_dropdown(d.field("type_and_source.source"), "test_network")
+        d.wait_RadioSelect("type_and_source.type", "network")
+        d.wait_DropdownSelect("type_and_source.source", "default")
+        d.set_DropdownSelect("type_and_source.source", "test_network")
         # Save the network settings
         b.click(d.apply_button())
         b.wait_not_present("#vm-subVmTest1-network-2-edit-dialog-modal-window")
@@ -629,9 +629,9 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         # Open the modal dialog
         b.click("#vm-subVmTest1-network-1-edit-dialog")
 
-        b.wait_val(d.field("type_and_source.type"), "bridge")
-        b.select_from_dropdown(d.field("type_and_source.type"), "direct")
-        source = b.val(d.field("type_and_source.source"))
+        d.wait_RadioSelect("type_and_source.type", "bridge")
+        d.set_RadioSelect("type_and_source.type", "direct")
+        source = d.get_DropdownSelect("type_and_source.source")
 
         # Save the network settings
         b.click(d.apply_button())
@@ -644,9 +644,9 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         # Open the modal dialog
         b.click("#vm-subVmTest1-network-1-edit-dialog")
 
-        b.wait_val(d.field("type_and_source.type"), "direct")
-        b.select_from_dropdown(d.field("type_and_source.type"), "bridge")
-        b.select_from_dropdown(d.field("type_and_source.source"), "virbr0")
+        d.wait_RadioSelect("type_and_source.type", "direct")
+        d.set_RadioSelect("type_and_source.type", "bridge")
+        d.set_DropdownSelect("type_and_source.source", "virbr0")
 
         # Save the network settings
         b.click(d.apply_button())
@@ -656,7 +656,7 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         b.wait_in_text("#vm-subVmTest1-network-1-source", "virbr0")
 
         b.click("#vm-subVmTest1-network-1-edit-dialog")
-        b.wait_in_text(d.field("type_and_source.source"), "virbr0")
+        d.wait_DropdownSelect("type_and_source.source", "virbr0")
         b.click(d.apply_button())
         b.wait_not_present("#vm-subVmTest1-network-1-edit-dialog-modal-window")
         b.wait_in_text("#vm-subVmTest1-network-1-source", "virbr0")
@@ -665,9 +665,9 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         # Open the modal dialog
         b.click("#vm-subVmTest1-network-1-edit-dialog")
 
-        b.wait_val(d.field("type_and_source.type"), "bridge")
-        b.select_from_dropdown(d.field("type_and_source.type"), "network")
-        b.select_from_dropdown(d.field("type_and_source.source"), "test_network")
+        d.wait_RadioSelect("type_and_source.type", "bridge")
+        d.set_RadioSelect("type_and_source.type", "network")
+        d.set_DropdownSelect("type_and_source.source", "test_network")
 
         # Save the network settings
         b.click(d.apply_button())
@@ -679,8 +679,7 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         # Remove test_network from the VM again
         self.deleteIface(1)
 
-        # Remove all Virtual Networks and confirm that trying to choose
-        # Virtual Networks type for a NIC disables the save button
+        # Remove all Virtual Networks
         m.execute("virsh net-dumpxml default > /tmp/net-default.xml")
         m.execute("virsh net-dumpxml test_network > /tmp/net-test-network.xml")
         m.execute("virsh net-destroy test_network; virsh net-destroy default")
@@ -709,15 +708,13 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         # Open the modal dialog
         b.click("#vm-subVmTest1-network-1-edit-dialog")
 
-        # And ensure that the network sources dropdown is disabled
-        b.wait_val(d.field("type_and_source.type"), "bridge")
-        b.select_from_dropdown(d.field("type_and_source.type"), "network")
-        b.click(d.apply_button())
-        b.wait_in_text(d.helper_text("type_and_source.source"), "No sources available")
-        b.wait_visible(d.apply_button() + ":disabled")
+        # And ensure that the network type is disabled
+        d.wait_RadioSelect("type_and_source.type", "bridge")
+        b.wait_visible(d.id("type_and_source.type", "network") + ":disabled")
         b.click(d.cancel_button())
 
-        # Ensure that when the source of a NIC was removed we can still change it
+        # Ensure that when the source of a NIC was removed, it is
+        # still shown in the Edit dialog, and we can still change it
 
         # Redefine deleted networks and attach an interface with source a deleted network
         next_mac = self.get_next_mac(next_mac)
@@ -733,18 +730,15 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
         self.performAction("subVmTest1", "forceOff")
 
         # Try to edit the interface changing the source to a non deleted network
-        b.click("#vm-subVmTest1-network-1-edit-dialog")
-        b.wait_visible("#vm-subVmTest1-network-1-edit-dialog-modal-window")
-        b.select_from_dropdown(d.field("type_and_source.type"), "network")
-        b.select_from_dropdown(d.field("type_and_source.source"), "test_network")
+        b.click("#vm-subVmTest1-network-2-edit-dialog")
+        b.wait_visible("#vm-subVmTest1-network-2-edit-dialog-modal-window")
+        d.wait_RadioSelect("type_and_source.type", "network")
+        d.wait_DropdownSelect("type_and_source.source", "default")
+        d.set_DropdownSelect("type_and_source.source", "test_network")
         b.click(d.apply_button())
-        b.wait_not_present("#vm-subVmTest1-network-1-edit-dialog-modal-window")
-        self.expandNicRow(1)
-        b.wait_in_text("#vm-subVmTest1-network-1-source", "test_network")
-
-        # Test detaching of disk on non-persistent VM
-        m.execute("virsh undefine subVmTest1")
-        b.wait_not_present("#vm-subVmTest1-network-1-edit-dialog")
+        b.wait_not_present("#vm-subVmTest1-network-2-edit-dialog-modal-window")
+        self.expandNicRow(2)
+        b.wait_in_text("#vm-subVmTest1-network-2-source", "test_network")
 
     def testNetworkAutostart(self):
         b = self.browser
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -93,7 +93,7 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         self.waitPageInit()
         b.click("#vm-subVmTest1-add-iface-button")  # open the Network Interfaces subtab
         b.wait_visible("#vm-subVmTest1-add-iface-dialog")
-        b.select_from_dropdown(d.field("type_and_source.type"), "bridge")
+        d.set_RadioSelect("type_and_source.type", "bridge")
         # dropdown options are not visible, but still part of DOM tree, so let's use _wait_present
         b._wait_present(d.field("type_and_source.source") + " option:nth-of-type(1):contains('0bridge')")
         b._wait_present(d.field("type_and_source.source") + " option:nth-of-type(2):contains('abridge')")
@@ -450,6 +450,25 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
             remove=False,
         ).execute()
 
+    def testNICAddNoDefault(self):
+        m = self.machine
+
+        self.createVm("subVmTest1", running=False)
+
+        # Delete the "default" virtual network and hide all real interfaces.
+        m.execute("virsh net-destroy default; virsh net-undefine default")
+        m.execute('mount -t tmpfs tmpfs /sys/class/net')
+        self.addCleanup(self.machine.execute, "umount /sys/class/net")
+
+        self.login_and_go("/machines")
+        self.waitPageInit()
+        self.goToVmPage("subVmTest1")
+
+        self.browser.click("#vm-subVmTest1-add-iface-button")
+        self.browser.wait_in_text(".pf-v6-c-modal-box .pf-v6-c-modal-box__header .pf-v6-c-modal-box__title",
+                                  "Add virtual network interface")
+        self.browser.wait_in_text(".pf-v6-c-modal-box", "There are no sources for virtual network interfaces")
+
     class NICEditDialog(DialogHelpers):
 
         def __init__(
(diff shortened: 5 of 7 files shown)


cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-26c86f56-20251202-155350-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-5168d18c-20251202-162035-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-dc7d51db-20251202-181824-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+        this.update();
+    }
+
+    rename_validation_state(path: string, from: number, to: number) {
+        const from_path = path + "." + String(from);
+        const to_path = path + "." + String(to);
+        if (from_path in this.validation_state) {
+            console.log("RENAME", from_path, to_path);
+            this.validation_state[to_path] = this.validation_state[from_path];
+            delete this.validation_state[from_path];
```
cockpituous: These 7 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-ea2d88b2-20251203-143118-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+        }
+        for (const k in this.validation_state) {
+            if (k.indexOf(from_path + ".") == 0) {
+                const to = to_path + k.substring(from_path.length);
+                console.log("RENAME.", k, to);
+                this.validation_state[to] = this.validation_state[k];
+                delete this.validation_state[k];
```
cockpituous: These 6 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-ea2d88b2-20251203-143118-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+        this.details = details;
+    }
+
+    toString() {
+        return this.title + ": " + String(this.details);
+    }
+
+    static fromError(title: string, err: unknown) {
+        if (err && typeof err == "object" && "message" in err && typeof err.message == "string")
+            return new DialogError(title, err.message);
```
cockpituous: These 3 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-ea2d88b2-20251203-143118-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-ea2d88b2-20251203-143118-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+    id(tag: string = "field"): string {
+        return "dialog-" + tag + "-" + this.path;
+    }
+
+    map<X>(func: (val: DialogValue<ArrayElement<T>>, index: number) => X): X[] {
+        const val = this.get();
+        if (Array.isArray(val)) {
+            return val.map((_, i) => func(this.sub(i as keyof T) as DialogValue<ArrayElement<T>>, i));
+        } else
+            return [];
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-9e650a26-20251203-151152-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+        if (Array.isArray(val)) {
+            val.forEach((_, i) => func(this.sub(i as keyof T) as DialogValue<ArrayElement<T>>, i));
+        }
+    }
+
+    remove(index: number) {
+        const val = this.get();
+        if (Array.isArray(val)) {
+            for (let j = index; j < val.length - 1; j++)
+                this.dialog.rename_validation_state(this.path, j + 1, j);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-9e650a26-20251203-151152-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-9e650a26-20251203-151152-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
cockpituous: These 6 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-e22b04b7-20251204-081542-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+            this.update();
+            return;
+        }
+
+        try {
+            cockpit.assert(this.values);
+            await func(this.values);
+        } catch (ex) {
+            console.error(String(ex));
+            this.error = ex;
```
cockpituous: These 3 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-e22b04b7-20251204-081542-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+            this as DialogState<unknown>,
+            () => this.values![tag],
+            (val) => {
+                console.log("TOP SET", tag, JSON.stringify(val));
+                this.values = { ...this.values, [tag]: val };
+                this.update();
+                if (this.online_validation)
+                    this.trigger_validation();
+                if (update_func)
+                    update_func(val);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-e22b04b7-20251204-081542-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+
+// Common elements
+
+export class DialogError {
+    title: string;
+    details: React.ReactNode;
+
+    constructor(title: string, details: React.ReactNode) {
+        this.title = title;
+        this.details = details;
```
cockpituous: These 3 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-e22b04b7-20251204-081542-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,910 @@
...
+    }
+
+    toString() {
+        return this.title + ": " + String(this.details);
```
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-e22b04b7-20251204-081542-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-e22b04b7-20251204-081542-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,918 @@
...
+    id(tag: string = "field"): string {
+        return "dialog-" + tag + "-" + this.path;
+    }
+
+    map<X>(func: (val: DialogValue<ArrayElement<T>>, index: number) => X): X[] {
+        const val = this.get();
+        if (Array.isArray(val)) {
+            return val.map((_, i) => func(this.sub(i as keyof T) as DialogValue<ArrayElement<T>>, i));
+        } else
+            return [];
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-ec962d21-20251204-104905-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,918 @@
...
+        if (Array.isArray(val)) {
+            val.forEach((_, i) => func(this.sub(i as keyof T) as DialogValue<ArrayElement<T>>, i));
+        }
+    }
+
+    remove(index: number) {
+        const val = this.get();
+        if (Array.isArray(val)) {
+            for (let j = index; j < val.length - 1; j++)
+                this.dialog.rename_validation_state(this.path, j + 1, j);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-ec962d21-20251204-104905-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-ec962d21-20251204-104905-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-dfd70a6b-20251208-133505-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,998 @@
...
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e63f3e8-20251209-142520-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e63f3e8-20251209-142520-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,998 @@
...
+
+    toString() {
+        return this.title + ": " + String(this.details);
+    }
+
+    static fromError(title: string, err: unknown) {
+        if (err && typeof err == "object" && "message" in err && typeof err.message == "string")
+            return new DialogError(title, err.message);
+        else
+            return new DialogError(title, String(err));
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e63f3e8-20251209-142520-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicAdd.tsx
```diff
@@ -73,275 +78,224 @@ function getRandomMac(vms: VM[]): string | undefined {
...
+            label={_("MAC address")}
+            hasNoPaddingTop
+            isInline
+        >
+            <Radio
+                id={v_set.id("off")}
+                name="mac-setting"
+                isChecked={!v_set.get()}
+                label={_("Generate automatically")}
+                onChange={() => v_set.set(false)}
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e63f3e8-20251209-142520-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicAdd.tsx
```diff
@@ -73,275 +78,224 @@ function getRandomMac(vms: VM[]): string | undefined {
...
-                <Form onSubmit={e => e.preventDefault()} isHorizontal>
-                    <NetworkTypeAndSourceRow
-                        vm={vm}
-                        idPrefix={idPrefix}
-                        dialogValues={{ ...this.state, availableSources: this.props.availableSources }}
-                        onValueChanged={this.onBodyValueChanged}
-                    />
+    const defaultBody = (
+        <>
+            <Form onSubmit={e => e.preventDefault()} isHorizontal>
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e63f3e8-20251209-142520-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-dd682c79-20251209-152815-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-474d21ed-20251211-131002-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-336a2323-20251215-091435-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/vm/nics/nicAdd.tsx
```diff
@@ -73,275 +79,224 @@ function getRandomMac(vms: VM[]): string | undefined {
...
+            label={_("MAC address")}
+            hasNoPaddingTop
+            isInline
+        >
+            <Radio
+                id={v_set.id("off")}
+                name="mac-setting"
+                isChecked={!v_set.get()}
+                label={_("Generate automatically")}
+                onChange={() => v_set.set(false)}
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-336a2323-20251215-091435-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicAdd.tsx
```diff
@@ -73,275 +79,224 @@ function getRandomMac(vms: VM[]): string | undefined {
...
-                            dialogValues={this.state}
-                            validation={this.state.validation}
-                            onValueChanged={this.onBodyValueChanged}
-                        />
-                    </Form>
+    const dlg = useDialogState(init, validate);
+
+    const defaultBody = (
+        <>
+            <Form onSubmit={e => e.preventDefault()} isHorizontal>
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-336a2323-20251215-091435-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1343 @@
...
+
+    at<TT extends T>(witness: TT): DialogValue<TT> {
+        cockpit.assert(Object.is(witness, this.get()));
+        return new DialogValue<TT>(
+            this.#dialog,
+            () => this.get() as TT,
+            (val) => {
+                this.set(val);
+            },
+            this.#path,
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-1d0f9e39-20251215-171555-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1343 @@
...
+    }
+
+    validate_async(debounce: number, func: (val: T) => Promise<string | undefined>): void {
+        const val = this.get();
+        if (!this.#dialog._probe_validation_cache(this.#path, val)) {
+            console.log("START VALIDATE ASYNC", this.#path, val);
+            this.#dialog._set_validation_timeout(
+                this.#path,
+                val,
+                debounce,
```
cockpituous: These 8 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-1d0f9e39-20251215-171555-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1343 @@
...
+                () => {
+                    console.log("BEGIN VALIDATE ASYNC", this.#path, val);
+                    const prom =
+                        func(val)
+                                .catch(
+                                    ex => {
+                                        console.error(ex);
+                                        return undefined;
```
cockpituous: These 7 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-1d0f9e39-20251215-171555-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1343 @@
...
+                                    }
+                                )
+                                .then(
+                                    result => {
+                                        if (this.#dialog._validation_state_is_current(this.#path, prom)) {
+                                            console.log("DONE VALIDATE ASYNC", this.#path, result);
+                                            this.#dialog._set_validation_result(this.#path, val, result);
+                                        } else {
+                                            console.log("OUTDATED", this.#path);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-1d0f9e39-20251215-171555-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1343 @@
...
+                                        }
+                                    }
+                                );
+                    this.#dialog._set_validation_promise(this.#path, val, prom);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-1d0f9e39-20251215-171555-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1343 @@
...
+            console.log("CACHE HIT", path, v);
+            this._set_validation(path, v);
+            return true;
+        } else
+            return false;
+    }
+
+    _get_validation_state(path: string): DialogValidationState | null {
+        if (path in this.#validation_state)
+            return this.#validation_state[path];
```
cockpituous: These 3 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-1d0f9e39-20251215-171555-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-1d0f9e39-20251215-171555-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+    id(tag: string = "field"): string {
+        return "dialog-" + tag + "-" + this.#path;
+    }
+
+    map<X>(func: (val: DialogValue<ArrayElement<T>>, index: number) => X): X[] {
+        const val = this.get();
+        if (Array.isArray(val)) {
+            return val.map((_, i) => func(this.sub(i as keyof T) as DialogValue<ArrayElement<T>>, i));
+        } else
+            return [];
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-3b71159d-20251216-075539-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-3b71159d-20251216-075539-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+
+    at<TT extends T>(witness: TT): DialogValue<TT> {
+        cockpit.assert(Object.is(witness, this.get()));
+        return new DialogValue<TT>(
+            this.#dialog,
+            () => this.get() as TT,
+            (val, is_aux) => {
+                this.#setter(val, is_aux);
+            },
+            this.#path,
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-3b71159d-20251216-075539-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+    }
+
+    validate_async(debounce: number, func: (val: T) => Promise<string | undefined>): void {
+        const val = this.get();
+        if (!this.#dialog._probe_validation_cache(this.#path, val)) {
+            console.log("START VALIDATE ASYNC", this.#path, val);
+            this.#dialog._set_validation_timeout(
+                this.#path,
+                val,
+                debounce,
```
cockpituous: These 8 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-3b71159d-20251216-075539-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+    get(): T {
+        return this.#getter();
+    }
+
+    set(val: T): void {
+        this.#setter(val, false);
+    }
+
+    set_aux(val: T): void {
+        this.#setter(val, true);
```
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: These 8 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+                val,
+                debounce,
+                () => {
+                    console.log("BEGIN VALIDATE ASYNC", this.#path, val);
+                    const prom =
+                        func(val)
+                                .catch(
+                                    ex => {
+                                        console.error(ex);
+                                        return undefined;
```
cockpituous: These 7 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+                                    }
+                                )
+                                .then(
+                                    result => {
+                                        if (this.#dialog._validation_state_is_current(this.#path, prom)) {
+                                            console.log("DONE VALIDATE ASYNC", this.#path, result);
+                                            this.#dialog._set_validation_result(this.#path, val, result);
```
cockpituous: These 5 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+                                        } else {
+                                            console.log("OUTDATED", this.#path);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1378 @@
...
+                                        }
+                                    }
+                                );
+                    this.#dialog._set_validation_promise(this.#path, val, prom);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-6408b0c9-20251216-134315-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+import {
+    FormSelect, FormSelectOption, type FormSelectProps,
+} from "@patternfly/react-core/dist/esm/components/FormSelect";
+import { Radio } from "@patternfly/react-core/dist/esm/components/Radio";
+
+const _ = cockpit.gettext;
+
+function debug(...args: unknown[]) {
+    if (window.debugging == "all" || window.debugging?.includes("dialog"))
+        console.debug("dialog:", ...args);
```
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+    get(): T {
+        return this.#getter();
+    }
+
+    set(val: T): void {
+        this.#setter(val, false);
+    }
+
+    set_aux(val: T): void {
+        this.#setter(val, true);
```
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+    id(tag: string = "field"): string {
+        return "dialog-" + tag + "-" + this.#path;
+    }
+
+    map<X>(func: (val: DialogValue<ArrayElement<T>>, index: number) => X): X[] {
+        const val = this.get();
+        if (Array.isArray(val)) {
+            return val.map((_, i) => func(this.sub(i as keyof T) as DialogValue<ArrayElement<T>>, i));
+        } else
+            return [];
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+        if (Array.isArray(val)) {
+            val.forEach((_, i) => func(this.sub(i as keyof T) as DialogValue<ArrayElement<T>>, i));
+        }
+    }
+
+    remove(index: number) {
+        const val = this.get();
+        if (Array.isArray(val)) {
+            for (let j = index; j < val.length - 1; j++)
+                this.#dialog._rename_validation_state(this.#path, j + 1, j);
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+        );
+    }
+
+    at<TT extends T>(witness: TT): DialogValue<TT> {
+        cockpit.assert(Object.is(witness, this.get()));
+        return new DialogValue<TT>(
+            this.#dialog,
+            () => this.get() as TT,
+            (val, is_aux) => {
+                this.#setter(val, is_aux);
```
cockpituous: These 7 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+            },
+            this.#path,
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+    }
+
+    validate(func: (val: T) => string | undefined): void {
+        const val = this.get();
+        this.#dialog._validate_value(this.#path, val, () => func(val));
+    }
+
+    validate_async(debounce: number, func: (val: T) => Promise<string | undefined>): void {
+        const val = this.get();
+        this.#dialog._validate_value_async(this.#path, val, debounce, () => func(val));
```
cockpituous: These 3 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+    #set_validation_state_timeout(
+        state: DialogValidationState,
+        val: unknown,
+        delay: number,
+        func: () => void,
+    ) {
+        if (state.timeout_id) {
+            debug("timeout cancel", state.path);
+            window.clearTimeout(state.timeout_id);
+            state.timeout_id = 0;
```
cockpituous: These 4 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-4e45c3b5-20251217-171357-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: These 7 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+       We use the #validation object as the round identifier, since it
+       is created fresh by each call to #trigger_validation.
+     */
+
+    #get_current_validation_round_id(): unknown {
+        return this.#validation;
+    }
+
+    #is_current_validation_round_id(id: unknown): boolean {
+        return Object.is(id, this.#validation);
```
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/common/dialog.tsx
```diff
@@ -0,0 +1,1550 @@
...
+       validation before running an action function, no debouncing
+       delay should be applied of course. We want to get on with
+       validation immediately.
+     */
+
+    #set_validation_state_timeout(
+        state: DialogValidationState,
+        val: unknown,
+        delay: number,
+        func: () => void,
```
cockpituous: These 5 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-0ef97292-20251218-063518-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

cockpituous (commented):
There are more than 10 code coverage comments, see [the full report here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html).

src/components/vm/nics/nicAdd.tsx
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicAdd.tsx
```diff
@@ -73,275 +79,224 @@ function getRandomMac(vms: VM[]): string | undefined {
...
 };
 
+function init_NetworkMacRow() {
+    return { set: false, val: "" };
+}
+
+function validate_NetworkMacRow(value: DialogValue<NetworkMacValue>) {
+    value.validate(v => {
+        if (v.set && v.val === "")
+            return _("Network MAC can not be empty.");
```
cockpituous, line 120: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicAdd.tsx
```diff
@@ -73,275 +79,224 @@ function getRandomMac(vms: VM[]): string | undefined {
...
                 {
-                    mac: (this.state.setNetworkMac
-                        ? this.state.networkMac
-                        : getRandomMac(await this.props.appState.getVms())),
-                    model: this.state.networkModel,
-                    type: this.state.networkType,
+                    mac: (
+                        values.mac.set
+                            ? values.mac.val
+                            : getRandomMac(await appState.getVms())
```
cockpituous, line 202: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicAdd.tsx
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicBody.tsx
```diff
@@ -21,290 +21,299 @@ import React from 'react';
...
-    const defaultModelType = dialogValues.networkModel;
+    const availableModelTypes = [
+        { value: 'virtio', label: 'virtio (Linux, perf)' },
+        { value: 'e1000e', label: 'e1000e (PCI)' },
+        { value: 'e1000', label: 'e1000 (PCI, legacy)' },
+        { value: 'rtl8139', label: 'rtl8139 (PCI, legacy)' }];
 
     if (osTypeArch == 'ppc64' && osTypeMachine == 'pseries')
-        availableModelTypes.push({ name: 'spapr-vlan' });
+        availableModelTypes.push({ value: 'spapr-vlan', label: 'spapr-vlan' });
```
cockpituous, line 53: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicBody.tsx
```diff
@@ -21,290 +21,299 @@ import React from 'react';
...
+                        </Button>
+                    </>
+                ),
+                disabled: !availableSources.direct || availableSources.direct.length == 0,
             },
         ];
     } else {
         // User session
-        if (dialogValues.availableSources.network.length > 0) {
+        if (availableSources.network && availableSources.network.length > 0) {
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicBody.tsx
```diff
@@ -21,290 +21,299 @@ import React from 'react';
...
             availableNetworkTypes = [
                 {
                     name: 'user',
                     desc: 'Userspace stack',
-                    detailParagraph: _("Provides a virtual LAN with NAT to the outside world.")
+                    detailHeadline: _("Provides a virtual LAN with NAT to the outside world."),
+                    disabled: false,
```
cockpituous: These 2 added lines are not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicBody.tsx
```diff
@@ -21,290 +21,299 @@ import React from 'react';
...
+        bridge: Object.keys(availableSources.device)
+                .filter(dev => availableSources.device[dev].type == "bridge")
+    };
+
+    if (network) {
+        // Make sure the current network source is available, no
+        // matter what.
+        if (network.type in _availableSourcesForType) {
+            const avail = _availableSourcesForType[network.type];
+            const src = getNetworkSource(network) || "";
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicBody.tsx
```diff
@@ -21,290 +21,299 @@ import React from 'react';
...
+            if (!avail.includes(src))
+                avail.push(src);
+        }
+    } else if (vm.connectionName == "system" && _availableSourcesForType.network.length == 0) {
```
cockpituous: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/components/vm/nics/nicBody.tsx
```diff
@@ -21,290 +21,299 @@ import React from 'react';
...
+    const _availableTypes = getAvailableNetworkTypes(vm, _availableSourcesForType);
+
+    let type: string;
+    let source: string;
+    let mode: string;
+
+    if (network) {
+        type = network.type;
+        source = getNetworkSource(network) || "";
+        mode = (type == "direct" ? (network.source.mode || "") : "bridge");
```
cockpituous, line 187: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2435-d596cf68-20251222-081408-fedora-43-devel/Coverage/lcov/github-pr.diff.gcov.html)

test/check-machines-nics
```diff
@@ -448,6 +448,32 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
             remove=False,
         ).execute()
 
+    def testNICAddNoDefault(self):
+        m = self.machine
+
+        self.createVm("subVmTest1", running=False)
+
+        # Delete the "default" virtual network and hide all real interfaces.
+        m.execute("virsh net-destroy default; virsh net-undefine default")
```
mvollmer, line 459: There should be a cleanup that recreates the "default" network if it still missing.

Venefilyn (commented):
I'll take a look at the code tomorrow and try it locally

---

> Demo: https://youtu.be/VSb6anWw-tc

Good demo! Thanks!

> * The "interface type" is selected with radio buttons instead of a dropdown. There are only three types and there is enough space to show them all. Instead of a info popover with questionably typography and content, the choices now have brief helper text to explain their purpose.

So this is tricky UX wise. The general rule is that when you have 2 or 3 items it is better to have a radio button. When you get to 4 items or more it starts taking up quite a lot of space and can then be moved to a dropdown.

Overall, I like the changes you made but if we ever suspect to get another network interface type we should keep it as a dropdown like we have in main atm. There is a way to make the dropdown better, see Select example here where some items have a description, but adding links is probably not possible. But we can likely add the links and more information when you've select an interface type
https://www.patternfly.org/components/menus/select#select-option-variants

> * When adding a interface and there are no sources at all, the dialog offers to create the "default" network (which really should always exist anyway) in a preparation step. This means that the "network" type is always enabled.

This is great, good to see it also added within the network interfaces component as well.

src/components/networks/networkList.tsx
```diff
@@ -35,6 +40,17 @@ export class NetworkList extends React.Component<NetworkListProps> {
...
 
+        async function createDefaultNetwork() {
+            try {
+                await networkCreateDefault();
+            } catch (ex) {
+                appState.addNotification({
+                    text: _("Failed to create \"default\" network"),
+                    detail: String(ex),
+                });
+            }
```
Venefilyn: Should be tested

src/components/vm/nics/nicBody.tsx
```diff
@@ -187,152 +205,84 @@ export function init_NetworkTypeAndSourceRow(
     };
 }
 
+export function NetworkTypeAndSource_has_sources(value: NetworkTypeAndSourceValue) {
```
Venefilyn: Complaining again that IMO JS functions should be camelCase :D

src/components/vm/nics/nicAdd.tsx
```diff
@@ -176,11 +173,20 @@ export const AddNIC = ({
 
     function validate() {
         validate_NetworkMacRow(dlg.field("mac"));
-        validate_NetworkTypeAndSourceRow(dlg.field("type_and_source"));
         if (dlg.values.type_and_source.type == "user")
             validate_PortForwards(dlg.field("portForwards"));
     }
 
+    async function createDefaultNetwork() {
+        if (await dlg.run_action(networkCreateDefault)) {
```
Venefilyn: `dlg.run_action` has return type `Promise<void>`. So checking it here seems weird. Are we testing it didn't raise?

`if` can be removed completely

src/components/vm/nics/nicAdd.tsx
```diff
@@ -228,33 +234,55 @@ export const AddNIC = ({
     }
 
     const dlg = useDialogState(init, validate);
+    const have_sources = NetworkTypeAndSource_has_sources(dlg.values.type_and_source);
 
-    const defaultBody = (
-        <>
-            <Form onSubmit={e => e.preventDefault()} isHorizontal>
-                <NetworkTypeAndSourceRow field={dlg.field("type_and_source")} />
+    let defaultBody;
```
Venefilyn: For TS type reasons I'd say we make this a React Node from the start


```suggestion
    let defaultBody = <></>;
```

src/components/networks/networkList.tsx
```diff
@@ -63,7 +79,28 @@ export class NetworkList extends React.Component<NetworkListProps> {
...
+                                             */
+                                            !appState.systemSocketInactive &&
+                                                <EmptyStateFooter>
+                                                    <Button
+                                                        variant="secondary"
+                                                        onClick={createDefaultNetwork}
+                                                    >
+                                                        {_("Create the \"default\" virtual network")}
+                                                    </Button>
+                                                </EmptyStateFooter>
```
Venefilyn: Split to a new component so we can reuse it in `nicsAdd` too. They're currently the same

src/components/vm/nics/nicAdd.tsx
```diff
@@ -228,33 +234,55 @@ export const AddNIC = ({
     }
 
     const dlg = useDialogState(init, validate);
+    const have_sources = NetworkTypeAndSource_has_sources(dlg.values.type_and_source);
```
Venefilyn: `has` and `have` both used here. Correct would be `has`


```suggestion
    const hasSources = NetworkTypeAndSource_has_sources(dlg.values.type_and_source);
```

src/components/vm/nics/nicBody.tsx
Venefilyn:
```suggestion
        if (availableSources.network?.length > 0) {
```

src/components/vm/nics/nicBody.tsx
```diff
@@ -157,7 +157,18 @@ export function init_NetworkTypeAndSourceRow(
...
+    if (network) {
+        // Make sure the current network source is available, no
+        // matter what.
+        if (network.type in _availableSourcesForType) {
+            const avail = _availableSourcesForType[network.type];
+            const src = getNetworkSource(network) || "";
+            if (!avail.includes(src))
+                avail.push(src);
+        }
+    }
```
Venefilyn:
```suggestion
    // Make sure the current network source is available, no
    // matter what.
    if (network && network.type in _availableSourcesForType) {
        const avail = _availableSourcesForType[network.type];
        const src = getNetworkSource(network) || "";
        if (!avail.includes(src))
            avail.push(src);
    }
```

src/components/vm/nics/nicBody.tsx
```diff
@@ -167,16 +178,23 @@ export function init_NetworkTypeAndSourceRow(
...
-        type = vm.connectionName == "session" ? "user" : "network";
+        type = "";
         source = "";
+        for (let i = 0; i < _availableTypes.length; i++) {
+            if (!_availableTypes[i].disabled) {
+                type = _availableTypes[i].name;
+                source = _availableSourcesForType[type][0];
+                break;
+            }
+        }
```
Venefilyn, line 208: Converting it to a `for-of` 
```suggestion
        for (const type of _availableTypes) {
            if (!type.disabled) {
                source = _availableSourcesForType[type.name][0];
                break;
            }
        }
```

src/components/vm/nics/nicBody.tsx
```diff
@@ -187,152 +205,84 @@ export function init_NetworkTypeAndSourceRow(
     };
 }
 
+export function NetworkTypeAndSource_has_sources(value: NetworkTypeAndSourceValue) {
+    return value.type != "";
+}
```
Venefilyn: Add documentation what this function should do. If the type is empty, that correlates to the network type not having sources? Reading this on its own is strange

Naming of the function and `NetworkTypeAndSourceValue` type is also kinda weird if this checks if it has sources to begin with

src/components/vm/nics/nicBody.tsx
```diff
@@ -187,152 +205,84 @@ export function init_NetworkTypeAndSourceRow(
...
     }
 
+    function makeTypeOption(t: NetworkTypeDescription): DialogRadioSelectOption<string> {
+        return {
+            value: t.name,
+            label: t.desc,
+            excuse: t.disabled ? _("no sources") : null,
+            explanation: t.detailHeadline,
+        };
+    }
```
Venefilyn:
```suggestion
    function makeTypeOption(type: NetworkTypeDescription): DialogRadioSelectOption<string> {
        return {
            value: type.name,
            label: type.desc,
            excuse: type.disabled ? _("no sources") : null,
            explanation: type.detailHeadline,
        };
    }
```

src/components/vm/nics/nicBody.tsx
```diff
@@ -187,152 +205,84 @@ export function init_NetworkTypeAndSourceRow(
...
         const sources = _availableSourcesForType[type] || [];
         field.sub("source").set(sources.length > 0 ? sources[0] : "");
     }
 
+    function makeTypeOption(t: NetworkTypeDescription): DialogRadioSelectOption<string> {
+        return {
+            value: t.name,
+            label: t.desc,
+            excuse: t.disabled ? _("no sources") : null,
+            explanation: t.detailHeadline,
```
Venefilyn: How does this look atm? I see that we had "No sources available" previously in `validate_NetworkTypeAndSourceRow`. Do we want to keep that or at least make it capitalized "No sources"?

src/components/vm/nics/vmNicsCard.tsx
```diff
@@ -40,7 +40,8 @@ const getNetworkDevices = async (): Promise<Record<string, NetworkDevice>> => {
 
         const devs: Record<string, NetworkDevice> = {};
         for (const dev of output.trim().split('\n')) {
-            devs[dev] = {};
+            if (dev != "")
```
Venefilyn: JS evalutates "" as false so can be simplified
```
$ "" ? "true" : "false"
"false" 
```

Unless we explicitly want `null` or `undefined`? But then `devs[dev]` wouldn't work anyway..

```suggestion
            if (dev)
```

src/libvirtApi/create-default-network.sh
```diff
@@ -0,0 +1,27 @@
...
+else
+    cat <<EOF | virsh -c qemu:///system net-define /dev/stdin
+<network>
+  <name>default</name>
+  <bridge name='virbr0'/>
+  <forward/>
+  <ip address='192.168.122.1' netmask='255.255.255.0'>
+    <dhcp>
+      <range start='192.168.122.2' end='192.168.122.254'/>
+    </dhcp>
```
Venefilyn: Are we guaranteed `192.168.122.0/24`?

test/check-machines-networks
```diff
@@ -1062,6 +1057,28 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
...
 
+    def testNetworksCreateDefault(self):
+        b = self.browser
+
+        self.login_and_go("/machines")
+        self.waitPageInit()
+
+        # Click on Networks card
+        b.wait_in_text("#card-pf-networks .pf-v6-c-card__header button", "Network")
+        b.click(".pf-v6-c-card .pf-v6-c-card__header button:contains(Network)")
```
Venefilyn: IMO change to OUIA ID for test selectors instead of relying on contains

test/check-machines-networks
```diff
@@ -1062,6 +1057,28 @@ class TestMachinesNetworks(machineslib.VirtualMachinesCase):
...
+
+        # Delete "default" network
+        b.click("#network-default-system-action-kebab")
+        b.click("#delete-network-default-system")
+        b.wait_visible("#delete-resource-modal")
+        b.click("#delete-resource-modal button:contains(Delete)")
```
Venefilyn: Ditto

test/check-machines-nics
```diff
@@ -448,6 +448,32 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
...
+        m.execute('mount -t tmpfs tmpfs /sys/class/net')
+        self.addCleanup(self.machine.execute, "umount /sys/class/net")
+
+        self.login_and_go("/machines")
+        self.waitPageInit()
+        self.goToVmPage("subVmTest1")
+        self.deleteIface(1)  # Remove default vNIC
```
Venefilyn: `deleteIface` expects string
```suggestion
        self.deleteIface("1")  # Remove default vNIC
```

test/check-machines-nics
```diff
@@ -802,8 +830,14 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
             self.browser.wait_in_text(".pf-v6-c-modal-box .pf-v6-c-modal-box__header .pf-v6-c-modal-box__title",
                                       "Add virtual network interface")
 
+        def handle_no_sources(self):
+            if self.expect_no_sources:
+                self.browser.wait_in_text(".pf-v6-c-modal-box", "There are no sources for virtual network interfaces")
+                self.browser.click("""button:contains('Create the "default" virtual network')""")
```
Venefilyn: Change to OUIA ID selector

Venefilyn (changes requested):
Ping me if you have questions about OUIA. PF has `ouiaId` parameter for several components, otherwise you can use `data-ouia-component-id` directly on the DOM element.

Code overall seems fine, but would like to refactor one of the empty states. Also curious about the shell script we send to the backend - feels hardcoded and I wanna know if we won't get edge-cases where the IP would be wrong etc.

mvollmer:
> > * The "interface type" is selected with radio buttons instead of a dropdown. There are only three types and there is enough space to show them all. Instead of a info popover with questionably typography and content, the choices now have brief helper text to explain their purpose.
> 
> So this is tricky UX wise. The general rule is that when you have 2 or 3 items it is better to have a radio button. When you get to 4 items or more it starts taking up quite a lot of space and can then be moved to a dropdown.

Yeah, but now it's only three. :-)

We also have this kind of dialog for LVM2 RAID:

image

It has six radio buttons, and honestly I thing it's still fine. The description under the whole group changes according to which radio you select. We can do this also here to make this more compact.

I don't super like how we explain why something is disabled, with the easy to overlook "(no sources)" addition. I think that would improve as well when we move the description to the bottom. Hmm.

src/components/vm/nics/nicBody.tsx
mvollmer: I find this too tricky. `false?.length` is `undefined` so we would compare undefined to a number.

src/components/vm/nics/nicAdd.tsx
mvollmer: This was changed in https://github.com/cockpit-project/cockpit/pull/22886, specifically for this use here.

src/components/vm/nics/nicAdd.tsx
mvollmer: That is not necessary. TypeScript infers `React.JSX.Element` either way (after the `if`). We can add a type to it maybe: `let defaultBody: React.ReactNode`.

src/components/vm/nics/nicBody.tsx
Venefilyn: The [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) essentially does what your code is already doing. It just makes it a bit easier to read.

If we just want to check that the length is more than `0` then we should only do that and use optional chaining for false-checking

src/components/vm/nics/nicAdd.tsx
Venefilyn: Currently TS infers it to `any` due to it starting `undefined` and getting a value later on. Can definitely use `let defaultBody: React.ReactNode` but you need to add `undefined` to it

```suggestion
    let defaultBody: React.ReactNode|undefined;
```

src/components/vm/nics/nicAdd.tsx
Venefilyn: Must've had outdated deps then, thanks!

src/components/vm/nics/nicBody.tsx
mvollmer: Roger. (But diff as displayed here is weird, no?)

src/components/vm/nics/nicBody.tsx
mvollmer: Yes, goo idea, but there is already a `type` declared that we need to set here, so we can't use "type" as the iteration variable.

src/components/vm/nics/nicBody.tsx
Venefilyn: Yes.. I tried using VSCode for review and change, sorry about that

src/components/vm/nics/nicBody.tsx
mvollmer: Yes, this is confusing!

src/components/vm/nics/nicBody.tsx
mvollmer: Roger.

src/components/vm/nics/nicBody.tsx
mvollmer: Yes, I don't like this part either.  We can be more specific, like "No virtual networks", "No bridges", "No network devices". What about that?

src/components/vm/nics/vmNicsCard.tsx
mvollmer: Roger.

src/libvirtApi/create-default-network.sh
mvollmer: No, this can go wrong... Normally, `/usr/share/libvirt/networks/default.xml` exists, but that might also go wrong.

I _think_ having a button to recreate the "default" network might get some people out of trouble if they have deleted it by accident or lost it some other way, but maybe Cockpit should stay out of it.

The user can already create virtual networks pretty easily, so I'll remove this whole machinery.

test/check-machines-nics
mvollmer: Hmm, it's never called with a string, however, always with a number. So I'd say the type annotation is wrong. Do we not check these mechanically somehow?

test/check-machines-networks
mvollmer: We do this everywhere in this file. Let's talk about OUIA separately, ok?

src/components/networks/networkList.tsx
mvollmer: This code has been removed. :-)

src/components/vm/nics/nicBody.tsx
mvollmer: Ignoring it again. :-P

src/components/networks/networkList.tsx
mvollmer: This has been removed.

src/components/vm/nics/nicAdd.tsx
mvollmer: Roger.

src/components/vm/nics/nicBody.tsx
mvollmer: Documented it.

src/components/vm/nics/nicBody.tsx
mvollmer: Done.

src/libvirtApi/create-default-network.sh
mvollmer: Done.

test/check-machines-nics
mvollmer: Changed the type in machineslib.py

src/components/vm/nics/nicBody.tsx
Venefilyn: Can always expand to `availableType`

src/components/vm/nics/nicAdd.tsx
mvollmer: > Currently TS infers it to any due to it starting undefined and getting a value later on. 

TS infers different types in different places. After the `if`, `defaultBody` is `React.JSX.Element`.  Before it it is `any`, but since this gets narrowed down later, TS will not complain about it. We get errors for things that are truly `any`.

> but you need to add undefined to it

No, that is also not necessary. :-)  `ReactNode` already includes `undefined`, but even the following is fine:

```
    let foo: string;
    if (condition)
        foo = "xxx";
    else
        foo = "yyy";
```

`foo` does not need to be initialized and it's type does not need to include `undefined`. Everything is still sound since TS does not let you use a uninitialized variable.

So, IMO adding explicit types to things is good and let's do it here, but we can also rely on TS doing pretty clever inference. When there are no errors, the code is well typed and well checked, regardless of how many type annotations there are. (Because our config for tsc is strict.)

More type annotations are a help for future modifications of the code, so that mistakes are caught closer to their source, so we should have them when things are not obvious.

src/components/vm/nics/nicBody.tsx
mvollmer: > If we just want to check that the length is more than 0 then we should only do that and use optional chaining for false-checking

Let me try to explain again what my objection is. When `avilableSources.network` evaluates to `undefined`, then `availableSource.network?.length` also evaluates to `undefined`. Consequently, we would be evaluating `undefined > 0` next. 

This does evaluate to `false`, but that's not obvious, and gets us into the dark corners of JS. JS evaluates all kinds stuff without raising errors, like `"foo" > 2`, but we shouldn't be doing that.

TS would actually reject `undefined > 0`, but TS believes that `availableSources.network` is always a `string[]`  and thus always true... which I actually consider a bug in the `Record` type, and I should learn more about why it does that.

Howver, `availableSource.network` should in fact always be present, so I'll go over this section and try to type it better.

mvollmer: > so I'll go over this section and try to type it better.

Oops, that turned into a larger rewrite. :-)

It is now an error to try to edit a interface with an unsupported type. That case was not really handled well earlier, I think. To catch and display that error the "Edit" dialog now uses `useDialogState_async` (bc the regular `useDialogState` doesn't handle errors nicely during initialization). And for good measure, the "Add" dialog now uses the same mechanism to handle the "No sources" situation. That also an error now, and the code is way simpler, imo.

src/components/vm/nics/nicBody.tsx
```diff
@@ -45,78 +43,94 @@ export const NetworkModelRow = ({
...
+            options={availableModelTypes}
+        />
     );
 };
 
+type NetworkType = "network" | "bridge" | "direct" | "user";
+
+export function isSupportedNetworkType(type: string): type is NetworkType {
+    return type == "network" || type == "bridge" || type == "direct" || type == "user";
+}
```
Venefilyn, line 68: To "clean" this up so we don't need to duplicate stuff we can change it around a bit. Currently it doesn't type-check as well cause you could do `type == "ntework"` and it would not throw a type error

Following https://stackoverflow.com/a/51529486 we can do this instead, which although doesn't help us with the above makes it so we don't need to maintain both places and can just use `Array.includes()`
```ts
const networkType = ["network", "bridge", "direct", "user"] as const;
type NetworkType = (typeof networkType)[number];

export function isSupportedNetworkType(type: string): type is NetworkType {
    return networkType.includes((type as NetworkType))
}
```

src/components/vm/nics/nicBody.tsx
```diff
@@ -45,78 +43,94 @@ export const NetworkModelRow = ({
...
+    excuse: string | null,
 }
 
-function getAvailableNetworkTypes(vm: VM, availableSources: AvailableSources) {
+interface AvailableSourcesForType {
+    network: string[];
+    bridge: string[];
+    direct: string[];
+    user: string[];
+}
```
Venefilyn, line 82:
```suggestion
type AvailableSourcesForType = {[key in NetworkType]: string[]}
```

src/components/vm/nics/nicBody.tsx
```diff
@@ -45,78 +43,94 @@ export const NetworkModelRow = ({
...
+
 interface NetworkTypeDescription {
-    name: string,
+    type: NetworkType,
     desc: string,
     detailHeadline?: React.ReactNode,
-    detailParagraph?: React.ReactNode,
-    externalDocs?: React.ReactNode,
-    disabled?: boolean,
+    excuse: string | null,
```
Venefilyn, line 74: Why `excuse: string|null` instead of `excuse?: string`

Venefilyn (approved):
Overall LGTM I think! There are some type suggestions to reduce the duplication and human factor

src/components/vm/nics/nicBody.tsx
mvollmer: Yes, I learned about this just recently! But I don't mind being extra explicit here. If the set of supported network types is expanded, we have a lot of code churn anyway.

src/components/vm/nics/nicBody.tsx
mvollmer: No strong reason, just that this is enough for the current uses of this interface. If we make it optional, I would still  add the "| null" so we can write

```
    excuse: availableSources.bridge.length == 0 ? _("No bridge devices") : null
```

instead of

```
    ...(_availableSources.bridge.length == 0 ? { excuse:  _("No bridges") } : { })
````

which I find more cumbersome and less readable.

Venefilyn: Would it not be possible to write `: undefined`?

PR #2444, opened 2025-12-08
consoles: Make console selector scrollable

mvollmer:
when it is a SimpleSelect instead of a set of toggle buttons. Otherwise the menu might not be useable when it is truly huge and starts overflowing the page.

Fixes RHEL-106754, again.


mvollmer:
Before:

image

After:

image

Imagine a scrollbar inside the menu in the after picture.

PR #2471, opened 2026-01-04
Rebuild our bundles during package build

martinpitt:
All of these are ported from c-podman, so nothing new.


mvollmer (approved):
Thanks!

PR #2465, opened 2025-12-29
Xmas Cleanups

mvollmer:

PR #2477, opened 2026-01-07
xml: Redo all the parsing with new helper functions

mvollmer:
- [x] #2465 

This is the kind of API that I like, and IMO it is easy enough to implement.

I have unfairly dismissed XPath in the past, probably because the docs that I found made it look very complex. I can be convince to give it another shot. But we really need to get rid of the 50 copies of `getElementsByTagName("foo")[0].childNodes[0].nodeValue` as well.

Also see the thread in https://github.com/cockpit-project/cockpit-machines/pull/2465#discussion_r2667362922

### DIFF
--- a/src/libvirtApi/domain.ts
+++ b/src/libvirtApi/domain.ts
@@ -982,7 +982,7 @@ function domainGetCapabilities({
     arch: optString,
     model: optString,
 }): Promise<DomainCapabilities> {
-    async function get() {
+    async function get(): Promise<DomainCapabilities> {
         const [capsXML] =
             await call<[string]>(
                 connectionName,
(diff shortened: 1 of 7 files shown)


src/libvirtApi/domain.ts
```diff
@@ -1001,6 +1001,7 @@ function domainGetCapabilities({
             supportsTPM: getDomainCapSupportsTPM(domCaps),
             interfaceBackends: getDomainCapInterfaceBackends(domCaps),
         };
+        console.log("CAPS", caps);
```
martinpitt: If you want to keep this, can you make this a `logDebug()`?

mvollmer: No, forgot to remove it, thanks!

mvollmer:
Oh, das war ja einfach! :-)

PR #2449, opened 2025-12-11
disks: Redo "Add disk" and "Insert media" dialog with new framework

mvollmer:
This untangles the complex logic in the "Add disk" dialog. It keeps the three modes cleanly separate and avoids overlapping their states and logic. For example, the "format" handling for the "Custom path" mode is different from that of "Use existing", but it was hard to see which code is for which.

PR #2478, opened 2026-01-07
app: Move AppState into its own file and make it global

mvollmer:
There are plans to make it a lot bigger. Doing the move now makes it easier to maintain long running branches that add to it.

Since the store is global, we make the state global too so that it is easy to use the AppState where we now use the store.  Now you know where this is going.

### DIFF
--- a/src/components/vm/disks/diskAdd.tsx
+++ b/src/components/vm/disks/diskAdd.tsx
@@ -49,7 +49,7 @@ import { VolumeCreateBody } from '../../storagePools/storageVolumeCreateBody.jsx
 import { domainAttachDisk, domainGet, virtXmlHotEdit, domainIsRunning } from '../../../libvirtApi/domain.js';
 import { storagePoolGetAll } from '../../../libvirtApi/storagePool.js';
 import { storageVolumeCreateAndAttach } from '../../../libvirtApi/storageVolume.js';
-import { useAppState } from '../../../app';
+import { appState } from '../../../state';
 
 const _ = cockpit.gettext;
 
@@ -492,7 +492,6 @@ export const AddDiskModalBody = ({
     vm: VM,
     supportedDiskBusTypes: string[],
 }) => {
-    const appState = useAppState();
     const [customDiskVerificationFailed, setCustomDiskVerificationFailed] = useState(false);
     const [customDiskVerificationMessage, setCustomDiskVerificationMessage] = useState<string | null>(null);
     const [dialogError, setDialogError] = useState<string | null>(null);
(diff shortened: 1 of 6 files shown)


src/components/vm/disks/diskAdd.tsx
```diff
@@ -546,7 +545,7 @@ export const AddDiskModalBody = ({
     useEffect(() => {
         // Get the list of VMs.
         appState.getVms().then(setVms);
-    }, [appState]);
+    }, []);
```
martinpitt, line 548: So was that never actually meant to be a continuous "get VMs whenever app state changes", but always a `useInit()`? If so, does that have an implicit assumption that modal bodies and other components which use appState aren't instantiated until the appState is initialized? That could do with a `cockpit.assert()`.

If that is correct, would you mind actually using `useInit()` for clarity and grep-ability? (you share my deep hate of `useEffect()` now, it seems! 😁 )

mvollmer: > So was that never actually meant to be a continuous "get VMs whenever app state changes", but always a `useInit()`? 

Yes, this is supposed to be a initialization action that runs when the modal opens.
I think I used `useEffect` just to blend in with the 6 `useEffect` uses below it. :-)

The real cleanup comes in https://github.com/cockpit-project/cockpit-machines/pull/2449, with the new Dialog Implementation Convenience Kit. With that, initializing a dialog can be any [async function](https://github.com/mvollmer/cockpit-machines/blob/bae37c6c2b29aa4730b1eaaa3a8534f4e4a775bd/src/components/vm/disks/diskAdd.tsx#L496), without any hook magic.

(It's a interesting question in general whether dialogs should follow along with changes in the system, and how that should look exactly. Right now, the dialogs don't react and we trust that we get useful errors during apply when, for example, the VM that we try to add a disk to has been deleted in the mean time.)

> If so, does that have an implicit assumption that modal bodies and other components which use appState aren't instantiated until the appState is initialized? That could do with a `cockpit.assert()`.

The `appState` object itself is guaranteed to be functional at all times. However, the actual application data such as the list of `VM` objects will only be initialized on demand. That's what this code is for, actually: `appState.getVms()` is a async function that triggers getting the list of VMs from libvirt and then waits for that to be done. (The dialog needs to know about all these VMs because it needs to know which pool volumes are used by which VMs in order to warn about using one that is already in use.)

> If that is correct, would you mind actually using `useInit()` for clarity and grep-ability?

Yes, will do.

> (you share my deep hate of `useEffect()` now, it seems! 😁 )

Oh, I always have. :-) But it's a necessary fundamental feature that React absolutely needs to have, so I feel it's unfair to bash useEffect when we really should be bashing our naive and excessive use of it. 

 (Well, hmm, useRef plus some way to run a function when a component is unmounted would probably be enough...)

mvollmer: > > If that is correct, would you mind actually using `useInit()` for clarity and grep-ability?
> 
> Yes, will do.

Actually, since you have approved and this is going away completely when the big plan moves forward, I'll keep it as it is...

PR #2480, opened 2026-01-09
xml: Clean up network parsing

mvollmer:
This was missed in the recent big parsing rewrite.

- Use the new helpers.

- Make it type clean.

- Rename interface.interface.dev to forward.dev


martinpitt (approved):
Thanks!

PR #2484, opened 2026-01-12
api: Don't poll VM usage in domainGetAll

### LINKED ISSUE #2428: domainGetAll() invokes `pollUsageNow` twice for virtual machines with multiple libvirtd connections
Calling `domainGetAll` multiple times invokes `pollUsageNow` every time which calls `pollVmUsage` for every vm in the redux state. So when first calling `domainGetAll` with the  "system" connection, and then the "user" connection with `pollVmUsage` will be invoked twice for all `system` virtual machines.

```
export async function domainGetAll({ connectionName } : { connectionName: ConnectionName }): Promise<void> {
    try {
        const [objPaths] = await call<[string[]]>(connectionName, '/org/libvirt/QEMU', 'org.libvirt.Connect', 'ListDomains', [0],
                                                  { timeout, type: 'u' });
        store.dispatch(deleteUnlistedVMs(connectionName, [], objPaths));
        await Promise.all(objPaths.map(path => domainGet({ connectionName, id: path })));
        pollUsageNow();
    } catch (ex) {
        console.warn('GET_ALL_VMS action failed:', String(ex));
        throw ex;
    }
}
```

mvollmer:
The calls to ensureUsagePolling already make sure that this is done during initialization.

Fixes #2428


martinpitt (approved):
Cheers!

PR #2485, opened 2026-01-12
install: Work around virt-install bug #975

mvollmer:
See https://github.com/virt-manager/virt-manager/issues/975


mvollmer:
I don't know where this bug is right now. It got fixed in Fedora 43 just a few days ago. Maybe we don't need this workaround anymore.

mvollmer:
Let's not bother unless we find places with the bug that can't update virt-install.

PR #2491, opened 2026-01-14
filesystems: Allow adding and removing while the VM is running

### LINKED ISSUE #1898: Adding a file share to a guest while vm is running works.
According to our [old broken test TestMachinesFilesystem.testBasic] (https://github.com/cockpit-project/cockpit-machines/blob/323/test/check-machines-filesystems#L134) starting a VM while the "add new shared filesystem" should disable the create button but that is not the case.

The test succeeded as the old Cockpit File component did trigger an invalid state and now it doesn't. Regardless of that issue, opening the dialog, configuration a valid path and then starting a vm via `virsh start subVmTest1` and then pressing "add" does add the share to the VM (without any errors)

So this actually works, but maybe we need to change the current implementation:

* Verify if we can add shares on the fly
* Allow adding shares on the fly and warn that a VM needs to be restarted
* Update the dialog state when a VM is started?

Alternatively we just ignore this case, from Cockpit's view you can't start a VM while adding a snapshot.

mvollmer:
This works now.

Fixes #1898 

- [x] https://github.com/cockpit-project/bots/pull/8649

### DIFF
--- a/src/components/vm/filesystems/vmFilesystemsCard.tsx
+++ b/src/components/vm/filesystems/vmFilesystemsCard.tsx
@@ -39,7 +39,7 @@ import { ModalError } from 'cockpit-components-inline-notification.jsx';
 import { FileAutoComplete } from "cockpit-components-file-autocomplete.jsx";
 import { useDialogs } from 'dialogs.jsx';
 
-import { virtXmlAdd, virtXmlRemove, virtXmlEdit } from "../../../libvirtApi/domain.js";
+import { virtXmlHotAdd, virtXmlHotRemove, virtXmlEdit } from "../../../libvirtApi/domain.js";
 import { vmId } from "../../../helpers.js";
 import { DeleteResourceButton } from '../../common/deleteResource.jsx';
 import { InfoPopover } from '../../common/infoPopover.jsx';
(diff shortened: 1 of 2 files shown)


mvollmer:
The failure on ubuntu-2404 comes with a libvirt segfault:

```
Jan 20 09:15:00 ubuntu kernel: rpc-libvirtd[8867]: segfault at 188 ip 000074b8bcad8973 sp 000074b8b71fb660 error 4 in libvirt.so.0.10000.0[74b8bca9f000+2cb000] likely on CPU 0 (core 0, socket 0)
Jan 20 09:15:00 ubuntu kernel: Code: 0f 1f 80 00 00 00 00 f3 0f 1e fa e9 77 ff ff ff 0f 1f 80 00 00 00 00 55 48 89 e5 41 55 41 89 d5 41 54 41 89 f4 53 48 83 ec 08 <48> 8b 9f 88 01 00 00 48 85 db 48 0f 44 df 48 8b 03 48 85 c0 74 0d
```

martinpitt (commented):
Cheers! LGTM, but that [Ubuntu apparmor/crash thing](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2491-09c65b0f-20260120-090552-ubuntu-2404/TestMachinesFilesystems-testBasicSessionConnection-ubuntu-2404-127.0.0.2-2201-FAIL.log.gz) needs Some Treatment™?

mvollmer:
> Cheers! LGTM, but that [Ubuntu apparmor/crash thing](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2491-09c65b0f-20260120-090552-ubuntu-2404/TestMachinesFilesystems-testBasicSessionConnection-ubuntu-2404-127.0.0.2-2201-FAIL.log.gz) needs Some Treatment™?

Yes, a naughty I guess...

PR #2495, opened 2026-01-15
api: Read qemu process start time from /proc

### LINKED ISSUE #1871: No uptime when a libvirt group user operate the system connection VM
Description of problem:
When a libvirt group user try to operate the system connection VM, there will be no uptime shown on the pop-up dialog

cockpit-322-1.fc40.x86_64
cockpit-machines-316-1.fc40.noarch
libvirt-dbus-1.4.1-4.fc40.x86_64

How reproducible:
100%

Steps to Reproduce:
1 Create a libvirt group user, then create a system connection VM
2 Log in with the libvirt group user, then try to operate the system connection VM

Actual results:
1 No uptime shown on the dialog after the step 2

Expected results:
1 There should be the uptime shown on the dialog or a hint for clarifying

Additional info:
It seems that libvirt group users may not have enough privilege for accessing the log in /var/log/libvirt/qemu

mvollmer:
Instead of reading libvirt log files to determine VM start time, use
"ps" to get the process creation time.  The PID files are accessible
also for non-admin users in the libvirt group, unlike the logs.
    
Fixes #1871

### DIFF
--- a/src/libvirtApi/domain.ts
+++ b/src/libvirtApi/domain.ts
@@ -122,6 +122,11 @@ function script(connectionName: ConnectionName, script: string): cockpit.Spawn<s
         });
 }
 
+function shlex_quote(str: string): string {
+    // yay, command line apis...
+    return "'" + str.replaceAll("'", "'\"'\"'") + "'";
+}
+
 /* Running virt-xml
 
    The virtXmlAdd, virtXmlEdit, etc family of functions can be used to
@@ -208,11 +213,6 @@ async function runVirtXml(
     // We don't pass the arguments for virt-xml through a shell, but
     // virt-xml does its own parsing with the Python shlex module. So
     // we need to do the equivalent of shlex.quote here.
-
-    function shlex_quote(str: string): string {
-        // yay, command line apis...
-        return "'" + str.replaceAll("'", "'\"'\"'") + "'";
-    }
 
     function encode_into(args: string[], key: string, val: unknown) {
         if (typeof val == "number" || typeof val == "string" || val === true) {
@@ -1005,31 +1005,30 @@ export async function domainGetStartTime({
     connectionName: ConnectionName,
     vmName: string,
 }): Promise<Date | null> {
-    const loggedUser = await cockpit.user();
+    const quotedName = shlex_quote(vmName);
+    let pidFile;
+    if (connectionName == "system")
+        pidFile = `/run/libvirt/qemu/${quotedName}.pid`;
+    else
+        pidFile = `\${XDG_RUNTIME_DIR:/run/user/$UID}/libvirt/qemu/run/${quotedName}.pid`;
 
-    /* The possible paths of logfiles path, as of libvirt 9.3.0 are defined in:
-        * https://gitlab.com/libvirt/libvirt/-/blob/v9.3.0/src/qemu/qemu_conf.c#L153
-        * There libvirt defines 3 possible directories where they can be located
-        * - /root/log/qemu/ - That is however only relevant for "qemu:///embed", and not cockpit's use case
-        * - /var/log/libvirt/qemu/ - For privileged (qemu:///system) VMs
-        * - ~/.cache/libvirt/qemu/logs - For non-privileged  (qemu:///session) VMs
-        */
-    const logFile = connectionName === "system" ? `/var/log/libvirt/qemu/${vmName}.log` : `${loggedUser.home}/.cache/libvirt/qemu/log/${vmName}.log`;
+    // We assume that the birth time of the PID file is (very close to)
+    // the start time of the QEMU process for the machine. Reading the
+    // actual start time out of /proc is too hairy and "ps
+    // --date-format" is not portable enough.
+    //
+    // If the birth time is not available, we use the meta data change
+    // time.
+
+    const cmd = `stat -c %W:%Z ${pidFile}`;
 
     try {
-        // Use libvirt APIs for getting VM start up time when it's implemented:
-        // https://gitlab.com/libvirt/libvirt/-/issues/481
-        const line = await script(connectionName, `grep ': starting up' '${logFile}' | tail -1`);
-
-        // Line from a log with a start up time is expected to look like this:
-        // 2023-05-05 11:22:03.043+0000: starting up libvirt version: 8.6.0, package: 3.fc37 (Fedora Project, 2022-08-09-13:54:03, ), qemu version: 7.0.0qemu-7.0.0-9.fc37, kernel: 6.0.6-300.fc37.x86_64, hostname: fedora
-
-        // Alternatively regex line.match(/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/g) can be used
-        const timeStr = line.split(': starting up')[0];
-        const date = new Date(timeStr);
-        return isNaN(date.getTime()) ? null : date;
+        const result = await script(connectionName, cmd);
+        const [btime, ctime] = result.trim().split(":")
+                .map(s => parseInt(s));
+        return new Date((btime || ctime) * 1000);
     } catch (ex) {
-        console.log("Unable to detect domain start time:", ex);
+        console.log("Unable to detect domain start time:", String(ex));
         return null;
     }
 }
--- a/test/check-machines-lifecycle
+++ b/test/check-machines-lifecycle
@@ -209,7 +209,8 @@ class TestMachinesLifecycle(machineslib.VirtualMachinesCase):
         b.click("#vm-subVmTest1-system-run")
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
         self.waitGuestBooted(args['logfile'])
-        self.performAction("subVmTest1", "off")
+        with b.wait_timeout(30):
+            self.performAction("subVmTest1", "off")
 
         # force shut off
         b.click("#vm-subVmTest1-system-run")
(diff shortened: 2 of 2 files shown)


mvollmer:
Claude got the location of session pid files slightly wrong, and messed up timezones.

src/scripts/get_vm_start_time.py
```diff
@@ -0,0 +1,87 @@
...
+        # Calculate process start time in seconds since boot
+        start_time_since_boot = start_time_ticks / clock_ticks_per_sec
+
+        # Calculate absolute start time
+        boot_time = time.time() - uptime_seconds
+        absolute_start_time = boot_time + start_time_since_boot
+
+        return absolute_start_time
+
+    except Exception as e:
```
martinpitt: This is sadly Claude's very common anti-pattern... In this case ignoring the exception is halfway ok, but not a blanket `Exception` (`OSError` sounds right), and not that widely scoped.

mvollmer: I have adjusted this. Only catch OSError and no catching in main.

src/libvirtApi/domain.ts
```diff
@@ -1005,29 +996,22 @@ export async function domainGetStartTime({
...
-    const logFile = connectionName === "system" ? `/var/log/libvirt/qemu/${vmName}.log` : `${loggedUser.home}/.cache/libvirt/qemu/log/${vmName}.log`;
-
     try {
-        // Use libvirt APIs for getting VM start up time when it's implemented:
-        // https://gitlab.com/libvirt/libvirt/-/issues/481
-        const line = await script(connectionName, `grep ': starting up' '${logFile}' | tail -1`);
+        const args = JSON.stringify({
+            connectionName,
+            vmName,
+        });
```
martinpitt: That feels a bit excessive -- why not pass these as plain positional values?

src/scripts/get_vm_start_time.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Get the start time of a VM by reading its qemu process PID from libvirt
+    and then calculating the start time from /proc.
+
+    Returns the start time as seconds since the epoch, or None if unable to determine.
+    """
+    try:
+        # Determine the PID file path based on connection type
+        if connection_name == "system":
+            pid_file = f"/var/run/libvirt/qemu/{vm_name}.pid"
```
martinpitt: `/run/` please.

src/scripts/get_vm_start_time.py
```diff
@@ -0,0 +1,75 @@
...
+        # Get system uptime
+        with open('/proc/uptime', 'r') as f:
+            uptime_seconds = float(f.read().split()[0])
+
+        # Calculate process start time in seconds since boot
+        start_time_since_boot = start_time_ticks / clock_ticks_per_sec
+
+        # Calculate absolute start time
+        boot_time = time.time() - uptime_seconds
+        absolute_start_time = boot_time + start_time_since_boot
```
martinpitt: Ugh -- can't we just get `os.stat(f"/proc/{pid}/stat").st_ctime` ?

Then invoking a whole python interpreter for this seems excessive -- it's just one read and one fsinfo channel.

src/libvirtApi/domain.ts
mvollmer: Yes, I agree.  The rest of c-machines does it like this for other programs like install_machine.py, so Claude copied the pattern. I'll change it.

src/scripts/get_vm_start_time.py
mvollmer: Yes, great idea and would be nice... but it's not documented, is it?  But it makes so much sense....

martinpitt: Could we call `ps --date-format '%s' -o lstart=`? Not sure if we already transitively depend on the `procps` package..

mvollmer: Also a good idea.

mvollmer: We could also just read the ctime or btime of /run/libvirt/qemu/name.pid...

martinpitt: @mvollmer that's a much better idea!

mvollmer: > We could also just read the ctime or btime of /run/libvirt/qemu/name.pid...

I did that, but left the get_vm_start_time script in place instead of using fsinfo. The latter doesn't report ctime, and I couldn't quickly see how to easily get the XDG_RUNTIME_DIR environment variable into JavaScript.

mvollmer: (If we keep the python script, we should also keep the much more robust /proc/pid/stat parsing approach. It's not any more expensive, really.)

martinpitt: No, but that could at least be a `cockpit.script()` oneliner instead of an expensive Python script. We already know on the JS side if that's a system or a user connection, so it can invoke `stat` on the right one.

martinpitt: > (If we keep the python script, we should also keep the much more robust /proc/pid/stat parsing approach. It's not any more expensive, really.)

The delta cost is negligible indeed, I'm worried about robustness. Like how does `os.sysconf_names['SC_CLK_TCK'])` behave on other architectures etc. Is all of that platform independent?

mvollmer:
I have pushed a commit that uses the `ps -o lstart=` method. I like that best.

src/libvirtApi/domain.ts
```diff
@@ -1005,29 +995,27 @@ export async function domainGetStartTime({
...
 
-    /* The possible paths of logfiles path, as of libvirt 9.3.0 are defined in:
-        * https://gitlab.com/libvirt/libvirt/-/blob/v9.3.0/src/qemu/qemu_conf.c#L153
-        * There libvirt defines 3 possible directories where they can be located
-        * - /root/log/qemu/ - That is however only relevant for "qemu:///embed", and not cockpit's use case
-        * - /var/log/libvirt/qemu/ - For privileged (qemu:///system) VMs
-        * - ~/.cache/libvirt/qemu/logs - For non-privileged  (qemu:///session) VMs
-        */
-    const logFile = connectionName === "system" ? `/var/log/libvirt/qemu/${vmName}.log` : `${loggedUser.home}/.cache/libvirt/qemu/log/${vmName}.log`;
+    const cmd = `ps --date-format %s -o lstart= $(cat ${pidFile})`;
```
martinpitt: (Ugh, `cat` -- what happened to just stating the pid file?)

Please quote the pidfile argument. It *should* not contain spaces, but eh..

martinpitt: FTR: removing `procps` from our VM leaves all the libvirt stack intact. In other words, if we use `ps` we need to prepare for it not being there. We do handle errors, and this isn't an important feature, so it's 👍

src/scripts/get_vm_start_time.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
```
martinpitt: This can be removed?

mvollmer: Yes, indeed.

src/libvirtApi/domain.ts
mvollmer: We want `pidFile` to be interpreted by the shell, to expand `$XDG_RUNTIME_DIR` etc.  The only unknown part is `vmName`, which is properly quoted.

mvollmer: Thanks for checking!

martinpitt: Yes, I know about XDG_RUNTIME_DIR - my question was why we can't just call `stat -c %W ${pidFile}`

mvollmer: Because that makes the assumption that the btime of the pid file is the creation time of the process. I think that's a very good and robust assumption, but still only an assumption. On the other hand `lstart` is exactly what we want, so why not use that?  Bc `ps` might not be installed?

martinpitt: Yes, and lots more moving parts. I don't have a too strong opinion about this, just wanted to discuss all the options first. But if you like this, good for me!

martinpitt (dismissed):
Thanks!

martinpitt:
Wait, lots of missing `#uptime` failures -- that approach might not actually work on Ubuntu and RHEL 8/9?

mvollmer:
> Wait, lots of missing `#uptime` failures -- that approach might not actually work on Ubuntu and RHEL 8/9?

Yes, ps is too old... Seems like only procps-ng 4 implements the -D/--date-format option... argh. How can Linux be that bad.

src/libvirtApi/domain.ts
mvollmer: > Bc `ps` might not be installed?

Or not support `--date-format`...

mvollmer:
Grr, `stat -c %W` always outputs 0 on rhel-8-10 and ubuntu-2204...

test/check-machines-lifecycle
```diff
@@ -220,7 +209,9 @@ class TestMachinesLifecycle(machineslib.VirtualMachinesCase):
         b.click("#vm-subVmTest1-system-run")
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
         self.waitGuestBooted(args['logfile'])
-        self.performAction("subVmTest1", "off")
+        print("booted")
```
martinpitt: debugging leftover?

mvollmer: oh, yes.

mvollmer:
> Sure, but we still need a naughty? Or is that an unrelated flake? I've seen it fail twice here now.

I think it's unrelated... The naughty is needed in #2491.

martinpitt:
Ah right, mixed up the PRs. Thanks!

PR #2513, opened 2026-01-26
test: wait on the vm detail page to load

jelly:
The curious flake:

https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2512-b79922ee-20260126-034701-arch/log.html

Seems like a misclick? Which is strange as we click by `id` but lets wait on page to load first, as we touched the test it should run retries.

### DIFF
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -228,6 +228,8 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         def plugAndUnplugNICs():
             mac_address = b.text("#vm-subVmTest1-network-1-mac")
 
+            # Edit dialog loads asynchronous, wait on it to load first so we don't misclick.
+            b.wait_visible("#vm-subVmTest1-network-1-edit-dialog:not(:disabled)")
             b.wait_in_text("#vm-subVmTest1-network-1-state", "up")
             b.wait_in_text("#vm-subVmTest1-iface-1-unplug", "Unplug")
 


mvollmer:
> Seems like a misclick? Which is strange as we click by `id`

Yeah, looks like the "Unplug" button has moved away and the "Edit" button has moved into its place while the test clicks on it. The "Edit" button does indeed appear asynchronously, after running some `find /sys/class/net ...` command.

So we could wait for the "Edit" button to appear?

test/check-machines-nics
```diff
@@ -222,6 +222,7 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         self.waitVmRow("subVmTest1")
         b.wait_in_text("#vm-subVmTest1-system-state", "Running")
         self.goToVmPage("subVmTest1")
+        b.wait_visible("#vm-subVmTest1-overview")
```
mvollmer, line 225: I think it is better to wait for the "Edit" button to appear next to the "Unplug" button that we want to click on, as that is the thing that messes up the layout and causes the misclick.  I think waiting for the overview does not change anything, or does it?

jelly:
> > Seems like a misclick? Which is strange as we click by `id`
> 
> Yeah, looks like the "Unplug" button has moved away and the "Edit" button has moved into its place while the test clicks on it. The "Edit" button does indeed appear asynchronously, after running some `find /sys/class/net ...` command.
> 
> So we could wait for the "Edit" button to appear?

Let's do that!

jelly:
Ignore the rawhide build failure, its unrelated.

test/check-machines-nics
jelly: Absolutely right, adjusted the test and kept this in as we never assert that we landed on the overview page, we just `b.go()` in `goToVmPage`.

martinpitt:
@jelly I think you forgot to re-push?

jelly:
> @jelly I think you forgot to re-push?

Repush where?

PR #2514, opened 2026-01-27
Resolve rawhide nodejs packaging changes

jelly:


khardix:
Clarification: The intention would be to require *both* `nodejs` and `%{_bindir}/node`; e.g. `Buildrequires: nodejs, %{_bindir}/node`. The first one select which "stream" you want (in this case, any one available), the other one that you want non-versioned `node` command from that stream.

The installation of `/usr/bin/node` should be solved transparently for the end-user via weak dependencies, but apparently koji does not do those, thus downstream packages will need to be explicit here. Sorry!

khardix:
@jelly @martinpitt It was later [pointed to me](https://pagure.io/packaging-committee/pull-request/1521#comment-227927) that the guidelines prescribe using `/usr/bin/node`, and that `%{_bindir}/node` is discouraged or even forbidden: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_file_and_directory_dependencies

So the original PR was actually the most correct one. There is now proposed [change](https://pagure.io/packaging-committee/pull-request/1521) to the guidelines that should clarify all that, but as one of the first people that have run into this, you get to see the whole mitigation dance. Sorry again!

martinpitt:
Thanks @khardix ! I updated that in https://github.com/cockpit-project/starter-kit/pull/1349 , and from there will propagate to the other projects.

PR #2515, opened 2026-01-27
vm: Expose life-cycle events and do good things with them.

mvollmer:
Previously we would reset all consoles to be "connected" when a VM was
shut off. Now we do this when a VM starts running, which results in the VNC console autoconnecting after reverting a snapshot.

The journey to this result was a bit convoluted and we had to expose libvirt life-cycle events, which allowed a very nice cleanup for the "onShutoff" feature.

Fixes RHEL-135549

### DIFF
--- a/src/components/vm/vmDetailsPage.tsx
+++ b/src/components/vm/vmDetailsPage.tsx
@@ -72,11 +72,13 @@ export const VmDetailsPage = ({
         };
     }, [vm.uuid]);
 
-    // We want to clear any VNC connection errors when the machine is shut off
+    // We want to autoconnect the VNC console when a machine starts (or is resumed).
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


mvollmer:
This interacts badly with how VncConsole wraps NoVNC. It's time to pull all the NoVNC state into VncState and manage it directly without hooks.

mvollmer:
> This interacts badly with how VncConsole wraps NoVNC. It's time to pull all the NoVNC state into VncState and manage it directly without hooks.

Nah, that's too much code churn. I think reconnecting after a VM has been paused is a better approach.

mvollmer:
Alright, after a couple more rounds around the block, I think I found a good way which also allows us to get rid of the hackish "onShutoff" event thing.

PR #2517, opened 2026-01-28
create: support virt-install --extra-args

Nykseli:
Hi! Was asked to add support for `virt-install --extra-arguments`. Submitting this in case this is a feature that you want to include as well.

Installers like Agama and Anaconda support variety of kernel arguments like inst.* which are sometimes needed for custom installations.

virt-install --extra-args are only supported when using the --location option, which is taken into an account to avoid confusing errors for the end user.

--------------------

Machines: Support for kernel boot arguments when creating VMs

Cockpit now supports the `--extra-args` option of `virt-install`. This allows you to pass arguments to the kernel when the installer is booted during the creation of a new virtual machine. Installers like Anaconda and Agama can be controlled in various ways via these arguments.

Thanks to [Nkyseli](https://github.com/Nykseli) for this contribution!

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -65,6 +65,15 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                             create_and_run=False,
                                                             pixel_test_tag="url"))
 
+        # boot argument check
+        runner.cancelDialogTest(TestMachinesCreate.VmDialog(self, sourceType='url',
+                                                            location=config.VALID_URL,
+                                                            memory_size=128, memory_size_unit='MiB',
+                                                            os_name=config.FEDORA_28,
+                                                            extra_arguments="foo=bar",
+                                                            create_and_run=False,
+                                                            pixel_test_tag="extra-args-url"))
+
         # OS input check
         runner.checkOsInputTest(TestMachinesCreate.VmDialog(self))
 
@@ -514,6 +523,15 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                                   os_name=config.FEDORA_28,
                                                                   os_short_id=config.FEDORA_28_SHORTID))
 
+        # test with --extra-args
+        runner.createDownloadAnOSTest(TestMachinesCreate.VmDialog(self, sourceType='os',
+                                                                  expected_memory_size=128,
+                                                                  expected_storage_size=128,
+                                                                  os_name=config.FEDORA_28,
+                                                                  os_short_id=config.FEDORA_28_SHORTID,
+                                                                  extra_arguments="foo=bar",
+                                                                  create_and_run=True))
+
         # name already used from a VM that is currently being created
         # https://bugzilla.redhat.com/show_bug.cgi?id=1780451
         runner.createDownloadAnOSTest(TestMachinesCreate.VmDialog(self, name='existing-name', sourceType='os',
@@ -820,6 +838,20 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                       storage_pool=NEW_VOLUME_RAW,
                                                       create_and_run=True,))
 
+        # test with --extra-args
+        dialog = TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                      name="extra-args-file",
+                                                      location=config.TREE_FILE,
+                                                      memory_size=128, memory_size_unit='MiB',
+                                                      storage_size=128, storage_size_unit='MiB',
+                                                      extra_arguments="foo=bar",
+                                                      delete=False,
+                                                      create_and_run=True)
+        runner.createTest(dialog)
+        testlib.wait(lambda: "--extra-args foo=bar" in self.machine.execute("ps aux | grep '[v]irt-install'"))
+        runner._deleteVm(dialog)
+        runner.checkEnvIsEmpty()
+
     def testCreateImportDisk(self):
         runner = TestMachinesCreate.CreateVmRunner(self)
         config = TestMachinesCreate.TestCreateConfig
@@ -898,6 +930,20 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                       storage_size=128, storage_size_unit='MiB',
                                                       create_and_run=False))
 
+        # test with --extra-args
+        dialog = TestMachinesCreate.VmDialog(self, sourceType='url',
+                                             name="extra-args-url",
+                                             location=config.TREE_URL,
+                                             memory_size=128, memory_size_unit='MiB',
+                                             storage_size=128, storage_size_unit='MiB',
+                                             extra_arguments="foo=bar",
+                                             delete=False,
+                                             create_and_run=True)
+        runner.createTest(dialog)
+        testlib.wait(lambda: "--extra-args foo=bar" in m.execute("ps aux | grep '[v]irt-install'"))
+        runner._deleteVm(dialog)
+        runner.checkEnvIsEmpty()
+
         # This functionality works on debian only because of extra qemu-block-extra dep.
         # Check error is returned if dependency is missing
         if m.image.startswith("debian"):
@@ -944,6 +990,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
         PATH_WITH_SPACE = '/var/lib/libvirt/novell with spaces.iso'
         ISO_URL = 'https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso?foo=bar'
         TREE_URL = 'https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/28/Server/x86_64/os'
+        TREE_FILE = '/var/lib/libvirt/pub/archive/fedora/linux/releases/28/Server/x86_64/os/'
 
         # LINUX can be filtered if 3 years old
         OLD_FILTERED_OS = 'Red Hat Enterprise Linux 8.3 (Ootpa)'
@@ -995,6 +1042,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                      connection="system",
                      offline_token=None,
                      offline_token_autofilled=True,
+                     extra_arguments=None,
                      pixel_test_tag=None):
 
             TestMachinesCreate.VmDialog.vmId += 1  # This variable is static - don't use self here
@@ -1049,6 +1097,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
             self.name_generated = name_generated
             self.offline_token = offline_token
             self.offline_token_autofilled = offline_token_autofilled
+            self.extra_arguments = extra_arguments
 
             self.pixel_test_tag = pixel_test_tag
 
@@ -1337,6 +1386,8 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                     if self.user_login:
                         user_login = virt_install_cmd_out.split("user-login=", 1)[1].split(",")[0].rstrip()
                         self.assertIn(user_login, self.user_login)
+                if self.extra_arguments:
+                    self.assertIn(f"--extra-args {self.extra_arguments}", virt_install_cmd_out)
                 return True
 
             testlib.wait(virt_install_correct, tries=180)
@@ -1470,6 +1521,18 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                     b.wait_visible("#create-and-edit-disabled-tooltip")
                 else:
                     b.wait_visible("#create-and-run:not([aria-disabled=true])")
+
+                if self.sourceType == "cloud":
+                    b.wait_not_present("#extra-argument")
+                else:
+                    b.wait_visible("#extra-argument")
+
+            if self.extra_arguments:
+                if (self.sourceType in ["file", "url"] and not self.sourceType.endswith(".iso")) \
+                or self.sourceType == "os":
+                    b.click("#pf-tab-1-automation")
+
+                b.set_input_text("#extra-argument", self.extra_arguments)
 
             return self
 
@@ -2090,7 +2153,8 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
             if dialog.sourceType != 'cloud':
                 self.browser.mouse(".pf-v6-c-tabs__list button[aria-disabled=true]:contains(Automation)",
                                    "mouseenter")
-                tooltip_text = "Automated installs are only available when downloading an image or using cloud-init."
+                tooltip_text = "Automated installs are only available when downloading an image,"
+                tooltip_text += " an install tree or using cloud-init."
                 self.browser.wait_in_text("div.pf-v6-c-tooltip__content", tooltip_text)
                 self.browser.mouse(".pf-v6-c-tabs__list button[aria-disabled=true]:contains(Automation)",
                                    "mouseleave")


src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1035,6 +1036,36 @@ const MemoryRow = ({
...
+    source: optString,
+    sourceType: string,
+}) => {
+    if ((sourceType !== URL_SOURCE && sourceType !== LOCAL_INSTALL_MEDIA_SOURCE) ||
+        (!source || source.endsWith(".iso"))) {
+        return null;
+    }
+
+    return (
+        <FormGroup label={_("Extra arguments")}
```
martinpitt: This isn't self-descriptive enough. Extra arguments for what? What's the format? That does not have any validation either. From the helper text below I take it this is for the kernel command line. It could say "Kernel options" here? And link to https://docs.kernel.org/admin-guide/kernel-parameters.html ?

martinpitt (commented):
Thanks!

This needs some UI polish, and desperately a test. Injecting kernel args is really difficult across virt-install, the target OS installer, the target OS'es grub/whatever bootloader config etc. I've been in this business long enough that it's practically guaranteed to fail somewhere 😁 

I know this is hard to test automatically, as we don't currently inspect the guest OSes. But it could at least check that the install succeeds and ends up at the `login:` prompt. It should be possible in general to then log in and type `cat /proc/cmdline` into the terminal -- @mvollmer WDYT?

src/components/create-vm-dialog/createVmDialog.tsx
Nykseli: "Kernel options" and having the link is a lot less confusing, I agree. I'll add those in.

When it comes to validation, I wasn't really sure what to do with it. The input can be almost anything and depends on the kernel version, what installer is used etc. So validating individual options is not really an option. But thinking about it again, I could look how the kernel validates it. It might have something simple that I could replicate here.

Nykseli:
> Injecting kernel args is really difficult across virt-install, the target OS installer, the target OS'es grub/whatever bootloader config etc. I've been in this business long enough that it's practically guaranteed to fail somewhere 😁
> 

Agreed. It's easy to input something that breaks the installation process and those situations are not always easy to debug. And the fact that `--extra-args` is only supported with `--location` doesn't help with the confusion. :sweat_smile: 
Personally I think using PXE or `.treeinfo` for custom installations would be preferable, but PXE is a lot of work to set up and `.treeinfo` is not always supported. So I also get why there's a need for this feature

> I know this is hard to test automatically, as we don't currently inspect the guest OSes. But it could at least check that the install succeeds and ends up at the `login:` prompt. It should be possible in general to then log in and type `cat /proc/cmdline` into the terminal -- @mvollmer WDYT?

I'll start looking into testing :slightly_smiling_face:

src/components/create-vm-dialog/createVmDialog.tsx
martinpitt: I didn't mean to actually validate what the particular options do. I just meant that passing something innocuous like `cockpit.test=42` actually ends up in /proc/cmdline, i.e. that plumbing the value through cockpit → virt-install → QEMU → OS installer → grub works.

Nykseli: Oh, I see. That's something that can only be done with tests, right? We can see what are passed to kernel from journal but we have to boot up the VM first.

Also, I'm not actually sure if these are meant to be saved to grub or if these arguments are just for installation. Need to look into that as well

Nykseli: Looks like arguments that are handled by dracut are not passed to grub but rest of them are

Nykseli:
Looked into testing a bit and one way it could be done is with Agama's  [unattended installation](https://agama-project.github.io/docs/overview/unattended). If you extract Leap 16 offline installation image contents to a http server, you can do an offline installation.
The downside is that it requires a bit of disk space (~4 gig image), the installation requires quite a bit more ram than the current testing images (preferably around 4 gigs), and the installation process takes a while. But this would properly test that the extra arguments are actually working

To use the unattended installation, you need to set `--extra-args` to something like this
```
root=live:http://192.168.0.105:8081/leap-16/LiveOS/squashfs.img inst.install_url=http://192.168.0.105:8081/leap-16/install hostname=leap-16-args inst.auto=http://192.168.0.105:8081/agama-config-leap16.json
```

<details>
<summary>agama-config-leap16.json</summary>

```json
{
  "hostname": {
    "transient": "leap-16-args",
    "static": "leap-16-args"
  },
  "root": {
    "password": "foobar",
    "hashedPassword": false
  },
  "software": {
    "patterns": [
      "selinux"
    ],
    "packages": []
  },
  "product": {
    "id": "openSUSE_Leap"
  },
  "storage": {
    "boot": {
      "configure": true
    },
    "drives": [
      {
        "search": {
          "condition": {
            "name": "/dev/vda"
          },
          "ifNotFound": "error"
        },
        "partitions": [
          {
            "filesystem": {
              "reuseIfPossible": false,
              "label": "",
              "path": "/",
              "mountOptions": [],
              "mkfsOptions": [],
              "type": {
                "btrfs": {
                  "snapshots": true
                }
              }
            },
            "size": {
              "min": 3221225472
            }
          },
          {
            "search": {
              "ifNotFound": "skip"
            },
            "delete": true
          }
        ]
      }
    ],
    "volumeGroups": [],
    "mdRaids": []
  },
  "network": {
    "connections": [
      {
        "id": "Wired Connection",
        "method4": "auto",
        "method6": "auto",
        "ignoreAutoDns": false,
        "status": "up",
        "autoconnect": true,
        "persistent": true
      }
    ]
  },
  "localization": {
    "language": "en_US.UTF-8",
    "keyboard": "us",
    "timezone": "Europe/Berlin"
  }
}
```

</details>

mvollmer:
> Looked into testing a bit and one way it could be done [...]

I think it's enough to verify that the text from the dialog ends up in the virt-install command line as a "--extra-args" argument. Whatever happens after that does not really need to be tested in our tests.

src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1035,6 +1036,36 @@ const MemoryRow = ({
...
+                       fieldId='extra-arguments' id='argument-group'>
+            <InputGroup>
+                <TextInput id='extra-argument' value={extraArguments || ''}
+                               className="extra-argument-input"
+                               onChange={(_, value) => onValueChanged('extraArguments', value.trim())} />
```
mvollmer: Let's not modify the value here in any way, we should store verbatim what the user has typed. Trimming makes it impossible to type a space at the end.

src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1035,6 +1036,36 @@ const MemoryRow = ({
...
+    sourceType,
+} : {
+    extraArguments: optString,
+    onValueChanged: OnValueChanged,
+    source: optString,
+    sourceType: string,
+}) => {
+    if ((sourceType !== URL_SOURCE && sourceType !== LOCAL_INSTALL_MEDIA_SOURCE) ||
+        (!source || source.endsWith(".iso"))) {
+        return null;
```
mvollmer: Isn't it a bit weird if the whole row pops in and out of existence depending on whether the user has types ".iso" suffix yet?

src/components/create-vm-dialog/createVmDialog.tsx
mvollmer: If this is for passing options to the installer, shouldn't we rather link to documentation to these and not to the kernel ones? Especially if they don't end up in the target installation. Or do they?

mvollmer:
So, hmm, I think this is not a great user experience. Ideally, we would implement specific use cases like "I want to supply Anaconda with a kickstart file", and Cockpit would offer that on the "Automation" tab iff it is supported by the installed OS (according to osinfo).  Plus others, plus maybe a row for "Extra installer options".   Same for Agama.

In any case, can we move this to the "Automation" tab and call it "Extra installer options"?

mvollmer:
`--extra-args` seem to work also with `--install`, not just with `--location`, so we should offer them also for "Download an OS".

src/components/create-vm-dialog/createVmDialog.tsx
Nykseli: It's a bit weird yes :sweat_smile: The validation is basically the same as in `install_machine.py` but I agree that the UI is a bit clunky

src/components/create-vm-dialog/createVmDialog.tsx
Nykseli: The arguments are for kernel, initramfs (like dracut, if the image uses it), and the installer. So it's honestly a bit of a mess to sometimes find what argument affects what 😅 

Arguments that kernel understand can end up in the target installation. At least agama puts them in grub config. But not at all sure if this is installer / bootloader specific behavior

Nykseli:
> So, hmm, I think this is not a great user experience. Ideally, we would implement specific use cases like "I want to supply Anaconda with a kickstart file", and Cockpit would offer that on the "Automation" tab iff it is supported by the installed OS (according to osinfo). Plus others, plus maybe a row for "Extra installer options". Same for Agama.
> 

Agama doesn't support `.treeinfo` so using osinfo to detect it is a bit tricky. But I agree that specific use cases is a much better user experience

> In any case, can we move this to the "Automation" tab and call it "Extra installer options"?

I think this is a good idea. It should make the "normal" installation UI less clunky

mvollmer:
> > So, hmm, I think this is not a great user experience. Ideally, we would implement specific use cases like "I want to supply Anaconda with a kickstart file", and Cockpit would offer that on the "Automation" tab iff it is supported by the installed OS (according to osinfo). Plus others, plus maybe a row for "Extra installer options". Same for Agama.
> 
> Agama doesn't support `.treeinfo` so using osinfo to detect it is a bit tricky. But I agree that specific use cases is a much better user experience

Yes, "osinfo support" is wishful dreaming of a perfect world even for Anaconda... :-)

> > In any case, can we move this to the "Automation" tab and call it "Extra installer options"?
> 
> I think this is a good idea. It should make the "normal" installation UI less clunky

Yes. We might need a version of the "Automation" tab that has only the "Extra installer options", but that is fine I'd say.

src/components/create-vm-dialog/createVmDialog.tsx
mvollmer: If we move this to the "Automation" tab, then that tab gets enabled/disabled as the use fills out the dialog. That is much more acceptable.

Nykseli:
Hi, and sorry I couldn't work on this for a while. We updated node from 22 to 24 which caused quite a bit of build issues with esbuild-wasm, and figuring that out took a while :sweat_smile:

Wasn't sure what's the best way to detect if the installation media actually supports `--extra-args` but since it's
probably mostly used with tree installations, I figured using `treeInstallable` from `src/getOSList.py` is probably
a good enough guess.

And this how the UI currently looks like with download os flow:

When you select an image that supports extra args, the automation tab is now activated

Screenshot From 2026-02-18 08-09-56

And since leap 16 doesn't support unattended installation, only the extra args field is displayed

Screenshot From 2026-02-18 08-10-10

If image that supports both unattended and extra args (for example Fedora 42), extra args field is displayed after the fields related to unattended installation

Screenshot From 2026-02-18 08-31-51


Is this the right direction? UI/UX is not my strong suit so feedback is appreciated

Also, I'll work on testing after the UI is more finalized

src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1491,25 +1523,30 @@ export class CreateVmModal extends React.Component<CreateVmModalProps, CreateVmM
...
         if ((this.state.sourceType == URL_SOURCE || this.state.sourceType == LOCAL_INSTALL_MEDIA_SOURCE) && this.state.os) {
             if (this.state.os.medias && this.state.sourceMediaID && this.state.sourceMediaID in this.state.os.medias)
                 unattendedDisabled = !this.state.os.medias[this.state.sourceMediaID].unattendedInstallable;
             else
                 unattendedDisabled = !this.state.os.unattendedInstallable;
+            if (this.state.source && !this.state.source.endsWith(".iso"))
+                extraArgsDisabled = !this.state.os.treeInstallable;
         } else if (this.state.sourceType == DOWNLOAD_AN_OS) {
             unattendedDisabled = !this.state.os || !this.state.os.unattendedInstallable;
+            extraArgsDisabled = !this.state.os || !this.state.os.treeInstallable;
```
mvollmer: I _think_ all OSes offered for "Download an OS" are tree-installable since that is one of the criteria to include them in the list. But maybe it is better not to rely on that even if true.

src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1592,6 +1632,15 @@ export class CreateVmModal extends React.Component<CreateVmModalProps, CreateVmM
                                      userPassword={this.state.userPassword}
                                      onValueChanged={this.onValueChanged} />
                 }
+                {((this.state.sourceType === URL_SOURCE || this.state.sourceType === LOCAL_INSTALL_MEDIA_SOURCE || this.state.sourceType === DOWNLOAD_AN_OS) &&
```
mvollmer: The logic is getting quite complicated here (and elsewhere), and is also duplicated inside `ExtraArgumentsRow` to some extend. Can we put it all into one place and compute whether or not the ExtraArgumentsRow should be displayed just once? Same for UnattendedRow.

mvollmer:
> And this how the UI currently looks like with download os flow:

Nice! I think we need to work on the text a bit... So now I would say that "Installer arguments" is not right after all, especially if we link to kernel documentation. We probably can't link to the exact right installer docs, hmm.

What about "Boot arguments" and the helper text says "Arguments passed to the booted kernel can often be used to control the installer. Refer to the documentation of the installer of $OS." and no link to the kernel docs. $OS would come from the os detection.

Nykseli:
Simplified the logic how things are displayed in automation tab and updated the extra argument field strings

src/components/create-vm-dialog/createVmDialog.tsx
mvollmer: No, at least RHEL is offered with "Download an OS" but it's not a install tree. (We download the "boot.iso" and boot from that via `--cdrom`, which of course doesn't support --extra-args.)

mvollmer:
> Simplified the logic how things are displayed in automation tab and updated the extra argument field strings

Brilliant, thanks!

mvollmer (changes requested):
Nice! Code looks good to me now. Let's add the the tests!

Nykseli:
Added some tests. Tested them on tumbleweed and fedora-43 locally. I wonder if that's enough coverage?

Also fixed an issue with `mirror.i3d.net` not resolving properly to `::1`  when using IPv6 which caused tests to timeout and fail

test/check-machines-create
```diff
@@ -1622,8 +1622,8 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                 self._fakeFedoraTree()
 
             # Make sure we don't run into long DNS timeouts when trying to contact this server
-            self.machine.write("/etc/hosts", "127.0.0.42 mirror.i3d.net", append=True)
-            self.machine.write("/etc/hosts", "::1 mirror.i3d.net", append=True)
+            self.machine.write("/etc/hosts", "127.0.0.42 mirror.i3d.net\n", append=True)
+            self.machine.write("/etc/hosts", "::1 mirror.i3d.net\n", append=True)
```
mvollmer, line 1675: Ouch! Thanks!

test/check-machines-create
```diff
@@ -834,6 +852,19 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
...
+        # test with --extra-args
+        runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='file',
+                                                      name="extra-args-file",
+                                                      location=config.TREE_FILE,
+                                                      memory_size=128, memory_size_unit='MiB',
+                                                      storage_size=128, storage_size_unit='MiB',
+                                                      extra_arguments="foo=bar",
+                                                      delete=False,
+                                                      create_and_run=True))
+        testlib.wait(lambda: f"--extra-args foo=bar" in self.machine.execute("ps aux | grep '[v]irt-install'"))
```
mvollmer: Do we need to check this here when it is already checked in virt_install_correct?

mvollmer:
> Added some tests. 

Thanks! I have triggered them.

> Tested them on tumbleweed and fedora-43 locally. I wonder if that's enough coverage?

No, they need to pass on all the OSes before we can merge this.
 
> Also fixed an issue with `mirror.i3d.net` not resolving properly to `::1` when using IPv6 which caused tests to timeout and fail

Thanks!

mvollmer:
Upps, some linting errors:

```
not ok 9 /static-code/test-ruff
# F541 [*] f-string without any placeholders
#    --> test/check-machines-create:850:30
#     |
# 848 |                                                       delete=False,
# 849 |                                                       create_and_run=True))
# 850 |         testlib.wait(lambda: f"--extra-args foo=bar" in self.machine.execute("ps aux | grep '[v]irt-install'"))
#     |                              ^^^^^^^^^^^^^^^^^^^^^^^
# 851 |         self.goToMainPage()
# 852 |         self.machine.execute("virsh destroy extra-args-file; virsh undefine extra-args-file")
#     |
# help: Remove extraneous `f` prefix
# 847 |                                                       extra_arguments="foo=bar",
# 848 |                                                       delete=False,
# 849 |                                                       create_and_run=True))
#     -         testlib.wait(lambda: f"--extra-args foo=bar" in self.machine.execute("ps aux | grep '[v]irt-install'"))
# 850 +         testlib.wait(lambda: "--extra-args foo=bar" in self.machine.execute("ps aux | grep '[v]irt-install'"))
# 851 |         self.goToMainPage()
# 852 |         self.machine.execute("virsh destroy extra-args-file; virsh undefine extra-args-file")
# 853 | 
# 
# F541 [*] f-string without any placeholders
#    --> test/check-machines-create:941:30
#     |
# 939 |                                                       delete=False,
# 940 |                                                       create_and_run=True))
# 941 |         testlib.wait(lambda: f"--extra-args foo=bar" in m.execute("ps aux | grep '[v]irt-install'"))
#     |                              ^^^^^^^^^^^^^^^^^^^^^^^
# 942 |         self.goToMainPage()
# 943 |         m.execute("virsh destroy extra-args-url; virsh undefine extra-args-url")
#     |
# help: Remove extraneous `f` prefix
# 938 |                                                       extra_arguments="foo=bar",
# 939 |                                                       delete=False,
# 940 |                                                       create_and_run=True))
#     -         testlib.wait(lambda: f"--extra-args foo=bar" in m.execute("ps aux | grep '[v]irt-install'"))
# 941 +         testlib.wait(lambda: "--extra-args foo=bar" in m.execute("ps aux | grep '[v]irt-install'"))
# 942 |         self.goToMainPage()
# 943 |         m.execute("virsh destroy extra-args-url; virsh undefine extra-args-url")
# 944 | 
```

@Nykseli, you can run all the linters locally via `./test/common/static-code`.

test/check-machines-create
Nykseli: I think virt_install_correct is only used in `createAndVerifyVirtInstallArgsUnattended`. So only `runner.createDownloadAnOSTest` ends up checking the extra-args

Nykseli:
Fixed the lint issues, sorry about that :sweat_smile:

test/check-machines-create
mvollmer: Ah, right. And `runner.createCloudBaseImageTest` has its own version. I was mostly worried that we are introducing more flakiness here, since checking virt-install arguments has traditionally been hairy... but we'll see. :-)

mvollmer: (I find these dialog helpers to be more obfuscation than actual helpers, tbh...)

mvollmer:
The Debians fail because they don't expect there to be a notification of failing to create the extra-args-url machine. I think there needs to be a call to `runner.checkEnvIsEmpty` to clean those up.

mvollmer:
This pixel test is suspicious: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2517-cf47e594-20260223-161912-fedora-43/pixeldiff.html#TestMachinesCreate-testCreateBasicValidation-iso

Installing from a Fedora 28 ISO should not offer "automation".  We have to check the new logic carefully.

mvollmer:
> Installing from a Fedora 28 ISO should not offer "automation". We have to check the new logic carefully.

So, I think the new code here is actually "more correct" than what we have in main right now. The detected osinfo for a Fedora 40 ISO (for example) clearly says that it supports unattended installs. The code in main would correctly deduce this, but then also only ever enable the "Automation" tab for "Download an OS" and "Cloud base image", never for a local ISO.

The code here shows the UnattendedRow whenever osinfo says that the detected OS supports unattended installs (regardless of mode) and always enables the Automation tab when it has something to show. This makes a lot of sense to me.

However! Unattended installs with a local or remote ISO just don't seem to work, at least not reliably.  We have recently disabled them for RHEL in #2375 for this reason.

`virt-install` doesn't seem to like `--cdrom` and `--unattended` together:

```
WARNING  Attempting unattended method=cdrom injection for a non-windows OS. If this doesn't work, try passing install media to --location
```

Using `--location` with my Fedora 40 Server ISO does seem to want to work, but I run into permission problems, as usual:

```
# virt-install --name fed --location /home/mvo/Downloads/Fedora-Server-netinst-x86_64-40-1.14.iso --unattended profile=jeos,admin-password-file=/home/mvo/password
Using fedora40 default --memory 4096
Using fedora40 default --disk size=20

Starting install...
Retrieving 'vmlinuz'                                                                                                                                                  |  14 MB  00:00:00     
Retrieving 'initrd.img'                                                                                                                                               | 142 MB  00:00:00     
Allocating 'fed.qcow2'                                                                                                                                                |  20 GB  00:00:00     
Removing disk 'fed.qcow2'                                                                                                                                             |         00:00:00     
ERROR    internal error: process exited while connecting to monitor: qemu: could not open kernel file '/root/.cache/virt-manager/boot/virtinst-1tytbgsq-vmlinuz': Permission denied
Domain installation does not appear to have been successful.
If it was, you can restart your domain by running:
  virsh --connect qemu:///session start fed
otherwise, please restart your installation.
```

What this means for this PR: Let's never show the UnattendedRow for local or remote ISOs. This is not a feature that we support in main, and this PR doesn't need to change that. We can in the future work on this and use `--location` with ISOs that osinfo says do support unattended installs. But not now. :-)

src/components/create-vm-dialog/createVmDialog.tsx
```diff
@@ -1491,25 +1516,32 @@ export class CreateVmModal extends React.Component<CreateVmModalProps, CreateVmM
...
+        let showUnattendedRow = false;
+        let showCloudInitRow = false;
+        let showExtraArgsRow = false;
         if ((this.state.sourceType == URL_SOURCE || this.state.sourceType == LOCAL_INSTALL_MEDIA_SOURCE) && this.state.os) {
             if (this.state.os.medias && this.state.sourceMediaID && this.state.sourceMediaID in this.state.os.medias)
-                unattendedDisabled = !this.state.os.medias[this.state.sourceMediaID].unattendedInstallable;
+                showUnattendedRow = this.state.os.medias[this.state.sourceMediaID].unattendedInstallable;
             else
-                unattendedDisabled = !this.state.os.unattendedInstallable;
+                showUnattendedRow = this.state.os.unattendedInstallable;
```
mvollmer: As explained [here](https://github.com/cockpit-project/cockpit-machines/pull/2517#issuecomment-3949792203), let's only ever set `showUnattendedRow` when `mode == DOWNLOAD_AN_OS`. This is what main does right now, and we shouldn't change that here. 

Please also double check my thinking. On main, the value stored into `unattendedDisabled` here does not reach the single use of `unattendedDisabled` below, and we can thus remove the code here.

mvollmer:
rhel-10-2 is flaking with my new `checkEnvIsEmpty` calls... I can't reproduce that locally. Maybe deleting the VM with cockpit itself is more reliable. I push one more FIXUP.

mvollmer:
> rhel-10-2 is flaking with my new `checkEnvIsEmpty` calls... I can't reproduce that locally. Maybe deleting the VM with cockpit itself is more reliable. I push one more FIXUP.

That seemed to have helped.

src/components/create-vm-dialog/createVmDialog.tsx
Nykseli: Oh I see. That makes sense to me too (assuming I fully understood it :sweat_smile:). Removed those lines

Nykseli:
> > Installing from a Fedora 28 ISO should not offer "automation". We have to check the new logic carefully.
> 
> So, I think the new code here is actually "more correct" than what we have in main right now. The detected osinfo for a Fedora 40 ISO (for example) clearly says that it supports unattended installs. The code in main would correctly deduce this, but then also only ever enable the "Automation" tab for "Download an OS" and "Cloud base image", never for a local ISO.
> 
> The code here shows the UnattendedRow whenever osinfo says that the detected OS supports unattended installs (regardless of mode) and always enables the Automation tab when it has something to show. This makes a lot of sense to me.
> 
> However! Unattended installs with a local or remote ISO just don't seem to work, at least not reliably. We have recently disabled them for RHEL in #2375 for this reason.
> 
> `virt-install` doesn't seem to like `--cdrom` and `--unattended` together:
> 
> ```
> WARNING  Attempting unattended method=cdrom injection for a non-windows OS. If this doesn't work, try passing install media to --location
> ```
> 
> Using `--location` with my Fedora 40 Server ISO does seem to want to work, but I run into permission problems, as usual:
> 
> ```
> # virt-install --name fed --location /home/mvo/Downloads/Fedora-Server-netinst-x86_64-40-1.14.iso --unattended profile=jeos,admin-password-file=/home/mvo/password
> Using fedora40 default --memory 4096
> Using fedora40 default --disk size=20
> 
> Starting install...
> Retrieving 'vmlinuz'                                                                                                                                                  |  14 MB  00:00:00     
> Retrieving 'initrd.img'                                                                                                                                               | 142 MB  00:00:00     
> Allocating 'fed.qcow2'                                                                                                                                                |  20 GB  00:00:00     
> Removing disk 'fed.qcow2'                                                                                                                                             |         00:00:00     
> ERROR    internal error: process exited while connecting to monitor: qemu: could not open kernel file '/root/.cache/virt-manager/boot/virtinst-1tytbgsq-vmlinuz': Permission denied
> Domain installation does not appear to have been successful.
> If it was, you can restart your domain by running:
>   virsh --connect qemu:///session start fed
> otherwise, please restart your installation.
> ```
> 
> What this means for this PR: Let's never show the UnattendedRow for local or remote ISOs. This is not a feature that we support in main, and this PR doesn't need to change that. We can in the future work on this and use `--location` with ISOs that osinfo says do support unattended installs. But not now. :-)

Thank you so much for figuring this all out! Didn't realize how unreliable that is :sweat_smile:

mvollmer:
I have force pushed new pixel test references. They look good now!

mvollmer:
And I had to rebase to main to resolve the `test/reference` conflict...

mvollmer (approved):
Nice, thanks!

mvollmer:
We did it! :-) Thanks a lot for this contribution!  I'll write some release notes as well.

mvollmer:
@Nykseli, there is one question open for me: Does this work with "Create and edit"? I think we might need to store the extra args as part of the Cockpit metadata in `install_machine.py`.

Nykseli:
> @Nykseli, there is one question open for me: Does this work with "Create and edit"? I think we might need to store the extra args as part of the Cockpit metadata in `install_machine.py`.
> 

@mvollmer Good question. I actually didn't even test that. I'll look into it later today and open a PR for it if needed

Issue #2520, opened 2026-01-29
testCreateDownloadAnOS gating failure on RHEL-10-2


jelly:
```
Wrote subVmTestCreate2 XML to TestMachinesCreate-testCreateDownloadAnOS-rhel-10-2-localhost-22-subVmTestCreate2.xml
Wrote subVmTestCreate2 log to TestMachinesCreate-testCreateDownloadAnOS-rhel-10-2-localhost-22-subVmTestCreate2.log
Traceback (most recent call last):
  File "/source/test/check-machines-create", line 498, in testCreateDownloadAnOS
    runner.createDownloadAnOSTest(TestMachinesCreate.VmDialog(self, sourceType='os',
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                              expected_memory_size=128,
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
                                                              os_short_id=config.FEDORA_28_SHORTID,
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                              create_and_run=True))
                                                              ^^^^^^^^^^^^^^^^^^^^^
  File "/source/test/check-machines-create", line 2068, in createDownloadAnOSTest
    .createAndVerifyVirtInstallArgsUnattended()
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/source/test/check-machines-create", line 1356, in createAndVerifyVirtInstallArgsUnattended
    testlib.wait(virt_install_correct, delay=3)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/test/common/testlib.py", line 2963, in wait
    val = func()
  File "/source/test/check-machines-create", line 1340, in virt_install_correct
    virt_install_cmd_out = self.machine.execute(virt_install_cmd)
  File "/source/bots/machine/machine_core/ssh_connection.py", line 342, in execute
    res = subprocess.run(command_line,
                         input=input.encode("UTF-8") if input else b'',
                         stdout=stdout, check=check)
  File "/usr/lib64/python3.14/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
[...]

Issue #2519, opened 2026-01-29
testCreatePXE gating retry/failure


jelly:
This test has some failures on RHEL 10 in gating and causes retries leading it to succeed but still lets investigate:

<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/294cc6f8-23a5-47be-859e-3003b0ac3cd9" />

This fails with:

```
Traceback (most recent call last):
  File "/source/test/check-machines-create", line 580, in testCreatePXE
    runner.createTest(TestMachinesCreate.VmDialog(self, name='pxe-guest',
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                  sourceType='pxe',
                                                  ^^^^^^^^^^^^^^^^^
                                                  location=f"type=direct,source={iface},source.mode=bridge",
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                  storage_pool=NO_STORAGE,
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^
                                                  create_and_run=True))
                                                  ^^^^^^^^^^^^^^^^^^^^^
  File "/source/test/check-machines-create", line 1988, in createTest
    self._deleteVm(dialog)
    ~~~~~~~~~~~~~~^^^^^^^^
  File "/source/test/check-machines-create", line 1826, in _deleteVm
    self.test_obj.performAction(dialog.name, "delete", connectionName=dialog.connection)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/test/machineslib.py", line 73, in performAction
    self.dropdownAction(f"#vm-{vmName}-{connectionName}-action-kebab", f"#vm-{vmName}-{connectionName}-{action}")
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/test/machineslib.py", line 58, in dropdownAction
[...]

PR #2521, opened 2026-01-30
test: Robustify check-machines-create

mvollmer:


mvollmer:
I was not able to reproduce #2519 nor #2520, but the fixes here are highly likely helpful.

I don't know what the opening and closing of the "Add disk" dialog is supposed to test for, so maybe we can also just remove those two clicks entirely.

test/check-machines-create
```diff
@@ -1905,6 +1905,8 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
             else:
                 b.wait_in_text(f"#vm-{name}-disks .pf-v6-c-empty-state", "No disks")
                 b.click(f"#vm-{name}-disks-adddisk")
+                # wait for the dialog to initialize and have a stable layout
+                b.wait_visible(f"#vm-{name}-disks-adddisk-dialog-add:not([disabled])")
```
jelly: after the cancel on the line below I would also `b.wait_not_present(".pf-v6-c-modal-box")` so we assert that the dialog has gone away

mvollmer: Good point, thanks!  I also had to switch to waiting for the "Source" input group to be visible. TestMachinesCreate.testCreateFileSource opens the dialog in a state where the Apply button stays disabled forever...

PR #2522, opened 2026-01-30
nics: Show all IP addresses in the table column

### LINKED ISSUE #2503: Network interface assigned IPs confusing for the user
With us fixing the show interface IPs in #2303 we then reshuffled the design to make it easier in #2324. Problem now is that it is confusing for the user despite us showing all IPs. This is because we now only show one IP in the table and then all of the IPs within the expanded content

Since we give no indication there are multiple IPs in the table row itself you cannot know if the interface has more than one IP or not, this might be confusing if you have two IPv4 on one interface and find out the second one is not showing in the table at a first glance

Image

Some ideas

- Show all IPs again
- Remove IP Address column in favor of something like network type

mvollmer:
Showing just one without any indication that there are more is too confusing. So let's just show them all like we used to do, but let's just show the addresses themselves without any individual "IPv4" or "IPv6" labels.  The form of an address is enough to see which version of IP it is for.

Fixes #2503

It used to look like this:

image

This was too messy layoutwise and we moved things into a expander section and only showed a single address in the table:

image

Only a single address is confusing when there are multiple, so this Pr proposes to make it look like this:

image

The "inet" and "inet6" labels are not necessary, and people probably have only few addresses. The buttons make it so that there is already place for two addresses without causing the row to get higher.

### DIFF
--- a/test/check-machines-nics
+++ b/test/check-machines-nics
@@ -53,18 +53,6 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
         if "rhel-8" not in m.image:
             b.wait_in_text("#vm-subVmTest1-network-1-ipv6-address-0", "fe80::")
         b.wait_in_text("#vm-subVmTest1-network-1-state", "up")
-        b.assert_pixels(
-            "#vm-subVmTest1-networks",
-            "vm-details-nics-card",
-            ignore=[
-                "#vm-subVmTest1-network-1-mac",
-                "#vm-subVmTest1-network-1-ipv4-address-0",
-                "#vm-subVmTest1-network-1-ipv6-address-0",
-            ],
-            mock={
-                'td[data-label="IP address"]': "192.168.122.NN",
-            },
-        )
 
         # Remove default interface
         self.deleteIface(1)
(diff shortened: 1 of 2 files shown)


test/check-machines-nics
```diff
@@ -56,11 +56,10 @@ class TestMachinesNICs(machineslib.VirtualMachinesCase):
             "vm-details-nics-card",
             ignore=[
                 "#vm-subVmTest1-network-1-mac",
-                "#vm-subVmTest1-network-1-ipv4-address-0",
-                "#vm-subVmTest1-network-1-ipv6-address-0",
             ],
             mock={
-                'td[data-label="IP address"]': "192.168.122.NN",
+                "#vm-subVmTest1-network-1-ipv4-address-0": "192.168.122.NN",
+                "#vm-subVmTest1-network-1-ipv6-address-0": "fe80::4d6b:b5b3:57b4:XXXX"
```
Venefilyn: IMO we should show two addresses of the same type, like 2x IPv4, 2x IPv6

Venefilyn (commented):
Overall LGTM, I think the tests should cover more though

test/check-machines-nics
mvollmer: Yes, good idea!

mvollmer: I have moved the pixel test to the place where we have two IPv4 addresses.

PR #2523, opened 2026-01-31
import/create: Add URL parameters to prepopulate the dialogs

ochosi:
This PR adds URL parameters to open a pre-populated VM import or create dialog.

This is in order to integrate Cockpit Image Builder better with Cockpit Machines. So when you build e.g. a qcow2 or ISO there should be a hyperlink to directly open the import dialog and make launching/installing it a one-click operation.

### DIFF
--- a/src/components/vms/hostvmslist.tsx
+++ b/src/components/vms/hostvmslist.tsx
@@ -33,6 +33,8 @@ import { WithDialogs } from 'dialogs.jsx';
 
 import { Progress, ProgressVariant } from "@patternfly/react-core/dist/esm/components/Progress";
 
+import { CreateVmModal } from '../create-vm-dialog/createVmDialog.jsx';
+
 import { VmActions } from '../vm/vmActions.jsx';
 
 import { vmId, rephraseUI, dummyVmsFilter, DOMAINSTATE } from "../../helpers.js";
@@ -44,6 +46,7 @@ import { VmUsesSpice } from '../vm/usesSpice.jsx';
 import { AggregateStatusCards } from "../aggregateStatusCards.jsx";
 import { appState } from "../../state";
 import { ensureUsagePolling } from '../../libvirtApi/common';
+import { useOn, usePageLocation } from "hooks";
 
 import { VMS_CONFIG } from "../../config.js";
 
@@ -127,6 +130,73 @@ const VmUsageMem = ({ vm } : {vm: VM | UIVM, }) => {
 };
 
 const _ = cockpit.gettext;
+
+/**
+ * Component that renders the Create/Import VM dialog when URL parameters are present.
+ * Uses declarative rendering based on URL state (the URL is the single source of truth).
+ *
+ * Supported URL formats:
+ * - Import VM: /machines#?action=import&source=/path/to/disk.qcow2&os=fedora40&name=my-vm
+ * - Create VM: /machines#?action=create&name=my-vm&type=url&source=http://example.com/install.iso&os=fedora40
+ *
+ * Parameters:
+ * - action: 'import' or 'create' (required)
+ * - name: VM name (optional)
+ * - os: OS shortId like 'fedora40', 'rhel9.4' (optional)
+ * - source: disk image path (import) or installation source URL/path (create)
+ * - type: source type - 'os', 'url', 'file', 'cloud', 'pxe' (create only)
+ */
+const CreateVmDialogOpener = ({ vms }: { vms: VM[] }) => {
+    // Subscribe to appState changes (for virtInstallCapabilities)
+    useOn(appState, "changed");
+
+    // Subscribe to URL changes - re-renders when location changes
+    const { options } = usePageLocation();
+
+    const vi_caps = appState.virtInstallCapabilities;
+    const action = options.action;
+
+    // Determine if dialog should be shown
+    const showDialog = (action === 'import' || action === 'create') &&
+                       vi_caps?.virtInstallAvailable;
+
+    if (!showDialog) return null;
+
+    // Parse URL parameters
+    const mode = action === 'import' ? 'import' : 'create';
+    const initialName = typeof options.name === 'string' ? options.name : undefined;
+    const initialOS = typeof options.os === 'string' ? options.os : undefined;
+    const initialSource = typeof options.source === 'string' ? options.source : undefined;
+    const initialSourceType = action === 'create' && typeof options.type === 'string'
+        ? options.type
+        : undefined;
+
+    // Close handler clears URL parameters
+    const handleClose = () => {
+        cockpit.location.replace(cockpit.location.path, {});
+    };
+
+    // Generate key from URL params to force re-mount when params change
+    const dialogKey = `${action}-${options.name}-${options.os}-${options.source}-${options.type}`;
+
+    return (
+        <CreateVmModal
+            key={dialogKey}
+            mode={mode}
+            onClose={handleClose}
+            nodeMaxMemory={appState.nodeMaxMemory}
+            vms={vms}
+            cloudInitSupported={vi_caps.cloudInitSupported}
+            downloadOSSupported={vi_caps.downloadOSSupported}
+            unattendedSupported={vi_caps.unattendedSupported}
+            unattendedUserLogin={vi_caps.unattendedUserLogin}
+            initialSource={initialSource}
+            initialOS={initialOS}
+            initialName={initialName}
+            initialSourceType={initialSourceType}
+        />
+    );
+};
 
 /**
  * List of all VMs defined on this host
@@ -248,6 +318,7 @@ export const HostVmsList = ({
 
     return (
         <WithDialogs key="vms-list">
+            <CreateVmDialogOpener vms={vms} />
             <Page className="pf-m-no-sidebar">
                 <PageSection hasBodyWrapper={false}>
                     <Gallery className="ct-cards-grid" hasGutter>
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -23,6 +23,7 @@ import os
 import re
 import string
 import time
+import urllib
 import xml.etree.ElementTree as ET
 from datetime import datetime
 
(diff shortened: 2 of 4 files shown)


mvollmer:
Thanks! This is exactly the kind of thing that Cockpit should be doing more of!

mvollmer:
Some brainstorming notes. Please don't rush into implementing these (unless you really want to).

 - When clicking on "launch", should the current tab switch to "Machines" and open the dialog? Or should a new tab be opened with a fresh Cockpit and a fresh "Machines" page?  Should the user be able to decide via left/middle clicking?
 - When not starting a fresh "Machines" page, what do we do if a dialog is already open in "Machines"? We could silently ignore the request to open a new one, we could forcefully close the currently open one, or we could even make it possible for the image builder to check whether "Machines" has a dialog open right now. I think it's ok to forcefully close any open dialog, since the user has navigated away from Machines.
 - Our current `Dialogs.show()` etc mechanism was introduced to have robust state handling for dialogs that are triggered by buttons deep in unstable DOM territory, such as actions in kebabs in table rows. But we don't have to use it for the "Create" and "Import" dialogs. And since we are now open them based on the URL, we can make the URL the primary record of whether they are open or not, "old school". We can probably also remove the `WithDialogs` on the vms list page, and don't have to worry about closing dialogs with `Dialogs.close()`
 - Do we then keep the path etc state also in the URL instead of in a `useState` (or whatever the dialog currently uses)? It's always best to only have one place with the state and not try to synchronize multiple.
 - Using the URL to store whether the dialog is open or not also means that the existing buttons to open them will change the URL to get them to open. I think that's clean.

mvollmer:
More storming brains:

- The dialogs themselves could wait for `vi_caps` to be ready and show a spinner until it is. In fact, the we don't need to retrieve vi_caps until a dialog is opened, hopefully. I think that simplifies the logic for opening them quite a bit. We already get the osinfo on-demand when the dialog opens, no?

ochosi:
> Some brainstorming notes. Please don't rush into implementing these (unless you really want to).
> 
>     * When clicking on "launch", should the current tab switch to "Machines" and open the dialog? Or should a new tab be opened with a fresh Cockpit and a fresh "Machines" page?  Should the user be able to decide via left/middle clicking?

For ISOs, both works. You can middle-click and it'll open a new tab. For qcow2s, the flow is slightly different because of the "copy to writeable location" intermediary step. I think for qcow2s it makes more sense to stay within the same window. So I actually think the current behavior is okayish, but I'll run it by our UX.

>     * When not starting a fresh "Machines" page, what do we do if a dialog is already open in "Machines"? We could silently ignore the request to open a new one, we could forcefully close the currently open one, or we could even make it possible for the image builder to check whether "Machines" has a dialog open right now. I think it's ok to forcefully close any open dialog, since the user has navigated away from Machines.

Right now, I just forcefully close any dialog. I guess my assumption was that Cockpit is supposed to be "stateless" so I wouldn't assume any specific state in any of the other plugins.
Is that correct - or how do you currently handle this in other plugins?

>     * Our current `Dialogs.show()` etc mechanism was introduced to have robust state handling for dialogs that are triggered by buttons deep in unstable DOM territory, such as actions in kebabs in table rows. But we don't have to use it for the "Create" and "Import" dialogs. And since we are now open them based on the URL, we can make the URL the primary record of whether they are open or not, "old school". We can probably also remove the `WithDialogs` on the vms list page, and don't have to worry about closing dialogs with `Dialogs.close()`
> 
>     * Do we then keep the path etc state also in the URL instead of in a `useState` (or whatever the dialog currently uses)? It's always best to only have one place with the state and not try to synchronize multiple.
> 
>     * Using the URL to store whether the dialog is open or not also means that the existing buttons to open them will change the URL to get them to open. I think that's clean.

Hmmm, those could all make sense to implement. I didn't think too much about that because I wanted the changes to be as non-invasive as possible.

>    * The dialogs themselves could wait for vi_caps to be ready and show a spinner until it is. In fact, the we don't need to retrieve vi_caps until a dialog is opened, hopefully. I think that simplifies the logic for opening them quite a bit. We already get the osinfo on-demand when the dialog opens, no?

That sounds clever! Want me to work that out?

src/components/vms/hostvmslist.tsx
```diff
@@ -128,6 +131,73 @@ const VmUsageMem = ({ vm } : {vm: VM | UIVM, }) => {
...
+
+    // Determine if dialog should be shown
+    const showDialog = (action === 'import' || action === 'create') &&
+                       vi_caps?.virtInstallAvailable;
+
+    if (!showDialog) return null;
+
+    // Parse URL parameters
+    const mode = action === 'import' ? 'import' : 'create';
+    const initialName = typeof options.name === 'string' ? decodeURIComponent(options.name) : undefined;
```
mvollmer: The calls to decodeURIComponent should not be necessary, this is already done by `cockpit.location`.

test/check-machines-create
```diff
@@ -2776,6 +2773,180 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
...
+
+        config = TestMachinesCreate.TestCreateConfig
+
+        # Create a test disk image to import
+        test_disk_path = "/var/lib/libvirt/images/test-import-disk.qcow2"
+        m.execute(f"qemu-img create -f qcow2 {test_disk_path} 1G")
+        self.addCleanup(m.execute, f"rm -f {test_disk_path}")
+
+        # URL-encode the disk path (/ becomes %2F)
+        encoded_path = test_disk_path.replace("/", "%2F")
```
mvollmer: Can we use some real encoding function, maybe `urllib.parse.urlencode`?

test/check-machines-create
```diff
@@ -2776,6 +2773,180 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
             b.wait_text(f"#vm-{vmName}-system-state", "Running")
             b.wait_not_present(f"#vm-{vmName}-system-state-error")
 
+    def testImportUrlParameters(self):
```
mvollmer, line 2777: I think we should combine the new tests, to reduce overhead.  Let's have only two tests: one for import (full and partial) and one for create (full and partial).

mvollmer:
> Very nice! Some small things only.

Ah, and the linting:

```
$ ./test/common/static-code 
not ok 9 /static-code/test-ruff
# F541 [*] f-string without any placeholders
#     --> test/check-machines-create:2922:27
#      |
# 2921 |         # Test with type=file (Local install media)
# 2922 |         self.login_and_go(f"/machines#?action=create&type=file")
#      |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# 2923 |
# 2924 |         b.wait_visible("#create-vm-dialog")
#      |
# help: Remove extraneous `f` prefix
# 2919 |         b.wait_not_present("#create-vm-dialog")
# 2920 | 
# 2921 |         # Test with type=file (Local install media)
#      -         self.login_and_go(f"/machines#?action=create&type=file")
# 2922 +         self.login_and_go("/machines#?action=create&type=file")
# 2923 | 
# 2924 |         b.wait_visible("#create-vm-dialog")
# 2925 | 
```

test/check-machines-create
```diff
@@ -2776,6 +2774,115 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
...
+        test_disk_path = "/var/lib/libvirt/images/test-import-disk.qcow2"
+        m.execute(f"qemu-img create -f qcow2 {test_disk_path} 1G")
+        self.addCleanup(m.execute, f"rm -f {test_disk_path}")
+
+        encoded_path = quote(test_disk_path, safe='')
+
+        # Test 1: Full parameters (source, os, name)
+        test_vm_name = "my-test-vm"
+        self.login_and_go(f"/machines#?action=import&source={encoded_path}&os={config.FEDORA_28_SHORTID}&name={test_vm_name}")
+
```
mvollmer, line 2795: We need a call to `self.waitPageInit()` here to get past the "HW virt disabled" screen. I did not know that! :-)

test/check-machines-create
```diff
@@ -25,6 +25,7 @@ import string
 import time
 import xml.etree.ElementTree as ET
 from datetime import datetime
+from urllib.parse import quote
```
mvollmer: Let's `import urllib` and use it as `usrlib.parse.quote`.  Seeing just a bare `quote(...)` in the code below is a bit mysterious otherwise.

test/check-machines-create
```diff
@@ -2776,6 +2774,115 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
...
+    def testCreateUrlParameters(self):
+        """Test that URL parameters can open the create dialog with pre-filled values (full and partial)"""
+        b = self.browser
+
+        config = TestMachinesCreate.TestCreateConfig
+
+        # Test 1: Full parameters (name, type=os, os)
+        test_vm_name = "my-create-vm"
+        self.login_and_go(f"/machines#?action=create&name={test_vm_name}&type=os&os={config.FEDORA_28_SHORTID}")
+
```
mvollmer, line 2842: `waitPageInit`

mvollmer (changes requested):
Very close!

mvollmer (dismissed):
Thanks!

test/check-machines-create
```diff
@@ -2776,6 +2774,117 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
...
+        # Test 1: Full parameters (source, os, name)
+        test_vm_name = "my-test-vm"
+        self.login_and_go(f"/machines#?action=import&source={encoded_path}&os={config.FEDORA_28_SHORTID}&name={test_vm_name}")
+        self.waitPageInit()
+
+        b.wait_visible("#create-vm-dialog")
+        b.wait_in_text(".pf-v6-c-modal-box .pf-v6-c-modal-box__header .pf-v6-c-modal-box__title",
+                       "Import a virtual machine")
+        b.wait_val("#vm-name", test_vm_name)
+        b.wait_val("#source-disk", test_disk_path)
```
mvollmer: This needs to be "#source-disk input" I believe.

mvollmer:
```diff
diff --git a/test/check-machines-create b/test/check-machines-create
index b9f9e0f7..2ba2a149 100755
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -2797,7 +2797,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
         b.wait_in_text(".pf-v6-c-modal-box .pf-v6-c-modal-box__header .pf-v6-c-modal-box__title",
                        "Import a virtual machine")
         b.wait_val("#vm-name", test_vm_name)
-        b.wait_val("#source-disk", test_disk_path)
+        b.wait_val("#source-disk input", test_disk_path)
         b.wait_attr_contains("#os-select-group input", "value", config.FEDORA_28)
 
         b.click(".pf-v6-c-modal-box__footer button:contains(Cancel)")
@@ -2809,7 +2809,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
         b.wait_visible("#create-vm-dialog")
         b.wait_in_text(".pf-v6-c-modal-box .pf-v6-c-modal-box__header .pf-v6-c-modal-box__title",
                        "Import a virtual machine")
-        b.wait_val("#source-disk", test_disk_path)
+        b.wait_val("#source-disk input", test_disk_path)
         b.wait_attr("#os-select-group input", "value", "")
 
         b.click(".pf-v6-c-modal-box__footer button:contains(Cancel)")
@@ -2823,7 +2823,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
         b.wait_in_text(".pf-v6-c-modal-box .pf-v6-c-modal-box__header .pf-v6-c-modal-box__title",
                        "Import a virtual machine")
         b.wait_val("#vm-name", test_vm_name)
-        b.wait_val("#source-disk", test_disk_path)
+        b.wait_val("#source-disk input", test_disk_path)
         b.wait_attr("#os-select-group input", "value", "")
 
         b.click(".pf-v6-c-modal-box__footer button:contains(Cancel)")
```

PR #2533, opened 2026-02-05
debian: Add missing fields and files to Debian copyright template

martinpitt:
- Add Upstream-Contact
 - Update top-level Copyright years
 - Explicitly give copyright for debian/ packaging
 - Add the modified PatternFly react-template copies, they are MIT
 - Add test/common/pixeldiff, which was copied from somewhere else
 - The -node tarball is unpacked into node/ directory, adjust the references to avoid confusion.
 - Remove redundant LGPL license text in the Files: paragraph, we already have a separate License: paragraph for it.

See https://alioth-lists.debian.net/pipermail/pkg-utopia-maintainers/2026-January/037817.html

Adopted from https://github.com/cockpit-project/cockpit/commit/528415563b21ccea

----

Pretty much exactly the same as https://github.com/cockpit-project/cockpit-files/pull/1302 and https://github.com/cockpit-project/cockpit-podman/pull/2425


packaging/debian/copyright.template
```diff
@@ -1,39 +1,41 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: cockpit-machines
+Upstream-Contact: https://github.com/cockpit-project/cockpit/wiki/Contributing
```
tomasmatus, line 3: this links to `cockpit-project/cockpit`. Although it is a wiki should we explicitly link to `cockpit-project/cockpit-machines`? It could potentially avoid machines related issues be reporter on cockpit.

packaging/debian/copyright.template
```diff
@@ -1,39 +1,41 @@
...
+Copyright: 2017 - 2026  Red Hat, Inc.
+License: LGPL-2.1-or-later
+
+Files: pkg/lib/cockpit-components-checkbox-select.tsx
+ pkg/lib/cockpit-components-multi-typeahead-select.tsx
+ pkg/lib/cockpit-components-typeahead-select.tsx
+ pkg/lib/cockpit-components-simple-select.tsx
+Comment: modified copies from PatternFly react-templates
+Copyright: (c) 2019 Red Hat, Inc.
+License: MIT
```
tomasmatus, line 36: what about the special flower `src/components/vm/consoles/VncConsole.tsx` we found in https://github.com/cockpit-project/cockpit-machines/pull/2532 which is also MIT?

martinpitt: Right, this PR came before that. Well spotted, thanks! Updated.

packaging/debian/copyright.template
martinpitt: As this is an upstream *contact*, I'd rather keep this. It has our ML, Matrix room etc., none of which are c-machines specific. The latter is already the `Source:` field below.

tomasmatus: makes sense!

PR #2556, opened 2026-02-26
install: Pass empty "extraArguments" to installation script

mvollmer:
To stop it from crashing. This was overlooked in the review of b5e01d8.


mvollmer:
> As penance: one test case that would have detected this regression. 👼

Added.

mvollmer:
The "XXX" is addressed in #2557. Kust #2557 is enough, but I would like to avoid rushing that, so if we need something quick to unbreak the release, this PR is it.

allisonkarlitskaya:
→ superseded by #2557

PR #2557, opened 2026-02-26
install: set and get extraArguments from metadata

Nykseli:
Fixes the issue that was introduced in https://github.com/cockpit-project/cockpit-machines/commit/b5e01d80349729de9776166c0ec8b27742940dd7

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -729,6 +729,28 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                                      storage_pool="iscsi-pool",
                                                                      storage_volume="unit:0:0:0"))
 
+        # Check "Download an OS" with separate install phase
+        runner.createThenInstallTest(TestMachinesCreate.VmDialog(self, sourceType='os',
+                                                                 os_name=config.FEDORA_28,
+                                                                 os_short_id=config.FEDORA_28_SHORTID,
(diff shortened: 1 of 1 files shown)


src/scripts/install_machine.py
```diff
@@ -303,6 +303,14 @@ def inject_metadata(xml):
             METADATA += f"<cockpit_machines:user_login>{args['userLogin']}</cockpit_machines:user_login>"
         if args['userPassword']:
             METADATA += f"<cockpit_machines:user_password>{args['userPassword']}</cockpit_machines:user_password>"
+
+    if (args['sourceType'] == "os" or
```
mvollmer: Do we need to guard this, actually? The less we duplicate complicated conditions like this, the less we will have surprises later.

Nykseli: I don't think it's strictly needed since `prepare_installation_source` doesn't pass that data to virt-install params even if the metadata is added here. Just felt like "playing it safe". Happy to clean it up

Nykseli:
@mvollmer Added couple of (simple) tests

Nykseli:
Tests were failing because they were checking the metadata even when metadata wasn't added. Should be fixed now

mvollmer (approved):
Thanks!

PR #2567, opened 2026-03-06
helpers: Don't throw error when qemu.conf can't be read

mvollmer:
Just log a warning and return the defaults.

### DIFF
--- a/test/check-machines-consoles
+++ b/test/check-machines-consoles
@@ -749,11 +749,13 @@ fullscreen=0
         # However, in the ws-container scenario, the 1006 failure will
         # never go away.  (TODO - debug this)
 
-        if self.machine.ws_container:
+        b = self.browser
+        m = self.machine
+
+        if m.ws_container:
             b.wait_in_text(".consoles-card", "Failed to connect")
             return
 
-        b = self.browser
         for _ in range(5):
             try:
                 b.wait_in_text(".consoles-card", "Authentication failed")
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ prepare-check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common test/reference
 # run the browser integration tests
 # this will run all tests/check-* and format them as TAP
 check: prepare-check
-	test/common/run-tests ${RUN_TESTS_OPTIONS}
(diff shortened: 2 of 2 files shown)


mvollmer:
I can't reproduce the TestMachinesConsoles.testLibvirtUserUnprivileged failure locally, but I think I have seen it once earlier. Let's see...

mvollmer:
> I can't reproduce the TestMachinesConsoles.testLibvirtUserUnprivileged failure locally, but I think I have seen it once earlier. Let's see...

Alright, this seems to be a know issue, and we have code already for retrying. (I can now also reproduce it locally, I think I might have used "arch" earlier, and it's green there...)

src/components/vm/consoles/vnc.tsx
```diff
@@ -460,6 +460,7 @@ export class VncActive extends React.Component<VncActiveProps, VncActiveState> {
     async getCredentials() {
         let password = this.props.consoleDetail.password || "";
         if (!password) {
+            console.log("Reading password from qemu.conf");
```
tomasmatus, line 463: leftover debug log?

mvollmer: Yes, that was indeed unintended, but I actually think it's okay to leave it in. It's a useful piece of information, no?

mvollmer:
I can't get this green with "retry of affected tests", but it's immediately green without.

test/check-machines-consoles
```diff
@@ -778,23 +798,36 @@ fullscreen=0
...
-                    if "Failed to connect" in b.text(".consoles-card"):
-                        print("Unexpected error message. Retrying...")
-                        b.click(".consoles-card button:contains(Retry)")
-                    else:
-                        raise
-
-            b.wait_in_text(".consoles-card", "Authentication failed")
+            self.waitForAuthenticationFailed()
+
+    @testlib.skipWsContainer("never works, debug this")
```
tomasmatus: this is new here since my previous review, right? Any explanation? Is the qemu password reading broken on ws?

mvollmer: Fair, this was sloppy.  Thanks for calling it out!

The error is always !006 and never "Authentication failed", just as with the testGlobalVNCPassword test.  We can actually run this test, and just accept that we will never see "Authentication failed" in the container scenario.  This can be done in the new waitForAuthenticationFailed, and explained centrally.

PR #2573, opened 2026-03-10
DBG - trigger some "affected tests" retries

mvollmer:
I expect this to fail for debian-testing with global timeouts, but succeed for arch.


mvollmer:
> I expect this to fail for debian-testing with global timeouts, but succeed for arch.

Yes, this has happened. I couldn't get over this in #2567 no matter how often I retried.

mvollmer:
`--capacity-factor 0.8` works.

PR #2420, opened 2025-11-24
test: Don't wait for virt-install to exit on Centos 10 after all

mvollmer:
Something has changed...

PR #2578, opened 2026-03-11
test: don't wait for virt-install to exit on rhel-10

tomasmatus:
This unblocks rhel-10-2 image refresh https://github.com/cockpit-project/bots/pull/8752

Same fix as previously for centos-10 #2420

### DIFF
--- a/test/check-machines-create
+++ b/test/check-machines-create
@@ -1419,7 +1419,7 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
             # well.  (On some OSes the created VM crashes in the
             # kernel, on others, the BIOS can't find a bootable
             # device. Maybe that's the difference.)
-            if self.create_and_run and self.machine.image.startswith(('opensuse-tumbleweed', 'rhel-10')):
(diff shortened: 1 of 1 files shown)


jelly:
Lets retry once.

tomasmatus:
looks like this will need the `--capacity-factor 0.5` commit from https://github.com/cockpit-project/cockpit-machines/pull/2567

But the testCreateDownloadAnOS fails here because we don't wait for virt-install to exit it seems. Whereas on the [image refresh](https://github.com/cockpit-project/bots/pull/8752) it is the opposite

tomasmatus:
let's see what happens when I run this against the newer rhel-10-2 image

tomasmatus:
triggered `rhel-10-2@bots#8752` to see if it works against the unmerged refresh which is blocked because of this

jelly:
Oh rhel-10-2 would always be red.. so we can't land this but must in lockstep.

jelly:
We need to land the rhel-10-2 image and then we can land this PR.

jelly:
All, green lets land rhel-10-2 image and land this!

jelly (approved):
Green means go!

PR #2581, opened 2026-03-13
disks: Use virt-xml to attach and edit disks

mvollmer:
This is a revival of  #394, finally.

------

### Machines: Pool volumes are now added as disks of type "file" or "block"

Previously, pool volumes were added with type "volume".  This has various problems, such as preventing to take external snapshots.  Cockpit now follows `virt-install`,`virt-xml`, and `virt-manager` and does not use type "volume" anymore for newly added disks.

### DIFF
--- a/src/components/storagePools/storageVolumeCreateBody.tsx
+++ b/src/components/storagePools/storageVolumeCreateBody.tsx
@@ -52,7 +52,7 @@ export function update_VolumeCreate(field: DialogField<VolumeCreateValue>, stora
 
 export function validate_VolumeCreate(field: DialogField<VolumeCreateValue>) {
     field.sub("name").validate(v => {
-        if (!v)
+        if (!v.trim())
             return _("Please enter new volume name");
     });
     field.sub("newSize").validate(v => {
--- a/src/components/vm/disks/diskAdd.tsx
+++ b/src/components/vm/disks/diskAdd.tsx
@@ -23,9 +23,9 @@ import {
     VolumeCreate, init_VolumeCreate, update_VolumeCreate, validate_VolumeCreate,
     type VolumeCreateValue,
 } from '../../storagePools/storageVolumeCreateBody.jsx';
-import { domainAttachDisk, domainGet, virtXmlHotEdit, domainIsRunning } from '../../../libvirtApi/domain.js';
+import { domainGet, virtXmlHotAdd, virtXmlHotEdit, domainIsRunning } from '../../../libvirtApi/domain.js';
 import { storagePoolGetAll } from '../../../libvirtApi/storagePool.js';
-import { storageVolumeCreateAndAttach } from '../../../libvirtApi/storageVolume.js';
+import { storageVolumeCreate } from '../../../libvirtApi/storageVolume.js';
 import { appState } from '../../../state';
 
 import {
@@ -55,8 +55,8 @@ function clearSerial(serial: string): string {
 
 function getFilteredVolumes(vmStoragePool: StoragePool, disks: Record<string, VMDisk>): StorageVolume[] {
     const usedDiskPaths = Object.getOwnPropertyNames(disks)
-            .filter(target => disks[target].source && (disks[target].source.file || disks[target].source.volume))
-            .map(target => (disks[target].source && (disks[target].source.file || disks[target].source.volume)));
+            .map(target => (disks[target].source && (disks[target].source.file || disks[target].source.volume || disks[target].source.dev)))
+            .filter(Boolean);
 
     const filteredVolumes = (vmStoragePool.volumes || []).filter(volume => !usedDiskPaths.includes(volume.path) && !usedDiskPaths.includes(volume.name));
 
@@ -387,7 +387,7 @@ function validate_CustomPath(field: DialogField<CustomPathValue>) {
     const { _fileInfoCache } = field.get();
     field.sub("file").validate_async(0, async file => {
         if (!file)
-            return _("Path can not be empty");
+            return _("Path cannot be empty");
 
         const { usesBackingFile } = await getFileInfo(file, _fileInfoCache);
         if (usesBackingFile)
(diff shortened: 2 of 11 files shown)


src/components/vm/disks/diskEdit.tsx
```diff
@@ -171,17 +173,26 @@ const EditDisk = ({
...
-                existingTargets
-            });
+                "disk",
+                { target: disk.target },
+                {
+                    target: newTarget,
+                    bus: values.bus.type,
+                    driver: { cache: values.cache },
+                    readonly: (values.access.mode == "readonly") ? "yes" : "no",
+                    xpath: [{ delete: "./address" }],
```
mvollmer: Maybe only do this when the bus has changed?

src/components/vm/disks/diskEdit.tsx
```diff
@@ -171,17 +173,26 @@ const EditDisk = ({
...
-                busType: values.bus.type,
-                cache: values.cache,
-                existingTargets
-            });
+                "disk",
+                { target: disk.target },
+                {
+                    target: newTarget,
+                    bus: values.bus.type,
+                    driver: { cache: values.cache },
```
mvollmer: This can just be `cache: ...` without `driver`.

src/components/vm/disks/diskEdit.tsx
mvollmer: Done.

src/components/vm/disks/diskEdit.tsx
mvollmer: Done.

mvollmer:
`TestMachinesDisks.testAddDiskPool` fails with 

```
> error: Disk failed to be added:
ERROR    Error attempting device hotplug: internal error:
unable to execute QEMU command 'device_add': Failed to get "write" lock
```

Hotplugging in general seems to work.

mvollmer:
> `TestMachinesDisks.testAddDiskPool` fails with
> 
> ```
> > error: Disk failed to be added:
> ERROR    Error attempting device hotplug: internal error:
> unable to execute QEMU command 'device_add': Failed to get "write" lock
> ```
> 
> Hotplugging in general seems to work.

This happens because the detection of which volumes are in use by which VMs is broken when the volume is used via type="block" and "source.dev". This already works for type="file" but now we add volumes from "disk" pools as type="block" so we need to extend that.

test/check-machines-storage-pools
```diff
@@ -572,34 +575,33 @@ class TestMachinesStoragePools(machineslib.VirtualMachinesCase):
...
+                    self.set_TextInput("volume.unit", self.unit)
 
                 if self.fmt:
-                    b.select_from_dropdown("#create-volume-dialog-format", self.fmt)
+                    self.set_DropdownSelect("volume.format", self.fmt)
 
             def cancel(self):
-                b.click(".pf-v6-c-modal-box__footer button:contains(Cancel)")
+                b.click(self.cancel_button())
                 b.wait_not_present("#create-storage-pool-dialog")
```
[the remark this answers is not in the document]
mvollmer: Excellent!

test/check-machines-disks
```diff
@@ -789,8 +757,8 @@ class TestMachinesDisks(machineslib.VirtualMachinesCase):
             if self.cache_mode:
                 b.wait_in_text(f"#vm-{self.vm_name}-disks-{self.expected_target}-cache", self.cache_mode)
 
-            if self.expected_serial:
-                b.wait_in_text(f"#vm-{self.vm_name}-disks-{self.expected_target}-serial", self.expected_serial)
+            if self.serial:
+                b.wait_in_text(f"#vm-{self.vm_name}-disks-{self.expected_target}-serial", self.serial)
```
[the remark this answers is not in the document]
mvollmer: True, but as long as it doesn't really happen in the test, we don't need all this extra code.

src/components/vm/disks/diskAdd.tsx
```diff
@@ -74,907 +77,679 @@ function getDiskUsageMessage(vms: VM[], storagePool: StoragePool, volumeName: st
...
+            const target = getNextAvailableTarget(existingTargets, values.additional_options.bus_type);
+
+            if (!target)
+                throw new DialogError(_("Failed to add disk"), _("Can not determine guest device name"));
+
+            const common = {
+                target,
+                cache: values.additional_options.cache_mode == "default" ? null : values.additional_options.cache_mode,
+                bus: values.additional_options.bus_type,
+                serial: values.additional_options.serial,
```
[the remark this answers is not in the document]
mvollmer: Very good!

src/components/storagePools/storageVolumeCreateBody.tsx
```diff
@@ -6,179 +6,147 @@
...
-                        value={volumeName || ""}
-                        validated={validationStateName}
-                        onChange={(_, value) => onValueChanged('volumeName', value)} />
-            <FormHelper fieldId={`${idPrefix}-name`} helperTextInvalid={validationStateName == "error" ? validationFailed.volumeName : null} />
-        </FormGroup>
-    );
-};
+export function validate_VolumeCreate(value: DialogField<VolumeCreateValue>) {
+    value.sub("name").validate(v => {
+        if (!v)
```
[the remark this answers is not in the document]
mvollmer: Nice. libvirt will indeed create a volume with name "   " if asked.

src/components/vm/disks/diskAdd.tsx
```diff
@@ -74,907 +77,679 @@ function getDiskUsageMessage(vms: VM[], storagePool: StoragePool, volumeName: st
...
+        device: "disk",
+        _fileInfoCache: {},
+    };
+}
+
+function validate_CustomPath(field: DialogField<CustomPathValue>) {
+    const { _fileInfoCache } = field.get();
+    field.sub("file").validate_async(0, async file => {
+        if (!file)
+            return _("Path can not be empty");
```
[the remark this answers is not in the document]
mvollmer: Ok.

src/components/vm/disks/diskEdit.tsx
```diff
@@ -171,17 +173,45 @@ const EditDisk = ({
...
+            newTarget = getNextAvailableTarget(existingTargets, values.bus.type);
+            if (!newTarget)
+                throw new DialogError(cockpit.format(_("No free targets for bus type $0."), values.bus.type));
+
+            // HACK - virt-xml will not reset the address when
+            // changing the bus type, and we need to add
+            // "xpath0.delete=./address" explicitly. However, not all
+            // versions of virt-xml that we care about support xpath
+            // yet, so we reset the address explicitly to something
+            // acceptable.
```
martinpitt, line 189: OOI, how bad is it? we just dropped Ubuntu 22.04, but I figure RHEL 8 is similarly old?

src/components/storagePools/storageVolumeCreateBody.tsx
```diff
@@ -52,7 +52,7 @@ export function update_VolumeCreate(field: DialogField<VolumeCreateValue>, stora
 
 export function validate_VolumeCreate(field: DialogField<VolumeCreateValue>) {
     field.sub("name").validate(v => {
-        if (!v)
+        if (!v.trim())
```
martinpitt, line 55: Is this some subtle difference between API and virt-xml wrt. volumes which are named like " "? I welcome this change, such a name is surely a typo. Should perhaps get a test somewhere, but not important.

src/components/vm/disks/diskAdd.tsx
```diff
@@ -387,7 +387,7 @@ function validate_CustomPath(field: DialogField<CustomPathValue>) {
     const { _fileInfoCache } = field.get();
     field.sub("file").validate_async(0, async file => {
         if (!file)
-            return _("Path can not be empty");
+            return _("Path cannot be empty");
```
martinpitt, line 390: I know the robodude suggested that, but does anybody think of the children^Wexisting translations? (They all get invalidated now). The internet seems to pretty much agree that "cannot" is right.

Anyway, not too hard to fix the fuzziness in weblate.

src/components/vm/disks/diskAdd.tsx
```diff
@@ -648,56 +648,68 @@ export const AddDisk = ({
...
-
             const common = {
-                permanent: values.permanent,
-                hotplug,
-                vmId: vm.id,
-                cacheMode: values.additional_options.cache_mode,
-                busType: values.additional_options.bus_type,
-                serial: values.additional_options.serial,
+                target,
+                cache: values.additional_options.cache_mode == "default" ? null : values.additional_options.cache_mode,
```
martinpitt: I'm surprised by this -- translating `"default"` into `null` is the *opposite* of what I'd expect when moving  from API to virt-xml CLI. I suppose/hope the tests cover this, but do you have an explanation for that? `null` has to be translated to something textual for the CLI arguments after all?

martinpitt (commented):
Cheers! Overall this is very clean. You already went through an interation with roboreview too. Most of the threads below are thinking aloud, but I would like to understand the default → null conversion.

src/components/vm/disks/diskEdit.tsx
mvollmer: Yes, I think it was just rhel 8.

src/components/storagePools/storageVolumeCreateBody.tsx
mvollmer: I think this was an oversight in https://github.com/cockpit-project/cockpit-machines/pull/2449

src/components/vm/disks/diskAdd.tsx
mvollmer: Yeah, hmm. So, `null` means there will be no "cache=..." at all in the command line. "cache" will be omitted.  This works because this is just adding a new disk, and virt-xml/libvirt will fill in the value for "cache".

But this would be wrong when editing disks, and there we indeed use "default" literally, which is inconsistent with what we do here for adding.  Hmm, I play with this a bit and if I can't find anything wrong with "cache=default" while adding, I'll change the code to do that.

mvollmer: So "cache=default" results in there not being any "cache" attribute in the disk XML, both when adding and when editing. So we really should be using "cache=default" also when adding. Thanks!

Done.

mvollmer: Aha, https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2581-7715d739-20260430-145036-arch/TestMachinesDisks-testAddDiskPool-arch-127.0.0.2-2401-FAIL.png

So apparently "cache=default" does not work with block devices: it needs to be omitted or explicitly be "cache=none"...

mvollmer: I have reverted to the original behavior, and added a comment.

martinpitt: Excellent, thanks!

martinpitt (approved):

PR #2606, opened 2026-04-07
disks: Use virt-xml to remove disks

mvollmer:
And also delete the volumes corresponding to `type="block"` disks.

- [x] #2581 
- [x] tests

### DIFF
--- a/src/components/vm/disks/vmDiskColumns.tsx
+++ b/src/components/vm/disks/vmDiskColumns.tsx
@@ -6,7 +6,7 @@
 import React from 'react';
 import cockpit from 'cockpit';
 
-import type { optString, VM, VMDisk, StoragePool } from '../../../types';
+import type { optString, VM, VMDisk } from '../../../types';
 import type { DeleteResourceModalProps } from '../../common/deleteResource';
 
 import { Button } from "@patternfly/react-core/dist/esm/components/Button";
@@ -18,7 +18,7 @@ import { KebabDropdown } from 'cockpit-components-dropdown.jsx';
 import { basename } from 'cockpit-path';
 
 import { useDialogs } from 'dialogs.jsx';
-import { domainDeleteStorage, domainDetachDisk, domainGet } from '../../../libvirtApi/domain.js';
+import { domainDeleteStorage, virtXmlHotRemove, domainGet } from '../../../libvirtApi/domain.js';
 import { MediaEjectModal } from './mediaEject.jsx';
 import { EditDiskAction } from './diskEdit.jsx';
 import { AddDisk } from './diskAdd.jsx';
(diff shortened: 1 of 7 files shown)


mvollmer:
Hmm, we don't get any coverage anymore for cockpit-machines...

src/components/vm/disks/vmDiskColumns.tsx
```diff
@@ -139,17 +143,16 @@ export const RemoveDiskModal = ({
...
 }) => {
+    const [info, setInfo] = useState<null | StorageDeleteInfo>(null);
+    useInit(async () => { setInfo(await getStorageDeleteInfoForDisk(vm, disk)) });
+
     const onRemoveDisk = async (deleteFile: boolean) => {
         await virtXmlHotRemove(vm, "disk", { target: disk.target }, disk.target in vm.inactiveXML.disks);
         await domainGet(vm);
 
-        if (deleteFile) {
+        if (deleteFile && info) {
```
martinpitt: Is this a race condition? is it possible to click the remove button before the async useInit detection is done? (this might hang for several seconds?)

src/components/vm/disks/vmDiskColumns.tsx
```diff
@@ -163,7 +166,7 @@ export const RemoveDiskModal = ({
 
     const dialogProps: DeleteResourceModalProps = {
         title: _("Remove disk from VM?"),
-        actionName: _("Remove"),
+        actionName: _("Remove disk"),
```
martinpitt: Spurious string change?

test/check-machines-create
```diff
@@ -704,6 +704,10 @@ class TestMachinesCreate(machineslib.VirtualMachinesCase):
                                                                  storage_pool="poolDisk",
                                                                  storage_volume=partition))
 
+        # Deleting the storage of the machine will remove the
+        # partition in "poolDisk". Recreate it.
+        self.machine.execute(f"virsh vol-create-as poolDisk {partition} 1024")
```
martinpitt, line 709: Should this better test that claim, or are you confident that this will fail if the vol already exists?

src/components/vm/disks/vmDiskColumns.tsx
mvollmer: No, it's not a race, good catch. I will make this more explicit by using `info` as a parameter .

src/components/vm/disks/vmDiskColumns.tsx
mvollmer: Can be argued, yes.  The secondary button has changed from "Remove and delete file" to "Remove disk and delete disk image" and so I have changed the primary button along with it... But it's not well motivated, so I revert that.

test/check-machines-create
mvollmer: I _assume_ it will fail... :-) I'll make sure.

mvollmer: Yes, it fails when the volume already exists:

```
error: Failed to create vol sda1
error: storage volume 'sda1' exists already
```

martinpitt (approved):
Cheers!

mvollmer:
> Hmm, we don't get any coverage anymore for cockpit-machines...

Solved in #2633, hopefully.

PR #2633, opened 2026-05-04
build: Fix coverage reporting

mvollmer:
We need to build the bundle with NODE_ENV=development.  This is achieved by building it outside the test VM with that option, and then switching rebuilding off inside the test VM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,9 +177,13 @@ else
 VM_INSTALL = --upload $(NODE_CACHE):/var/tmp --build $(TARFILE) --script $(CURDIR)/test/vm.install
 endif
 
+ifeq ($(NODE_ENV),development)
+VM_NO_REBUILD = --build-options '--define "rebuild_bundle 0"'
+endif
+
 # build a VM with locally built distro pkgs installed
 $(VM_IMAGE): $(TARFILE) $(NODE_CACHE) packaging/debian/rules packaging/debian/control packaging/arch/PKGBUILD bots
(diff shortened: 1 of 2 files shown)


mvollmer:
This is how cockpit itself does it. Maybe it would be better to just `--define "NODE_ENV development"` or something like that.

mvollmer:
Hmm, TestMachinesCreate.testCreateThenInstall reliably fails in TF in the new test code from #2606, but it has passed in #2606... I'll debug that separately.

mvollmer:
Coverage has been produced: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2633-f3d263d7-20260504-165915-fedora-43-devel/Coverage/index.html

(And, yeah, storage pool coverage is pretty low...)

martinpitt (approved):
Looks ok, just not sure how this breaks [testCreateThenInstall](https://artifacts.dev.testing-farm.io/1f7ba90e-d2e1-456b-b714-38d55fbec340/)? It fails on all TF targets, that doesn't look like a coincidence?

mvollmer:
> Looks ok, just not sure how this breaks [testCreateThenInstall](https://artifacts.dev.testing-farm.io/1f7ba90e-d2e1-456b-b714-38d55fbec340/)? It fails on all TF targets, that doesn't look like a coincidence?

Good question. it did pass when I made that change... I don't think it's a coincidence, I am debugging.

PR #2634, opened 2026-05-05
test: Enable "Delete VM" only once all disk info has been retrieved

mvollmer:


mvollmer:
We get failures on TF in this area that could be explained by the dialog not actually listing the block devices for some reason.

mvollmer:
Right, there is a race condition in the "Delete VM" dialog now: The list of disks to remove is populated while it is already open and while the "Delete" button is enabled. If the test clicks fast enough, no disk images will be removed.

The new code here in the test removes that race by explicitly waiting for the disk to appear in the dialog. There is also new code to remove the race completely by only enabling the "Delete" button after the list of disks is complete.

martinpitt (dismissed):
Excellent, thanks!

mvollmer:
rhel-9-9 doesn't seem to want to cooperate.

mvollmer:
> rhel-9-9 doesn't seem to want to cooperate.

Reproduced locally, but not reliably.

mvollmer:
> > rhel-9-9 doesn't seem to want to cooperate.
> 
> Reproduced locally, but not reliably.

Grr. libvirt gets stuck on rhel-9-9 somtimes so that volume lookups by path don't work anymore. Even things like `virsh vol-name ...` will fail. At the same time `virsh vol-list` will happily output that same path.

Restarting a couple of daemons gets it unstuck.

martinpitt (approved):
Thanks! Strong deja-vu on this one, I think you fixed the same thing a few days ago on disk deletion?

But let's retry [these](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2634-e7981357-20260506-175915-ubuntu-2404/log.html) [flakes](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2634-e7981357-20260506-175917-ubuntu-stable/log.html)

mvollmer:
> Thanks! Strong deja-vu on this one, I think you fixed the same thing a few days ago on disk deletion?

I think it's the same PR! :-) I had to adjust it to be reliable on rhel-9-9.

PR #2641, opened 2026-05-07
Spawn cleanups

jelly:


src/libvirtApi/network.ts
```diff
@@ -187,12 +187,18 @@ function networkUpdateStaticHostEntries({
     if (isNetworkActive)
         flags |= Enum.VIR_NETWORK_UPDATE_AFFECT_LIVE;
 
+    const doc = document.implementation.createDocument('', '', null);
```
allisonkarlitskaya, line 190: I'm good with everything else, but someone else will need to comment on this.  I think I'd have preferred a "pure data" quoting-based approach...

allisonkarlitskaya: Something like

```js
[...macAddress].map(c => `&#${c.codePointAt(0)};`).join('');
```

is a big hammer here...

jelly: Hmm well a lot of machines code uses this pattern of using the DOM API and it should be safe. @mvollmer should get the final say here I'd say :)

mvollmer (approved):
Thanks!

martinpitt (approved):
Thanks!

PR #2650, opened 2026-05-18
HACKING: Show how to run a tiny test VM

martinpitt:
Let's start a debugging tips/snippets collection.

---

I am transferring my personal notes into our projecs. This is the only one I had for c-machines, but hopefully it'll grow?


jelly (approved):
This is useful but maybe even more useful would be to boot an alpine image as you can test life cycle events on that image.

But that maybe requires us to pull the the image when creating the image with `image-customize` into /var/lib/libvirt/images/ instead of doing it dynamically during the test?

martinpitt:
Good idea, perhaps @mvollmer can add that as a second snippet? This tiny one is nice as you can easily create several dozen sub-VMs in our standard test VMs and don't need any actual on-disk image, so I'd like to keep it. But of course having a "real" bootable OS makes sense in some situations as well.

PR #2662, opened 2026-05-29
Bump @novnc/novnc to 1.7

mvollmer:
Version 1.7 is now implemented as ESM modules, which makes it possible for us to bundle it even though it contains a top-level await.

This needs support for top-level await in the Browser, however, and we need to output the cockpit-machines bundle as a ESM module itself.

Scripts with type="module" are "defer" by default, so we don't need to (and actually can't) wait for the "DOMContentLoaded" event.

XXX - there is no [@types ](https://www.npmjs.com/package/@types/novnc__novnc) for novnc 1.7 yet, so this is blocked on that in any case.

### DIFF
--- a/build.js
+++ b/build.js
@@ -96,6 +96,7 @@ function watch_dirs(dir, on_change) {
 const context = await esbuild.context({
     ...!production ? { sourcemap: "linked" } : {},
     bundle: true,
+    format: "esm",
     entryPoints: ['./src/index.js'],
     // Allow external font files which live in ../../static/fonts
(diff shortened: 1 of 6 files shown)


mvollmer:
This is blocked on `@types/novnc__novnc` version 1.7 being released.

mvollmer:
> This is blocked on `@types/novnc__novnc` version 1.7 being released.

The necessary change from 1.6 to 1.7 seems to just be to replace

```
declare module "@novnc/novnc/lib/rfb" {
```

with

```
declare module "@novnc/novnc" {
```

so if we get desperate we could could just install our own copy of it and make this change.

PR #2668, opened 2026-06-05
consoles: replace deprecated substr() with slice()

jelly:


src/components/vm/consoles/common.tsx
```diff
@@ -36,15 +36,15 @@ export function connection_address() {
     if (cockpit.transport.host == "localhost") {
         const app = cockpit.transport.application();
         if (app.startsWith("cockpit+=")) {
-            address = app.substr(9);
+            address = app.slice(9);
```
mvollmer, line 39: Hmm, I would have replaced it with `substring`.... only bc I didn't know that `slice` also works with strings... Ok, I guess `slice` is the best thing to use going forward, right?

mvollmer (approved):

src/components/vm/consoles/common.tsx
jelly: I would be happy to have done that :) [substring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) is baseline available.

PR #2682, opened 2026-06-22
Bump to Patternfly 6.5

mvollmer:


mvollmer:
Draft until we have the new pixel references.

src/components/vm/hostdevs/hostDevAdd.tsx
```diff
@@ -154,7 +154,7 @@ const DevRow = ({
                 <Tbody>
                     {selectableDevices.map((dev, rowIndex) => {
                         return (
-                            <Tr key={"row-" + rowIndex}>
+                            <Tr key={"row-" + rowIndex} data-row={rowIndex}>
```
Venefilyn, line 157: If it's only for tests we should use ouia stuff IMO

Venefilyn (approved):
LGTM

PR #2705, opened 2026-07-09
Drop unneeded admin session cleanup

jelly:
It's not entirely clear if this breaks but running this reproducer on debian-trixie reproduces the same issue

```
systemctl reset-failed user@1002.service 2>/dev/null
systemctl stop user@1002.service 2>/dev/null

for i in 1 2 3 4 5 6; do
    echo "--- attempt $i ---"
    systemctl start user@1002.service
    loginctl terminate-user 1002 2>/dev/null || true
    loginctl kill-user 1002 2>/dev/null || true
    pkill -9 -u admin | grep -v "systemd --user" || true
    ps aux | grep admin
while pgrep -u admin; do sleep 0.2; done
    sleep 0.3
done

systemctl show user@1002.service -p Result -p ActiveState
```

### DIFF
--- a/test/machineslib.py
+++ b/test/machineslib.py
@@ -494,13 +494,6 @@ class VirtualMachinesCase(VirtualMachinesCaseHelpers, storagelib.StorageHelpers,
         else:
             m.execute("systemctl reset-failed libvirtd; systemctl try-restart libvirtd")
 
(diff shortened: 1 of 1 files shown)


tomasmatus (approved):
sounds reasonable, thanks

PR #2709, opened 2026-07-14
all: Stop using cockpit.user

mvollmer:
It is deprecated and cockpit.info.user is nicer since we can assume that it is always available.

### DIFF
--- a/src/components/common/machinesConnectionSelector.tsx
+++ b/src/components/common/machinesConnectionSelector.tsx
@@ -6,8 +6,6 @@
 
 import React from 'react';
 
-import { appState } from '../../state';
-
 import { LIBVIRT_SYSTEM_CONNECTION, LIBVIRT_SESSION_CONNECTION, rephraseUI } from '../../helpers.js';
 import { FormGroup } from "@patternfly/react-core/dist/esm/components/Form";
 import { Radio } from "@patternfly/react-core/dist/esm/components/Radio";
@@ -36,8 +34,7 @@ export const MachinesConnectionSelector = ({
     showInfoHelper,
     isReadonly,
 }: MachinesConnectionSelectorProps) => {
-    cockpit.assert(appState.loggedUser);
-    if (appState.loggedUser.id == 0)
+    if (cockpit.info.user.uid == 0)
         return null;
 
     return (
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -7,6 +7,7 @@ import "cockpit-dark-theme";
 import 'patternfly/patternfly-6-cockpit.scss';
 import 'polyfills'; // once per application
 
+import cockpit from 'cockpit';
 import React from 'react';
 import { createRoot } from 'react-dom/client';
 
@@ -25,6 +26,7 @@ function appMain(): void {
 
 (function() {
     document.addEventListener("DOMContentLoaded", async function() {
+        await cockpit.init();
         await load_config();
         appMain();
     });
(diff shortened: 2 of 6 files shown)


src/components/common/machinesConnectionSelector.tsx
```diff
@@ -36,8 +34,7 @@ export const MachinesConnectionSelector = ({
     showInfoHelper,
     isReadonly,
 }: MachinesConnectionSelectorProps) => {
-    cockpit.assert(appState.loggedUser);
-    if (appState.loggedUser.id == 0)
+    if (cockpit.info.user.uid == 0)
```
allisonkarlitskaya, line 37: `cockpit.info` is async initialized, and (I think) only happens after the user has called (and awaited) `cockpit.init()`.  Is there any guarantee that the user in question has done that?

src/index.tsx
```diff
@@ -25,6 +26,7 @@ function appMain(): void {
 
 (function() {
     document.addEventListener("DOMContentLoaded", async function() {
+        await cockpit.init();
```
allisonkarlitskaya, line 29: ....and there it is.  I guess that happens for *any* page that might be loaded?  Is there any way to navigate directly to a different URL?

allisonkarlitskaya (approved):
Nice cleanup, thanks!

See the comment below before merging.

src/components/common/machinesConnectionSelector.tsx
allisonkarlitskaya: Also: this is slightly inconsistent: you compare uid == 0 here and username == "root" in another part of this patch.  But that's a pre-existing issue, obviously.

mvollmer: > `cockpit.info` is async initialized, and (I think) only happens after the user has called (and awaited) `cockpit.init()`. Is there any guarantee that the user in question has done that?

Yes, this happens in `index.tsx` before the app is mounted.

This is what makes cockpit.info nicer: It's a hassle to account for the asynchronicity of cockpit.user together with having it globally available. cockpit.info does this by lying to TS, which I think is the correct compromise here. Once we can use top-level await, this lie should no longer be necessary, but we might need to change to cockpit.info API again...

mvollmer: > Also: this is slightly inconsistent: you compare uid == 0 here and username == "root" in another part of this patch. But that's a pre-existing issue, obviously.

Yes, I noticed as well. but decided to leave it alone...

src/index.tsx
mvollmer: No, this is guaranteed to happen before any React component renders.

mvollmer:
ubuntu-2404 seems to be completely broken by this... weird.

mvollmer:
> ubuntu-2404 seems to be completely broken by this... weird.

`cockpit.init()` fails on ubuntu-2404 with "not-supported".  cockpit-files has a workaround:

```
    try {
        await cockpit.init();
    } catch (exp: any) { /* eslint-disable-line @typescript-eslint/no-explicit-any */
        /* Remove this when we take a dependency on Cockpit 336 ('info' channel in the bridge) */
        if (exp.problem === 'not-supported') {
            const user = await cockpit.user();
            cockpit.info.user = {
                fullname: user.full_name,
                group: user.groups[0],
                uid: user.id,
                ...user,
            };
        }
    }
```

Grr.

mvollmer:
> Grr.

I don't feel like having the same workaround in c-machines. I'd rather keep using cockpit.user, maybe with a top-level await. I'll try that out.

PR #2745, opened 2026-08-12
test: Drop Ubuntu 24.04 test workarounds

jelly:
Cockpit no longer tests on Ubuntu 24.04.

### DIFF
--- a/test/check-machines-consoles
+++ b/test/check-machines-consoles
@@ -568,19 +568,14 @@ fullscreen=0
 
         assert_state("This machine has a SPICE graphical console that can not be shown here.")
 
-        if self.machine.image in ["ubuntu-2404"]:
-            # virt-xml in older Ubuntus does not have the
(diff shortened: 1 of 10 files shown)


test/check-machines-consoles
```diff
@@ -568,19 +568,14 @@ fullscreen=0
 
         assert_state("This machine has a SPICE graphical console that can not be shown here.")
 
-        if self.machine.image in ["ubuntu-2404"]:
```
mvollmer, line 571: Cool, this means we can now rely on --convert-to-vnc always being present.  Let's clean up the code as well.

mvollmer (dismissed):
Let's remove the code that checks for `--convert-to-vnc` etc,

mvollmer:
We should really find a way to test convert-to-vnc on rhel and centos.

mvollmer (approved):
Thanks! Please merge only when you agree with my commit here.

PR #2760, opened 2026-08-26
Update mainframe emulation

SludgeGirl:
Mainframes emulate previous versions of POWER by using host-model as the mode and model as the Power version. This set of changes allow for individual selection of mode and model on ppc architectures to have compatibility for how it handles emulation

### DIFF
--- a/src/components/vm/overview/cpuModal.tsx
+++ b/src/components/vm/overview/cpuModal.tsx
@@ -4,14 +4,14 @@
  * Copyright (C) 2023 Red Hat, Inc.
  */
 
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import cockpit from 'cockpit';
 
 import type { optString, VM } from '../../../types';
 
 import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form";
-import { FormSelect, FormSelectOption, FormSelectOptionGroup } from "@patternfly/react-core/dist/esm/components/FormSelect";
+import { FormSelect, FormSelectOption } from "@patternfly/react-core/dist/esm/components/FormSelect";
 import {
     Modal, ModalBody, ModalFooter, ModalHeader
 } from '@patternfly/react-core/dist/esm/components/Modal';
@@ -57,6 +57,15 @@ export const CPUModal = ({
     models: string[],
 }) => {
     const Dialogs = useDialogs();
+    const [cpuArch, setCpuArch] = useState<string | null>(null);
+
+    useEffect(() => {
+        cockpit.spawn(["uname", "-m"])
+                .then((output) => {
+                    setCpuArch(output);
+                })
+                .catch((error) => console.log(error));
+    }, []);
 
     function getInt(val: optString, def: number) {
         return val ? parseInt(val) : def;
@@ -302,26 +311,30 @@ export const CPUModal = ({
             </FormGroup>
 
             <FormGroup id="cpu-model-select-group" label={_("Mode")}>
-                <FormSelect value={cpuModel || cpuMode}
-                            aria-label={_("Mode")}
-                            onChange={(_event, value) => {
-                                if ((value == "host-model" || value == "host-passthrough")) {
-                                    setCpuMode(value);
-                                    setCpuModel(undefined);
-                                } else {
-                                    setCpuModel(value);
-                                    setCpuMode("custom");
-                                }
-                            }}>
+                <FormSelect value={cpuMode}
+                        aria-label={_("Mode")}
+                        onChange={(_event, value) => {
+                            setCpuMode(value);
+                        }}>
                     <FormSelectOption key="host-model"
-                                      value="host-model"
-                                      label="host-model" />
+                                value="host-model"
+                                label="host-model" />
                     <FormSelectOption key="host-passthrough"
-                                      value="host-passthrough"
-                                      label="host-passthrough" />
-                    <FormSelectOptionGroup key="custom" label={_("custom")}>
-                        {models.map(model => <FormSelectOption key={model} value={model} label={model} />)}
-                    </FormSelectOptionGroup>
+                                value="host-passthrough"
+                                label="host-passthrough" />
+                    <FormSelectOption key="custom"
+                                value="custom"
+                                label="custom" />
+                </FormSelect>
+            </FormGroup>
+            <FormGroup id="cpu-model-select-group" label={_("Model")}>
+                <FormSelect value={cpuModel}
+                        aria-label={_("Model")}
+                        onChange={(_event, value) => {
+                            setCpuModel(value);
+                        }}>
+                    <FormSelectOption key="none" value="none" label="none" />
+                    {models.map(model => <FormSelectOption key={model} value={cpuArch && cpuArch.startsWith("ppc") ? model.toLowerCase() : model} label={model} />)}
                 </FormSelect>
             </FormGroup>
         </Form>
--- a/src/libvirt-xml-parse.ts
+++ b/src/libvirt-xml-parse.ts
@@ -458,9 +458,7 @@ export function parseDumpxmlForCpu(cpuElem: Element | undefined): VMCpu {
     }
 
     cpu.mode = get_attr(cpuElem, 'mode') || "custom";
-    if (cpu.mode === 'custom') {
-        cpu.model = get_text(cpuElem, 'model');
-    }
+    cpu.model = get_text(cpuElem, 'model');
 
     const topologyElem = get_child(cpuElem, 'topology');
     if (topologyElem) {


src/components/vm/overview/cpuModal.tsx
```diff
@@ -301,8 +311,45 @@ export const CPUModal = ({
                 </FormSelect>
             </FormGroup>
 
-            <FormGroup id="cpu-model-select-group" label={_("Mode")}>
-                <FormSelect value={cpuModel || cpuMode}
+            {cpuArch && cpuArch.startsWith("ppc")
```
mvollmer: Would it be wrong to use your new code always, instead of only for ppc?

mvollmer:
> Mainframes emulate previous versions of POWER by using host-passthrough as the mode and model as the Power version. 

Do they use `mode == "custom"` at all?  Do we maybe need to offer a different set of models for "host-passthrough" vs "custom"?

src/components/vm/overview/cpuModal.tsx
```diff
@@ -301,8 +311,45 @@ export const CPUModal = ({
...
+                // Running vm's of ppc can be a little weird.
+                // Since it has a special mode for emulating previous versions of
+                // the Power models. This accounts for this by allowing cpu mode
+                // to be selectable on it's own since this is often set to host-model
+                // as opposed to the usual custom mode when emulating other cpu's
+                ? <FormGroup id="cpu-model-select-group" label={_("Mode/Model")}>
```
mvollmer: I'd say the two FormSelects should get their own label each:

```
Mode:   [ ..... | v ]
Model:  [ ..... | v ]
```

mvollmer:
Do the domain capabilities contain anything interesting?  Can you show the output of `virsh domcapabilities`?

src/components/vm/overview/cpuModal.tsx
SludgeGirl: No this would be fine, I just wrote this assuming everything else would prefer to keep the current behavior

SludgeGirl:
> > Mainframes emulate previous versions of POWER by using host-model as the mode and model as the Power version.
> 
> Do they use `mode == "custom"` at all? Do we maybe need to offer a different set of models for "host-passthrough" vs "custom"?

So it only uses host-model (sorry I wrote this wrong in my initial comment) for different power versions, as these are run through a special mode. Anything else would need to be done through custom since these would be run via an emulator

> Do the domain capabilities contain anything interesting? Can you show the output of `virsh domcapabilities`?

Sure:

```xml
$ virsh domcapabilities
<domainCapabilities>
  <path>/usr/bin/qemu-system-ppc64</path>
  <domain>kvm</domain>
  <machine>pseries-11.0</machine>
  <arch>ppc64le</arch>
  <vcpu max='4096'/>
  <iothreads supported='yes'/>
  <os supported='yes'>
    <enum name='firmware'/>
    <firmwareFeatures supported='yes'>
      <enum name='secureBoot'/>
      <enum name='enrolledKeys'/>
    </firmwareFeatures>
    <loader supported='yes'>
      <enum name='type'>
        <value>rom</value>
        <value>pflash</value>
      </enum>
      <enum name='readonly'>
        <value>yes</value>
        <value>no</value>
      </enum>
      <enum name='secure'>
        <value>no</value>
      </enum>
    </loader>
    <varstore supported='no'/>
  </os>
  <cpu>
    <mode name='host-passthrough' supported='yes'>
      <enum name='hostPassthroughMigratable'>
        <value>off</value>
      </enum>
    </mode>
    <mode name='maximum' supported='no'/>
    <mode name='host-model' supported='yes'>
      <model fallback='allow'>POWER11</model>
      <vendor>IBM</vendor>
    </mode>
    <mode name='custom' supported='yes'>
      <model usable='unknown' vendor='IBM'>POWER10</model>
      <model usable='unknown' vendor='IBM'>POWER11</model>
      <model usable='unknown' vendor='IBM'>POWER7</model>
      <model usable='unknown' vendor='IBM'>POWER8</model>
      <model usable='unknown' vendor='IBM'>POWER9</model>
    </mode>
  </cpu>
  <memoryBacking supported='yes'>
    <enum name='sourceType'>
      <value>file</value>
      <value>anonymous</value>
      <value>memfd</value>
    </enum>
  </memoryBacking>
  <devices>
    <disk supported='yes'>
      <enum name='diskDevice'>
        <value>disk</value>
        <value>cdrom</value>
        <value>lun</value>
      </enum>
      <enum name='bus'>
        <value>scsi</value>
        <value>virtio</value>
        <value>usb</value>
        <value>sata</value>
        <value>nvme</value>
      </enum>
      <enum name='model'>
        <value>virtio</value>
        <value>virtio-transitional</value>
        <value>virtio-non-transitional</value>
      </enum>
    </disk>
    <graphics supported='yes'>
      <enum name='type'>
        <value>sdl</value>
        <value>vnc</value>
        <value>egl-headless</value>
        <value>dbus</value>
      </enum>
    </graphics>
    <video supported='yes'>
      <enum name='modelType'>
        <value>vga</value>
        <value>cirrus</value>
        <value>virtio</value>
        <value>none</value>
        <value>bochs</value>
      </enum>
    </video>
    <hostdev supported='yes'>
      <enum name='mode'>
        <value>subsystem</value>
      </enum>
      <enum name='startupPolicy'>
        <value>default</value>
        <value>mandatory</value>
        <value>requisite</value>
        <value>optional</value>
      </enum>
      <enum name='subsysType'>
        <value>usb</value>
        <value>pci</value>
        <value>scsi</value>
      </enum>
      <enum name='capsType'/>
      <enum name='pciBackend'>
        <value>default</value>
        <value>vfio</value>
      </enum>
      <enum name='iommufd'>
        <value>no</value>
      </enum>
    </hostdev>
    <rng supported='yes'>
      <enum name='model'>
        <value>virtio</value>
        <value>virtio-transitional</value>
        <value>virtio-non-transitional</value>
      </enum>
      <enum name='backendModel'>
        <value>random</value>
        <value>egd</value>
        <value>builtin</value>
      </enum>
    </rng>
    <filesystem supported='yes'>
      <enum name='driverType'>
        <value>path</value>
        <value>virtiofs</value>
      </enum>
    </filesystem>
    <tpm supported='yes'>
      <enum name='model'>
        <value>tpm-spapr</value>
        <value>spapr-tpm-proxy</value>
      </enum>
      <enum name='backendModel'>
        <value>passthrough</value>
        <value>emulator</value>
        <value>external</value>
      </enum>
      <enum name='backendVersion'>
        <value>1.2</value>
        <value>2.0</value>
      </enum>
    </tpm>
    <redirdev supported='yes'>
      <enum name='bus'>
        <value>usb</value>
      </enum>
    </redirdev>
    <channel supported='yes'>
      <enum name='type'>
        <value>pty</value>
        <value>unix</value>
      </enum>
    </channel>
    <crypto supported='yes'>
      <enum name='model'>
        <value>virtio</value>
      </enum>
      <enum name='type'>
        <value>qemu</value>
      </enum>
      <enum name='backendModel'>
        <value>builtin</value>
        <value>lkcf</value>
      </enum>
    </crypto>
    <interface supported='yes'>
      <enum name='backendType'>
        <value>default</value>
        <value>passt</value>
      </enum>
    </interface>
    <panic supported='yes'>
      <enum name='model'>
        <value>pseries</value>
        <value>pvpanic</value>
      </enum>
    </panic>
    <console supported='yes'>
      <enum name='type'>
        <value>null</value>
        <value>vc</value>
        <value>pty</value>
        <value>dev</value>
        <value>file</value>
        <value>pipe</value>
        <value>stdio</value>
        <value>udp</value>
        <value>tcp</value>
        <value>unix</value>
        <value>qemu-vdagent</value>
        <value>dbus</value>
      </enum>
    </console>
  </devices>
  <features>
    <gic supported='no'/>
    <vmcoreinfo supported='no'/>
    <genid supported='no'/>
    <backingStoreInput supported='yes'/>
    <backup supported='yes'/>
    <async-teardown supported='yes'/>
    <s390-pv supported='no'/>
    <ps2 supported='no'/>
    <tdx supported='no'/>
    <sev supported='no'/>
    <sgx supported='no'/>
    <launchSecurity supported='no'/>
  </features>
</domainCapabilities>
```

jelly:
Related to this, some Fedora testers wanted to run an ARM VM on x86_64 for debugging so maybe this should be generalized?:

Quote from Matrix:

> Best I can tell, cockpit-machines doesn't (yet) support emulating architectures that differ from your host. Probably worth filing an RFE at https://github.com/cockpit-project/cockpit-machines

mvollmer:
> Related to this, some Fedora testers wanted to run an ARM VM on x86_64 for debugging so maybe this should be generalized?:

Not _really_ related: ARM vs x86 is the architecture, not the cpu mode. It would be an option during creation only.

mvollmer:
> So it only uses host-model (sorry I wrote this wrong in my initial comment) for different power versions, as these are run through a special mode. Anything else would need to be done through custom since these would be run via an emulator

If there is no "custom" choice, I think we can keep the current UI, with a single dropdown.  Just put the choices for "custom" under the "host-model" header:

image

I don't like how the domcapabilities don't describe the situation correctly. They still say that "custom" is supported and list the alternatives for that. I'd say this is arguably a bug in libvirt. *shrug*

src/components/vm/overview/cpuModal.tsx
```diff
@@ -57,6 +57,15 @@ export const CPUModal = ({
     models: string[],
 }) => {
     const Dialogs = useDialogs();
+    const [cpuArch, setCpuArch] = useState<string | null>(null);
```
mvollmer, line 60: Can we use `vm.arch` here, or do we really need the host architecture?

SludgeGirl:
> > So it only uses host-model (sorry I wrote this wrong in my initial comment) for different power versions, as these are run through a special mode. Anything else would need to be done through custom since these would be run via an emulator
> 
> If there is no "custom" choice, I think we can keep the current UI, with a single dropdown. Just put the choices for "custom" under the "host-model" header:
> image
> 
> I don't like how the domcapabilities don't describe the situation correctly. They still say that "custom" is supported and list the alternatives for that. I'd say this is arguably a bug in libvirt. _shrug_

If we're going for this approach, wouldn't it make more sense to just do that for ppc and leave the rest of the current behavior as it is, since this was initial just to give greater control over this for ppc. I'm not sure of any reason you'd really need to individually select these options for another other arch. Is it worth just adding an exception for ppc to load the POWER options under host-model and just keep it all contained to the same dropdown?

mvollmer:
> If we're going for this approach, wouldn't it make more sense to just do that for ppc and leave the rest of the current behavior as it is, 

Yes, that's what I meant. On non-power archs, we keep the current UI as it is. On power archs, we drop "custom" from the dropdown and put the choices for "custom" under the "host-model". During apply of the dialog, we dtrt.

In any case, we should have the same general UI for all archs. And since power doesn't need both custom and host-model, I'd say the current UI is good enough for power as well.

PR #2772, opened 2026-08-31
test: Replace "set_val/set_input_text" hack with "paste"

mvollmer:
This is simpler and more robust.


Venefilyn (commented):
LGTM but needs passing tests

mvollmer:
> LGTM but needs passing tests

debian-trixie is hopeless right now with "retries of affected tests", that needs to be investigated separately.

Venefilyn (approved):

Venefilyn:
> > LGTM but needs passing tests
> 
> debian-trixie is hopeless right now with "retries of affected tests", that needs to be investigated separately.

Ack, approved

PR #2781, opened 2026-09-04
[rhel-8] all: Command line hardening

mvollmer:


mvollmer:
The TestMachinesCreate.testCreateCloudBaseImage failure seems real, it happens on both rhel-8-10 and on centos-stream-8 on TF.

I can't run any of these tests locally...

mvollmer:
> I can't run any of these tests locally...

I managed to run it in the old tasks container, and it was a silly paren nesting mistake...

jelly (approved):
Thanks!

PR #2783, opened 2026-09-06
docs: fix two typos in HACKING.md

Avicennasis:
## Summary

Two one-word typo fixes in `HACKING.md`. Documentation only - no code, no user-facing strings, no translatable text touched.

- **HACKING.md** (line 10): `links your checkout to the location were cockpit-bridge looks for packages` -> `... the location where cockpit-bridge looks for packages`
- **HACKING.md** (line 61): `styleint is executed as part of test/static-code` -> `stylelint is executed as part of test/static-code` (matches the parallel `eslint is executed as part of ...` sentence 20 lines above, and the tool is Stylelint everywhere else in the section)

Both survive because `.github/workflows/codespell.yml` runs `codespell src`, so it never looks at top-level Markdown - and `were` is a real word that codespell would not flag anyway.

### Deliberately NOT changed

Nothing under `src/` was touched. In particular I left `src/components/vm/vmReplaceSpiceDialog.tsx` alone: the translatable string `_("Please see $0 how to reconfigure your VM manually.")` on the line below the comment there is an existing msgid with translations in all 21 `po/*.po` catalogues, and rewording it would invalidate every one of them.

No functional changes - documentation only.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -41,7 +41,7 @@ set to upload code changes to `~/.local/share/cockpit/` instead of
 Cockpit Machines uses [ESLint](https://eslint.org/) to automatically check
 JavaScript code style in `.jsx` and `.js` files.
 
-eslint is executed as part of `test/static-code`, aka. `make codecheck`.
+eslint is executed as part of `test/common/static-code`, aka. `make codecheck`.
 
 For developer convenience, the ESLint can be started explicitly by:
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -58,7 +58,7 @@ Rules configuration can be found in the `.eslintrc.json` file.
 Cockpit uses [Stylelint](https://stylelint.io/) to automatically check CSS code
 style in `.css` and `scss` files.
 
-styleint is executed as part of `test/static-code`, aka. `make codecheck`.
+stylelint is executed as part of `test/static-code`, aka. `make codecheck`.
```
tomasmatus: I noticed that `test/static-code` is outdated, these days the path is `test/common/static-code`. If you can, update it, otherwise I can add it myself.

tomasmatus (commented):
Looks good, just a minor comment

Avicennasis:
Thanks — updated in 2b7ae63.

One note: the stale path appeared **twice** in `HACKING.md`, not once — the eslint line (44) as well as the stylelint line (61) you spotted. Both said `test/static-code`, so I corrected both rather than leave one wrong; happy to drop the eslint one if you'd rather keep this PR to the single line.

tomasmatus (approved):
Thanks

PR #2661, opened 2026-05-28
create: Reimplement dialog

mvollmer:
Ideas for improvements:

 - Clear out source when mode is changed.
 - Instead of tooltips for root password and user name, use "explanation" under the text input.
 - Instead of mutual validation, disable user password when user name is empty.
 - Use the regular validation machinery to disable apply when the offline token is missing/invalid.

PR #2785, opened 2026-09-07
Dbg create redo dialog

mvollmer:
Let's debug the test failures in #2661 a bit.

- The pixel test shows a focused TypeaheadSelect, although the assert_pixels function does explicitly unfocus everything...


mvollmer:
Seems to work, I propose that independently.

PR #2788, opened 2026-09-07
test: Wait a bit after filtering the OSes

mvollmer:
The TypeaheadSelect component might asynchronously put the focus back into the input element after a subsequent pixel test has removed it.


mvollmer:
The unexpected focus state has become a problem in #2661, but I think it was always flaky.

mvollmer:
The debian-trixie flakes are something else.

tomasmatus (approved):
* tmatus waits 1 second before clicking

PR #2793, opened 2026-09-10
api: Correctly reuse DBusClient objects

mvollmer:
By making the object that caches them global instead of local...


jelly (approved):
Thanks!

PR #2799, opened 2026-09-16
DBG - figure out what's going on with sd_bus_atttach_event on debian-trixie

mvollmer:
Tests fail on debian-trixie quite often during login with this backtrace:

```
Exception in callback _Transport._read_ready()
handle: <Handle _Transport._read_ready()>
  File "/usr/lib/python3.13/asyncio/events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/transports.py", line 110, in _read_ready
    self._protocol.data_received(data)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 192, in data_received
    result = self.consume_one_frame(self.buffer)
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 142, in consume_one_frame
    self.frame_received(data[start:end])
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 96, in frame_received
    self.control_received(data)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 106, in control_received
    self.channel_control_received(channel, command, message)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/router.py", line 223, in channel_control_received
    endpoint.do_channel_control(channel, command, message)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/channel.py", line 166, in do_channel_control
    self.do_control(command, message)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/channel.py", line 144, in do_control
    self.do_open(message)
    ~~~~~~~~~~~~^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/channels/dbus.py", line 248, in do_open
    self.bus.attach_event(None, 0)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/_vendor/systemd_ctypes/librarywrapper.py", line 194, in <lambda>
    setattr(cls, name, lambda *args: func(*args))
                                     ~~~~^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/_vendor/systemd_ctypes/librarywrapper.py", line 73, in errcheck
    raise OSError(-result, f"{func.__name__}: {os.strerror(-result)}")
OSError: [Errno 22] sd_bus_attach_event: Invalid argument
```

It's not clear why `sd_bus_attach_event` returns EINVAL.

Example test run: 
https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/pull-2661-511db09b-20260914-130351-debian-trixie/log.html


mvollmer:
These failures seem to happen only in  cockpit-machines...

mvollmer:
In normal operation, _all_ calls to sd_bus_attach_event fail with EBUSY. The standard buses (including the internal one) seem to already be attached to a event loop.
Thus, we could just skip that call except for newly created Bus objects. That would be a workaround for a obscure bug, but worth considering.

mvollmer:
> In normal operation, _all_ calls to sd_bus_attach_event fail with EBUSY. The standard buses (including the internal one) seem to already be attached to a event loop.

In fact, Bus.new already calls attach_event unless explicitly told not to, which we don't do.

So the second call in dbus.py seems superfluous in any case.

mvollmer:
> In fact, Bus.new already calls attach_event

Without checking for errors, I might add. :-)

mvollmer:
So we expect a EBUSY but get EINVAL. This means that the code that returns EINVAL runs before the one that would return EBUSY. The only possibility for that is `assert_return(bus, -EINVAL);`, which means we are passing a NULL bus, which seems impossible as well.

mvollmer:
> which means we are passing a NULL bus, which seems impossible as well.

This has been confirmed. `self.bus.value` is None when calling `self.bus.attach_event`. One way this can happen is when `Bus._default_user` fails. Then we have a `Bus` object, but `.value` is still NULL and the next time someone calls Bus.default_user they get a bus that holds a NULL pointer without any exception.

mvollmer:
Here is a annotated debug log:

```
OPEN internal None
ATTACH internal None 720585088 16
```
This is the first channel that is opened and this means that sd_bus_attach_event failed with EBUSY. This is expected: the internal bus is constructed with `Bus.new()` which already attaches the bus to the event loop.
The number 720585088 is the address of the sd_bus instance from `self.bus.value`.
```
OPEN internal None
ATTACH internal None 720585088 16
OPEN None org.freedesktop.hostname1
ATTACH None org.freedesktop.hostname1 720672288 16
OPEN internal None
ATTACH internal None 720585088 16
OPEN internal None
ATTACH internal None 720585088 16
OPEN system org.freedesktop.systemd1
ATTACH system org.freedesktop.systemd1 720672288 16
OPEN system org.freedesktop.hostname1
ATTACH system org.freedesktop.hostname1 720672288 16
OPEN system org.freedesktop.login1
ATTACH system org.freedesktop.login1 720672288 16
OPEN internal None
ATTACH internal None 720585088 16
```
More of the same
```
OPEN session org.freedesktop.systemd1
GET ON THE BUS ERROR [Errno 111] sd_bus_default_user: Connection refused
```
This is the first open a session bus channel. There is no "ATTACH" message for it, but we see that `Bus.default_user()` failed to connect. This gets turned into a "protocol-error" and delivered back to JavaScript. (I don't what JS does with it.)
```
OPEN system org.freedesktop.timedate1
ATTACH system org.freedesktop.timedate1 720672288 16
OPEN internal None
ATTACH internal None 720585088 16
OPEN system org.freedesktop.login1
ATTACH system org.freedesktop.login1 720672288 16
```
More system and internal bus stuff.
```
OPEN session org.libvirt
ATTACH session org.libvirt None 22
```
This is the second open of a session bus channel. Calling attach_event fails with EINVAL. The "None" in the "ATTACH" line means that we have passed a NULL pointer to sd_bus_attach_event.

See https://github.com/allisonkarlitskaya/systemd_ctypes/pull/89 for a potential fix. 

The "Connection refused" error might be a real problem, and due to a more fundamental bug, but one step at a time.

mvollmer:
The fix in https://github.com/allisonkarlitskaya/systemd_ctypes/pull/89 works: https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/pull-2799-df35caa1-20260916-115206-debian-trixie/log.html

mvollmer:
I think that concludes this.

